PR #101, opened 2014-03-17
Added support for remote repositories (with and without authentication)

GeertvanHorrik:
When no local .git folder is available, it is now possible to execute the logic on remote branches:

gitversion -url [repositoryurl]

to use authentication:

gitversion -url [repositoryurl] -u [username] -p [password]

PR #269, opened 2014-10-05
Improved support for dynamic repositories

GeertvanHorrik:
New PR for #245 which was buggy


dazinator:
There are current issues with Dynamic Repositories, see #101 and #336 - and I was wondering if this PR would perhaps also aid in fixing those.

If so an ideas when this will get merged?

JakeGinnivan:
It needs to be rebased and the test needs to be made to work offline (i.e create a fake repo then clone it). if someone gets to that before me, great. Otherwise I hope to do it in the next few days

Issue #323, opened 2014-12-09
GitFlow: Include commits in PreReleaseTag for release, hotfix and feature branches


yannisgu:
For release/hotfix branches currently the PreReleaseTag  looks like this:

     -beta.1

and for feature-branches
   
      -feature-branch-name

The problem is that for every commit, the NuGet version will be the same.  My proposition is to do the same as on the develop branch: Add the number of commits (build metadata) in the pre-release tag. 

The change in the code will be trivial, but I want to hear your opinion, before I send a PR.

JakeGinnivan:
This config approach is done. Check out the 3.0.0 branch

@remyvd I think that is a separate issue. Mind opening a new issue if it is still a problem with 3.0.0

PR #325, opened 2014-12-10
WIP: Continuous deployment and delivery versioning mode

yannisgu:
As discussed #323 GitVersion should support different versioning modes which can be configured in the GitVersionConfig.yaml.

The modes are continuous deployment and continuous delivery.

As @JakeGinnivan says:
> Continuous deployment means you want each build to have a unique nuget version, continuous delivery means you do not.

At the moment the only difference is the pre-release tag. In continuous delivery the scheme is `<configured tag>.1` or if there is an applicable tag `<last-tag-name>-<last-tag-number>+1`. 
In continuous deployment the scheme is `<configured tag>.<nr of commits>`

What is still open:
- [ ] GitHub flow, how to handle those there?
- [ ] Feature branches, how is the scheme here?
- [ ] Are hotfix  branches "the same" as release branches?
- [ ] Support branches?
- [ ] Maybe find a better place than GitFlowVersionFInder for setting the `CurrentBranchConfig` property or maybe even find a better place for this property  than the `context`
- [ ] Do we need more than just the pre-release tag in the versioning mode?
- [ ] Can code be dropped?

Happy to see your comments and thoughts :smile:

### DIFF
--- a/GitVersionCore.Tests/ConfigProviderTests.cs
+++ b/GitVersionCore.Tests/ConfigProviderTests.cs
@@ -64,6 +64,7 @@ branches:
         var config = ConfigurationProvider.Provide(gitDirectory, fileSystem);
         config.Branches["develop"].VersioningMode.ShouldBe(VersioningMode.ContinuousDeployment);
         config.Release.VersioningMode.ShouldBe(VersioningMode.ContinuousDelivery);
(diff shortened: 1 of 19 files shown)


JakeGinnivan:
@yannisgu moving to #338 which is a branch in the main repo, so we can all submit PR's to this effort. This is a great start

PR #327, opened 2014-12-18
Failing test case for feature branch name with a number

orjan:
This is just a test to reproduce my issue with feature branches based on
the jira key e.g. feature/JIRA-123-a-description.

I'm not sure what a valid assert should look like, I have 3 examples I could live with, but the current output is broken from my point of view. What do you think would be a valid assert for this case?


JakeGinnivan:
Closing, lets continue discussion at #337

PR #333, opened 2015-01-03
Use defined project intermediate directory if it exists

### LINKED ISSUE #332: Defined project intermediate directory is not respected
Fix for #146  causes a directory in $(ProjectDir)\obj\$(Configuration) to be created, even if the project has specified a different intermediate output directory.

I believe the value of $(IntermediateOutputPath) should be respected if it exists (which it does by default in VS projects).

bdhess:
Fixes #332

### DIFF
--- a/GitVersionTask/AssemblyInfoBuilder/UpdateAssemblyInfo.cs
+++ b/GitVersionTask/AssemblyInfoBuilder/UpdateAssemblyInfo.cs
@@ -142,7 +142,7 @@
             }
             else
             {
-                tempDir = Path.Combine(IntermediateOutputPath, "obj", Configuration);
+                tempDir = Path.Combine(IntermediateOutputPath, "obj", assemblyInfo);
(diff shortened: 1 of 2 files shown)


gep13:
@SimonCropp when you have a second, can you take a look at this one?  Thanks!

bdhess:
Sorry about that, and for the delay getting back to this.
@grufffta @SimonCropp - ready for another look?  The build appears to be clean now.  And thanks!

JakeGinnivan:
This looks ok to me, but looks like a breaking change. @SimonCropp thoughts?

SimonCropp:
Yeah in the strict sense this is breaking change

JakeGinnivan:
Do we care?

Should we take this into v3 which has the config changes and other breaking stuff. Or just merge and put into next minor release?

SimonCropp:
i think next minor is OK. i doubt many people are using the task on its own. and even if they do it will fail at buid time

PR #337, opened 2015-01-06
Feature branch with jira key

JakeGinnivan:
Replacing #327 with a possible solution to the failing test


JakeGinnivan:
While looking into this issue I noticed that we could have SemVer which has a - in the tag. These should be replaced with `.`. We probably need to look into this more and strip out all `.`'s for SemVer v1 version numbers.

JakeGinnivan:
It also brings up an interesting issue about the difference between GitFlow and GitHubFlow.

For branch `feature/JIRA-123`
GitFlow: `1.1.0-JIRA.123+5`
GitHubFlow: `1.0.1-feature.JIRA.123+5`

Thoughts @orjan @andreasohlund @SimonCropp @gep13

orjan:
I'm not sure that the jira identifier "JIRA-123" should be splitted since an identifier might contain - http://semver.org/ [0-9A-Za-z-] so I would prefer
1.1.0-JIRA-123.5

JakeGinnivan:
@orjan you are right. I didn't realise that hyphens were allowed in a pre-release tag. That seems strange to me, but that means I do not have to replace the - with .

JakeGinnivan:
@orjan as for the .5 vs +5

That will be a configuration option in the new version hopefully. You will have to use the variables or format the version manually at the moment.

orjan:
@JakeGinnivan great news regarding +5 .5 I'll guess we're all having different use cases.

orjan:
@JakeGinnivan both GitFlow and GitHubFlow is looking good but I cannot see why the should differ.
I'll guess #325 will handle customised prerelease tags?

gep13:
@JakeGinnivan so is the current branch tagged at 1.0.0, or the NextVersion.txt contains 1.0.0, or in the YAML file it has the 1.0.0?

In which case, GitFlow is returning correctly, right?

Just want to make sure we are on the same page :smile_cat:

JakeGinnivan:
GitHubFlow bumps patch, GitFlow bumps minor as patch is done with hotfix branches.

gep13:
> GitHubFlow bumps patch, GitFlow bumps minor as patch is done with hotfix branches.

Now that, I should have known.  So what is the difference that you are referring to? The inclusion of the word feature?

JakeGinnivan:
Oh. yeah

GitFlow understands feature branches. GithubFlow doesn't.

gep13:
> Oh. yeah

> GitFlow understands feature branches. GithubFlow doesn't.

Right, sorry, now I follow! I thought you were saying that there was a problem with the differences being there.  But you are saying that these differences are expected, and you were just highlighting them.

I will step away from the keyboard now...

JakeGinnivan:
Updated based on feedback.

For branch feature/JIRA-123
GitFlow: `1.1.0-JIRA-123+5`
GitHubFlow: `1.0.1-JIRA-123+5` - got rid of the feature. seemed silly

andreasohlund:
+1

PR #338, opened 2015-01-07
[WIP] Branch specific configuration

JakeGinnivan:
The goal of this PR is to remove the whole branch specific code and to enable a better "branch-configuration". To do this we need basically three things:
* Branch-specific configuration: _Done by adding a CurrentBranchConfig property to GitVersionContext_
* Implement different version calculation strategies like NextVersionConfigStrategy, MergeMessageVersionStrategy or LastTagVersionStrategy. For all branches the version is then calculated like this

```
var versionStrategies = new IVersionStrategy[] { new NextVersionConfigStrategy(), new MergeMessageVersionStrategy(), new LastTagVersionStrategy() };
var version = versionStrategies.Select(s => s.CalculateVersion(configuration, branchConfiguration)).Max();
```
* Implement different variable providers (continous deployment => every commit has different NuGet Version, continous delivery) to enable following code:
```
var variables = variableProvider.Provide(branchConfiguration.Mode);
```

### Outstanding tasks

 - [x] Branch-specific configuration
 - [x] Update variable provider
 - [ ] Fix broken tests
 - [x] Legacy configuration error messages
 - [ ] Documentation - https://github.com/ParticularLabs/GitVersion/wiki/Branch-Specific-Configuration
 - [ ] Discuss: continuous deployment 4 part stable version not including 4th part if it is 0
 - [ ] Add `useBranchNameAsTag` rather than using "useBranchName" magic string. `useBranchNameAsTag` being true should cause the `tag` setting to be ignored.


yannisgu:
@JakeGinnivan the description is kind of outdated. Maybe change it to something like this:

The goal of this PR is to remove the whole branch specific code and to enable a better "branch-configuration". To do this we need basically three things:
* Branch-specific configuration: _Done by adding a CurrentBranchConfig property to GitVersionContext_
* Implement different version calculation strategies like NextVersionConfigStrategy, MergeMessageVersionStrategy or LastTagVersionStrategy. For all branches the version is then calculated like this

```
var versionStrategies = new IVersionStrategy[] { new NextVersionConfigStrategy(), new MergeMessageVersionStrategy(), new LastTagVersionStrategy() };
var version = versionStrategies.Select(s => s.CalculateVersion(configuration, branchConfiguration)).Max();
```
* Implement different variable providers (continous deployment => every commit has different NuGet Version, continous delivery) to enable following code:
```
var variables = variableProvider.Provide(branchConfiguration.Mode);
```

yannisgu:
I started offline with this version finders.. Will try to quickly create a PR

JakeGinnivan:
No worries. 

I have made a start on the variable provider stuff. I think we need to introduce an effective configuration which flattens everything out based on the current branch and apply fallbacks etc. Will give that a spin

JakeGinnivan:
I think breaking changes are a good idea. Mainly:

 - Config file changes. We could migrate like we do now, but could be complex. Maybe we need to introduce a fileVersion or something to the config file so we can version it properly (i.e upgrade to v3 of the config file to light up all these features).
 - Removing options from the MSBuild tasks and only support the configuration file.

What do yo think @SimonCropp @andreasohlund

SimonCropp:
not sure i like putting in a version number into the file. would prefer to gradually add options and give good errors when we obsolete old ones

+1 on removing MSBuild option support

yannisgu:
This should be really easy to implement: Throw an exception in the old properties of `Config` and handle them nicely.

JakeGinnivan:
Am working on another PR which has a legacy config class, reads in everything legacy then provides migration information in a error message.

Sent from my Windows Phone

JakeGinnivan:
Added a wiki page so we can write the doco as we go

https://github.com/ParticularLabs/GitVersion/wiki/Branch-Specific-Configuration

JakeGinnivan:
I propose getting rid of `ClassicVersion` and maybe `AssemblyFileSemVer` (we have config to change what is included in the assembly version, lets just have a single variable). Thoughts @SimonCropp @andreasohlund

JakeGinnivan:
BTW, I think this will be v3. So I am ok with breaking changes

andreasohlund:
+1 for removing them

JakeGinnivan:
Ok, I have been putting some thought into how this could work. Here is my current thinking of new architecture. Everything is generated using http://www.plantuml.com/plantuml, I will include source and the png.

### Version Calculation
image

```
database Config

NextVersionCalculator -> BaseVersionCalculator: GetBaseVersion
note right: see below for how base is calculated
BaseVersionCalculator -> NextVersionCalculator: 2.0.0|25e4ba

Config -> NextVersionCalculator: Branch increment strategy (patch)

alt base version has pre-release tag
    NextVersionCalculator -> NextVersionCalculator: Increment the number on the tag
else 
    Config -> NextVersionCalculator: pre-release tag for current branch (alpha)
    NextVersionCalculator -> NextVersionCalculator: Increment using correct strategy: 2.0.1
    NextVersionCalculator -> NextVersionCalculator: 2.0.1-alpha.1
end

NextVersionCalculator -> CommitCounter: 25e4ba | current branch
note right
     Calculate number of commits since 
     timestamp of 25e4ba on current branch
end note
CommitCounter -> NextVersionCalculator: 10

NextVersionCalculator -> BuildMetadataCreator: 10 | Context
BuildMetadataCreator -> NextVersionCalculator: 10.branch.develop.sha.5e42ba

NextVersionCalculator -> NextVersionCalculator: 2.0.1-alpha.1+10
```

### Base Version Calculation
image

```
NextVersionCalculator -> BaseVersionCalculator: GetBaseVersion
BaseVersionCalculator -> ConfigNextVersionBaseVersion: GetVersion
BaseVersionCalculator -> MergeMessageBaseVersion: GetVersion
BaseVersionCalculator -> LastTagBaseVersion: GetVersion
BaseVersionCalculator -> ReferenceBranchTagBaseVersion: GetVersion
ConfigNextVersionBaseVersion -> BaseVersionCalculator: -
MergeMessageBaseVersion -> BaseVersionCalculator: 2.0.0|25e4ba
LastTagBaseVersion -> BaseVersionCalculator: 2.0.0-beta.2|bec75d (from release branch)
ReferenceBranchTagBaseVersion -> BaseVersionCalculator: 1.9.0|a2c1e3 (reference branch is master)
BaseVersionCalculator -> BaseVersionCalculator: Max of BaseVersion calculation strategies
BaseVersionCalculator -> NextVersionCalculator: 2.0.0|25e4ba
```

I am going to have a crack to see if this will actually work out and wanted to draw out what I had in my head before coding it. If you see any major issues shout, otherwise I will let you know how I go

yannisgu:
In continous deployment mode, if the version is not a pre-release the SemVer contains 4 digits: `Major.Minor.Path.Commits`. In my opinion if the commits number is 0 (which is the normal case in GitFlow), it should not be included, are there any thoughts about it?

BREAKING CHANGES.md
```diff
@@ -0,0 +1,5 @@
+v3.0.0
+ - `AssemblyFileSemVer` variable removed, AssemblyVersioningScheme configuration value makes this variable obsoluete
```
SimonCropp: obsoluete

JakeGinnivan:
@andreasohlund there is one major downside with this approach, it understands merging and *reachable* versions.

Develop no longer tracks master to bump the version of develop. We also do not have any enforcing if you do GitFlow wrong. I think this is a simpler approach and should work fine.

```
         *  hotfix-1.2.1       -----------C--      
         *                    /              \     
         *  master           A----------------F-----H-------N
         *                    \                    / \     /
         *  hotfix-1.3.1       \                  /   ----L
         *                      \                /         \
         *  release-1.3.0        \        -D----G---        \
         *                        \      /          \        \
         *  develop                -----B----E-------I-----M--O--P
         *                                    \           /
         *  feature                            -------J-K-
```
In this case though when release-1.3.0 is merged into master, J used to bump to 1.4.0. This will no longer happen. If you rebase `feature` then it will bump to 1.4.0 because release-1.3.0 has been merged into develop.

Can you forsee any issues with this?

andreasohlund:
I can't see any major drawbacks, +1 for proceeding

PR #339, opened 2015-01-07
Context assigns branch configuration

JakeGinnivan:


JakeGinnivan:
@yannisgu thoughts?

JakeGinnivan:
@nulltoken you might know better than me about this. We seem to have a race condition in commit counting. Rerunning that test which is failing sometimes fails and sometimes passes..

Could there be an issue with commits and the merge commit having the same timestamp?

JakeGinnivan:
GitVersionCore.Tests.IntegrationTests.GitFlow.SupportBranchScenarios.SupportIsCalculatedCorrectly was failing until a rebuild

nulltoken:
@JakeGinnivan If you're referring to `DevelopBasedVersionFinderBase.NumberOfCommitsOnBranchSinceCommit()`, I don't *think* so.

>Could there be an issue with commits and the merge commit having the same timestamp?

I hope not. Actually relying on timestamps only would be weak. We should always account for the topology (parent/child relationships between commits).

How about surrounding the test code with a tight loop in order to see if we can make it fail more frequently? We may get some more actionable results from a ton of failures.

JakeGinnivan:
@nulltoken doesn't fail that often.. but is intermittent.

```
System.Exception : Test failed 2/100 times with errors:


    ExecuteGitVersion(repository).ToString("f")
        should be
    "2.1.0-unstable.0+0"
        but was
    "2.1.0-unstable.1+1"


    ExecuteGitVersion(repository).ToString("f")
        should be
    "2.1.0-unstable.0+0"
        but was
    "2.1.0-unstable.1+1"
```

nulltoken:
@JakeGinnivan Are you able to Console Out the result of the above method as well and see if that correlates?

JakeGinnivan:
@nulltoken nothing relevant is outputted..

```
Created git repository at 'C:\Users\jake.ginnivan\_code\GitVersion\GitVersionCore.Tests\bin\Debug\TestRepositories\f4151a15-b124-447c-b05a-a30eb692478c'
Running against branch: develop
GitFlow version strategy will be used
Running against branch: support/1.0.0
GitFlow version strategy will be used
Running against branch: release/1.2.0
GitFlow version strategy will be used
Running against branch: support/1.0.0
GitFlow version strategy will be used
Running against branch: hotfix/1.2.1
GitFlow version strategy will be used
Running against branch: support/1.0.0
GitFlow version strategy will be used
```

x100

            var versionOnMasterFinder = new VersionOnMasterFinder();
            var tip = context.CurrentCommit;

            var f = new CommitFilter
            {
                Since = tip,
                Until = context.Repository.FindBranch("master").Tip,
                SortBy = CommitSortStrategies.Topological | CommitSortStrategies.Time
            };

            var c = context.Repository.Commits.QueryBy(f);
            var numberOfCommitsSinceRelease = c.Count();

This is the commit counting code, for some reason we sometimes get an incorrect count. Normally it returns 0, but sometimes it returns 1..

JakeGinnivan:
Progress


```
*   commit 79e7cdf (HEAD, tag: 2.0.0, master, develop)
|\  Merge: ae0f676 2dec116
| | Author: A. U. Thor <thor@valhalla.asgard.com>
| | Date:   0 seconds ago
| | 
| |     Merge branch 'release-2.0.0'
| |   
| * commit 2dec116 (release-2.0.0)
| | Author: A. U. Thor <thor@valhalla.asgard.com>
| | Date:   0 seconds ago
| | 
| |     Test Commit
| |   
| * commit 2b82b01
|/  Author: A. U. Thor <thor@valhalla.asgard.com>
|   Date:   0 seconds ago
|   
|       Test Commit
|  
* commit ae0f676 (tag: 1.1.0)
  Author: A. U. Thor <thor@valhalla.asgard.com>
  Date:   0 seconds ago




*   commit f162942 (HEAD, develop)
|\  Merge: 5ad6616 e5fedc9
| | Author: A. U. Thor <thor@valhalla.asgard.com>
| | Date:   0 seconds ago
| | 
| |     Merge branch 'release-2.0.0'
| |       
| | *   commit 40fcaaa (tag: 2.0.0, master)
| | |\  Merge: 5ad6616 e5fedc9
| |/ /  Author: A. U. Thor <thor@valhalla.asgard.com>
|/| /   Date:   1 second ago
| |/    
| |         Merge branch 'release-2.0.0'
| |   
| * commit e5fedc9 (release-2.0.0)
| | Author: A. U. Thor <thor@valhalla.asgard.com>
| | Date:   1 second ago
| | 
| |     Test Commit
| |   
| * commit c87bb9c
|/  Author: A. U. Thor <thor@valhalla.asgard.com>
|   Date:   1 second ago
|   
|       Test Commit
|  
* commit 5ad6616 (tag: 1.1.0)
  Author: A. U. Thor <thor@valhalla.asgard.com>
  Date:   1 second ago
  
      Test Commit
```

Extra commit!

JakeGinnivan:
How does that even happen...

JakeGinnivan:
@nulltoken if you are not subscribed to thread.. check above

JakeGinnivan:
@nulltoken  Looks like head has not properly moved when tag is applied. Or tag is creating a new commit somehow.

JakeGinnivan:
Graph was produced with

    public void DumpGraph()
    {
        var output = new StringBuilder();

        ProcessHelper.Run(
            o => output.AppendLine(o),
            e => output.AppendLineFormat("ERROR: {0}", e),
            null,
            "git",
            @"log --graph --abbrev-commit --decorate --date=relative --all",
            RepositoryPath);

        Console.Write(output.ToString());
    }

nulltoken:
@JakeGinnivan I'm already watching this thread :wink: 

So the problem isn't related to the way we count,.

nulltoken:
I think I got it. This is indeed a race concurrency issue, but at the git level.

If, within the same second, one creates two commit, with the same parents, the same message and the same tree, they will both bear the same sha.

In a real life scenario, the second graph would be the expected one (two different merge commits).

If you want to enforce and make the test bulletproof (without thread.sleeping for one second), I think that adding one commit on develop at the very beginning of the test would be enough. This way the merge commits will always bear different parents.

JakeGinnivan:
Magic!

BaseGitFlowRepositoryFixture now creates a commit on develop after branching, test reliably passes.

Tree has changed to

```
*   commit 0123d14 (HEAD, develop)
|\  Merge: c9d130e 44e9c2b
| | Author: A. U. Thor <thor@valhalla.asgard.com>
| | Date:   0 seconds ago
| | 
| |     Merge branch 'release-2.0.0'
| |       
| | *   commit f38eb04 (tag: 2.0.0, master)
| | |\  Merge: 6e51d11 44e9c2b
| | |/  Author: A. U. Thor <thor@valhalla.asgard.com>
| |/|   Date:   0 seconds ago
| | |   
| | |       Merge branch 'release-2.0.0'
| | |    
| * | commit 44e9c2b (release-2.0.0)
| | | Author: A. U. Thor <thor@valhalla.asgard.com>
| | | Date:   0 seconds ago
| | | 
| | |     Test Commit
| | |    
| * | commit 5bb8e98
|/ /  Author: A. U. Thor <thor@valhalla.asgard.com>
| |   Date:   0 seconds ago
| |   
| |       Test Commit
| |   
* | commit c9d130e
|/  Author: A. U. Thor <thor@valhalla.asgard.com>
|   Date:   0 seconds ago
|   
|       Test Commit
|  
* commit 6e51d11 (tag: 1.1.0)
  Author: A. U. Thor <thor@valhalla.asgard.com>
  Date:   0 seconds ago
  
      Test Commit
```

JakeGinnivan:
Which doesn't really look right. but think that it is git log issue.

nulltoken:
Have you chased down all the potential similar issues? Maybe one way to put them under the light would be add a temporary `Thread.Sleep(1000)` in `MergeNoFF()`.

JakeGinnivan:
Appears ok with the Thread.Sleep in there. Found another test which already fixed that race condition.

Wondering if we should thread.Sleep in MergeNoFF - or somehow detect the condition which could cause it and throw or thread.sleep on demand?

nulltoken:
I tend to not feel at ease with Thread.Sleep() calls as they make the tests slower.

Regarding the detection of the root cause condition, I'm not sure there's a magic answer.

The way I've worked until now is by creating my test case by hand in git command line and then ensuring that the automated ones were reproducing the same topology.

I've more than once relied on the hack of adding an additional commit on a branch (to force the merge sha to differ). Most of the time, that wasn't hacking the test *per se* as it often makes sense to have a `develop` branch ahead of master, for instance.

JakeGinnivan:
Yeah, I put it in the GitFlow base fixture, so as long as we use those base fixtures it should be fine.

JakeGinnivan:
So after all that, @yannisgu can you have a look at the PR

GitVersionCore.Tests/Fixtures/BaseGitFlowRepositoryFixture.cs
```diff
@@ -24,5 +39,21 @@ void SetupRepo(Action<IRepository> initialMasterAction)
         initialMasterAction(Repository);
 
         Repository.CreateBranch("develop").Checkout();
+        Repository.MakeACommit();
+    }
+
+    public void DumpGraph()
```
yannisgu: This was only for test purpose, not?

JakeGinnivan: yep, test purposes. But is useful if we need to debug issues in the future

GitVersionCore/GitVersionContext.cs
```diff
@@ -9,6 +11,8 @@
     /// </summary>
     public class GitVersionContext
     {
+        readonly bool IsContextForTrackedBranchesOnly;
```
yannisgu: No idea what this field is, but is it correct that the default changed from true to false?

GitVersionCore/GitVersionContext.cs
```diff
@@ -77,5 +81,23 @@ IEnumerable<Branch> GetBranchesContainingCommit(string commitSha)
...
+            {
+                CurrentBranchConfig = new BranchConfig();
+            }
+            else if (matchingBranches.Length == 1)
+            {
+                CurrentBranchConfig = matchingBranches[0].Value;
+            }
+            else
+            {
+                throw new Exception(string.Format("Multiple branch configurations match the current branch name of ''. Matching configurations: '{0}'", string.Join(", ", matchingBranches.Select(b => b.Key))));
```
yannisgu: Current branch name of what? :smiley:

GitVersionCore/GitVersionContext.cs
JakeGinnivan: It is always assigned through ctor. The initial value was unused

GitVersionCore/GitVersionContext.cs
JakeGinnivan: Nice catch, fixed

yannisgu:
Ok seems good :+1: 

Maybe a small test would be helpfull?

JakeGinnivan:
During the course of this PR tests were failing.

Also planning more changes which will require tests. This was more a refactoring step.

Did you see I took your PR and turned it into a branch in the main repo. This PR is targeting that. If you have further changes also submit PRs to that branch then we can build it up in pieces.

yannisgu:
OK

Yes I have seen that, sounds good to me!

PR #341, opened 2015-01-08
Configuration Updates

JakeGinnivan:
@yannisgu this branch is probably useful to you when I finish it. I think without the core configuration changes contained in this PR it will make updating the branch finders much harder (I think that is also why most of your tests are failing).

Thoughts on the direction in this PR?


GitVersionTask/GetVersion.cs
```diff
@@ -98,36 +90,16 @@ public override bool Execute()
         {
             try
             {
-                CachedVersion versionAndBranch;
+                Tuple<CachedVersion, GitVersionContext> versionAndBranch;
```
JakeGinnivan: Not a massive fan of this, but it is a temporary hack to allow me to keep the MSBuild tasks in a working and compatible state.

With the Context being responsible for calculating effective configuration it needs to be exposed with the version for use in the variable provider.

As we get further down this refactoring hopefully more of this can all move into Core

yannisgu:
I like the idea of having a merged `EffectiveConfiguration` object. 

However I would like to have a more dynamic solution here with reflection. Instead of calculating every property one by one in `CalculateEffectiveConfiguration`, I would add a constructor to `EffectiveConfiguration` which takes a `EffectiveConfiguration` and a `BranchConfiguration`. Then both objects gets merged by reflection (iterating over properties of `EffectiveConfiguration`, if it exists on `BranchConfiguration` and is not null, get from it).

JakeGinnivan:
Could do. It would take 3 things, Config (which is the top level config), BranchConfig and EffectiveConfiguration (which is the defaults/fallback for everything).

Will give that a go

JakeGinnivan:
@yannisgu this and #342 are ready to go I think.

#342 needs to be merged first. Once we merge them I think we rebase your branch on top and it should help you get tests passing

yannisgu:
Yeah looks good :+1: 

:shipit:

PR #344, opened 2015-01-10
Config based versioning

JakeGinnivan:
Start of new internals which does not use any branch finders or know anything about branches.

Still a bunch of integration tests which need to be fixed when you switch over to use new stuff in `GitVersionFinder` (uncomment commented line)

PR #346, opened 2015-01-12
Config based versioning improvements

JakeGinnivan:
Builds on #344


JakeGinnivan:
At this point there are a heap of failing tests around commit counting. I think we need to introduce the commit counting strategy which @yannisgu talked about in his PR, but didn't get to that.

But most of the semantic versions are correct with this PR. Looks like we should be able to make his work

JakeGinnivan:
@SimonCropp @andreasohlund @yannisgu this has a green build now

The new config based versioning is *not* turned on. It still needs a line uncommented to switch over to the new approach. There are a heap of failing unit tests once switched, but think that should be a different PR to fix.

The failing tests are mainly around commit counting and where pre-release tags start counting from (some branches are 0, others start at 1).

Would love a review on the approach. I know it is a massive PR so might be worth pulling code down and stepping through the new stuff

GitVersionCore.Tests/GitVersionContextTests.cs
```diff
@@ -50,5 +51,26 @@ public void UsesBranchSpecificConfigOverTopLevelDefaults()
             context.Configuration.VersioningMode.ShouldBe(VersioningMode.ContinuousDeployment);
             context.Configuration.Tag.ShouldBe("alpha");
         }
+
+        [Test]
+        public void CanFindParentBranchForInheritingIncrementStrategy()
+        {
+            var config = new Config();
+            config.Branches["develop"].Increment = IncrementStrategy.Major;
+            config.Branches["feature[/-]"].Increment = IncrementStrategy.Inherit;
```
JakeGinnivan: This solves the issue with feature branches which are taken off master should bump patch but feature branches off develop should bump minor

GitVersionCore.Tests/IntegrationTests/GitFlow/DevelopScenarios.cs
```diff
@@ -6,29 +6,43 @@
...
 {
     [Test]
-    public void WhenDevelopBranchedFromMaster_MinorIsIncreased()
+    public void WhenDevelopBranchedFromTaggedCommitOnMasterVersionDoesNotChange()
     {
         using (var fixture = new EmptyRepositoryFixture(new Config()))
         {
             fixture.Repository.MakeATaggedCommit("1.0.0");
             fixture.Repository.CreateBranch("develop").Checkout();
+            // TODO Should actually be 1.0.0+0
```
JakeGinnivan: I think we have a bug at the moment.

If I tag master, then branch develop from that tag it gets an unstable pre-release tag. It shouldn't because that commit is tagged.

JakeGinnivan:
The overall architecture implemented in this PR is documented at https://github.com/ParticularLabs/GitVersion/pull/338#issuecomment-69468768

yannisgu:
How cool is that? :smile:

GitVersionCore.Tests/GitVersionContextTests.cs
yannisgu: Isn't `IncrementStrategy.Inherit` default for feature branches?

JakeGinnivan: Yeah, I just thought I would make it explicit in the test. The test doesn't make sense without that knowledge

yannisgu: Ah I see

JakeGinnivan:
@yannisgu so what do you recon. Is this along the lines of what you were thinking?

Next steps if this is ok is to merged. Then I will switch the code over so we have failing tests, we can then try and PR to get all the tests green. That is where the commit counting bits will need to come in I think, that was the main source of failing tests from memory

GitVersionCore.Tests/IntegrationTests/GitFlow/DevelopScenarios.cs
SimonCropp: Agreed this seems like a bug

GitVersionCore.Tests/Mocks/MockCommit.cs
```diff
@@ -6,12 +6,17 @@
...
     public MockCommit(ObjectId id = null)
     {
         idEx = id ?? new ObjectId(Guid.NewGuid().ToString().Replace("-", "") + "00000000");
-        MessageEx = "";
+        MessageEx = "Commit " + commitCount++;
         ParentsEx = new List<Commit> { null };
-        CommitterEx = new Signature("Joe", "Joe@bloggs.net", DateTimeOffset.Now);
+        CommitterEx = new Signature("Joe", "Joe@bloggs.net", when);
+        // Make sure each commit is a different time
+        when = when.AddSeconds(1);
```
SimonCropp: No issues with this and running test multithreaded? Can think of any but just wanted to ensure static shared state was ok

GitVersionCore.Tests/Mocks/MockCommitLog.cs
```diff
@@ -8,7 +9,10 @@ public class MockCommitLog : ICommitLog, ICollection<Commit>
 
     public IEnumerator<Commit> GetEnumerator()
     {
-        return Commits.GetEnumerator();
+        if (SortedBy == CommitSortStrategies.Reverse)
```
SimonCropp: This seems weird. If reverse then we dont reverse?

GitVersionCore/BranchNameComparer.cs
```diff
@@ -0,0 +1,18 @@
+namespace GitVersion
+{
+    using System.Collections.Generic;
+    using LibGit2Sharp;
+
+    class BranchNameComparer : IEqualityComparer<Branch>
+    {
+        public bool Equals(Branch x, Branch y)
+        {
+            return x.Name == y.Name;
```
SimonCropp: Since we have caseless running this comparison will be case insensitive. Is this correct? If so then the get hashcode should do a lower otherwise the equals does not matc the behaviour a dictionary will exibit

GitVersionCore/GitVersionContext.cs
```diff
@@ -81,30 +81,83 @@ IEnumerable<Branch> GetBranchesContainingCommit(string commitSha)
...
-
-            Configuration = new EffectiveConfiguration(configuration.AssemblyVersioningScheme, versioningMode, configuration.TagPrefix, tag, nextVersion);
+            var incrementStrategy = currentBranchConfig.Value.Increment ?? IncrementStrategy.Patch;
+            var assemblyVersioningScheme = configuration.AssemblyVersioningScheme;
+            var gitTagPrefix = configuration.TagPrefix;
+            Configuration = new EffectiveConfiguration(assemblyVersioningScheme, versioningMode, gitTagPrefix, tag, nextVersion, incrementStrategy, currentBranchConfig.Key);
         }
 
-        BranchConfig GetBranchConfiguration(KeyValuePair<string, BranchConfig>[] matchingBranches)
+        KeyValuePair<string, BranchConfig> GetBranchConfiguration(Branch currentBranch)
```
SimonCropp: Wouldnt a tuple make more sense here?

GitVersionCore/GitVersionFinder.cs
```diff
@@ -18,6 +18,7 @@ public SemanticVersion FindVersion(GitVersionContext context)
                 throw new Exception("NextVersion.txt has been depreciated. See https://github.com/ParticularLabs/GitVersion/wiki/GitVersionConfig.yaml-Configuration-File for replacement");
             }
 
+            //return new NewNextVersionCalculator().FindVersion(context);
```
SimonCropp: What is the comment for?

GitVersionCore/VersionCalculation/BaseVersionCalculator.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace GitVersion.VersionCalculation
```
SimonCropp: Is there value this being in a sub namespace?

JakeGinnivan: Can we create a .DotSettings files which does that.

It is a pain fighting tooling

GitVersionCore.Tests/Mocks/MockCommit.cs
JakeGinnivan: Not really, if two tests get the same time that is not an issue. It is within the same test we want it to bump

GitVersionCore.Tests/Mocks/MockCommitLog.cs
JakeGinnivan: By default libgit2 starts at latest commit and goes back. Or at least seems to, I naively put reverse = oldest first. @nulltoken what should I do here

GitVersionCore/BranchNameComparer.cs
JakeGinnivan: hrrm, don't need case insensitivity. Should we disable it for this method? Possible we don't even need this if Branch supports equals and gethashcode properly. I didn't check. @nulltoken ? :)

GitVersionCore/GitVersionContext.cs
JakeGinnivan: Possible, but we already have a Dictionary so didn't bother converting between. Also was internal to the class so was not too concerned

GitVersionCore/GitVersionFinder.cs
JakeGinnivan: It turns all this new code on. When this is uncommented a whole heap of tests fail. I wanted to make sure I put the code in without affecting anything existing.

I will uncomment this on the other branch once this PR is ok

GitVersionCore/BranchNameComparer.cs
nulltoken: `Branch`  supports `Equals()`, `==`, `!=` and `GetHashCode()`. The `CanonicalName` and the `Id` of the pointed at `GitObject` will be used to assert the equality. `CanonicalName`s will be compared in a case sensitive way. Those things matter on non case insensitive file systems.

Although if you decide to name two branches `master` and `mAsTeR`, you may deserve to encounter some problems :wink:

gep13: > Although if you decide to name two branches master and mAsTeR, you may deserve to encounter some problems :wink:

Or to encounter a papercut on your eye!

GitVersionCore.Tests/Mocks/MockCommitLog.cs
nulltoken: I'm not sure to understand why you'd like it reversed. Could you please me a little bit more context?

FWIW, `CommitSortStrategies.Reverse` will take more time as the whole log has to be walked before yielding the first result. The default mimics the git.git `git log` behavior that makes sense most of the time and allow to stream the results as the log is walked.

JakeGinnivan: Yeah, so I think this is right.

`Commits.Add()` will add to the bottom of the list internally. So to replicate the commit log we call `.Reverse()`, but if you pass that flag I don't reverse.

I think this is all good

GitVersionCore/GitVersionFinder.cs
JakeGinnivan: Uncommented this. Tests are now failing.

JakeGinnivan:
Turned on the new code. This will cause the build to fail

Have deleted the branch comparer. Other comments (like namespace) we can fix on the other branch if we can figure out how to make resharper/visual studio do that by default.

Are we happy with overall approach of this PR? Is the architecture with the base version finder then apply tags/counting/metadata etc on top of the base version?

GitVersionCore/VersionCalculation/NewNextVersionCalculator.cs
```diff
@@ -0,0 +1,74 @@
...
+        public SemanticVersion FindVersion(GitVersionContext context)
+        {
+            var baseVersion = baseVersionFinder.GetBaseVersion(context);
+
+            if (baseVersion.ShouldIncrement) IncrementVersion(context, baseVersion);
+
+            if (baseVersion.ShouldUpdateTag && !baseVersion.SemanticVersion.PreReleaseTag.HasTag() && !string.IsNullOrEmpty(context.Configuration.Tag))
+            {
+                var tagToUse = context.Configuration.Tag;
+                if (tagToUse == "useBranchName")
```
yannisgu: I would prefer a boolean property "UseBranchNameAsTag" instead of magic strings.

JakeGinnivan: I thought about that, but it is a lot harder.

If we specify defaults, then you override do we clear out the `tag` option. If the user sets `useBranchNameAsTag` to true and also sets `tag` do we throw, or ignore the tag?

yannisgu: I would throw when the users sets `useBranchNameAsTag`, though ignoring it would also be ok.

GitVersionCore/VersionCalculation/FallbackBaseVersionStrategy.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace GitVersion.VersionCalculation
+{
+    using System.Linq;
+    using GitVersion.VersionCalculation.BaseVersionCalculators;
+
+    public class FallbackBaseVersionStrategy : BaseVersionStrategy
```
yannisgu: This one should probably be in the BaseVersionCalculators folder next to the other version strategies?

GitVersionCore/VersionCalculation/NewNextVersionCalculator.cs
JakeGinnivan: Added as a task on the main PR

yannisgu:
@JakeGinnivan yeah I like the architecture,  I'm really looking forward now on how this plays with the new variable provider!

JakeGinnivan:
Ok, will merge this. Then we can start trying to fix tests

PR #350, opened 2015-01-15
Pr/dynamic

JakeGinnivan:
Replaces #269 

This is #269 rebased plus fixed the test so it runs all the time and not against remote repo

### DIFF
--- a/GitVersion.sln.GhostDoc.xml
+++ b/GitVersion.sln.GhostDoc.xml
@@ -23,6 +23,7 @@
       <Protected>false</Protected>
       <Private>false</Private>
       <Inherited>true</Inherited>
+      <InheritedFromReferences>true</InheritedFromReferences>
       <EnableTags>false</EnableTags>
       <TagList />
     </IncludeScopes>
--- a/GitVersionCore/BuildServers/GitHelper.cs
+++ b/GitVersionCore/BuildServers/GitHelper.cs
@@ -7,7 +7,7 @@ namespace GitVersion
 
     public static class GitHelper
     {
-        const string MergeMessageRegexPattern = "refs/heads/pull(-requests)?/(?<issuenumber>[0-9]*)/merge";
+        const string MergeMessageRegexPattern = "refs/heads/(pr|pull(-requests)?/(?<issuenumber>[0-9]*)/merge)";
 
         public static void NormalizeGitDirectory(string gitDirectory, Authentication authentication)
         {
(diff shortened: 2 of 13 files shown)


dazinator:
@JakeGinnivan 
When I ran this, after the following line:
image

I checked the _dynamicrepository directory expecting to see a .git repo.

Instead I see an empty folder:

image

Is this correct? I was kind of expecting to see my git repo located there!

JakeGinnivan:
@GeertvanHorrik help! :)

JakeGinnivan:
@dazinator do you have hidden folders turned on? the .git folder could be hidden. The code says not to check anything out

dazinator:
Ah gees - I am using a new laptop today - and yes.. do not show hidden files is turned on.

Damn you Windows!

*Hits self on head with large object*

image

dazinator:
Cool - this seems to be running well based on my own local testing. I have not tried plugging it in to a team city build yet, I'm curious to know if it will allow the checkout on agent scenario. I will try that next. Cheers Guys!

GeertvanHorrik:
We do a bare clone (not a checkout since we don't need that). That means that you only have the hidden .git folder. For another project, GitLink, I have created this:

https://github.com/CatenaLogic/GitLink/blob/develop/src/GitLink/TemporaryFilesContext.cs

Maybe good to copy/paste to ensure the removing of temporary files? Or are they already being deleted? (I am not sure, done so many projects I simply forgot).

dazinator:
@JakeGinnivan @GeertvanHorrik  - I found a problem.

Because we don't checkout the repo - the files don't exist on disk. This includes the "NextVersion.txt" file which GitVersion expects to be on disk in the repo directory - so it won't ever find that file at the moment! 

I'm not that familiar with Git (as you well know) - would it be possible to query the git repo to see if has such a file and then if it does, check it out? I think that's the only bit that's missing.

SimonCropp:
ok going merge this. @dazinator can u raise a new issue for the nextversion issue and we will work on that separately.

GeertvanHorrik:
It should be supported, see https://github.com/ParticularLabs/GitVersion/blob/ca002ed570b03f2f3a610523ab79adf34848fdc1/GitVersionExe/GitPreparer.cs#L105

dazinator:
Thanks @GeertvanHorrik - from looking at that code - it seems that nextversion txt checkout only happens if you pass in a targetbranch argument - which I'm not doing so that may explain it. I'll raise a seperate issue

SimonCropp:
@dazinator thanks

SimonCropp:
@dazinator thanks

PR #353, opened 2015-01-18
Build warning fixes

JakeGinnivan:


JakeGinnivan:
Bump

orjan:
It's looking good from my point of view. The changes is good looking, the same amount of tests are failing and the warnings are gone except "warning : Fody/JetBrainsAnnotations: Reference to JetBrains.Annotations not found". But that one is present for the master as well.

PR #354, opened 2015-01-18
Fix typo

JakeGinnivan:


gep13:
I am assuming the failing build is due to the on-going work that you are doing, so happy to merge this change just now.

JakeGinnivan:
Yep, all PR's to the Config based branch will fail now because I have turned the new code on. 

The goal is to get that build green now.

gep13:
> The goal is to get that build green now.

Understood.  I am going to be away most of the day today, so won't be able to help out :cry: Just waiting until daughter wakes up from nap, and then heading out.

PR #355, opened 2015-01-18
Fixing tests

JakeGinnivan:
Taking a run at fixing a bunch of the failing tests. All of these tests are mainly failing because of commit counting strategies.

I have simplified this a great deal (in my mind). GitVersion now uses this basic rule:

`Count commits from the source of the base version`, if a base version doesn't have a source (say NextVersion from Config) it will use the source from the next base version.

For example, if the base version comes from a tag then it will be counted from that tag. If it is from a version in a branch name then it will be number of commits since that branch was created. If it comes a merge message it will count from when that merge was done.

The current commit counting confuses me. Would love a decent review of the commit counts in this @andreasohlund @SimonCropp @gep13 @nulltoken if possible :)

### DIFF
--- a/GitVersionCore.Tests/IntegrationTests/GitFlow/GitFlowFeatureBranchTests.cs
+++ b/GitVersionCore.Tests/IntegrationTests/GitFlow/GitFlowFeatureBranchTests.cs
@@ -17,7 +17,7 @@ public class GitFlowFeatureBranchTests
             fixture.Repository.Checkout("feature/JIRA-123");
             fixture.Repository.MakeCommits(5);
 
-            fixture.AssertFullSemver("1.1.0-JIRA-123+5");
+            fixture.AssertFullSemver("1.1.0-JIRA-123.1+5");
         }
     }    
 }
--- a/GitVersionCore.Tests/IntegrationTests/GitFlow/PatchScenarios.cs
+++ b/GitVersionCore.Tests/IntegrationTests/GitFlow/PatchScenarios.cs
@@ -13,13 +13,13 @@ public class PatchScenarios
             // create hotfix
             fixture.Repository.CreateBranch("hotfix-1.2.1").Checkout();
 
+            fixture.AssertFullSemver("1.2.1-beta.1+0");
+            fixture.Repository.MakeACommit();
             fixture.AssertFullSemver("1.2.1-beta.1+1");
-            fixture.Repository.MakeACommit();
-            fixture.AssertFullSemver("1.2.1-beta.1+2");
             fixture.Repository.ApplyTag("1.2.1-beta.1");
-            fixture.AssertFullSemver("1.2.1-beta.1+2");
+            fixture.AssertFullSemver("1.2.1-beta.1+0");
             fixture.Repository.MakeACommit();
-            fixture.AssertFullSemver("1.2.1-beta.2+3");
+            fixture.AssertFullSemver("1.2.1-beta.2+1");
 
             // Merge hotfix branch to master
             fixture.Repository.Checkout("master");
--- a/GitVersionCore.Tests/IntegrationTests/GitFlow/ReleaseBranchTests.cs
+++ b/GitVersionCore.Tests/IntegrationTests/GitFlow/ReleaseBranchTests.cs
@@ -16,7 +16,9 @@ public class GitFlowReleaseBranchTests
             fixture.Repository.CreateBranch("release-2.0.0");
             fixture.Repository.Checkout("release-2.0.0");
 
-            fixture.AssertFullSemver("2.0.0-beta.1+5");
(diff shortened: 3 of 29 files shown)


GitVersionCore.Tests/IntegrationTests/GitFlow/GitFlowFeatureBranchTests.cs
```diff
@@ -17,7 +17,7 @@ public void ShouldNotUseNumberInFeatureBranchAsPreReleaseNumber()
             fixture.Repository.Checkout("feature/JIRA-123");
             fixture.Repository.MakeCommits(5);
 
-            fixture.AssertFullSemver("1.1.0-JIRA-123+5");
+            fixture.AssertFullSemver("1.1.0-JIRA-123.1+5");
```
JakeGinnivan: I think this new behaviour is correct. the .1 is the same as we put on -beta, -unstable etc.

GitVersionCore.Tests/IntegrationTests/GitFlow/PatchScenarios.cs
```diff
@@ -13,13 +13,13 @@ public void PatchLatestReleaseExample()
             // create hotfix
             fixture.Repository.CreateBranch("hotfix-1.2.1").Checkout();
 
-            fixture.AssertFullSemver("1.2.1-beta.1+1");
+            fixture.AssertFullSemver("1.2.1-beta.1+0");
```
JakeGinnivan: Good example of new commit counting. We have not commited to the new branch, so why was it 1?

GitVersionCore.Tests/IntegrationTests/GitFlow/ReleaseBranchTests.cs
```diff
@@ -16,7 +16,9 @@ public void CanTakeVersionFromReleaseBranch()
             fixture.Repository.CreateBranch("release-2.0.0");
             fixture.Repository.Checkout("release-2.0.0");
 
-            fixture.AssertFullSemver("2.0.0-beta.1+5");
+            fixture.AssertFullSemver("2.0.0-beta.1+0");
```
JakeGinnivan: Same here, haven't commited to the release branch yet, so commit count is 0

GitVersionCore.Tests/IntegrationTests/GitHubFlow/ReleaseBranchTests.cs
```diff
@@ -71,7 +75,7 @@ public void WhenReleaseBranchIsMergedIntoMasterHighestVersionIsTakenWithIt()
             fixture.Repository.Checkout("master");
             fixture.Repository.MergeNoFF("release-1.0.0", Constants.SignatureNow());
 
-            fixture.AssertFullSemver("2.0.0+11");
+            fixture.AssertFullSemver("2.0.0+5");
```
JakeGinnivan: I assume this is because we are not as smart with commit counting. It is counting commits along the branch, not including the commits which were merged in. Is this an issue?

JakeGinnivan:
image

Also got tests showing that PR's off develop bump minor but PR's off master bump patch (PR's inherit the incrementing config from the branch they were branched from

// @SimonCropp @andreasohlund @gep13 @yannisgu a review of this would be grand.

GitVersionCore.Tests/IntegrationTests/PullRequestScenarios.cs
```diff
@@ -0,0 +1,99 @@
...
+            fixture.Repository.MakeACommit();
+
+            fixture.Repository.Checkout("master");
+            fixture.Repository.MergeNoFF("feature/Foo");
+            fixture.Repository.CreateBranch("pull/2/merge").Checkout();
+            fixture.Repository.Checkout("master");
+            fixture.Repository.Reset(ResetMode.Hard, "HEAD~1");
+            fixture.Repository.Checkout("pull/2/merge");
+
+            fixture.DumpGraph();
```
yannisgu: Isn't this only for debug purpose? 
Or is it intentional logging? Then it should probably be added to all tests?

JakeGinnivan: It has no side effects but could probably be removed.

JakeGinnivan: Just noticed test is wrong. Should be PullRequest.2

JakeGinnivan:
Anyone have any issues with the general commit counting changes in this.

There is still work to be done on the main branch to get all tests green but don't want to do any more on this. 

Otherwise I can just merge and continue until I get green then you can take it for a spin and review the final changes?

andreasohlund:
The logic of the new commit counting seems sound to me. +1 for proceeding

PR #363, opened 2015-02-11
More test fixes

JakeGinnivan:


JakeGinnivan:
Apart from the pull requests not taking the PR number and putting into the pre-release tag most issues are sorted.

A few of the more complex samples are blowing up too which I need to investigate but it would be helpful if people could start testing this and finding problems

PR #388, opened 2015-03-14
Fixes #387

### LINKED ISSUE #387: Allow specification of target SHA for dynamic repositories
GitLink also supports target sha (a build might be running later and otherwise use a different commit to calculate the version). By default it should use the latest, otherwise it must use the specified commit SHA.

GeertvanHorrik:
Works on both offline and online repositories. If commit id is not found, it will fall back to latest commit (current behavior). Also includes unit tests.

### DIFF
--- a/GitVersionCore/GitVersionContext.cs
+++ b/GitVersionCore/GitVersionContext.cs
@@ -40,7 +40,7 @@
             {
                 Logger.WriteWarning("No specific commit specified or found, falling back to latest commit on specified branch");
 
-                CurrentCommit = repository.Head.Tip;
+                CurrentCommit = currentBranch.Tip;
             }
 
             if (currentBranch.IsDetachedHead())
--- a/GitVersionCore/GitVersionContext.cs
+++ b/GitVersionCore/GitVersionContext.cs
@@ -11,12 +11,12 @@
     {
         readonly Config configuration;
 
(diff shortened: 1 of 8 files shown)


GitVersionCore/GitVersionContext.cs
```diff
@@ -11,12 +11,12 @@ public class GitVersionContext
     {
         readonly Config configuration;
 
-        public GitVersionContext(IRepository repository, Config configuration, bool isForTrackingBranchOnly = true)
-            : this(repository, repository.Head, configuration, isForTrackingBranchOnly)
+        public GitVersionContext(IRepository repository, Config configuration, bool isForTrackingBranchOnly = true, string commitId = null)
```
JakeGinnivan: I think this should be on the config with the branch and such? Otherwise looks good

GeertvanHorrik: To me, Config is something you read from the yaml file. This is not something you want to read from the yaml file (because it's too dynamic). I can easily move it to the config class though, just let me know what you prefer.

JakeGinnivan: I think you are right actually. Was thinking config included other arguments and such.

JakeGinnivan:
@SimonCropp is this something that should be on the Task? I can't see it being used but just wanted to check

PR #390, opened 2015-03-14
Fixes #389

GeertvanHorrik:
Supports release tags coming from develop branch when no commits are made back to develop.

PR #392, opened 2015-03-15
Pr/git flow with develop

JakeGinnivan:
Builds on #390, switches to configuration based approach rather than hard coding branch name


JakeGinnivan:
Thanks @GeertvanHorrik for the initial PR. 

@andreasohlund this PR now contains a breaking change between beta.1 and stable. The branch configurations now use the `-` naming like the overall configuration rather than camel casing.

andreasohlund:
Got it, looks good to me!

Issue #385, opened 2015-03-10
Invalid version number detected in merge message


markpjohnson:
I am looking into using GitVersion 3.0, but it detects the following message as a versioned release: *Merge branch 'release/Sprint_2.0_Holdings_Computed_Balances'*

My team used to use descriptive release branches instead of versioned release branches, and even though we would start at version 1.0, GitVersion 3.0 parses our merge messages and finds version 2.0.

JakeGinnivan:
Not sure how to fix this unless we drop support for parsing versions in format `major.minor`. PR #383 is changing this code, suggestions or thoughts for a fix would be handy.

markpjohnson:
I think it would be better if the regex check for the semantic version parser were more strict. For example, if the branch were *release/2.0* instead of the branch containing *2.0*, then I would expect the version parser to extract a 2.0 version. 

It looks like the best change would be to modify the regex pattern in the SemanticVersion.TryParse method.

JakeGinnivan:
Did you want to have a go and submit a pull request to the release/3.0.0 branch?

JakeGinnivan:
See #395 for another case and another test case

orjan:
Adding another failing test case
```
        [Test]
        public void AMergeCommitIncludingAThirdPartyLibraryVersionShouldNotBeParsedAsAVersion()
        {
            var mergeCommitWithIpNumber = @"Merge pull request #1 in FOO/bar from feature/ISSUE-1 to develop

* commit '38560a7eed06e8d3f3f1aaf091befcdf8bf50fea':
  Updated jQuery to v2.1.3";

            var parents = GetParents(true);

            AssertMergeMessage(mergeCommitWithIpNumber, null, parents);
        }

```

JakeGinnivan:
Yeah. this is definitely an issue. If someone doesn't pick this up I will try do it this weekend. PR's welcome though :)

orjan:
The previous version of the merge message based parse was pretty fine grained with:
```
            var knownMergePrefixes = new[] { "Merge branch 'hotfix-", "Merge branch 'hotfix/", "Merge branch 'release-", "Merge branch 'release/" };
```
and some other if:s and but:s. The question is whether we should
* Revert to the old approach
* Make it configurable
* Try fix the current generic version?

JakeGinnivan:
I am not sure really. I am tempted to try and make it generic with something like this:

```
[...]

PR #397, opened 2015-03-18
[SPIKE] A more strict approach for finding versions in commit messages

orjan:
It's hard to satisfy the tests without making the assumption that the relevant version is located at the first line. I don't the like the implementation but I think the added tests is correct?
 
References #385

### DIFF
--- a/GitVersionCore.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
+++ b/GitVersionCore.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
@@ -67,6 +67,25 @@
             AssertMergeMessage(message + "\n ", expectedVersion, parents);
         }
 
+        [TestCase(@"Merge pull request #1 in FOO/bar from feature/ISSUE-1 to develop
+
+* commit '38560a7eed06e8d3f3f1aaf091befcdf8bf50fea':
+  Updated jQuery to v2.1.3")]
+        [TestCase(@"Merge pull request #45 in BRIKKS/brikks from feature/NOX-68 to develop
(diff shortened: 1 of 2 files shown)


orjan:
I have a couple of issues with this implementation.

In order to support:
```csharp
        [TestCase("Finish Release-0.12.0", true, "0.12.0")] //Support Syntevo SmartGit/Hg's Gitflow merge commit messages for finishing a 'Release' branch
        [TestCase("Finish 0.14.1", true, "0.14.1")] //Support Syntevo SmartGit/Hg's Gitflow merge commit messages for finishing a 'Hotfix' branch
```

and not library updates in merge commits with an aggregated commit history like:
```
Merge pull request #1 in FOO/bar from feature/ISSUE-1 to develop
* commit '38560a7eed06e8d3f3f1aaf091befcdf8bf50fea':
  Updated jQuery to v2.1.3
```

We'll need to make some assumptions:
https://github.com/orjan/GitVersion/commit/db7beab257660a09586fdc2efa29ede40ee16e28#diff-ff51c8573866921bbeb1edf0998dc6f8R47
```"^.*?(-|/|'|Finish )(?<PossibleVersions>\d+\.\d+\.\d+)"```
* The version is located at the first commit line
* and starts with ', /, - or Finish

This makes the feature hard to understand for the users, why is my message picked up as a version or why not is it picked up. This makes the implementation error prone and fussy e.g. why if Finish a magical word?

Another option would be to make this feature optional and allow users to specify how to find the version in a commit message via configuration. But that will cause breaking changes.

Thoughts @JakeGinnivan?

JakeGinnivan:
This is fine, it passes all the tests so I am happy.

Issue #404, opened 2015-03-22
It would be nice to redirect the json output to a file rather than console


anobleperson:
Some script languages make it hard to extract information sent to the console (looking at you Powershell!) so it would be nice to send the output to a file instead.

gep13:
@anobleperson I am just curious, what exactly is hard about doing this in PowerShell?  I do this internally at work, and I don't have an issue doing it.  I am not back in the office until Tuesday, but I can certainly dig out the code when I get back if it would help.  Also, assuming you are running on a build server, why not simply read the generated version number information from Environment Variables?

SimonCropp:
Isnt it a case of using outfile? http://www.computerperformance.co.uk/powershell/powershell_file_outfile.htm#Example_1_-_Starting_with_Out-File

anobleperson:
@SimonCropp  Unfortunately it's not that simple. We run it as a separate process ($process = [Diagnostics.Process]::Start($startInfo)), so then we need to capture the output, which means either hooking the async screen updates, or waiting until the process completes and capturing the output. If there is an error then something goes weird in the output stream and it waits for a timeout (30mins) before continuing.

@gep13 I'd be interested in how you execute the commands.

gep13:
@andreasohlund I checked in-house, and we are using the approach documented here: http://jake.ginnivan.net/blog/2014/05/25/simple-versioning-and-release-notes/ by @JakeGinnivan without any problems.

JakeGinnivan:
Whats the status of this? Can we close it or create a `up-for-grabs` issue with a clear plan?

gep13:
For my point of view, and my understanding of how GitVersion, GitReleaseNotes, and GitReleaseManager will interact, I think it would still be a nice to have.

JakeGinnivan:
@gep13 yeah, I agree. Could you create an up-for-grabs issue with well laid out requirements?

gep13:
Yip, let me try to do that tonight.

JakeGinnivan:
Bump, just putting together release notes. Also related discussion in #534

Issue #403, opened 2015-03-22
ShowVariable switch has incorrect help and fails to work


anobleperson:
In the code the actual switch is /v, and even when used the mechanism for validating the variable names is now out of date.

JakeGinnivan:
Fixed by #482

PR #405, opened 2015-03-22
Fixes #400 #403 #404

### LINKED ISSUE #400: No command line setting to Include untracked branches when calculating versions
TFS build for Git checks out the remote branch, which means there are no local branches, hence for GitVersion to work it needs to use untracked branches. There is no command line option for that.

anobleperson:
Sorry about the spam of pull requests, hopefully I've got it right this time, just let me know.


GitVersionExe/Program.cs
```diff
@@ -124,9 +125,21 @@ static int Run()
...
+                            jsonOutput =part;
                             break;
                     }
+                    if (string.IsNullOrWhiteSpace(arguments.JsonOutputFile))
+                    {
+                        Console.WriteLine(jsonOutput);
+                    }
+                    else
+                    {
+                        var outfile = new FileInfo(arguments.JsonOutputFile);
```
JakeGinnivan: This should be able to just specify a filename and it writes it to the working directory.

GitVersionExe/Arguments.cs
```diff
@@ -33,5 +33,9 @@ public Arguments()
         public string UpdateAssemblyInfoFileName;
 
         public bool ShowConfig;
+
+        public bool IncludeUntrackedBranches;
```
JakeGinnivan: i can't see this being used anywhere?

anobleperson: It's being used in the Context creation... last diff at: https://github.com/anobleperson/GitVersion/commit/539c5e493712492d9df34de0a2a6f9d5fbaf84b9

GitVersionExe/Program.cs
anobleperson: Do you mean the FileInfo, or GitVersion itself? I'm likely missing the obvious, but what call should write to the working directory. Out of interest, writing to the working directory is not adequate because it can be called from different locations (thanks TFS!)

JakeGinnivan:
I didn't have time to look into this properly. I have just pushed Beta 2. I will look at this soon hopefully.

Hopefully 3.0.0 stable will not be too far off and we can get this in there

anobleperson:
No worries, I’ve been (and still are) on leave.  Let me know any questions.

cheers,
ash.

On 7 April 2015 at 4:42:04 pm, Jake Ginnivan (notifications@github.com) wrote:

I didn't have time to look into this properly. I have just pushed Beta 2. I will look at this soon hopefully.

Hopefully 3.0.0 stable will not be too far off and we can get this in there

JakeGinnivan:
Hey @anobleperson 

I have pulled 17a93cd into master, but the rest I am unsure about the approach. What we need is to add TFS as a known build server, then branch normalisation will be applied and it should fix up all the issues with branches not being local branches.

See https://github.com/ParticularLabs/GitVersion/blob/master/GitVersionCore/BuildServers/TeamCity.cs

anobleperson:
Sorry, been a bit distracted. I tried this approach last week, but it looks like the build server approach needs git credentials to work, which tfs  doesn't provide. Our scripts run after everything is fetched, so I think we'd need to check in the password which wouldn't be ideal. The problem is it only checks out master locally, then checks out the revision, not the branch, so everything is remote.  Is there a better forum to discuss this, or do you just want to do it here?

Cheers,
Ash.

JakeGinnivan:
Lets move the conversation to https://github.com/GitTools/GitVersion/issues/518 - I have created a dedicated issue to figure out team build.

Issue #406, opened 2015-03-23
GitVersionTasks throws ArgumentNullException when building on TeamCity.


GeirGrusom:
> Error occurred: System.ArgumentNullException: Value cannot be null.
Parameter name: path
   at System.IO.Directory.GetParent(String path)
   at GitVersion.ConfigurationProvider.GetConfigFilePath(String gitDirectory) in c:\BuildAgent\work\5106452ed7682238\GitVersionCore\Configuration\ConfigurationProvider.cs:line 56
   at GitVersion.ConfigurationProvider.Provide(String gitDirectory, IFileSystem fileSystem) in c:\BuildAgent\work\5106452ed7682238\GitVersionCore\Configuration\ConfigurationProvider.cs:line 11
   at GitVersionTask.WriteVersionInfoToBuildLog.InnerExecute() in c:\BuildAgent\work\5106452ed7682238\GitVersionTask\WriteVersionInfoToBuildLog.cs:line 53
   at GitVersionTask.WriteVersionInfoToBuildLog.Execute() in c:\BuildAgent\work\5106452ed7682238\GitVersionTask\WriteVersionInfoToBuildLog.cs:line 31 

Project using GitFlow. Builds fine locally, but fails on TeamCity. ~~It seems that SolutionDirectory is null~~.
I have done no configuration other than installing the NuGet pacakge and removing AssemblyInfoAttributes.

Server is building using Microsoft Build Tools 2013.

GeirGrusom:
It seems that the cause is that the build agent doesn't actually check out sources, but rather gets the repository, excluding the `.git` folder.

The obvious fix would be to check out sources on the build-agent.

A new error arises:

PR #407, opened 2015-03-24
Check for null when looking for a git directory and show an appropriate ...

GeirGrusom:
...error. Fixes issue #406.

By fixing I mean it throws an exception describing the issue rather than propagating a null further.


JakeGinnivan:
Thanks for the PR

PR #410, opened 2015-03-31
Added support for -nofetch command line argument and 'NoFetch' MSBuild T...

Martaver:
...ask parameter, to suppress fetching from remotes during execution.

Up for review and discussion. I had no problem feeding in the noFetch argument, but I couldn't find an obvious way to test whether or not GitVersion was trying to fetch from a remote or not given the current test fixtures available.  Any pointers?

### DIFF
--- a/GitVersionCore/BuildServers/AppVeyor.cs
+++ b/GitVersionCore/BuildServers/AppVeyor.cs
@@ -25,10 +25,7 @@
                 throw new WarningException("Failed to find .git directory on agent.");
             }
 
-            if (!noFetch)
-            {
-                GitHelper.NormalizeGitDirectory(gitDirectory, authentication);
-            }
+            GitHelper.NormalizeGitDirectory(gitDirectory, authentication, noFetch);
         }
 
         public override string GenerateSetVersionMessage(string versionToUseForBuildNumber)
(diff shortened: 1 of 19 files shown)


GitVersionCore/BuildServers/AppVeyor.cs
```diff
@@ -18,14 +18,17 @@ public override bool CanApplyToCurrentContext()
...
-        public override void PerformPreProcessingSteps(string gitDirectory)
+        public override void PerformPreProcessingSteps(string gitDirectory, bool noFetch)
         {
             if (string.IsNullOrEmpty(gitDirectory))
             {
                 throw new WarningException("Failed to find .git directory on agent.");
             }
 
-            GitHelper.NormalizeGitDirectory(gitDirectory, authentication);
+            if (!noFetch)
```
JakeGinnivan: To save this duplication can we pass it down to NormalizeGitDirectory, otherwise it will be easy to forget this step with new Build Servers

GitVersionExe.Tests/ArgumentParserTests.cs
```diff
@@ -211,4 +211,11 @@ public void can_log_to_console()
         var arguments = ArgumentParser.ParseArguments("-l console -proj foo.sln");
         arguments.LogFilePath.ShouldBe("console");
     }
+
+    [Test]
+    public void nofetch_true_when_defined()
```
JakeGinnivan: Can you make sure arguments after this get parsed properly. The parser is not that great and switches have created issues in the past

JakeGinnivan:
Thanks for the PR

PR #422, opened 2015-04-23
Fix #411

### LINKED ISSUE #411: StackOverflowException when two feature branches are pointing to the same commit
I made a mistake and set two feature branches to point to the same commit. When running `gitversion` on one of the branches it crashes with a `StackOverflowException`. Of course this is my mistake because two feature branches shouldn't point to the same commit, but maybe you can detect and break the recursion (or whatever the exception triggers).

Run the following commands to reproduce the issue:

    mkdir stackoverflowtest
    cd .\stackoverflowtest
    git init
    git commit --allow-empty -m "Commit 1"
    git checkout -b develop
    git checkout -b feature/feature1
    git commit --allow-empty -m "Commit 2"
    git commit --allow-empty -m "Commit 3"
    git checkout -b feature/feature2
    gitversion # This will take some time and then crash with a StackOverflowException
    cd ..
    rm -Force -Recurse .\stackoverflowtest

The version of GitVersion is used was 3.0.0-Beta1.

yannisgu:
Fixes #411.

Not: I rebased it on top of #423, so #423 must be merged first

### DIFF
--- a/GitVersionCore.Tests/IntegrationTests/FeatureBranchScenarios.cs
+++ b/GitVersionCore.Tests/IntegrationTests/FeatureBranchScenarios.cs
@@ -63,7 +63,7 @@ public class FeatureBranchScenarios
             fixture.Repository.CreateBranch("feature/feature2");
             fixture.Repository.Checkout("feature/feature2");
 
-            fixture.AssertFullSemver("1.0.1-test.1+5");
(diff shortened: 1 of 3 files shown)


yannisgu:
Build is failing, but it's not related.. Will have a look if I can fix it

PR #427, opened 2015-04-23
This changes AssemblyVersioningScheme.MajorMinorPatchMetadata to Assembl...

### LINKED ISSUE #391: Classic version and MajorMinorPatchMetadata
VariableProvider resets the CommitsSinceTag:

```
                // For continuous deployment the commits since tag gets promoted to the pre-release number
                semanticVersion.PreReleaseTag.Number = semanticVersion.BuildMetaData.CommitsSinceTag;
                semanticVersion.BuildMetaData.CommitsSinceTag = null;
```

So the AssemblySemVer will never have the right last number?

1) Why is it reset to null? Am I not understand the VersioningMode correctly?
2) A possible fix would be this change in AssemblyVersionsGenerator:

```
return string.Format("{0}.{1}.{2}.{3}", sv.Major, sv.Minor, sv.Patch, sv.BuildMetaData.CommitsSinceTag ?? sv.PreReleaseTag.Number ?? 0);
```

JakeGinnivan:
...yVersioningScheme.MajorMinorPatchTag

The issue with metadata is that it is not part of the semantic version at all, so why are we even allowing it to be put into the assembly version. On the other hand in continuous delivery we promote metadata to the tag to allow the version to change every commit. By using the tag, not the metadata we can create meaningful AssemblySemVer values.

Fixes #391

### DIFF
--- a/GitVersionCore/AssemblyVersioningScheme.cs
+++ b/GitVersionCore/AssemblyVersioningScheme.cs
@@ -2,7 +2,7 @@
 {
     public enum AssemblyVersioningScheme
     {
-        MajorMinorPatchMetadata,
+        MajorMinorPatchTag,
         MajorMinorPatch,
         MajorMinor,
         Major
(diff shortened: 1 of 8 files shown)


JakeGinnivan:
@GeertvanHorrik this sorts one of your issues. It is a breaking change between beta.2 and beta.3 or RTM but I think it's worth it

JakeGinnivan:
@GeertvanHorrik can you review this change

GeertvanHorrik:
Looks good to me!

PR #432, opened 2015-04-30
Add NoFetch option to GitVersionTask.targets

MarkZuber:
I need the ability in our builds to set the NoFetch property for the GitVersionTask.  This isn't exposed as configurable anywhere when the GitVersionTask nupkg is referenced into a csproj and it's run automatically.


MarkZuber:
This is my first attempt at a pull request via GitHub, so please excuse me if I'm submitting this in the wrong way.  Happy to re-do this differently if there's a better process.  thanks :D

PR #439, opened 2015-05-12
Pick up GitVersionConfig.yaml when using a dynamic repository

eggapauli:
I implemented a test that shows that `GitVersionConfig.yaml` isn't picked up correctly when a dynamic repository is used.  
In 7c27ad0ef91172fe6c0112bc5feec69341d88a82 you said that the config will be picked up from the original repo. but I think this isn't true since `ConfigurationProvider.GetConfigFilePath` is called with the path to the dynamic repository.  
Checking out `GitVersionConfig.yaml` in the dynamic repository is probably the easiest thing to resolve this (you removed this in the above mentioned commit).  
Or do you have some better idea?

### DIFF
--- a/GitVersionExe.Tests/GitPreparerTests.cs
+++ b/GitVersionExe.Tests/GitPreparerTests.cs
@@ -5,6 +5,7 @@ using GitVersion;
 using LibGit2Sharp;
 using NUnit.Framework;
 using Shouldly;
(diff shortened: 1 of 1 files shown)


JakeGinnivan:
I have worked through this issue (I think) in #451. Would you be able to have a look @eggapauli ?

PR #442, opened 2015-05-14
Resolve incorrect generation of meta count when there are multiple remote branches 

MeirionHughes:


MeirionHughes:
Added a failing test which https://github.com/MeirionHughes/GitVersion/commit/f61179cbd95fde509ab4d018dbae332cc8b40850 will fix.

PR #447, opened 2015-05-28
Showvariable paramter fix

KyleOndy:
Working on #403. I am unfamiliar with reflection so I am not sure I am even on the right path here. I will look into writing some tests shortly.

### DIFF
--- a/GitVersionExe/ArgumentParser.cs
+++ b/GitVersionExe/ArgumentParser.cs
@@ -185,7 +185,7 @@ namespace GitVersion
                     throw new WarningException("assemblyversionformat switch removed, use AssemblyVersioningScheme configuration value instead");
                 }
 
-                if ((IsSwitch("v", name)) && VersionParts.Contains(value.ToLower()))
+                if ((IsSwitch("showvariable", name)) && VersionParts.Contains(value.ToLower()))
                 {
                     arguments.ShowVariable = value.ToLower();
                     continue;
--- a/GitVersionExe/ArgumentParser.cs
+++ b/GitVersionExe/ArgumentParser.cs
@@ -4,7 +4,6 @@ namespace GitVersion
     using System.Collections.Generic;
     using System.Collections.Specialized;
     using System.Linq;
-    using System.Reflection;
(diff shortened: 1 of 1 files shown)


GitVersionExe/ArgumentParser.cs
```diff
@@ -12,7 +12,7 @@ public class ArgumentParser
     {
         static ArgumentParser()
         {
-            var fields = typeof(VariableProvider).GetFields(BindingFlags.Public | BindingFlags.Static);
+            var fields = typeof(VersionVariables).GetProperties();
             VersionParts = fields.Select(x => x.Name.ToLower()).ToArray();
```
JakeGinnivan: Should actually be able to just do `new VariableProvider().AvailableVariables` now.

KyleOndy:
I see I failed the CI build. I'll take a better look at this tonight.

JakeGinnivan:
`Using directive is not required by the code and can be safely removed`

`using System.Reflection` is not needed anymore

GitVersionExe/ArgumentParser.cs
KyleOndy: I see an 'AvailableVariables' property contined in VersionVariables. To access that property I need to instantiate a new VersionVariables instance.

JakeGinnivan:
One additional thing which would be good is to make a better error message when the variable doesn't exist.

At the moment:  
.\GitVersionExe\bin\Debug\GitVersion.exe /showvariable SemVer2  
`Failed to parse arguments: /showvariable SemVer2`

I would like

```
Variable SemVer2 doesn't exist, available variables:
....
list them all
....
```

Do you think you could do that as part of this PR? Otherwise this looks good

JakeGinnivan:
@KyleOndy any thoughts on extending this PR to list available variables before I merge?

KyleOndy:
Let me clean up what I have and get some tests written.

JakeGinnivan:
@KyleOndy how did you go?

JakeGinnivan:
Superseded by #482 

@KyleOndy Thanks for the contribution anyways :)

PR #451, opened 2015-05-28
Dynamic repo config file

JakeGinnivan:
A more complete attempt at #439

Also reduces a lot of the additional special code which GitVersionTask uses, bringing it closer to being just a simple shim onto core


eggapauli:
I can't give you any feedback on 022f2c4 because I'm not familiar with this part of GitVersion. The other commits look fine, well done.

Issue #292, opened 2014-11-10
support for other names for tags and develop branch


PeterLehmann:
I do like this tool, so version info is keep in git, but on this project has a other naming conv. for tags and develop branch, so we fail back to githubflow and gitflow(what are are runnign with some diff.).
so we will be lovy that you can define what prefix version tags has e.g. version-{x.y.z}
and or what the name for main develop branch is called.
i don't knwon if config or passing it into task/command is the best way.

JakeGinnivan:
This is fully supported in 3.0.

Phew.. that was *a lot* of work...

asbjornu:
Great work! Exactly how is this supported? Via `.yml` configuration?

JakeGinnivan:
Yep, via the config. You can use 'GitVersion init' to generate a sample config file or 'gitversion /showConfig' to see the effective config and just override what you want to.

Sent from my Windows Phone
________________________________

PR #456, opened 2015-06-03
Failing inheritance fixes

JakeGinnivan:
Builds on @rcknight's pull request at #429. This fixed the perf issues when inheriting branch config and also fixes the failing test in #429 

@asbjornu @GeertvanHorrik fyi and code review would be grand

@nulltoken could you have a look at the libgit2 stuff?

### DIFF
--- a/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/GitVersionCore/BranchConfigurationCalculator.cs
@@ -117,8 +117,8 @@ namespace GitVersion
             else
                 errorMessage = "Failed to inherit Increment branch configuration, ended up with: " + string.Join(", ", possibleParents.Select(p => p.Name));
 
-            var hasDevelop = repository.Branches["develop"] != null;
-            var branchName = hasDevelop ? "develop" : "master";
+            var developBranch = repository.Branches.FirstOrDefault(b => Regex.IsMatch(b.Name, "^develop", RegexOptions.IgnoreCase));
+            var branchName = developBranch != null ? developBranch.Name : "master";
 
             Logger.WriteWarning(errorMessage + Environment.NewLine + Environment.NewLine + "Falling back to " + branchName + " branch config");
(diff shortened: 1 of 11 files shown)


GitVersionCore/BranchConfigurationCalculator.cs
```diff
@@ -117,8 +131,8 @@ public class BranchConfigurationCalculator
             else
                 errorMessage = "Failed to inherit Increment branch configuration, ended up with: " + string.Join(", ", possibleParents.Select(p => p.Name));
 
-            var hasDevelop = repository.Branches["develop"] != null;
-            var branchName = hasDevelop ? "develop" : "master";
+            var developBranch = repository.Branches.FirstOrDefault(b => Regex.IsMatch(b.Name, "^develop", RegexOptions.IgnoreCase));
```
GeertvanHorrik: Why using regex for this one? Maybe it's even better to make this a setting (I see some people using "dev" instead on the inet).

GeertvanHorrik:
I think it looks good, but I am not 100 % into GitVersion (yet), so might have overseen something.

GitVersionCore/BranchConfigurationCalculator.cs
asbjornu: A setting for this is already available (as per #292), no?

GeertvanHorrik: Then we should use it :)

JakeGinnivan: I think this can be removed. The new logic should allow not having this fallback. Let me look into it

Sent from my Windows Phone

JakeGinnivan: This can't be removed at the moment. If the code gets to the point where it cannot figure out which branch to inherit from then this code needs to be able to pick between master and develop.

An example:

```
* [master, develop]
|
* [feature/foo]
```

In this scenario there is no way to know if GitVersion should inheirit the config from master or develop, this code makes sure that develop gets selected. It isn't great but I can't think of a better solution?

Issue #461, opened 2015-06-12
Version is not incremented by commits within beta branches (release and hotfix branches)


FabianTrottmann:
We create nuget packages with the version we got from GitVersion.exe. The problem we have, is that none of the generated versions fits our needs. We used to use LegacySemVerPadded because it is the right format to create a valid nuspec and nuget package (1.8.0-beta0001). But now we noticed, that when we create release or hotfix branches (GitFlow branching model) and create commits within these branches, the version is not incremented by commits. This leads to deployment problems, because we cannot deploy the same version (which actually is not the same version) twice.

The strange thing is, that this behavior does not appear within the develop branch (unstable). It seems that in the develop branch the SemanticVersionPreReleaseTag.Number is incremented. Where as in release and hotfix branches this number always remains one.

We could now use ClassicVersionWithTag, but then we have the problem with the format when we want to create nuget packages (1.8.0.9-beta.1 is not allowed).

gep13:
@FabianTrottmann just wanted to check...

What version of GitVersion are you currently using?

FabianTrottmann:
@gep13 GitVersion.Portable.2.0.1

MeirionHughes:
try 3.0.3: choco GitVersion.Portable -pre -y

gep13:
I would definitely suggest that you try out the current beta version, as @MeirionHughes suggests.  @JakeGinnivan has been putting in a ton of work to ensure that GitVersion is far more maintainable and support much more workflows than it did in the 2.0.x release.  It could (and hopefully will be the case) that the issue that you are seeing is already corrected.

If not, then I would suggest you raise a PR against the beta version, not directly against master.

FabianTrottmann:
@MeirionHughes @JakeGinnivan 
The version remains the same (not incremented) as it is in 2.0.1. I'll raise a PR against the beta version.
Thanks for the quick responses

MeirionHughes:
How about if you change the branch config to use continuous delivery? (not the one its set to by default). I think that is supposed to bump the version per commit.

JakeGinnivan:
Yeah, try using `gitversion init` it explains the config. also read https://github.com/ParticularLabs/GitVersion/blob/master/Docs/versionIncrements.md which explains why things are like they are.

In short if we incremented every commit then you would not be doing SemVer

JakeGinnivan:
[...]

PR #462, opened 2015-06-12
appended legacySemVer with commits since tag to get the version incre…

FabianTrottmann:
Relates to #461. It increments LegacySemVerPadded by commits and therefore creates the corresponding NuGetVersionV2.


FabianTrottmann:
noticed failing tests, but just wanted to give you the idea, what i would need

JakeGinnivan:
This is by design. You can change the mode GitVersion runs in so it increments every commit in v3

PR #464, opened 2015-06-15
Fix tracking branches in dynamic repositories

eggapauli:
When running GitVersion with a dynamic repository using a new branch it cannot find a remote tracking branch. I dug a bit deeper and found that in `GitHelper.CreateMissingLocalBranchesFromRemoteTrackingOnes` the `remoteTrackingReference` is a `DirectReference`, but I think it should be a `SymbolicReference` for this to work.

I'm not sure how to fix this as I'm not familiar with LibGit2Sharp. Any advice?

### DIFF
--- a/GitVersionCore/BuildServers/GitHelper.cs
+++ b/GitVersionCore/BuildServers/GitHelper.cs
@@ -185,7 +185,9 @@ namespace GitVersion
 
             foreach (var remoteTrackingReference in repo.Refs.FromGlob(prefix + "*").Where(r => r.CanonicalName != remoteHeadCanonicalName))
             {
-                var localCanonicalName = "refs/heads/" + remoteTrackingReference.CanonicalName.Substring(prefix.Length);
+                var remoteTrackingReferenceName = remoteTrackingReference.CanonicalName;
+                var branchName = remoteTrackingReferenceName.Substring(prefix.Length);
+                var localCanonicalName = "refs/heads/" + branchName;
 
                 if (repo.Refs.Any(x => x.CanonicalName == localCanonicalName))
                 {
@@ -203,6 +205,9 @@ namespace GitVersion
                 {
                     repo.Refs.Add(localCanonicalName, new ObjectId(symbolicReference.ResolveToDirectReference().TargetIdentifier), true);
                 }
+
+                var branch = repo.Branches[branchName];
+                repo.Branches.Update(branch, b => b.TrackedBranch = remoteTrackingReferenceName);
             }
         }
 
--- a/GitVersionExe.Tests/GitPreparerTests.cs
+++ b/GitVersionExe.Tests/GitPreparerTests.cs
@@ -208,5 +208,41 @@ public class GitPreparerTests
         }
     }
 
+    [Test]
+    public void UsingDynamicRepositoryWithFeatureBranchWorks()
+    {
+        var repoName = Guid.NewGuid().ToString();
+        var tempPath = Path.GetTempPath();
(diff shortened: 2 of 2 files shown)


gep13:
Is this still happening on the V3 beta branch, or is this only on v2?

eggapauli:
I tried this on master (97dade8), which is ahead of 3.0.0-Beta3

gep13:
Just to confirm, the 3.0.0 release hasn't been merged yet:

https://github.com/ParticularLabs/GitVersion/tree/release/3.0.0

Can you try on this?

eggapauli:
Sorry, I don't understand, running `git merge-base master release/3.0.0` outputs 24fa6b5 which is essentially release/3.0.0, so what do you mean by it hasn't been merged yet?

gep13:
To the best of my knowledge, although I hadn't checked, the beta code hasn't been merged to master, but I could be wrong.  @JakeGinnivan could confirm.

You could well be right, and I am just a little behind :smile_cat:

eggapauli:
Yeah, GitHub also says that release/3.0.0 is 0 commits ahead of master and 4 commits behind it :-)

JakeGinnivan:
Sorry for confusion, I am just using release branch for releasing betas. Am keeping master in sync/ahead

Sent from my Windows Phone

JakeGinnivan:
Ping @nulltoken, thoughts on the libgit2sharp question?

nulltoken:
> I dug a bit deeper and found that in GitHelper.CreateMissingLocalBranchesFromRemoteTrackingOnes the remoteTrackingReference is a DirectReference, but I think it should be a SymbolicReference for this to work.

@JakeGinnivan Thanks for the ping!

@eggapauli Could you please elaborate further regarding what makes you think that?

eggapauli:
I was wrong. I think the problem was that a local ref was added, but it wasn't set up to track the remote ref.
I tried to do this in the second commit. Again, I'm not familiar with libgit2sharp and found [your answer](http://stackoverflow.com/questions/22662543/create-a-local-branch-to-track-a-remote-branch-with-libgit2sharp) on how to update a branch to track a remote one.

GitVersionCore/BuildServers/GitHelper.cs
```diff
@@ -203,6 +205,9 @@ static void CreateMissingLocalBranchesFromRemoteTrackingOnes(Repository repo, st
                 {
                     repo.Refs.Add(localCanonicalName, new ObjectId(symbolicReference.ResolveToDirectReference().TargetIdentifier), true);
                 }
+
+                var branch = repo.Branches[branchName];
+                repo.Branches.Update(branch, b => b.TrackedBranch = remoteTrackingReferenceName);
```
nulltoken: Would GV be run on a development environment, this would blindly overwrite the potentially existing tracking configuration for every considered branch. Is this expected?

/cc @JakeGinnivan

eggapauli: I think these statements are only executed for newly created branches, otherwise they are skipped due to the if statement in [line 192](#discussion-diff-32599197R192).

JakeGinnivan: Correct, this normalisation code should only ever happen on the build server.

JakeGinnivan: Even without that if statement which was highlighted. Are there any other scenarios this could happen?

JakeGinnivan: (other than normalisation)

PR #468, opened 2015-06-16
Use authentication in `GitPreparer.GetRemoteReference`

eggapauli:
You already had a helper for this (`GitHelper.GetRemoteTipsUsingUsernamePasswordCredentials`), but it was internal. I made it public.

I saw somewhere else that you used `repository.Network.Remotes.Single()`, which fails when there are multiple remotes, so I'm not sure if it's ok that I also used it.

### DIFF
--- a/GitVersionCore/BuildServers/GitHelper.cs
+++ b/GitVersionCore/BuildServers/GitHelper.cs
@@ -1,5 +1,6 @@
 namespace GitVersion
 {
+    using System;
     using LibGit2Sharp;
     using System.Collections.Generic;
     using System.Linq;
@@ -162,6 +163,20 @@ namespace GitVersion
 
             Logger.WriteInfo(string.Format("Checking local branch '{0}' out.", fakeBranchName));
             repo.Checkout(fakeBranchName);
+        }
+
+        internal static IEnumerable<DirectReference> GetRemoteTipsUsingUsernamePasswordCredentials(Repository repo, string repoUrl, string username, string password)
+        {
+            // This is a work-around as long as https://github.com/libgit2/libgit2sharp/issues/1099 is not fixed
+            var remote = repo.Network.Remotes.Add(Guid.NewGuid().ToString(), repoUrl);
+            try
+            {
+                return GetRemoteTipsUsingUsernamePasswordCredentials(repo, remote, username, password);
+            }
+            finally
+            {
+                repo.Network.Remotes.Remove(remote.Name);
+            }
         }
 
         static IEnumerable<DirectReference> GetRemoteTipsUsingUsernamePasswordCredentials(Repository repo, Remote remote, string username, string password)
--- a/GitVersionCore/GitPreparer.cs
+++ b/GitVersionCore/GitPreparer.cs
@@ -172,7 +172,7 @@ namespace GitVersion
 
                 if (newHead == null)
                 {
-                    var remoteReference = GetRemoteReference(repository, targetBranch, repositoryUrl);
+                    var remoteReference = GetRemoteReference(repository, targetBranch, repositoryUrl, authentication);
                     if (remoteReference != null)
                     {
                         repository.Network.Fetch(repositoryUrl, new[]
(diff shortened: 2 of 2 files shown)


JakeGinnivan:
@nulltoken can you have a quick look over this one?

GitVersionCore/BuildServers/GitHelper.cs
```diff
@@ -164,7 +164,7 @@ static void CreateFakeBranchPointingAtThePullRequestTip(Repository repo, Authent
             repo.Checkout(fakeBranchName);
         }
 
-        static IEnumerable<DirectReference> GetRemoteTipsUsingUsernamePasswordCredentials(Repository repo, Remote remote, string username, string password)
+        public static IEnumerable<DirectReference> GetRemoteTipsUsingUsernamePasswordCredentials(Repository repo, Remote remote, string username, string password)
```
nulltoken: Why public rather than `internal`?

GitVersionCore/GitPreparer.cs
```diff
@@ -202,10 +202,11 @@ private static Reference GetLocalReference(Repository repository, string branchN
             return repository.Refs.FirstOrDefault(localRef => string.Equals(localRef.CanonicalName, targetBranchName));
         }
 
-        private static DirectReference GetRemoteReference(Repository repository, string branchName, string repositoryUrl)
+        private static DirectReference GetRemoteReference(Repository repository, string branchName, Authentication authentication)
         {
             var targetBranchName = branchName.GetCanonicalBranchName();
-            var remoteReferences = repository.Network.ListReferences(repositoryUrl);
+            var remote = repository.Network.Remotes.Single();
```
nulltoken: Will this method only be used on a CI server? Could it also be used on a developer environment? If that's the case, there are chances `Single()` will fail at runtime.

GitVersionCore/GitPreparer.cs
```diff
@@ -202,10 +202,11 @@ private static Reference GetLocalReference(Repository repository, string branchN
...
+            var remoteReferences = GitHelper.GetRemoteTipsUsingUsernamePasswordCredentials(repository, remote, authentication.Username, authentication.Password);
```
nulltoken: I just logged https://github.com/libgit2/libgit2sharp/issues/1099 in order to ease this kind of use case.
This shouldn't need a remote to work. One should be able to do this over an anonymous url.

GitVersionCore/GitPreparer.cs
nulltoken: Are you sure the remote you're after will always exist? If that's not the case, I'd go a safer way, creating a temporary remote (with a random name) and the provided url, list remotes refs through it, then drop this temp remote.

nulltoken:
> @nulltoken can you have a quick look over this one?

Sir! Done, Sir!

GitVersionCore/BuildServers/GitHelper.cs
eggapauli: Right, I overlooked that GitVersionCore internals are visible in GitVersionExe.

GitVersionCore/GitPreparer.cs
eggapauli: I think it depends on the `NoFetch` argument. If `NoFetch` isn't set (default) then `GitHelper.NormalizeGitDirectory` calls `GitHelper.EnsureOnlyOneRemoteIsDefined` which throws when the number of remotes != 1.

It's only called from within `GitPreparer.CreateDynamicRepository` which is normally only run on the build server but theoretically could run on a developer's machine.

OK, i will update the PR to create a temp remote, thanks for the advice.

eggapauli:
I now added an overload of `GitHelper.GetRemoteTipsUsingUsernamePasswordCredentials` that creates a temporary remote and calls the original `GetRemoteTipsUsingUsernamePasswordCredentials` with this remote. I used a Guid for the remote name. Although it's not perfect, it should be ok most of the time.

nulltoken:
:cool: Looks :+1: to me.

JakeGinnivan:
Thanks for your help @nulltoken!

nulltoken:
@JakeGinnivan As always, feel free to ping me!

Issue #466, opened 2015-06-15
Merge commit message picks up message not related to GitVersion


Sacrelicious:
Using -beta0003, we have an old commit in our repo with the message something along the lines of `feature/fix-for-08.14-push` which leads to GitVersion updating the version to 8.14, which is obviously not what we want in this case.
I will be submitting a PR shortly to address this issue.

W/o looking at their repo I can't say for sure, but I think it's possible that this is the root cause of the issue reported in #443

Issue #443, opened 2015-05-18
GitVersionTask is generating an unexpectedly high version number


asbjornu:
As requested in shawnmclean/Mandrill-dotnet#98, I've tried to add GitVersionTask to make versioning in the project more stringent and predictable. Before adding GitVersionTask (version 3 beta 2), I created a `develop` branch and then a `feature/git-version` branch, which is the branch I've added it in. When I build that branch, the version number for some reason ends up as `4.0.1.0`. The full `GitVersionInformation` class is as following:

```csharp
[System.Runtime.CompilerServices.CompilerGenerated]
static class GitVersionInformation
{
    public static string Major = "4";
    public static string Minor = "0";
    public static string Patch = "1";
    public static string PreReleaseTag = "git-version.1";
    public static string PreReleaseTagWithDash = "-git-version.1";
    public static string BuildMetaData = "67";
    public static string FullBuildMetaData = "67.Branch.feature/git-version.Sha.5fdf5674d81fe68fa1c9496e0a128b8e1f6ed15b";
    public static string MajorMinorPatch = "4.0.1";
    public static string SemVer = "4.0.1-git-version.1";
    public static string LegacySemVer = "4.0.1-git-version1";
    public static string LegacySemVerPadded = "4.0.1-git-version0001";
    public static string AssemblySemVer = "4.0.1.0";
    public static string FullSemVer = "4.0.1-git-version.1+67";
    public static string InformationalVersion = "4.0.1-git-version.1+67.Branch.feature/git-version.Sha.5fdf5674d81fe68fa1c9496e0a128b8e1f6ed15b";
    public static string BranchName = "feature/git-version";
    public static string Sha = "5fdf5674d81fe68fa1c9496e0a128b8e1f6ed15b";
    public static string NuGetVersionV2 = "4.0.1-git-version0001";
    public static string NuGetVersion = "4.0.1-git-version0001";

}
```

The last tag on the `master` branch is `v2.0.1`. I'd expect the version of a feature branch to be `2.1.*` or something similar and at most `3.*`, so I wonder where those extra two major versions come from. Ideas?

JakeGinnivan:
Could you post the logs for this, it should tell you where the high version is coming from

asbjornu:
@JakeGinnivan How do I get verbose logs with GitVersionTask when building in Visual Studio?

JakeGinnivan:
Unsure, @SimonCropp how can we get info's showing?

[...]

PR #470, opened 2015-06-16
Enable config based ignoring of commits/messages that incorrectly inc…

Sacrelicious:
…rement version

The latest changes to the MergeMessage branching strategy, while very
helpful, have pulled in very old messages from our repo that are
formatted in a way that cause the version number to be unexpectedly
high.
This changes allows configuration of commit SHAs to ignore, or strings
to ignore inside merge messages, so you can "fix" old commits in your
repo that you don't want to effect the version generation.
This is related to #466 and a possible fix for #443

### DIFF
--- a/GitVersionCore/VersionCalculation/BaseVersionCalculators/MergeMessageBaseVersionStrategy.cs
+++ b/GitVersionCore/VersionCalculation/BaseVersionCalculators/MergeMessageBaseVersionStrategy.cs
@@ -1,6 +1,7 @@
 ﻿namespace GitVersion.VersionCalculation.BaseVersionCalculators
 {
     using System.Linq;
+    using System.Security.Cryptography.X509Certificates;
     using System.Text.RegularExpressions;
     using LibGit2Sharp;
 
@@ -38,6 +39,16 @@
         private static SemanticVersion Inner(Commit mergeCommit, EffectiveConfiguration configuration)
         {
             if (mergeCommit.Parents.Count() < 2)
+            {
+                return null;
+            }
+
+            if (configuration.CommitsToIgnore != null && configuration.CommitsToIgnore.Contains(mergeCommit.Sha))
+            {
+                return null;
+            }
+
+            if (configuration.MergeMessagesToIgnore != null && configuration.MergeMessagesToIgnore.Any(x => mergeCommit.Message.Contains(x)))
             {
                 return null;
             }
--- a/GitVersionCore/VersionCalculation/BaseVersionCalculators/MergeMessageBaseVersionStrategy.cs
+++ b/GitVersionCore/VersionCalculation/BaseVersionCalculators/MergeMessageBaseVersionStrategy.cs
@@ -1,7 +1,6 @@
 ﻿namespace GitVersion.VersionCalculation.BaseVersionCalculators
 {
     using System.Linq;
-    using System.Security.Cryptography.X509Certificates;
     using System.Text.RegularExpressions;
     using LibGit2Sharp;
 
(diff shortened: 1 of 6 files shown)


GitVersionCore/VersionCalculation/BaseVersionCalculators/MergeMessageBaseVersionStrategy.cs
```diff
@@ -42,6 +42,16 @@ private static SemanticVersion Inner(Commit mergeCommit, EffectiveConfiguration
                 return null;
             }
 
+            if (configuration.CommitsToIgnore != null && configuration.CommitsToIgnore.Contains(mergeCommit.Sha))
```
JakeGinnivan: Wondering if we should do something a little more generic, what about bad tags or other things? Thoughts?

Sacrelicious: My thought on tags specifically is that it's pretty easy to delete/fix a bad tag, but fixing a bad commit/message from way back in the repo is nigh on impossible.

As to the larger point, how do you think that would be accomplished? Just have a `ThingsToIgnore` collection and just do 
```
if(mergeCommit.Sha.In(ThingsToIgnore) || mergeCommit.Tag.In(ThingsToIgnore) || ... etc. etc.)
````
?

I'm not super familiar w/ how the `Commit` object works, so there might be a better way than that.

JakeGinnivan: That's true. Leave me with this, want to think it through before I put it in the config (cause then we are stuck with it).

I will merge this before the next release.

Sacrelicious: Sounds good!

JakeGinnivan:
I am wondering if we should add something to the `gitversion init` helpers. Something along the lines of:

 - Solve GitVersion generating the wrong version? -> 
 - Ignore a particular commit or merge message:
   - List the discovered base merge messages.
 - Choose to ignore one, it adds the config

Recon it's worth it?

JakeGinnivan:
I can help out with the `init` stuff if you think it would be worth it

JakeGinnivan:
I have been thinking about this. I think we should just have `commits-to-ignore` in the config. 

Then in https://github.com/GitTools/GitVersion/blob/release/3.0.0/GitVersionCore/VersionCalculation/BaseVersionCalculator.cs#L24 add another `if` which checks to see if the `Source` on that base version matches any sha's in that config.

Sacrelicious:
I'm not opposed to that, although I do have 2 caveats:
1) if the string that's incrementing the version incorrectly is part of a branch name, it could be in a number of merge messages (merged into a feature branch, then merged into develop, then into release, then into master) resulting in a bunch of commits that would need to be excluded to prevent 1 bad version name from being used.
2) (Unless I'm missing it) There would need to be more details available in logs/console about the commit the final version number is derived from so the user will know which commit(s) to exclude.

Sacrelicious:
And I'm definitely all for doing something w/ the `init` stuff.

JakeGinnivan:
Beta 4 has much better logs so it lists everything you would need.

Also the init configuration could have an option to exclude commits which are causing incorrect version bumps, one option could be to exclude commits with commit message. So config is simple and init has the options?

Sent from my Windows Phone

mauroservienti:
I just faced this while trying to add GitVersion to an existing repo with a `legacy` history. What happened is that existing TAGs were causing `GitVersion` to generate a wrong `SemVer` version.
In my scenario, an optimal setting would be something like 'IgnoreAnythingBeforeSHA` XYZ.

JakeGinnivan:
Ignore before a time would be better than a sha, otherwise it needs another scan of the repo to find all sha's which are before a specific sha. At least specific shas are a simple equality check.

gep13:
:+1: for ignore before a time.

mauroservienti:
Time works for me as well

asbjornu:
Some sort of `ignore-before-*" mechanism should solve the problem I've explained in #443, so getting something like that in v3 before RTM would be great!

PR #474, opened 2015-06-18
Replace or update assembly info in files

### LINKED ISSUE #473: AssemblyInfo files not compatible using both CommandLine and Task
Using the GitVersionTask requires that there not be any conflicting attributes in the appropriate AssemblyInfo.cs in order to correctly update the assembly's version info, however, when using the command line tool, the AssemblyInfo.cs files won't get updated unless those attributes exists (b/c it uses a regex to find that attribute). This means that you cannot use both the Task and command line on the same projects, for instance, we would like to use the Task when building projects on dev machines, but use the command line when running on a build server.

Pull request for this issue coming shortly.

Sacrelicious:
I realize that the implementation of this is somewhat repetitious, but I
couldn't figure out a good way to implement this that wasn't also super
complicated.

Fixes #473

### DIFF
--- a/GitVersionExe/AssemblyInfoFileUpdate.cs
+++ b/GitVersionExe/AssemblyInfoFileUpdate.cs
@@ -2,7 +2,10 @@ namespace GitVersion
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.IO;
+    using System.Linq;
+    using System.Text.RegularExpressions;
     using GitVersion.Helpers;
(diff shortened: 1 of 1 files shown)


JakeGinnivan:
It would be good if we could factor this logic into core. Having two implementations is not great.

Ideas on how we could do that?

JakeGinnivan:
Also, getting some more tests around these changes would be good. Are you able to look at adding a test or two?

Issue #465, opened 2015-06-15
Hotfix resets PreReleaseTag.Number to 1


mjamin:
From the gitflow examples: 

![](http://i.imgur.com/LSXkGOf.png)

`develop` might've been at `1.3.0-unstable.2` before the hotfix, but as soon as the hotfix is merged, GitVersion resets the number to `1.3.0-unstable.1`.

We build NuGet packages off of our develop branch (e.g. `1.3.0-unstable0002`) and if the counter resets, people won't get any updates.

Edit: using ContinuousDeployment on develop

JakeGinnivan:
Would you have time to look into this, or at least create a failing test? I am short on time at the moment (moving back to Australia in two weeks).

mjamin:
Behaviour with the default config of the latest GitVersion:

1. `master` is at `1.2.0`.
2. I continue to work on `develop` and add a commit. The version is now `1.3.0-unstable.1`.
3. Now I create a hotfix based on `1.2.0` and merge it back to both branches. `master` is now at `1.2.1`, but `develop` is at `1.3.0-unstable.0`.

This is because the commits are counted since the hotfix merge commit due to the `MergeMessageBaseVersionStrategy` (commit message is `Merge branch 'hotfix/1.2.1' into develop`).

If I disable that merge message stragety it works as desired in my case, because now `HighestTagBaseVersionStrategy` delivers the version and since we only tag on `master` all commits on `develop` since the last release are counted.

But I have no idea what to do with that information :)

Should BaseVersionStrategies be configureable in some way? Should I avoid using a version in my hotfix branch name? Does the merge message strategy need to be adjusted? Are is the `MetaDataCalculator` simply not clever enough?

mjamin:
I just discovered the `HotfixBranchScenarios` Fixture and the `PatchLatestReleaseExample` seems to be the exact case I am describing. It also asserts that the version of `develop` is `1.3.0-unstable.0` after merging the hotfix, although it was `1.3.0-unstable.1` before. So this seems to be by design.

But this leads to multiple different releases of e.g. `1.3.0-unstable.1`.

JakeGinnivan:
Ok, so I have partially fixed the problem, now am not sure where to go (I can sort the implementation, but I just cannot figure out the rules..). 

Lets take this commit graph:

```
      Test Commit for file 'd423d32f-d2a8-4dfd-a945-094d72c74f91'
*   commit 1141e4e (HEAD, develop)
[...]

PR #478, opened 2015-06-20
Feature/dont reset count when version will be the same

JakeGinnivan:


JakeGinnivan:
@mjamin @GeertvanHorrik could either of you have a bit of a look through and see if the changes to commit counting in tests does indeed solve #465 as you would expect?

mjamin:
Looks good!

JakeGinnivan:
Might try and release beta 4 today.

PR #479, opened 2015-06-21
Init improvements

JakeGinnivan:
Looks like this:

```
$ gitversion init
GitVersion init will guide you through setting GitVersion up to work for you

Which would you like to change?

0) Save changes and exit
1) Exit without saving

2) Run getting started wizard

3) Branch specific configuration
4) Branch Increment mode (per commit/after tag) (Current: ContinuousDelivery)
5) Assembly versioning scheme (Current: MajorMinorPatch)
6) Setup build scripts

> 6

What build server are you using?

Want to see more? Contribute a pull request!

0) Back
1) AppVeyor

> 1

GitVersion doesn't support modifying existing appveyor config files. We will generate appveyor.gitversion.yml instead

What sort of config template would you like generated?

0) Back
1) Generate basic (gitversion + msbuild) configuration
2) Generate with NuGet package publish

> 2
INFO [06/21/15 12:40:07:17] AppVeyor sample config file written to D:\_code\shouldly\appveyor.gitversion.yml
```


JakeGinnivan:
// @gep13 what do you think?

GitVersionCore/Configuration/Init/BuildServer/AppVeyorSetup.cs
```diff
@@ -0,0 +1,122 @@
...
+                    steps.Enqueue(new EditConfigStep());
+                    return StepResult.Ok();
+            }
+            return StepResult.InvalidResponseSelected();
+        }
+
+        void GenerateBasicConfig(string workingDirectory, IFileSystem fileSystem)
+        {
+            WriteConfig(workingDirectory, fileSystem, @"install:
+  - choco install gitversion.portable -pre
```
gep13: Just to check... Are you aware that on the latest version of Chocolatey, this is no longer a silent operation, the user will be prompted whether they want to continue with the installation or not.  As a result, we might want to include the ```-y``` parameter, so automatically proceed with the installation.

GitVersionCore/Configuration/Init/BuildServer/AppVeyorSetup.cs
```diff
@@ -0,0 +1,122 @@
...
+  - ps: gitversion /l console /output buildserver /updateAssemblyInfo
+
+build:
+  project: <your sln file>");
+        }
+
+        void GenerateNuGetConfig(string workingDirectory, IFileSystem fileSystem)
+        {
+            WriteConfig(workingDirectory, fileSystem, @"install:
+  - choco install gitversion.portable -pre -y
```
gep13: Ah, I notice you have ```-y``` here, any reason you didn't do it above?

GitVersionCore/Configuration/Init/BuildServer/AppVeyorSetup.cs
```diff
@@ -0,0 +1,122 @@
...
+            var prompt = new StringBuilder();
+            if (AppVeyorConfigExists(workingDirectory, fileSystem))
+            {
+                prompt.AppendLine("GitVersion doesn't support modifying existing appveyor config files. We will generate appveyor.gitversion.yml instead");
+                prompt.AppendLine();
+            }
+
+            prompt.Append(@"What sort of config template would you like generated?
+
+0) Back
```
gep13: This is a general comment, but I looked at this option for a while before realising that it means "Go back".  I think the addition of the "Go" would help a lot.  Or perhaps I am just slow :smile_cat:

gep13:
Just a couple comments, but overall, looks really good to me! :smile:

PR #482, opened 2015-06-25
Update ArgumentParser to use variables from VersionVariables

arledesma:
I wasn't seeing any new progress in #447.

This changes VersionVariables.AvailableVariables to static and bubbles up WarningException.Message to console output

I went ahead and kept /v for backwards compatibility and added the documented /showvariable


JakeGinnivan:
Nice! Thanks for the contribution

Issue #137, opened 2014-03-30
Add version bumping strategy 'commit message'


JakeGinnivan:
I would like to bump the major or minor by having certain keywords in my commit message..

It would be done via two configuration properties, both accepting a RegEx value with the keys and their default values being:

``` yaml
major-version-bump-message: (\+semver:.?breaking|breaking:)
minor-version-bump-message: (\+semver:.?(feature|minor)|(feature|minor):)
```

The match should be case insensitive. 

This will match `Api change doing blah +semver: breaking` and `breaking: Api change doing blah`  
And minor can be `new feature description +semver: feature` or `feature: new feature description`

andreasohlund:
I like the `+gv: breaking` , pretty similar to: https://github.com/boennemann/semantic-release#how-does-this-work

That said since we're actually talking about a semver breaking change should we consider: `semver: breaking change` to be less gitversion specific?

pvandervelde:
+1 for the +semver:breaking option. Maybe the other option should then also be +semver:feature?

JakeGinnivan:
Why not make this configurable.

major-version-bump-message: <regex>

:)

andreasohlund:
Yea, `+semver: feature` sounds good.

>Why not make this configurable.

at least as the default option :)

JakeGinnivan:
So what about

``` yaml
major-version-bump-message: \+semver: breaking
minor-version-bump-message: \+semver:(feature|minor)
```

andreasohlund:
+1

JakeGinnivan:
Updated issue description. It is up for grabs so anyone can grab it :)

robdmoore:
Great idea!

roryprimrose:
Are there any special rules around using this? I've tried an empty commit against develop (local), a commit with a change to the readme.md in a feature branch (also local) and I still see the existing versioning scheme. I've tried with commit messages of

+semver: minor

and then tried

Version bump to 2.4
+semver: minor

Still no joy. Are there special conditions that cause this comment to be actioned?

roryprimrose:
Hmmm, I think this is my lack of understanding. I'm trying to bump minor on a develop branch after I have created a release-2.3 branch which is going through a release candidate phase. 

[...]

PR #489, opened 2015-07-10
Add support for indicating change severity in commit messages

thogil:
Adds two new increment strategies: `CommitMessage` and `MergeMessage`.

While active, the version increment is determined by reading commit/merge commit messages for markers describing the change severity. All new commits introduced since the base version are scanned.

For example, "+semver: breaking" will force a major version bump.

### New Branch Configuration Options

* `major-version-bump-message`: a regex used for parsing breaking changes in commit messages. Defaults to `\+semver:\s?(breaking|major)`.
* `minor-version-bump-message`: a regex used for parsing minor changes in commit messages. Defaults to `\+semver:\s?(feature|minor)`.
* `message-fallback`: the increment strategy to fall back into if no change severity indicators are found in commit messages, while using the `CommitMessage` or `MergeMessage` strategies. Defaults to `Patch`.

### Potential Issues
* I changed the default increment strategy for the `master` branch to `CommitMessage`. This will behave the same as it did prior to this change if the user does not have markers in this commit messages. However this could certainly be considered a breaking change.
* This change does not work for new repositories, as the base version of a new repository will be determined by the `FallbackBaseVersionStrategy`, which never increments versions. Certainly, If I am using GitHub Flow, this is not what I would expect. Changing the behaviour of `FallbackBaseVersionStrategy` when using commit message increments would be an option, but I do not know if this would make sense for other work flows.
* Breaking changes always increment the major version. This may not be desirable for < 1.0.0 builds. We could add an option to treat breaking changes as minor version bumps for 0.x.x builds.

Addresses #137.

### DIFF
--- a/src/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
+++ b/src/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
@@ -2,6 +2,10 @@
 mode: ContinuousDelivery
 tag-prefix: '[vV]'
 continuous-delivery-fallback-tag: ci
+major-version-bump-message: '\+semver:\s?(breaking|major)'
+minor-version-bump-message: '\+semver:\s?(feature|minor)'
+patch-version-bump-message: '\+semver:\s?(fix|patch)'
+commit-message-incrementing: Enabled
 branches:
   master:
     mode: ContinuousDelivery
--- a/src/GitVersionCore/IncrementStrategyFinder.cs
+++ b/src/GitVersionCore/IncrementStrategyFinder.cs
@@ -1,0 +1,107 @@
+﻿namespace GitVersion
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text.RegularExpressions;
+    using VersionCalculation.BaseVersionCalculators;
+    using LibGit2Sharp;
+
+    public enum CommitMessageIncrementMode
+    {
+        Enabled,
+        Disabled,
+        MergeMessageOnly
+    }
+
+    public static class IncrementStrategyFinder
+    {
+        public const string DefaultMajorPattern = @"\+semver:\s?(breaking|major)";
+        public const string DefaultMinorPattern = @"\+semver:\s?(feature|minor)";
+        public const string DefaultPatchPattern = @"\+semver:\s?(fix|patch)";
+
+        public static VersionField? DetermineIncrementedField(GitVersionContext context, BaseVersion baseVersion)
+        {
+            var commitMessageIncrement = FindCommitMessageIncrement(context, baseVersion);
+            var defaultIncrement = context.Configuration.Increment.ToVersionField();
+
+            // use the default branch config increment strategy if there are no commit message overrides
+            if (commitMessageIncrement == null)
+            {
+                return baseVersion.ShouldIncrement ? defaultIncrement : (VersionField?)null;
+            }
+
+            // cap the commit message severity to minor for alpha versions
+            if (baseVersion.SemanticVersion < new SemanticVersion(1) && commitMessageIncrement > VersionField.Minor)
+            {
+                commitMessageIncrement = VersionField.Minor;
+            }
+
+            // don't increment for less than the branch config increment, if the absense of commit messages would have
+            // still resulted in an increment of configuration.Increment
+            if (baseVersion.ShouldIncrement && commitMessageIncrement < defaultIncrement)
+            {
+                return defaultIncrement;
+            }
+
+            return commitMessageIncrement;
+        }
+
+        private static VersionField? FindCommitMessageIncrement(GitVersionContext context, BaseVersion baseVersion)
+        {
+            if (context.Configuration.CommitMessageIncrementing == CommitMessageIncrementMode.Disabled)
+            {
+                return null;
+            }
+            
+            var commits = GetIntermediateCommits(context.Repository, baseVersion.BaseVersionSource, context.CurrentCommit);
+
+            if (context.Configuration.CommitMessageIncrementing == CommitMessageIncrementMode.MergeMessageOnly)
+            {
+                commits = commits.Where(c => c.Parents.Count() > 1);
+            }
+
+            var majorRegex = CreateRegex(context.Configuration.MajorVersionBumpMessage ?? DefaultMajorPattern);
+            var minorRegex = CreateRegex(context.Configuration.MinorVersionBumpMessage ?? DefaultMinorPattern);
+            var patchRegex = CreateRegex(context.Configuration.PatchVersionBumpMessage ?? DefaultPatchPattern);
+
+            var increments = commits
+                .Select(c => FindIncrementFromMessage(c.Message, majorRegex, minorRegex, patchRegex))
+                .Where(v => v != null)
+                .Select(v => v.Value)
+                .ToList();
+
+            if (increments.Any())
+            {
+                return increments.Max();
+            }
+
+            return null;
+        }
+        
+        private static IEnumerable<Commit> GetIntermediateCommits(IRepository repo, Commit baseCommit, Commit headCommit)
+        {
+            var filter = new CommitFilter
+            {
+                Since = headCommit,
+                Until = baseCommit,
+                SortBy = CommitSortStrategies.Topological | CommitSortStrategies.Reverse
+            };
+
+            return repo.Commits.QueryBy(filter);
+        }
+
+        private static VersionField? FindIncrementFromMessage(string message, Regex major, Regex minor, Regex patch)
+        {
+            if (major.IsMatch(message)) return VersionField.Major;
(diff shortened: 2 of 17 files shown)


JakeGinnivan:
Nice, I will hold off on this until 3.0 has been released (which should be soon)

thogil:
I've rebased onto master after the 3.0.2 release. I would like some feedback on those potential issues before a merge, though.

JakeGinnivan:
Awesome, will look through it later today.

JakeGinnivan:
Sorry for the delay, just started back at work last week since moving back to Australia. 

Rather than changing config I think that maybe we change the increment strategy to have something like this (obv this code would not compile/work, but hopefully give you the idea)

```
public static VersionField DetermineIncrementedField(GitVersionContext context, BaseVersion baseVersion)
{
    var incrementFromCommitMessage = FindIncrementsFromMessagesSince(context, baseVersion.Source); //baseVersion.Source is the commit used for commit counting, it is a good from filter of the commits we need to scan
    if (incrementingCommitOrMergeMessages.Any()) {
        var max = incrementFromCommitMessage.Max();
        var configIncrement = SimpleIncrement(context.Configuration.Increment);
        return (max >  ? max : configIncrement);
    }
    return SimpleIncrement(context.Configuration.Increment);
}
```

This way we don;t have to update any config, it will just light up?

JakeGinnivan:
> This change does not work for new repositories, as the base version of a new repository will be determined by the FallbackBaseVersionStrategy, which never increments versions. Certainly, If I am using GitHub Flow, this is not what I would expect. Changing the behaviour of FallbackBaseVersionStrategy when using commit message increments would be an option, but I do not know if this would make sense for other work flows.

If we use my approach the commit messages override the increment from the config, solving this issue

JakeGinnivan:
Actually, I think my approach fixes all the issues listed under potential issues?

What do you think?

thogil:
You are essentially proposing that we always respect commit message severities, rather than have the user specify it via new increment strategies. I have no particular objection to this, but it does not fix all of the problems.

This code overrides the normal increment strategy when it finds relevant commit messages. However, the increment strategy is ignored entirely if the version strategy returns a `BaseVersion` with `ShouldIncrement` set to false. The `FallbackBaseVersionStrategy` always does this, and as such commit messages will never increment the version while that strategy is in effect (e.g. a clean repository with not yet any releases). Is this expected behaviour? As soon as the user does something which changes the version strategy (e.g. tags v0.2.0), commit messages will begin to affect the version number.

There is also still the issue of alpha (< 1.0.0) versions. There could easily be many commits tagged as  breaking changes, but many users would not expect their version to be bumped up to or beyond 1.0.0 automatically. We could maybe do with having an option to treat breaking changes as minor version increments while the base version is < 1.0.0.

JakeGinnivan:
> However, the increment strategy is ignored entirely if the version strategy returns a BaseVersion with ShouldIncrement set to false.

If we made this strategy always run before the normal increment code. If it finds an increment message then it increments the version, even if the setting branch says not to. 

 > There is also still the issue of alpha (< 1.0.0) versions

Agreed, if the version we are incrementing is < 1.0 we should just bump minor instead.

JakeGinnivan:
What do you think @thogil?

thogil:
Sounds reasonable. I'll have a look at updating the PR later today.

thogil:
The code now looks for major, minor and patch commit messages. If it finds one, it will use that increment, regardless of the base version `ShouldIncrement` property.

There are branch configuration values for the three message regexes, which default to:
1. `\+semver:\s?(breaking|major)` e.g. `+semver: breaking`
1. `\+semver:\s?(feature|minor)` e.g. `+semver: feature`
1. `\+semver:\s?(fix|patch)` e.g. `+semver: fix`

There is also a `commit-message-version-bump-mode` setting, which controls which messages are read. It defaults to `All`, but can also be `MergeMessageOnly` or `None` (disable this feature).

If no commit message increments are found, it falls back to the default behaviour. If the version would have been incremented by `branchConfig.Increment`, had there been no commit messages, then it will always increment by at least `branchConfig.Increment`. It is also capped at minor for versions < 1.0.0.

thogil:
@JakeGinnivan I don't know why the AppVeyor build is failing. The only thing which looks like an error is coming from Fody, but that error also exists in the (passing) master build. The build just appears to halt silently.

JakeGinnivan:
I'll check it out tomorrow

Sent from my Windows Phone

src/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
```diff
@@ -9,18 +9,30 @@ branches:
     increment: Patch
     prevent-increment-of-merged-branch-version: true
     track-merge-target: false
+    major-version-bump-message: '\+semver:\s?(breaking|major)'
```
JakeGinnivan: This config shouldn't be per branch I don't think? Can it be pulled up to root?

src/GitVersionCore/IncrementStrategyFinder.cs
```diff
@@ -0,0 +1,107 @@
+﻿namespace GitVersion
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text.RegularExpressions;
+    using VersionCalculation.BaseVersionCalculators;
+    using LibGit2Sharp;
+
+    public enum CommitMessageVersionMode
```
JakeGinnivan: Could you explain the purpose of this class? Don't get why we need it

src/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
thogil: Yep, sure.

src/GitVersionCore/IncrementStrategyFinder.cs
thogil: It allows the user to either disable the commit message parsing, or to limit it to only look at merge messages. Limiting to merge messages may be useful for repo maintainers who want to specify the severity themselves when merging a pull request.

thogil: Although the PR submitter could still include merges in their own branch, so it is not a very hard restriction.

JakeGinnivan: Ok I get it.

Maybe name the setting `commit-message-incrementing: [enabled|disabled|merge-message-only]`

The enum can them become `CommitMessageIncrementing`

Thoughts on naming?

thogil: Is there a way to get the enum values to format like that in the yaml? `merge-message-only` instead of `MergeMessageOnly`?

JakeGinnivan: I think you will need to attribute up the Enum members with YmlMember() attributes. Let me try and find some docs

JakeGinnivan: Give `[YamlMember('merge-message-only')]` a go.. If not, unsure..

SimonCropp:
try updating to  https://www.nuget.org/packages/Visualize.Fody/0.4.4

src/GitVersionCore/IncrementStrategyFinder.cs
thogil: There does not appear to be a way to do it in YamlDotNet. `YamlMember` cannot be applied to enum members, and we are already using the hyphenate convention.

JakeGinnivan: Ok, no worries then. MergeMessageOnly it is

src/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
```diff
@@ -9,18 +12,21 @@ branches:
     increment: Patch
     prevent-increment-of-merged-branch-version: true
     track-merge-target: false
+    commit-message-incrementing: Enabled
```
JakeGinnivan: Do we need this to be configurable per branch?

The default should be set at the root level, you can leave it configurable per branch but it should be an override to whatever is specified in the root

JakeGinnivan:
Awesome!

One more thing, you are going to have to rewrite history to remove `For example, "+semver: breaking" will force a major version bump.` from your commit message :P we are currently building v4. 

So it works :P

thogil:
Oh, good point ;)

JakeGinnivan:
I am heading out now. I have just put up a PR myself. This will be in 3.1 which should be released tomorrow/saturday

JakeGinnivan:
Thanks for this, its an awesome addition!

JakeGinnivan:
Really should have added docs for this.. didn't think about that at the time, now I am doing the release thinking we should have.. I might put initial docs in then move the tag :P 

After I have put the initial docs in if you want to review/add anything feel welcome to

Issue #495, opened 2015-07-13
SemanticVersion.TryParse() vs. TagPrefix


Tungsten78:
First off, great work on this tool!  You guys have put together something great.

I've been trying to integrate GitVersion (v3.0.0-beta.4) into our build process and have been running into issues with respect to the tag strategies.  Our build server tags our repository with a build tag of the form "build/product/branch/1.2.3".  This allows our team to easily track commits back to the most recent build.

The issue stems from GitVersion using these tags within TrackMergeTargetBaseVersionStrategy and TaggedCommitVersionStrategy.   Given the default prefix of [vV] we expected that our tags would be ignored.

I was looking at the commit that added TagPrefix (https://github.com/JakeGinnivan/GitVersion/commit/31d42088106e6ca9100260c74e18b8bfb0022f65) and it seems from the unit tests that the intent was to be more flexible when parsing tags (instead of only allowing an optional v).  What's odd is that regardless of what we use for TagPrefix the result is always successful.

Consider (tagprefix changed): 
```c#
    [Test]
    public void CanSpecifyTagPrefixes()
    {
        using (var fixture = new EmptyRepositoryFixture(new Config { TagPrefix = "SomethingOfftheWall" }))
        {
            const string TaggedVersion = "version-1.0.3";
            fixture.Repository.MakeATaggedCommit(TaggedVersion);
            fixture.Repository.MakeCommits(5);
            fixture.AssertFullSemver("1.0.4+5");
        }
    }
```
Here's a failed test to illustrate what I was expecting:
```c#
[Test]
    public void AreTagsNotAdheringToTagPrefixIgnored()
    {
        using (var fixture = new EmptyRepositoryFixture(new Config { TagPrefix = "" }))
        {
            string TaggedVersion = "version-1.0.3";
            fixture.Repository.MakeATaggedCommit(TaggedVersion);
            fixture.Repository.MakeCommits(5);

            fixture.AssertFullSemver("0.1.0+5");    //Fallback version + 5 commits since tag

            TaggedVersion = "bad/1.0.3";
            fixture.Repository.MakeATaggedCommit(TaggedVersion);

            fixture.AssertFullSemver("0.1.0+6");   //Fallback version + 6 commits since tag
        }
    }
```

[...]

PR #496, opened 2015-07-14
Address tag parsing

JakeGinnivan:
Opening PR to discuss changes highlighted in #495

### DIFF
--- a/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
+++ b/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
@@ -1,6 +1,6 @@
 ﻿assembly-versioning-scheme: MajorMinorPatch
 mode: ContinuousDelivery
-tag-prefix: 
+tag-prefix: '[vV]|'
 continuous-delivery-fallback-tag: ci
 branches:
   master:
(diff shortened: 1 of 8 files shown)


JakeGinnivan:
There is a small issue (which is causing the test failure) that a tag without the prefix will fail to pass.

Changing the regex from `var match = Regex.Match(version, string.Format("^({0})(?<version>.*)$", tagPrefixRegex));` to `var match = Regex.Match(version, string.Format("^({0})?(?<version>.*)$", tagPrefixRegex));` (making the prefix optional) fixed the issue.

Thanks gain for the changes

JakeGinnivan:
Merged into master via 50833b4900ad653ddabfade567d58c52ef669110

Tungsten78:
@JakeGinnivan, I committed another pass to better illustrate what I was trying to accomplish. The intent was to have the TagPrefix denote whether to treat as optional.  This being the default should handle most legacy usages with the exception of those already carrying around the old default of "[vV]" in their GitVersionConfig.yaml.   Technically, this would be a breaking change to those people.

I apologize for having left loose ends with the failed test.  I'm only reiterating because I feel the partial inclusion of the "[vV]|" is leaving the code base a little off (since the ? within the regex makes this redundant).

JakeGinnivan:
I get you. The only issue is if someone has run `git init` it dumps the default config into the config file (I probably should fix that).

Do you see an issue with leaving the default as `[vV]` and making it optional in the parser. The inclusion of the `^$` should fix the issues you describe?

Tungsten78:
Yes as merged it corrects my issue. It is only the remnants of what I was trying to accomplish that have left a bad taste in my mouth. I agree with the backwards compatibility issue. I will rework my pull tomorrow as a cleanup only. Thanks.

JakeGinnivan:
Thanks, sorry I missed part of what you were doing

Tungsten78:
@JakeGinnivan Here is the cleanup I promised.

JakeGinnivan:
@Tungsten78 could you submit a PR for this?

PR #502, opened 2015-07-15
Docs update

JakeGinnivan:
Ported wiki docs and wrote a bunch of new docs


JakeGinnivan:
@gep13 @robdmoore thoughts on the docs updates.

I like having them in the repo as people can PR. Later we can build a github.io page from that folder too. You can browse and have a read at https://github.com/JakeGinnivan/GitVersion/tree/DocsUpdate

gep13:
@JakeGinnivan so, does that mean we are not going to be using readme.io anymore?

gep13:
I have had a quick skim through this, but will need to dig into it deeper.  From what I read, it looks good though! :+1:

gep13:
@JakeGinnivan did you see my question above? Cheers!

JakeGinnivan:
Ah yeah. I think for the moment not using readme.io.

Having docs outside the repo makes it hard to keep docs and pr's in sync. I also like having atom open just editing locally, it is easier than editing in the browser.

JakeGinnivan:
My ideal scenario is having a deployment step which would push the docs into a new readme.io version. Then you take advantage of the versioned docs and can just build docs during a release. Then somehow having a process for updating the docs as well outside a release

gep13:
>@JakeGinnivan said...
My ideal scenario is having a deployment step which would push the docs into a new readme.io version. Then you take advantage of the versioned docs and can just build docs during a release. Then somehow having a process for updating the docs as well outside a release

Totally agree, this would be ideal!

JakeGinnivan:
Future Jake's problem :P

Issue #505, opened 2015-07-16
PR from code changes on same branch


RaphHaddad:
As a GitVersion user when a PR is sent to the base repository with code on the same branch GitVersion should not fail.

please see these builds:
[0.0.38](https://ci.appveyor.com/project/GitTools/gitversion/build/0.0.38)
[0.0.40](https://ci.appveyor.com/project/GitTools/gitversion/build/0.0.40)
[0.0.41](https://ci.appveyor.com/project/GitTools/gitversion/build/0.0.41)
[0.0.42](https://ci.appveyor.com/project/GitTools/gitversion/build/0.0.42)

Also please see the discussion on issue https://github.com/GitTools/GitVersion/issues/378

JakeGinnivan:
Quick bit of hunting around suggests this may be due to libgit2sharp. See https://github.com/libgit2/libgit2sharp/pull/927

I will see if I can create a failing test

JakeGinnivan:
I can't figure a failing test.. Instead when an unexpected exception occurs I will dump the commit graph to the log (without commit messages). This will allow us to reproduce the graph in a test and hopefully fix the problem

See https://github.com/GitTools/GitVersion/pull/500/files#diff-cdf777fc03b2d4c69c5026eaafd6c2f8R103

JakeGinnivan:
@RaphHaddad if you get a chance could you try to submit a PR which fails like it did before. Then paste the commit log, then we can get a test in

RaphHaddad:
will do.

RaphHaddad:
I can' seem to get it to fail anymore lol.

JakeGinnivan:
Did you ever delete your fork? The other PR's say the remote repo is not available which is odd

RaphHaddad:
Yes. I deleted all my old forks (at the time I was frustrated on why the build wasn't working) :stuck_out_tongue:

JakeGinnivan:
Lets reopen if we can reproduce

PR #509, opened 2015-07-16
adding fake code change to test issue #505

RaphHaddad:


JakeGinnivan:
Oh well, if we see it again we should have more info

PR #510, opened 2015-07-16
TagPrefix cleanup

Tungsten78:
Cleaning up a few leftovers that resulted from a partial merge of pull request #496.  As before, TagPrefix is always considered optional.

### DIFF
--- a/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
+++ b/GitVersionCore.Tests/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
@@ -1,6 +1,6 @@
 ﻿assembly-versioning-scheme: MajorMinorPatch
 mode: ContinuousDelivery
-tag-prefix: '[vV]|'
+tag-prefix: '[vV]'
 continuous-delivery-fallback-tag: ci
 branches:
   master:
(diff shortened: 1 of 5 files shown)


JakeGinnivan:
@SimonCropp this PR seems to be causing a PEVerify error. https://ci.appveyor.com/project/GitTools/gitversion/build/0.0.60#L204

Any ideas? Wondering if it's the const?

Tungsten78:
Not sure what this error is:

MSBUILD : error : Fody: PEVerify of the assembly failed. [C:\projects\gitversion\GitVersionCore\GitVersionCore.csproj]

Tungsten78:
I'm seeing the error on my side with a solution rebuild.  If I remove the const it does indeed go away.

There's a bit more info in the error:
```
1>MSBUILD : error : [IL]: Error: [C:\Projects\GitVersion\GitVersion\GitVersionCore\bin\Release\GitVersionCore.dll : GitVersion.Config::DebuggerDisplay][offset 0x00000026] Unable to resolve token.(Error: 0x80131869)
```

@JakeGinnivan Would you rather me just rid the const and hardcode the "[vV]" as it was?

Tungsten78:
I updated the branch to use an ```internal const``` instead.  This gets us past the validation error and is still accessible to the test library because of the existing ```[assembly:InternalsVisibleTo("GitVersionCore.Tests")]```

PR #523, opened 2015-07-21
Fixes version calculation issue introduced when fixing #465

JakeGinnivan:
The logic for figuring out conflicting sources was wrong. It was close enough to look right in most scenarios but caused master to calculate with the wrong version after merging a release branch and tagging (did not increment to next patch release).


gep13:
wow! That AppVeyor build was wicked fast!

JakeGinnivan:
Yeah, since the VM upgrades they fly. It's awesome

gep13:
I knew there had been some improvements, had hadn't actually "witnessed" one.  Truly amazing!  @FeodorFitsner has a lot to be proud of!

PR #525, opened 2015-07-21
Added generated classes to their own namespace to avoid conflicts.

distantcam:
Fixes #448 by putting generated classes in a namespace named after the project file.

PR #526, opened 2015-07-21
Add GitVersionInformationAttribute to make reflection easier

asbjornu:
Right now, it's a bit hard to do reflection on an assembly to dig out all of the juicy GitVersion data, since the data is embedded within the static class `GitVersionInformation`. To get to this class, you'd have to do `Assembly.GetTypes()`, which is prone to throwing exceptions. It would be more reflection-friendly to have the information inside an `[assembly: Attribute]`.

So I've created a `GitVersionInformationAttribute` that should remedy this. I've not removed any of the previously generated data, but I've taken the liberty of removing extraneous whitespace in the generated `.cs` file. I've also expanded on the `AssemblyInfoBuilderTests.VerifyCreatedCode()` test to verify the contents of the generated assembly and assert whether or not the `GitVersionInformationAttribute` actually contains the information we want to or not.

### DIFF
--- a/src/GitVersionTask.Tests/AssemblyInfoBuilderTests.cs
+++ b/src/GitVersionTask.Tests/AssemblyInfoBuilderTests.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Linq;
+using System.Reflection;
 using System.Runtime.CompilerServices;
 using ApprovalTests;
 using GitVersion;
@@ -30,13 +31,21 @@ public class AssemblyInfoBuilderTests
         Approvals.Verify(assemblyInfoText);
 
         var compilation = CSharpCompilation.Create("Fake.dll")
-            .WithOptions(new CSharpCompilationOptions(OutputKind.NetModule))
+            .WithOptions(new CSharpCompilationOptions(OutputKind.DynamicallyLinkedLibrary))
             .AddReferences(MetadataReference.CreateFromFile(typeof(object).Assembly.Location))
             .AddSyntaxTrees(CSharpSyntaxTree.ParseText(assemblyInfoText));
-        
 
-        var emitResult = compilation.Emit(new MemoryStream());
-        Assert.IsTrue(emitResult.Success, string.Join(Environment.NewLine, emitResult.Diagnostics.Select(x => x.Descriptor)));
+
+        using (var stream = new MemoryStream())
+        {
+            var emitResult = compilation.Emit(stream);
+            
+            Assert.IsTrue(emitResult.Success, string.Join(Environment.NewLine, emitResult.Diagnostics.Select(x => x.Descriptor)));
+
+            stream.Seek(0, SeekOrigin.Begin);
+            var assembly = Assembly.Load(stream.ToArray());
+            VerifyGitVersionInformationAttribute(assembly, versionVariables);
+        }
     }
 
     [Test]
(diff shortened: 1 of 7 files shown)


src/GitVersionTask.Tests/AssemblyInfoBuilderTests.cs
```diff
@@ -30,13 +31,21 @@ public void VerifyCreatedCode()
...
-        
 
-        var emitResult = compilation.Emit(new MemoryStream());
-        Assert.IsTrue(emitResult.Success, string.Join(Environment.NewLine, emitResult.Diagnostics.Select(x => x.Descriptor)));
+
+        using (var stream = new MemoryStream())
+        {
+            var emitResult = compilation.Emit(stream);
+            
+            Assert.IsTrue(emitResult.Success, string.Join(Environment.NewLine, emitResult.Diagnostics.Select(x => x.Descriptor)));
```
JakeGinnivan: FYI shouldly would give much better error messages:

Current: `false should be true (some descriptor)`  
Shouldly: `emitResult.Success should be true. Additional Context: some descriptor`

src/GitVersionTask/AssemblyInfoBuilder/AssemblyInfoBuilder.cs
```diff
@@ -32,26 +38,56 @@ static class GitVersionInformation
...
+            .Trim();
 
         return assemblyInfo;
     }
 
-    string GenerateVariableMembers(IEnumerable<KeyValuePair<string, string>> vars)
+
+    static string GenerateStaticVariableMembers(IList<KeyValuePair<string, string>> vars)
     {
         var members = new StringBuilder();
```
JakeGinnivan: `return string.Join(Environment.NewLine, vars.Select(variable => string.Format("    public static string {0} = \"{1}\";", variable.Key, variable.Value)))`

One liner for this if you want

JakeGinnivan:
#525 puts the different classes into namespaces which prevents conflicts when using `InternalsVisibleTo`. I am going to merge that PR first, if you could rebase this and put this new attribute under the assemblies namespace

src/GitVersionTask/AssemblyInfoBuilder/AssemblyInfoBuilder.cs
```diff
@@ -1,11 +1,15 @@
...
+using System.Linq;
 using System.Text;
+
 using GitVersion;
 
 public class AssemblyInfoBuilder
 {
     public string GetAssemblyInfoText(VersionVariables vars)
     {
+        var v = vars.ToArray();
```
JakeGinnivan: FYI, `.ToList()` is actually more efficient because an IEnumerable length is not fixed the array buffer has to be expanded multiple times, lists have that functionality built in. Not that it really matters in most code but just a FYI :)

JakeGinnivan:
Nice addition, have rebased and pushed to #531 so it can be merged

asbjornu:
@JakeGinnivan Awesome! :smile:

src/GitVersionTask.Tests/AssemblyInfoBuilderTests.cs
asbjornu: Yea, it's not my assert. I thought I'd just leave as much code I didn't write untouched to focus the PR on one thing so it would be easier to review. :smile:

src/GitVersionTask/AssemblyInfoBuilder/AssemblyInfoBuilder.cs
asbjornu: Yea, it's just that I didn't want all that extraneous whitespace. :wink:

src/GitVersionTask/AssemblyInfoBuilder/AssemblyInfoBuilder.cs
asbjornu: Ah yes. I just like arrays better due to the built in syntactic sugar. Not that I used that here, but habits die hard. :)

Issue #121, opened 2014-03-24
Commented AssemblyVersion attributes cause confusing error


distantcam:
While converting a project to use GitVersion I commented out the `Assembly*Version` attributes but still received the following error when compiling.

> File contains assembly version attributes with conflict with the attributes generated by GitVersionProperties\AssemblyInfo.cs

Checking the source code GitVersion is still picking up the text, even though it's commented out, and reporting the error.

asbjornu:
As I mentioned in a comment to 41f84f3d89b1b70e495501c9707583fb60231263; the current implementation is a bit naïve, as `AssemblyVersion`, `AssemblyFileVersion` and `AssemblyInformationalVersion` can be mentioned in comments, class names or otherwise without actually being `[assembly:]` attributes. It shouldn't be a high priority to fix, but we just bumped into this problem and it would be nice if nothing had to be done in our code to make GitVersionTask happy. :)

tofutim:
haha i saw this today

PR #527, opened 2015-07-21
Used regex to determine if version attributes are still there.

distantcam:
I've added some tests for the InvalidFileChecker, and then upgraded it to use Regexes to search for the attributes.

Please check this to make sure I've covered as many different scenarios of code as possible.


distantcam:
Hopefully fixes #121

PR #537, opened 2015-07-26
Updated the generated output to include an "auto-generated" header

### LINKED ISSUE #535: Generated files are missing a standard auto-generated header
Several source analysis tools look for an "auto-generated" tag in the header of generated files in order to suppress analysis for these files. One example of such a tool is StyleCop. Here is an example of a suitable header. The specific part of the header which the tools look for is the substring `<auto-generated`. The remaining part is informational, but frequently included in such headers.

```csharp
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
//     GitVersion
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------
```

:memo: I'm planning to submit a pull request soon with a proposed implementation of this. But if someone wants to beat me to it, that's fine too. :smile:

sharwell:
Fixes #535


src/GitVersionTask/AssemblyInfoBuilder/AssemblyInfoBuilder.cs
```diff
@@ -10,7 +10,17 @@ public string GetAssemblyInfoText(VersionVariables vars, string assemblyName)
         var v = vars.ToList();
 
         var assemblyInfo = string.Format(
-@"using System;
+@"//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by a tool.
+//     GitVersion
```
sharwell: :memo: I did not include the version of GitVersion on this line because it's not already part of the output. Many tools do include this information; it's something to consider especially for cases where this file will not be part of source control.

Issue #313, opened 2014-11-20
Support Jenkins CI


serra:
See the comment thread in #288.

I'll see if I can get something to work in my environment. I'll open a PR then.

* Implement Jenkins.cs (compare: [TeamCity.js][tc]), with tests. Check for existence of `JENKINS_URL` to confirm that we're running from within the context of a job in Jenkins. See [wiki/Command-Line-Tool#directly-on-build-server](https://github.com/ParticularLabs/GitVersion/wiki/Command-Line-Tool#directly-on-build-server)
* choose parameter name to set [see variables in wiki](https://github.com/ParticularLabs/GitVersion/wiki/Variables)
* find out how to set parameter values from within a job

Current implementations write a message to the build output. This message should then be picked up by the build server. In Team City, this concept is called [Service Messages][svcmsg].

AFAIK a similar concept is *not* available in Jenkins CI, which would mean the Jenkins implementation would be conceptually different. Before picking this up, I'll wait for some answers to my [post to the Jenkins newsgroup](https://groups.google.com/forum/#!topic/jenkinsci-users/b7n1iAjJ7pA).

 [svcmsg]: http://confluence.jetbrains.com/display/TCD65/Build+Script+Interaction+with+TeamCity#BuildScriptInteractionwithTeamCity-AddingorChangingaBuildParameterfromaBuildStep
 [tc]: https://github.com/ParticularLabs/GitVersion/blob/master/GitVersionCore/BuildServers/TeamCity.cs
 [oldquestion]: https://groups.google.com/forum/#!searchin/jenkinsci-users/service$20messages/jenkinsci-users/7ktpX1RDpiY/Sy8jmf5cIRkJ

#### Todo

There does not appear to be a similar concept to service message in Jenkins. There are however several Jenkins plugins that add similar functionality to Jenkins, e.g. environment script plugin mentioned in the comments thread. We require use of this (or similar) plugin:

 * [x] select a plugin - Going for [EnvInject]
  * [x] Evaluate [EnvironmentScript]. Cannot be used; can only be run from .sh ish shells, which we cannot assume to be present on our .net build slaves IMO
  * [x] Evaluate [EnvInject], e.g. from sample [here](http://stackoverflow.com/questions/10625259/how-to-set-environment-variables-in-jenkins). This one would work.
 * [ ] implement `/buildserver` with tests for Jenkins CI
 * [x] set up jenkins job that runs agains my fork, building the propertiesfile_output branch. Build steps:
[...]

PR #551, opened 2015-07-30
Support for Jenkins CI

serra:
For #313.


JakeGinnivan:
Awesome. Minor thing, but there is a mix of `Assert.` and shouldly. It would be good to standardise on shouldly. Let me know if you have time otherwise I will be doing a pass at some point to remove all nUnit style assertions

serra:
OK, will address that today.

PR #552, opened 2015-08-03
Fixed path too long problem.

baltie:
In solutions containing a node_modules folder somewhere, there was a very good chance that you would get a PathTooLongException in DirectoryDateFinder.GetLastDirectoryWrite() since the fully qualified name in one of its sub directories would likely be larger than 260 chars. This change determines last write time in a way that doesn't throw this exception.

It was not possible to create a failing unit test without calling CreateDirectoryW in Windows API directly since .NET refuses to create directories with full path names longer than 260 characters.


JakeGinnivan:
Nice change, thanks!

baltie:
Thanks! :)

PR #555, opened 2015-08-04
Namespace name cannot start with numbers

cristianiorga:
Updating the assembly info on a project which file name starts with numbers will fail or when building a WPF project because the GenerateTemporaryTargetAssembly task will create a temporary project with a random name which can occasionally start with a number (see https://msdn.microsoft.com/en-us/library/bb397861.aspx).

The fix removes numbers from the start of the assembly name


JakeGinnivan:
#574 fixes this issue by using RootNamespace instead.

Thanks for the PR though!

Issue #561, opened 2015-08-10
Create VSO Build vNext Build Step for GitVersion


onovotny:
More info can be found here:
https://github.com/Microsoft/tfs-cli/blob/master/docs/buildtasks.md

The idea is to package GitVersion as a VSO custom build step so it can be more easily used in VSO.

PR #560, opened 2015-08-10
Add support for VSO agent #546

onovotny:

Issue #553, opened 2015-08-03
GitVersion not workig for detached head - TFS Build vNext


onovotny:
Some of the time my TFS vNext builds are failing due to blank version numbers being returned from GitVersion (yes, I could check for that, but I want the version!)

```
WARN [08/03/15 8:17:59:29] Could not determine assembly version: It looks like the branch being examined is a detached Head pointing to commit 'c28dce0'. Without a proper branch name GitVersion cannot determine the build version.
```
Seems like the build agents are checking out specific commits in a way that's confusing GitVersion? I do have a GitVersinConfig.yaml with the following:
```
assembly-versioning-scheme: MajorMinorPatch
next-version: 3.0.0
```

onovotny:
So I'm not sure why VSO is in a detached head state, maybe it's doing a checkout of a remote repo?

One workaround for that was to do the following in my VSO build script:
```xml     
<!-- ensure branch is checked out -->
<Exec Command="git checkout -b $(BUILD_SOURCEBRANCHNAME) $(BUILD_SOURCEVERSION)" />
```
VSO does know the branch and commit id, so I just manually call checkout again before the GitVersionTask runs.

onovotny:
So not sure if anything can ultimately be done in GitVersion to work with a detached head, but perhaps the workaround can be added to the docs on using GitVersion with VSO.

pascalberger:
Maybe you can raise an issue over at the https://github.com/Microsoft/vso-agent repo. I assume it's where the git checkout happens.

JakeGinnivan:
Checking out a detached head is very common, we have explicit support for build servers and when GitVersion detects it is running on a build server it normalises the git directory to checkout a branch and also turn remote branches into local ones.

https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/ExecuteCore.cs#L13 calls https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/GitPreparer.cs#L34 which controls this.

We need to add Team Build vNext to https://github.com/GitTools/GitVersion/tree/master/src/GitVersionCore/BuildServers - another discussion is happening at https://github.com/GitTools/GitVersion/issues/546 about this.

onovotny:
So given the fixups needed for VSO, I think the agent task is really the best way: https://github.com/GitTools/GitVersion/pull/562.

There was a bunch of PowerShell voodoo to normalize stuff as I couldn't use any fetching.

PR #562, opened 2015-08-10
Support for Visual Studio Online Build vNext - VSO Build Step

onovotny:
This is an implementation for #561. It should be done and has been tested with CI builds, manual builds and pull request builds.

Needs code review.


onovotny:
Not sure about that build failure- - it's an access denied error on the copy of GitVersion.exe as part of the build? 

Anyway, @JakeGinnivan @pascalberger, I think this is ready for you to take a look at. I could not get a git fetch working from within GitVersion due to not knowing the right temp session creds, but I was able to powershell my way around it, I think...

I've tested this with regular CI builds, pull requests and manual queue builds. You'll need to use the tfx tools to upload the `build\GitVersionVsoTaskBuild` directory to your VSO instance until the VSO gallery is available. We can also ask the VSO team if this can get published another way somehow before that.

onovotny:
The docs/video on how to install TFX are here: https://github.com/Microsoft/tfs-cli/blob/master/docs/buildtasks.md. Basically npm install tfx-cli -g, then you can do tfx login and then tfx upload and point to the dir.

Delete doesn't seem to work well due to agent caching of the tasks, so I've just incremented the patch version every time I make a change and re-upload it. That won't be necessary for the real package, just for dev.

pascalberger:
Thanks! 

Why all the handling of git branches in the PowerShell script? As mentioned by @JakeGinnivan in [this comment](https://github.com/GitTools/GitVersion/issues/553#issuecomment-129333679) I understand this should already happen inside gitversion.exe.

Do you have deliberately added all the changes from #560 again? Doesn't make much sense IMHO. I would keep this separated (#560 with the necessary changes in gitversion and this PR with only the VSO build task)

docs/more-info/build-server-setup/tfs-build-vnext.md
```diff
@@ -0,0 +1,26 @@
+# Visual Studio Online (Build vNext) Setup
+## Basic Usage
+In [Visual Studio Online](https://www.visualstudio.com/) build vNext (the web based build system) you can add a build step as follows:
+
+* **Build Step:** Command Line
```
pascalberger: Why command line when this PR is about a custom build step?

docs/more-info/build-server-setup/tfs-build-vnext.md
```diff
@@ -0,0 +1,26 @@
...
+	* Might need to fully qualify the path.
+	* Tip: Use a script and/or another command step to call `NuGet Install GitVersion.CommandLine` first so you don't have to check in the exe.
+* **Tool:**  `GitVersion.exe`
+* **Arguments:** `/output buildserver /updateassemblyinfo true`
+
+Then in your build parameters simply [add a placeholder](#nuget-in-teamcity) of the GitVersion variables you would like to use.
+
+GitVersion writes build parameters into VSO, so they will automatically be passed to your build scripts to use.
```
pascalberger: I don't think this works at the moment

onovotny:
So the reason all of the changes are there is that it's a branch off of a branch. I needed my fixes there to continue my work here. This is the only PR that should be needed; the other one can die.

The reason I had to do it all in PowerShell is that the C# code relies on authentication and VSO needs per-session tokens. I've pinged the team but have not gotten any response yet if any usable username/password is even available for use by the task to do the fetch/pull. This powershell seems to work so the fetch isn't needed. VSO does a full clone but does the checkout of a commit leaving the detached head. My script recreates the local branches and then sets the selected branch at the designated commit.

docs/more-info/build-server-setup/tfs-build-vnext.md
```diff
@@ -0,0 +1,26 @@
...
+## Basic Usage
+In [Visual Studio Online](https://www.visualstudio.com/) build vNext (the web based build system) you can add a build step as follows:
+
+* **Build Step:** Command Line
+	* Might need to fully qualify the path.
+	* Tip: Use a script and/or another command step to call `NuGet Install GitVersion.CommandLine` first so you don't have to check in the exe.
+* **Tool:**  `GitVersion.exe`
+* **Arguments:** `/output buildserver /updateassemblyinfo true`
+
+Then in your build parameters simply [add a placeholder](#nuget-in-teamcity) of the GitVersion variables you would like to use.
```
pascalberger: This seems like copy/past from TeamCity documentation and not accurate for this PR

docs/more-info/build-server-setup/tfs-build-vnext.md
```diff
@@ -0,0 +1,26 @@
...
+
+GitVersion writes build parameters into VSO, so they will automatically be passed to your build scripts to use.
+
+## GitVersion Build Step for VSO
+Visual Studio Online has support for custom build steps. This is planned but TBD. For now, the command line does work.
+
+ 
+## Running inside TeamCity
```
pascalberger: This seems like copy/past from TeamCity documentation and not accurate for this PR

docs/more-info/build-server-setup/tfs-build-vnext.md
onovotny: That's old and needs to change. That refers to using GitVersion.exe as a standalone. That won't get past the detached head issues that the VSO build step does. I'll update the docs.

docs/more-info/build-server-setup/tfs-build-vnext.md
onovotny: It was and it did work originally....will update! That's on the to do notes at the top :)

pascalberger:
OK, I see the point about the authentication. But this would also mean that the NuGet package cannot be used successfully with VSO? Maybe this would also be useful information to add to the docs.

onovotny:
Correct.... for now, this VSO build step is the only way to get it working.... unless someone else runs a similar script before running the NuGet package version.

The GitVersionTask would work too, long as this build step runs first as it'll prep the local clone.

pascalberger:
Just thinking if it maybe would make sense to split the task into two tasks:
- One for the branch handling
- One for running gitversion.exe

Like this for NuGet or GitVersionTask scenarios only the branch handling task can be included. Not that executing gitversion twice will do much harm, but it might clutter the logs.

What do you think?

onovotny:
Maybe... but then people would need to remember to add two tasks/build steps to their build definition. There's something to be said by keeping it simple. Just drop this one step in and it'll do whatever it needs to just work.

pascalberger:
OK, thats a point :)

src/GitVersionCore.Tests/BuildServers/VsoAgentTests.cs
```diff
@@ -0,0 +1,26 @@
...
+public class VsoAgentTests
+{
+    [Test]
+    public void Develop_branch()
+    {
+        var versionBuilder = new VsoAgent();
+        var vsVersion = versionBuilder.GenerateSetVersionMessage("0.0.0-Unstable4");
+        //  Assert.AreEqual("##vso[task.setvariable variable=GitBuildNumber;]0.0.0-Unstable4", vsVersion);
+
+        Assert.Null(vsVersion);
```
JakeGinnivan: Minor thing, but using Shouldly will give much better error messages. `vsVersion.ShouldBe(null)`

I think I am going to move to xUnit so there is not the option for using the inbuilt assertions, not going to block PR though, I will cleanup later

src/GitVersionVsoTask/Run-GitVersion.ps1
```diff
@@ -0,0 +1,66 @@
...
+
+$localBranch = $branch.Trim().Replace("refs/heads/", "").Replace("refs/","")
+
+Set-Location $loc
+
+$haveLocalBranch = $false
+# VSO checks out the commit as a detached HEAD
+# loop through and checkout all branches and the one we want
+# then reset merge to get to the desired commit id
+foreach ($remoteBranch in . git branch -r) {
```
JakeGinnivan: Not convinced about this, we should try and use the normalise git directory features in gitversion. With the /nofetch option it should skip the fetch then just normalise?

JakeGinnivan:
Awesome work @onovotny 

I just would like to see us use the inbuilt normalisation features in GitVersion, which fix up PR branches and a bunch of other things.

onovotny:
I did try to use the existing normalize method first, but was blocked by not having credentials available. As VSO appears to do a full clone, fetch/pull doesn't seem to be needed for it anyway. The question is if you want to special-case the existing method either for vso or a special marker user/pass to run a similar set of logic there, without the fetch-just checkouts and reset....

Just lmk

src/GitVersionVsoTask/Run-GitVersion.ps1
onovotny: With the no-fetch option, the state is a detached head as VSO checks out the commit, not the branch. There still needs cleanup. I based my script on the TeamCity GitVersion docs for the case where the clone uses SSH instead of HTTPS.

onovotny:
@JakeGinnivan the built-in normalize method doesn't apply for nofetch:
https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/BuildServers/GitHelper.cs#L16-L20

I don't know enough about how that method works to know if removing that early return and simply bypassing the `repo.Netowrk.Fetch()` and `repo.Netowrk.ListReferences()` will work?

The other possibly VSO-specific thing is that PR's appear as /refs/pull/<id>/merge. They're not under /refs/heads/pulls/... When I run `git branch -r`, the /refs/pull ones do not appear there, but there's still a valid commit it provides to checkout. I handle that logic in the PowerShell, not sure how to reconcile that with the normalize.

I do agree though that it's preferable to have it all in one place. The question was how much of this is VSO-specific cleanup/behavior. I didn't want to start putting server-specific code in the normalize method.

Those were my thoughts anyway.

onovotny:
There are two things remaining that I'm not sure how you want to handle @JakeGinnivan 
1. Set the version in the `task.json` to match the major.minor.patch of GitVersion. Is there a script that can parse/read/modify this on build?
2. Zip up the output of the `build\GitVersionVsoTaskBuild` directory and put it somewhere for people to download until the VSO gallery is ready for submissions (then ppl can just get it there). Not sure how you want to do that or where to put the zip?

gep13:
@onovotny in the interim, we could add that zip file as an entry on the releases page:

https://github.com/GitTools/GitVersion/releases

onovotny:
@gep13 yup, still needs to be created though; not sure if there's a preferred script/tool to zip up the artifact and make it available however you do for the nupkgs.

gep13:
Currently we are using AppVeyor build steps, either PowerShell or CMD, to do the build orchestration:

https://github.com/GitTools/GitVersion/blob/master/appveyor.yml

Another step, likely PS, to zip up the folder, and then publish as an artefact would be a good starting point I think.

onovotny:
Perfect, I see 7z in there, so I'll add another cmd to do the same for this task.

Next question, any suggestions on how best to update the task.json with the correct version number?

gep13:
>@onovotny said...
Next question, any suggestions on how best to update the task.json with the correct version number?

Off the top of my head, nothing springs to mind, aside from a regular expression replacement.

onovotny:
Thank you powershell. I think all of my tasks are done. The task.json version now matches Major.Minor.Patch and is available as an artifact. The only remaining thing is to put it on the /releases tab and update the download link in the VSO vNext setup docs with that link.

pascalberger:
A smaller issue: The build step currently requires access to the git command line tool. This should be at least documented as a requirement for the agents. Or a demand in the build step would be even better, but I don't know if it is possible to detect git capabilities of an agent.

JakeGinnivan:
I still think branch normalisation will work, then we can drop the powershell script entirely. I will try to investigate over the weekend

JakeGinnivan:
if we use the inbuilt GitVersion branch normalisation then we won't need the git command line tool on the build agent

onovotny:
If the in-built normalization can be made to work then awesome. I just wasn't sure what needed to be done there w/o needing any fetches/ops that would required creds.

If a script is still needed, I'm fairly certain I can convert it to use LibGit2Sharp instead of git.exe. That library is already distributed and used by the build agent to do its checkouts, so I could use it in PS. I don't want to do that work if it's not needed though.

JakeGinnivan:
I am currently giving this a spin, will report back in about an hour once I know more

johnkattenhorn:
This is awesome work guys, I dock my hat to you all! What's the word Jake ? Is this good to go :-)

JakeGinnivan:
Nearly, have a branch at https://github.com/JakeGinnivan/GitVersion/commits/vso-build-step which builds on this PR.

I want to figure out how I can get PR's detected properly in GitVersions branch normalisation. Tomorrow. But for now I am heading to bed :)

johnkattenhorn:
Cool, I can't wait, I'll be a test subject if you want @JakeGinnivan when your ready as I'm working on CI over the next couple of days.

JakeGinnivan:
Feel free to pull down my branch and give it a spin.

I reckon I'll get this merged tomorrow and released on the weekend. Was great work by Oren to get it to this point so don't want to delay it.

Sent from my Windows Phone

JakeGinnivan:
Ok, so there is a slight issue with the integration.

VSO does not force merge commits. Meaning the reason pull requests do not detect is because the branch is fast forwarded so the HEAD which is building matches an actual branch, so we never go searching for a pull request ref.

The fact that VSO doesn't force merge commits might have some other issues (multiple heads pointing at the same commit etc). I need to investigate. I have also reached out to some MS people to see if it's planned to allow forcing merge commits on branches.
I will still merge/release this and just put some notes up for things to look out for. I might even try and change the logic in normalisation to cater for fast forward merges in PR merge targets.

dazinator:
Could we add some logic to check for multiple heads pointing to the same commit, in order to check for these fast forward merges? It would be nice to handle fast forward merges - for the instances where people (or VSO) don't use `no-ff`

JakeGinnivan:
I do have some logic, but it is still non deterministic. There is enough logged that when it blows up, it gives us enough context of the scenario that we can figure how to improve it.

Sent from my Windows Phone

onovotny:
I think I have a solution but it's VSO specific. VSO passes the branch name in BUILD_SOURCEBRANCH and the commit id in BUILD_SOURCEVERSION.

The source branch is passed in the form refs/heads/foo, refs/heads/features/bar, etc for normal branches. For PR's it's refs/pull/<id>merge. It's not a real/committed branch.

The workaround is to see if the specified source branch matches any actual branches and create it if not. I did have that part in my script but it could be done in code too.

Maybe the `BuildServerBase` class and `IBuildServer` interface could have a `PrepForNormalization()` method where agent-specific pre-normalization code like this could run? The data can be pulled from the environment variables and local branches created for PR's based on the commit before the regular normalization happens?

JakeGinnivan:
See #571

johnkattenhorn:
So not off a great start, I can't even get tfx-cli to run at all. I assume the I needed npm installed in order to grab the tfx-cli package as I found the installation instructions https://github.com/Microsoft/tfs-cli.

I've tried running npm install from command prompt as both administrator and non-administrator and the install runs but then running tfx after the install gets me 'tfx' is not recognised.

I'll log an issue at the tfs-cli repo and see what I've missed.

pascalberger:
@johnkattenhorn Have you installed tfx-cli globally: `npm install -g tfx-cli`?

johnkattenhorn:
Yep, tried global, non-global all the same from Windows 10 Command Prompt and Admin, Non-Admin, Powershell.

I had this type of problem before with node and npm but don't know enough about npm to debug. I can see a node_module in my roaming profile but nothing in the path and no .exe / .cmd etc.

johnkattenhorn:
Here is a screenshot.

image

johnkattenhorn:
Someone helping me on tfx-cli, I'll report back shortly. I can pull the latest straight from master now can't I as the latest bits are merged right ?

pascalberger:
Strange. Works fine for me on Windows 10. After global install of the tfx-cli module you should see the module installed in <RoamingProfile>\npm\node_modules\tfx-cli\ (it contains no exe or cmd)

You can download the latest version from the master branch from the [AppVeyor build](https://ci.appveyor.com/project/GitTools/gitversion/branch/master/artifacts).

johnkattenhorn:
Bingo, seems that the installation of node from a full VS2015 install doesn't add the NPM directory to the path.

Doing a
    set PATH=%PATH%;C:\Users\johnk\AppData\Roaming\npm

Does the trick, thanks @pascalberger, i'll pull from there.

johnkattenhorn:
Investigating now, I think I remember @JakeGinnivan post something about the libraries.

image

pascalberger:
Maybe this is related to #573. I have only tested yet with earlier versions of the VSO task where I haven't experienced this error.

johnkattenhorn:
Looks like it could be @pascalberger, good find! @SimonCropp, @JakeGinnivan what do you think ? Should we be modifying the VSO BuildTask to include the external dependencies ?

johnkattenhorn:
Yep, if I rebuild from source then I get a Nativebinaries directory under the GitVersionVsoTaskBuild so I'm going to attempt to upload this custom task over the top of the existing.

johnkattenhorn:
Yep that is working fine, now to put together a custom build step for Octopus Packing ....

SimonCropp:
@johnkattenhorn yeah i think i fixed that on a later commit.  the native binaries should now be shipped with all the packaged things. let me know if any cant find the dependencies

pascalberger:
@SimonCropp I get the same exception with current head version. I created #589 for this.

SimonCropp:
@pascalberger not the same exception as in 589. the one here is about not being bale to find the libgit native files. 589 is about git.exe

SimonCropp:
@pascalberger sorry i stand corrected. i see u have two errors there

PR #563, opened 2015-08-11
Enable GitVersionTask to work with NuGet v3

onovotny:
NuGet v3 does not allow .props/.targets files directly in the `\build` directory. They must be in a subdirectory.

This PR puts the targets file in `\Build\portable-net+sl+win+wpa+wp` and `\lib\dotnet` so it'll work with any NuGet v2 or v3 client. The relative paths in the .targets file has been adjusted to match the new location.

Without this, the targets are never included in UWP or other project.json based projects (like modern PCL, etc).


gep13:
@SimonCropp is this similar to the changes that had to be made in Fody?

onovotny:
@gep13 yes, this is similar: https://github.com/Fody/Fody/blob/master/NuGet/Nuget.csproj#L50-L51

SimonCropp:
@gep13 yes.   @onovotny this is very good work

@JakeGinnivan this is good for a merge from my perspective

PR #579, opened 2015-08-17
Fix command for uploading of the task

pascalberger:
The correct command for uploading tasks is `tfx build tasks upload`


JakeGinnivan:
Nice catch, thanks

PR #588, opened 2015-08-19
Quote source path

pascalberger:
Quote source path to support paths containing spaces


pascalberger:
Closed. #591 is an updated PR

PR #593, opened 2015-08-19
Add generic error message if more than one remote is found

pascalberger:
Add generic error message if more than one remote is found since this does not only apply for TeamCity


gep13:
Looks good to me! :+1:

PR #594, opened 2015-08-20
fix version cache key in msbuild task

### LINKED ISSUE #488: Version number is not cached between projects in a solution
As mentioned in #437, when GitVersionTask (version 3 beta 4) is installed in a solution, I get the following informational message when building every project in the solution.
```
INFO [07/08/15 14:57:03:21] Version not in cache. Calculating version.
```
It would speed the build up noticeably if the version number was successfully cached between projects.

SimonCropp:
fixes #488

### DIFF
--- a/src/GitVersionTask/VersionAndBranchFinder.cs
+++ b/src/GitVersionTask/VersionAndBranchFinder.cs
@@ -27,17 +27,20 @@ public static class VersionAndBranchFinder
         var gitDir = GitDirFinder.TreeWalkForDotGitDir(directory);
         using (var repo = RepositoryLoader.GetRepo(gitDir))
         {
-            var ticks = DirectoryDateFinder.GetLastDirectoryWrite(directory);
-            var key = string.Format("{0}:{1}:{2}", repo.Head.CanonicalName, repo.Head.Tip.Sha, ticks);
+            var ticks = DirectoryDateFinder.GetLastDirectoryWrite(gitDir);
+            var key = string.Format("{0}:{1}:{2}",gitDir, repo.Head.CanonicalName, repo.Head.Tip.Sha);
 
+            Logger.WriteInfo("CacheKey: " + key );
             CachedVersion result;
             if (versionCacheVersions.TryGetValue(key, out result))
             {
                 if (result.Timestamp != ticks)
                 {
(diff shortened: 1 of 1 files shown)


src/GitVersionTask/VersionAndBranchFinder.cs
```diff
@@ -27,17 +27,20 @@ public static VersionVariables GetVersion(string directory, Authentication authe
         var gitDir = GitDirFinder.TreeWalkForDotGitDir(directory);
         using (var repo = RepositoryLoader.GetRepo(gitDir))
         {
-            var ticks = DirectoryDateFinder.GetLastDirectoryWrite(directory);
-            var key = string.Format("{0}:{1}:{2}", repo.Head.CanonicalName, repo.Head.Tip.Sha, ticks);
+            var ticks = DirectoryDateFinder.GetLastDirectoryWrite(gitDir);
+            var key = string.Format("{0}:{1}:{2}",gitDir, repo.Head.CanonicalName, repo.Head.Tip.Sha);
```
JakeGinnivan: Ticks is no longer used?

JakeGinnivan: `var key = string.Format("{0}:{1}:{2}:{3}",gitDir, repo.Head.CanonicalName, repo.Head.Tip.Sha, ticks);` ?

SimonCropp: we dont want ticks in the key. they are used in the second step. a static dictionary is bascially a memory leak. keying on ticks is not great since we gradually add more to the dictionary on every change.

JakeGinnivan: Makes sense, if tip sha has changed, recalc. Got it

asbjornu: The dictionary should probably be changed to a [`ConcurrentDictionary`](https://msdn.microsoft.com/en-us/library/dd287191.aspx) too, to avoid race conditions and concurrency issues if (and with Roslyn; when) the build is parallelized.

SimonCropp: @asbjornu agreed. make it so

Issue #566, opened 2015-08-13
Using ref/heads in a Dynamic Repository /b appears fail and fallback to default without warning


RichiCoder1:
See Title.

I'm using GitVersion with TeamCity, and because I'm using SSH I have to fallback to dynamic repositories. I set everything up and it works great on develop (since it's the default), but passing `refs/heads/release/2.2.0` appeared to fail and just default to Develop. No warning or anything. Using `release/2.2.0` seemed to work just fine. Also not sure how to pass in "tag", which is not technically a branch.

Would be nice to see a failure instead of just defaulting in these cases. Also would like guidance on the recommended branch format, and how to point GitVersion at tags.

JakeGinnivan:
Can you post the GitVersion logs when it fails?

RichiCoder1:
Here's the results when running `tags/2.2.0`: https://gist.github.com/RichiCoder1/b294bccc2fdf5ee32788
Looks like the tag is simply ignored.

RichiCoder1:
Here's the results when running `refs/heads/release/2.2.0`: https://gist.github.com/RichiCoder1/57963d963b4d4c2e66f2

Same result. The branch appears to simple be ignored.

JakeGinnivan:
Can you try 3.1.0 and see if it is still and issue. If it is I will need the new gist, I have improved the logging

PR #600, opened 2015-08-22
Removed normalisation related code from dynamic repositories. It uses…

JakeGinnivan:
… NormalizeGitDirectory fully now


JakeGinnivan:
Should fix #566

Issue #611, opened 2015-08-25
GitVersion fails to pull a dymanic repository when on TeamCity


RichiCoder1:
Here's the error message:
```
C:\Source\core [release/2.2.0 +1 ~4 -0 !]> .\tools\GitVersion.CommandLine\tools\GitVersion.exe /url "******" /u ******* /p ********* /b "refs/heads/release/2.2.0" /c ea0f3605343b5703b98815e8f34c04dda91f9564 /output json
INFO [08/25/15 16:14:47:55] Working directory: C:\Source\core
INFO [08/25/15 16:14:47:56] Applicable build agent found: 'TeamCity'.
ERROR [08/25/15 16:14:47:56] An unexpected error occurred:
System.Exception: Dynamic Git repositories must have a target branch (/b)
   at GitVersion.GitPreparer.CreateDynamicRepository(String targetPath, Authentication authentication, String repositoryUrl, String targetBranch, Boolean noFetch)
   at GitVersion.GitPreparer.Initialise(Boolean normaliseGitDirectory, String currentBranch)
   at GitVersion.ExecuteCore.ExecuteGitVersion(IFileSystem fileSystem, String targetUrl, String dynamicRepositoryLocation, Authentication authentication, String targetBranch, Boolean noFetch, String workingDirectory, String commitId)
   at GitVersion.SpecifiedArgumentRunner.Run(Arguments arguments, IFileSystem fileSystem)
   at GitVersion.Program.VerifyArgumentsAndRun()
```

Source at fault: https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/ExecuteCore.cs#L20

TeamCity returns `string.Empty` which cases `currentBranch ?? targetBranch` to evaluate to `string.Empty`. Can be fixed by instead writing:
```csharp
gitPreparer.Initialise(buildServer != null, string.IsNullOrWhiteSpace(currentBranch) ? targetBranch : currentBranch); 
```
This is a somewhat critical bug as it does block using 3.1.0 at all, though we're experimenting with a work around where we kill `$env.TEAMCITY_VERSION` for the lifetime of the git version run.

EDIT: Workaround worked :). World not as ending.

furrot:
Any chance you'll publish the build soon for this? Can't figure out a workaround.

JakeGinnivan:
Yeah, probably tomorrow

JakeGinnivan:
Actually, am just doing it now

JakeGinnivan:
[...]

PR #612, opened 2015-08-26
re #611 - Ensures branch with dynamic repositories and buildserver

Philo:
When using a dynamic repository with a buildserver the current branch will for most buildserver types return null or empty, this causes no branch to be set, dynamic repositories require a branch.

The fix ensures that if we are using dynamic repositories with a buildserver that if the buildserver does not return a branch we use the targetBranch specified via /b

### DIFF
--- a/src/GitVersionCore.Tests/BuildServers/BuildServerBaseTests.cs
+++ b/src/GitVersionCore.Tests/BuildServers/BuildServerBaseTests.cs
@@ -44,10 +44,5 @@ public class BuildServerBaseTests
         {
             return new string[0];
         }
-
-        public override string GetCurrentBranch()
-        {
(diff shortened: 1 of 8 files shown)


JakeGinnivan:
With this we no longer need the `currentBranch ?? targetBranch` as well right? Might be worth cleaning that up at the same time

Philo:
Indeed, I can do a small tidy up of the current branch resolution with ```ExecuteCore``` and perhaps standardise the return value of ```IBuildServer.GetCurrentBranch()``` to return null by default unless overridden in a specific build server implementation (like VSO).

This update will however mean that if using a dynamic repository with VSO the environment variable BUILD_SOURCEBRANCH will override the value of /b, but /b will still need to be specified as an argument.

PR #615, opened 2015-08-30
Fix effective configuration using minor version bump regex instead of patch regex

TomGillen:
Oops?


TomGillen:
This would have caused `+semver: patch` to be ignored - which typically would result in a patch increment anyway, so its not likely to be something someone would run into.

Issue #620, opened 2015-09-01
Update default branch config


onovotny:
Default branch config should support features and dev. Also, develop is currently catching too many items, like developer.

PR #622, opened 2015-09-01
Default config updates to recognize more defaults

onovotny:
This addresses #620

### DIFF
--- a/src/GitVersionCore/Configuration/ConfigurationProvider.cs
+++ b/src/GitVersionCore/Configuration/ConfigurationProvider.cs
@@ -32,7 +32,7 @@ namespace GitVersion
 
             ApplyBranchDefaults(config, GetOrCreateBranchDefaults(config, "master"), defaultTag: string.Empty, defaultPreventIncrement: true);
             ApplyBranchDefaults(config, GetOrCreateBranchDefaults(config, "release[/-]"), defaultTag: "beta", defaultPreventIncrement: true);
-            ApplyBranchDefaults(config, GetOrCreateBranchDefaults(config, "feature[/-]"), defaultIncrementStrategy: IncrementStrategy.Inherit);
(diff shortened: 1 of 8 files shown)


JakeGinnivan:
We need to do a config upgrade for all these.

If someone has provided config overrides for these different branch configurations we can't have the new configs taking over

src/GitVersionCore/Configuration/ConfigurationProvider.cs
```diff
@@ -31,15 +31,15 @@ public static void ApplyDefaultsTo(Config config)
             var configBranches = config.Branches.ToList();
 
             ApplyBranchDefaults(config, GetOrCreateBranchDefaults(config, "master"), defaultTag: string.Empty, defaultPreventIncrement: true);
-            ApplyBranchDefaults(config, GetOrCreateBranchDefaults(config, "release[/-]"), defaultTag: "beta", defaultPreventIncrement: true);
```
JakeGinnivan: Add another parameter to this which is `migrate:` and then it takes an array. You can then add `migrate: new[] { "release[/-]" }` and ApplyBranchDefaults should rename the configs to be migrated, then apply the defaults.

Just one idea on how to solve.

onovotny: I guess I'm not sure how the migration should work? Should apply branch defaults look for the old keys and rename them?

JakeGinnivan: Yeah exactly. So I have:

```
branches:
  develop:
    tag: beta
```

The effective config will have both `develop` and `dev(elop)?...`. We need to look for develop, rename it to the new one then delete it

onovotny:
Okay this should be working now. Also, the CanReadDocument test is now CanReadDocumentAndMigrate as that proves that it works.

Issue #619, opened 2015-09-01
Enable config for amount of padding


onovotny:
The padded variables should have a configurable option for the amount of padding

PR #623, opened 2015-09-01
Make padding configurable and add padded BuildMetaData

onovotny:
This addresses #619


JakeGinnivan:
We need to update the docs to include this new variable. Also the GitVersion task .targets file needs to be extended. https://github.com/GitTools/GitVersion/blob/master/src/GitVersionTask/NugetAssets/GitVersionTask.targets#L44

Issue #618, opened 2015-09-01
NullReferenceException in FindCommitBranchWasBranchedFrom


michaelnoonan:
Apologies for the terse Issue, I was out of time today and am hoping to reproduce the issue locally tomorrow.

What I've noticed is this `NullReferenceException` started after merging from `master` back into our `release/3.1.0` branch. Now running `gitversion.exe` fails with the NRE, and `GitVersionTask` is doing the same thing in Visual Studio making it horribly slow.

Both `GitVersion.Portable` and `GitVersionTask` are `3.1.0`.

I've worked around this temporarily by adding a Tag to force the SemVer. I haven't found another workaround that gets back to normal operation (without the Tag hack).

Full log file as requested...

```
2015-09-01 15:55:50		INFO [09/01/15 15:55:50:78] Working directory: c:\dev\OctopusDeploy\source
2015-09-01 15:55:50		INFO [09/01/15 15:55:50:81] Project root is: c:\dev\OctopusDeploy
2015-09-01 15:55:50		INFO [09/01/15 15:55:50:85] Using latest commit on specified branch
2015-09-01 15:55:51		INFO [09/01/15 15:55:51:06] Running against branch: release/3.1.0 (7025eb3f4a737e85403ce6d24bcfd3a08dd5d640)
2015-09-01 15:55:51		INFO [09/01/15 15:55:51:10] Begin: Calculating base versions
2015-09-01 15:55:51		  INFO [09/01/15 15:55:51:47] Fallback base version: 0.1.0 with commit count source b69a057dd90cf0f567819efcf8b88e3806161d72
2015-09-01 15:55:53		  INFO [09/01/15 15:55:53:94] Git tag '3.0.21': 3.0.21 with commit count source 3fcb35e57ea3fda043d433e4526ac2ddc80c4cc0
2015-09-01 15:55:53		  INFO [09/01/15 15:55:53:97] Git tag '3.0.20': 3.0.20 with commit count source ea8a3f0b96c77b1ea5340e67a698a1e57f0a502c
2015-09-01 15:55:53		  INFO [09/01/15 15:55:53:99] Git tag '3.0.19': 3.0.19 with commit count source 16713d9a7700a750a0c271bb6ac75133a0df4ec1
2015-09-01 15:55:54		  INFO [09/01/15 15:55:54:90] Merge message 'Merge remote-tracking branch 'origin/master' into release/3.1.0

Conflicts:
	source/Octopus.Server/packages.config
[...]

PR #625, opened 2015-09-02
FindCommitBranchWasBranchedFrom and GetBranchesContainingCommit ignore branches with no Tip

michaelnoonan:
A change to ignore branches with no `Tip` from `FindCommitBranchWasBranchedFrom()`. Whilst this change works on the problem branch in my repo, I'm concerned if it has unintended side-effects.

Writing a test for this proved difficult - do you have any suggestions?

Fixes GitTools/GitVersion#618


michaelnoonan:
We're changing approach... See #618 for more detail.

PR #626, opened 2015-09-02
Should not checkout branch on update

### LINKED ISSUE #624: Mixed up branches when running in TeamCity
Running GV 3.1.1 on TC I'm having a issue with GV using the branch from the previous build see attached image. I've also seen this we're "develop" got the version from a feature-branch pointing at the same commit. I've tried to reproduce this with a failing integration test but have failed and even locally against the same repository but have failed. @JakeGinnivan do you have a clue what could be wrong or how to mimic the TC behavior so I can reproduce the issue locally and provide a fix?

image

2015-08-28 23_02_25-brikks __ development __ build release _ overview teamcity

On hint could be, `HEAD points at branch 'refs/heads/feature/BRIKKS-4446` from the gist below:
https://gist.github.com/orjan/40c8fc5ac36e10e381c8

orjan:
An update on a local branch cannot checkout the updated branch, since that
will cause GV to calculate the version on the updated branch instead of
the intended.

Fixes #624

### DIFF
--- a/src/GitVersionCore.Tests/GitHelperTests.cs
+++ b/src/GitVersionCore.Tests/GitHelperTests.cs
@@ -100,5 +100,34 @@
                 }
             }
         }
+
+        [Test]
+        public void ShouldNotChangeBranchWhenNormalizingTheDirectory()
+        {
+            using (var fixture = new EmptyRepositoryFixture(new Config()))
+            {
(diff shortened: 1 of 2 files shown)


JakeGinnivan:
Oops :)

Issue #630, opened 2015-09-03
[Feature Request] Allow creation of source file containing assembly version information


Philo:
Using /updateassemblyinfo true|<file> will update the existing version information stored within either **/AssemblyInfo.cs or <file>, however I think it would be useful to provide an option to actually create <file> and populate it with assembly version information if the file does not exist.

```
GitVersion.exe /updateassemblyinfo VersionAssemblyInfo.cs
```

will:

* check for the existence of VersionAssemblyInfo.cs, if found will update existing version attributes
* if VersionAssemblyInfo.cs does not exist, create the file and add the version attributes to it
 - Inspect the filename extension to .cs or .vb to determine whether the generated file should contain VB and C# code

In general a common/global assembly info file would not be committed to source control, and for large projects managing individual AssemblyInfo.* files would be a pain.

Thoughts?

robdmoore:
If GitVersion is creating the file then wouldn't that mean it probably hasn't been referenced in the .csproj? Hence the file won't be used if it's created.

The only time this wouldn't be the case is if it's in the gitignore, but then you would break the local clone and build (without gitversion) scenario.

Philo:
In our workflow we have build tasks that would be executed locally against a freshly cloned repository, if a new developer was starting on a code base.  They would be running this initial script (usually the same script executed on our build server) before beginning work, this would create the ```CommonAssemblyInfo.cs``` file along with any other setup required.

We never commit the "CommonAssemblyInfo.cs" file for our projects, it is always generated via a script or build server workflow task.

In this scenario I have to add a build workflow step to create the file prior to executing GitVersion, my preferred solution would be that we remove the version attributes from our "CommonAssemblyInfo.cs" and add a new "VersionAssemblyInfo.cs" file that is generated by GitVersion.

__**/Properties/AssemblyInfo.cs__ (within each project)
```c#
using System.Reflection;

[assembly: AssemblyTitle("Product.Stuff")]
```

__CommonAssemblyInfo.cs__
```c#
using System.Reflection;
using System.Runtime.InteropServices;

[assembly: AssemblyDescription("My product rocks")]
[assembly: AssemblyCompany("Best Company Ever")]
[assembly: AssemblyProduct("My Super Product")]
[assembly: AssemblyCopyright("Copyright 2015")]
[assembly: ComVisibleAttribute(false)]
```

__VersionAssemblyInfo.cs__
```c#
//------------------------------------------------------------------------------ 
// <auto-generated> 
[...]

PR #631, opened 2015-09-03
re #630 - optionally allows /updateassemblyinfofilename to be generated

Philo:
Adds argument /ensureassemblyinfo, when specified and /updateassemblyinfofilename is specified and does not exist we will generate a template of the assembly info file containing only *Version
attributes in either cs, fs, vb allowing the version values to be replaced.


gep13:
@JakeGinnivan what are your thoughts on this one?  Personally I was thinking that this would GitVersion away from it's core functionality.

JakeGinnivan:
Yeah unsure about it.

It only supports adding an assemblyInfo file in the root of the repo which is probably too limiting. I also don't like the idea of a file missing until you run GitVersion in your repo. Any reason the Task doesn't work for you, because it has this functionality because it is inside the MSBuild pipeline?

Philo:
The primary reason for not considering the MSBuild task is a lack of support for dynamic repositories, could this be implemented, perhaps by adding repo url, branch and commit id to the build pipeline somehow?

Coming from a rake build background I'm not a fan of modifying msbuild xml as it tends is likely to lead to confusion within the team as it is less transparent than an "external" build script/process.

The specific implementation would solve a specific issue we have with our build process and using GitVersion, that we would have to commit a file to the repo that we have historically always avoided adding as it is auto-generated from our build scripts (rake).

I'm happy to discuss further, either make refinements or consider looking at implementing dynamic repositories within the MSBuild task assuming there are no blocking issues with doing so.

Philo:
@JakeGinnivan, I have tried the GitVersionTask as suggested, and while it stamps the version on the assemblies as I would need am I still left with 2 issues.

* No support for dynamic repositories (our usual CI environment is Teamcity, I used appveyor for testing)
* With the GitVersionTask I can not create a valid symbols package due to the auto-generated version assembly file not being included in the project itself.  (see #636)

A very typical case for using GitVersion is for library packages or mini frameworks (one of my core reasons for wanting to use it to manage versioning across a team), it is important for us to be able to generate symbols packages for debugging.  Relying on something like dotPeek would be an alternative but I wouldnt want to have to dependent on an external tool.

In the worst case I can live without symbols, based on @SimonCropp reasons.

I can appreciate not wanting to pollute GitVersionExe with an answer to every problem, however this is really just an extension to the existing /updateassemblyinfo task.

At the moment if you specify __/updateassemblyinfo file_that_doesnt_exist.cs__ then the fallback is to search for AssemblyInfo.cs files and just update them, I would argue this is not the correct fallback as the user has specifically requested that an individual file be updated, if that file doesnt exist then either an error is thrown or nothing is updated.

JakeGinnivan:
Yeah completely understand. I think it's fine we bring this in, the issue is it doesn't quite fit in with what we have not:

 - `/updateassemblyinfo file.cs` actually is not an absolute path, it looks for all assemblyinfo.cs files
 - `/ensureAssemblyInfo` creates a single file in the root of the repository, it is too restrictive

I think maybe it should be a config setting:

```
assemblyInfoFiles:
  - src\Proj\Properties\VersionInfo.cs
  - src\Proj2\Properties\VersionInfo.cs
```

Then we change gitversion to append those to the files found by the scan and then it should always create them if they do not exist.

Thoughts?

dazinator:
Is doing this stuff as a "pre-build" event, and also ensuring the assembly info file is included in the proj file (by checking the proj file and modifying if necessary) a bad idea? I'm not a fan of modifying files during a build but would that cover the scenario of git "ignored" assembly info files, and the issue with missing file from proj preventing production of symbol source package?

Philo:
@JakeGinnivan, I agree about ```/ensureassemblyinfo``` being far too restrictive, I actually never thought about this and was focused on re-using as much of the existing file modification code as possible and not altering or breaking any existing execution scenarios.

I'll take a look at adding to the config options for specifying these as additional assembly info files.

dazinator:
In other words, why dont we update the proj file to include the auto generated assembly info file?

JakeGinnivan:
I have tested `/updateassemblyinfo path/to/versionfile.cs` works just fine. If you have a single version file just make a change that if it cannot be found that path is created relative to the working directory

The config can be done later

JakeGinnivan:
Do you want me to create a new issue to track this?

Philo:
@jakeginnivan, I will try and get some time in the next few days to update this

PR #641, opened 2015-09-13
Improve formatting and fix some typos

ermshiperete:


.gitattributes
```diff
@@ -20,4 +20,7 @@
 *.ico binary
 *.snk binary
 *.pdb binary
-*.svg binary
\ No newline at end of file
+*.svg binary
+
+# Don't check for whitespace errors in the doc pages
+*.md -whitespace
```
JakeGinnivan: Wouldn't it be better for force line normalisation? What does this achieve, and does github diff take this into account?

docs/reference/intro-to-semver.md
```diff
@@ -10,18 +10,18 @@ Version Promiscuity is the opposite problem, **JsonLibrary** releases _v1.1.0_ w
 
 SemVer introduces conventions about breaking changes into our version numbers so we can safely upgrade dependencies without fear of unexpected breaking changes while still allowing us to upgrade downstream libraries to get new features and bug fixes. The convention is quite simple.
 
-{major}.{minor}.{patch}-{tag}+{buildmetadata}  
-{major} = Only incremented if the release has breaking changes (includes bug fixes which have breaking behavioural changes
-{minor} = Incremented if the release has new non-breaking features
-{patch} = Incremented if the release only contains non-breaking bug fixes
-{tag} = Optional -{tag} denotes a pre-release of the version preceeding
-{buildmetadata} = Options +{buildmetadata} contains additional information about the version, but DOES NOT AFFECT the semantic version preceding it.
+**{major}.{minor}.{patch}-{tag}+{buildmetadata}**
```
JakeGinnivan: Line break

ermshiperete: What do you mean? Do you want an additional line break here? I guess that might make sense.

.gitattributes
ermshiperete: Markdown uses two blanks at the end of the line to force a linebreak. This option turns off checking for whitespace errors so that a local git client won't complain when you try to commit a file with such lines.
I don't know what GitHub does, I hope it honors those settings (as it should) - if not then it's no worse than before on GitHub, but at least it is an improvement for local git clients.

docs/reference/intro-to-semver.md
JakeGinnivan: Ignore, there are the spaces, the difftool wasn't showing them unlike the other lines

Issue #73, opened 2014-01-23
Problem using nuget pack -includedependencies with assemblies using GFV


thirkcircus:
Background
-----------------
I have two assemblies in a solution, AssemblyA and AssemblyB.  AssemblyB has a project reference to AssemblyA.
I'm using GFV and publishing develop branch nuget packages from TC for both assemblies.

I end up with the following for both assemblies (from the temp AssemblyInfo written out by GFV:
[assembly: AssemblyVersion("2.8.0")]
[assembly: AssemblyFileVersion("2.8.0")]
[assembly: AssemblyInformationalVersion("2.8.0-unstable12 Branch:'develop' Sha:'c415f7a3dbadad7b72c8af4b7ae8993d1ef35710'")]
The actual compiled assemblies both have AssemblyVersion 2.8.0 **.0**  (I guess this is a standard .net thing - to append ".0" to get a four part version number.)
When running nuget pack I'm using %GitFlowVersion.NugetVersion% as the package version and I'm telling nuget to IncludeReferencedProjects - so AssemblyB.nupkg will have a dependency on AssemblyA.nupkg

Problem
------------
When I install AssemblyB.nupkg nuget fails to resolve the AssemblyA.nupkg dependency.  
> Unable to resolve dependency 'AssemblyA (= 2.8.0.0)'.

And that's because the AssemblyB nupkg contains the following:
```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
    <metadata>
        <id>AssemblyB</id>
        <version>2.8.0-Unstable0012</version>
        <dependencies>
            <dependency id="AssemblyA" version="2.8.0.0" />
        </dependencies>
    </metadata>
</package>
```

What I really want is the dependency to specify the GitFlowVersion.NugetVersion instead of the assemblyversion from the referenced project.  A workaround is not to includereferencedprojects when packing but I don't want that.

How does NServiceBus handle this?  How do you get your unstable NServiceBus.Host package to depend on NServiceBus? eg 
```xml
<?xml version="1.0" encoding="utf-8"?>
<package xmlns="http://schemas.microsoft.com/packaging/2011/10/nuspec.xsd">
    <metadata>
        <id>NServiceBus.Host</id>
        <version>4.0.0-unstable3109</version>
        <title>NServiceBus Host (unstable)</title>
        <dependencies>
[...]

PR #658, opened 2015-10-02
Issue #73: proposal for AssemblyInformationalVersioningScheme

MarkZuber:
Similar to the AssemblyVersioningScheme setting, this adds the ability to configure an AssemblyInformationalVersioningScheme in the yml file.  The default remains "full informational" as it is today.

The new option is for NugetVersion which will put the same tag into the AssemblyInformationalVersion attribute that it would for the NuGet package version.   

This should allow IncludeReferencedProjects to work according to the Nuget docs:

From nuget docs (https://docs.nuget.org/create/versioning)

If building a package from a project (.csproj or .vbproj), use the AssemblyInformationalVersionAttribute to specify the version.

[assembly: AssemblyInformationalVersion("1.0.1-alpha")]

NuGet will pick up this value instead of the one specified in the AssemblyVersion attribute (this attribute does not support SemVer which is why a different attribute was needed).


JakeGinnivan:
Ditto, a few things which will make your life easier:

1. Work on feature branches
1. Have two remotes configured, origin pointing at your fork, upstream pointing at the main repo
1. rebase your feature branch on upstream/master
1. Never commit to master, just update it with new stuff from upstream.

To get yourself in this state you want to:

```
git checkout -b feature/assemblyInfoUpdates
git fetch upstream
git rebase upstream/master    # if you hit conflicts, you resolve just like a merge. Make sure you git rebase --continue after though, not commit
git checkout master
git reset --hard upstream/master   # this will fix your master so it doesn't have any merge commits or commits on it and can be updated with a fast forward merge
git checkout feature/assemblyInfoUpdates
git push origin feature/assemblyInfoUpdates
```

Then you can create a new pull request from your feature branch. Let me know if you need more info Mark, git has a bit of a learning curve :)

MarkZuber:
Hey Jake,

Thanks.  That was super helpful.  I appreciate your patience :D

Updated pull request from my feature branch has been submitted:
https://github.com/GitTools/GitVersion/pull/660

I've verified locally with my exe that it solves the issue.  And I've run
all unit tests (including new ones) and verified that they've passed.

thanks!
mark

JakeGinnivan:
Superseded by #660

Issue #368, opened 2015-02-15
Allow configuring AssemblyInformationalVersion contents


JakeGinnivan:
Add a new branch configuration option `assembly-informational-version-format` which supports string interpolated values from variables.

For example `{majorMinorPatch}.branch.{branch}` would use the Branch and MajorMinorPatch variables from the variable provider to create the informational version. This way if you are using ClickOnce (which expects that attribute to contain a valid version) you can override it

robdmoore:
Great idea!

deyvsh:
+1

veccie:
Maybe this link can be of use: https://mhusseini.wordpress.com/2014/05/03/fast-named-formats-in-c/

I've made a quick example to test it which you can see below.

The only thing I had to change was the Regex pattern, but then it also works great.

    private static readonly Regex RegexFormatArgs = new Regex(@"({)([^}]+)(})", RegexOptions.Compiled);

Example:
```
    public class SemanticVersion
    {
        public int Major;
        public int Minor;
        public int Patch;
        public string MajorMinorPatch;
        public string PreReleaseTag;
        public int CommitsSinceTag;
        public string Branch;
        public string Sha;
        public string ShaShort;
        public DateTimeOffset CommitDate;

        public string ToString(string format)
        {
            return NamedFormat.Format(format, this);
        }
    }

    SemanticVersion version = new SemanticVersion
    {
        Major = 2,
        Minor = 4,
        Patch = 1,
        MajorMinorPatch = "2.4.1",
        PreReleaseTag = "beta",
        CommitsSinceTag = 8,
        Branch = "release",
        Sha = "82ff5cc4c3f3ea9aa2711c163fb1edf68abd858b",
        ShaShort = "82ff5cc",
        CommitDate = DateTimeOffset.Now
    };

    Debug.WriteLine(version.ToString("{majorMinorPatch}-{PreReleaseTag}.{CommitsSinceTag}.{shaShort}"));
    Debug.WriteLine(version.ToString("{MajorMinorPatch}.{CommitsSinceTag}"));
    Debug.WriteLine(version.ToString("{MajorMinorPatch}-{PreReleaseTag}.{CommitsSinceTag}+Branch.{branch}.Sha.{sha}"));
    Debug.WriteLine(version.ToString("{Major}.{minor}-{PreReleaseTag}"));
```
[...]

PR #660, opened 2015-10-02
Add support and tests for AssemblyInformationalVersion configurability

MarkZuber:


JakeGinnivan:
Thanks for rebasing this @MarkZuber 

Just wondering if it would be better to instead call it `assembly-informational-override` then specify a GitVersion variable. For instance `assembly-informational-override: NuGetVersion`, but you can also go `assembly-informational-override: FullSemVer`

Another option is to have `assembly-informational-format: {NuGetVersion}, sha: {Sha}`, then you can format however you want. I would suggest maybe using https://www.nuget.org/packages/netfx-System.StringFormatWith/ if you wanted to go down this road, if that doesn't do the job then have a look at http://haacked.com/archive/2009/01/14/named-formats-redux.aspx/ for some code to use.

asbjornu:
I'd prefer a `String.Format`-based solution with named variables. :)

MarkZuber:
OK.  I've looked at StringFormatWith and it works great except the code
pulled in from the nuget package is older .net and has a few conflicts as
is.  It's one file, so i pulled that code in (with appropriate
attributions) into the gitversioncore project.

all tests currently pass and i'm writing some additional tests now to cover
a few additional formatting cases.  hope to submit an updated pull request
later today.

thanks for the feedback and input.

-mark

MarkZuber:
Quick ping.  Does everything look acceptable?  Would love to get this into mainline and get a real drop.  Please let me know if there are any other changes you'd like me to make.  thanks!

pascalberger:
Does this also fix #368?

src/GitVersionCore/SemanticVersionFormatValues.cs
```diff
@@ -0,0 +1,116 @@
+﻿namespace GitVersion
+{
+    public class SemanticVersionFormatValues
```
JakeGinnivan: Why can't we use the `VersionVariables` class for this?

src/GitVersionCore/GitVersionCore.csproj
```diff
@@ -47,6 +47,7 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Web" />
```
JakeGinnivan: What is needed from this reference. Am worried this could cause issues porting to .net core/mono which is happening in v4

src/GitVersionCore.Tests/ConfigProviderTests.cs
```diff
@@ -111,12 +112,27 @@ public void CanWriteOutEffectiveConfiguration()
     }
 
     [Test]
+    public void CanUpdateAssemblyInformationalVersioningScheme()
+    {
+        const string text = @"
+assembly-versioning-scheme: MajorMinor
+assembly-informational-format: '{NugetVersion}'";
```
JakeGinnivan: I would also like to see a test which uses multiple format values. i.e `'{Major}.{Minor}.{Patch}'`

JakeGinnivan:
Looking really good @MarkZuber, just a few comments on there. Thanks!

src/GitVersionCore.Tests/ConfigProviderTests.cs
MarkZuber: done.

src/GitVersionCore/GitVersionCore.csproj
MarkZuber: will verify and report back.

src/GitVersionCore/SemanticVersionFormatValues.cs
MarkZuber: We need this class in order to generate the VersionVariables class since the types we're using int he formatting need to come from a constructed object.  And if we exposed the VersionVariables object as what's valid for formatting, then someone could do {InformationalVersion} and get themselves into an infinite loop.  So this gives us type safety of what's allowed to be formatted as well as getting the formatting done to construct the VersionVariables type.  I agree it's redundant on most fields (since VersionVariables now just becomes a passthrough/holder for most of the values calculated in the SemanticVersionFormatValues class.  Open to suggestions to minimize here if you have them.

src/GitVersionCore/GitVersionCore.csproj
MarkZuber: StringFormatWith.cs needs them for the DataBinder class which can throw HttpException.  This is using the code you recommended to do the formatting, which definitely works well.  Another option might be to use C#6 string format inlining e.g. $"{foo} {bar}" instead of string.Format("{0} {1}", foo, bar).  But i wasn't sure you wanted to require C#6 semantics for the code to compile.

try
            {
                if (string.IsNullOrEmpty(format))
                {
                    return (DataBinder.Eval(source, expression) ?? "").ToString();
                }
                return DataBinder.Eval(source, expression, "{0:" + format + "}");
            }
            catch (HttpException)
            {
                throw new FormatException("Failed to format '" + expression + "'.");
            }

MarkZuber:
Ping...  I think i've answered/fixed all of the issues requested on the feedback.  Would love to get this merged so we can pick up an official drop to use on our build servers.  thanks!

PR #661, opened 2015-10-04
Feature/659 fix ip should not be matched as versions

### LINKED ISSUE #659: 'Version from merge commit' may consider ip from remote-url
Funny. A merge commit resulting from a standard pull for a remote url specified by ip resulted in a version derived from the ip.

The message of the merge commit was something like

    Merge branch 'develop' of http://10.0.6.3/gitblit/r/.... into develop

This resulted in version `10.1.0` reported on a feature branch.

Interestingly, i tagged with `v1.1.0` two commits behind ans was expecting this version to be reported. I was clearly not aware of the merge commit specified above which was 330 commits behind. 

So this took me a while to figure out. The buildserver output helped then and made it crystal clear where this was coming from.

mkoertgen:
Fixes #659

### DIFF
--- a/src/GitVersionCore.Tests/GitVersionCore.Tests.v2.ncrunchproject
+++ b/src/GitVersionCore.Tests/GitVersionCore.Tests.v2.ncrunchproject
@@ -22,6 +22,6 @@
   <UseCPUArchitecture>AutoDetect</UseCPUArchitecture>
   <MSTestThreadApartmentState>STA</MSTestThreadApartmentState>
   <BuildProcessArchitecture>x86</BuildProcessArchitecture>
-  <AdditionalFilesToInclude>..\Packages\LibGit2Sharp.0.19.0.0\lib\net40\NativeBinaries\**.*;Resources\**.*</AdditionalFilesToInclude>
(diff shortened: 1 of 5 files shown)


gep13:
I recently ran into this same issue on a repository at work, so I would be keen to get this merged in.

@JakeGinnivan what are your thoughts?  The timeout issue that @mkoertgen mentioned here: https://github.com/GitTools/GitVersion/issues/659#issuecomment-145349745 concerns me a little bit.

JakeGinnivan:
I've never seen it timeout, but the logs have timing info in them, so should be easy to spot what I'd taking the time.

Other than that, pr looks good.

Sent from my Windows Phone

gep13:
@mkoertgen can you provide the logs for a GitVersion run where there is a timeout?

mkoertgen:
Sorry, i did not mean to mention a timeout, just timing infos with `GitVersion /output buildserver`

JakeGinnivan:
This will not be a performance bottleneck in comparison to other things. Negative lookbehinds can be slow, but the strings we have are small and compared to some of the git operations we do, it will be negligible.

gep13:
:+1: I will try to give the new version a try this morning on the repo that was giving me problems, and will report back as well

gep13:
:+1: this is now working on the internal repo that was exhibiting the same behaviour :smile_cat:

PR #665, opened 2015-10-05
Update build version from VSO Agent task.

### LINKED ISSUE #663: Set VSO Build Number in Task
See https://github.com/Microsoft/vso-agent-tasks/commit/815986b0f28ae56a522943127c3a18c016e0666b

The command is now available. Need to remove the note added in #577 and update docs as well.

pascalberger:
Fixes #663 

VSO task always updates the build version, as does GitVersion in TeamCity. Updated docs and testcases too.


pascalberger:
I assume the failing check is a problem with AppVeyor and not with the code. it also succeeded in an earlier version with the same code (I first added another commit, then later rested it back).

PR #669, opened 2015-10-06
Allow building on Linux

ermshiperete:


ermshiperete:
This change should really go on the release/4.0.0 branch - would work on master as well, but it's better to put it on the branch that potentially will work with Mono. I'll resubmit a new PR and close this one.

PR #671, opened 2015-10-06
Allow building and running on Linux

ermshiperete:


src/GitVersionExe/ArgumentParser.cs
```diff
@@ -52,7 +54,7 @@ public static Arguments ParseArguments(List<string> commandLineArguments)
 
             List<string> namedArguments;
             var arguments = new Arguments();
-            if (firstArgument.StartsWith("-") || firstArgument.StartsWith("/"))
+            if (firstArgument.StartsWith("-") || (firstArgument.StartsWith("/") && Path.DirectorySeparatorChar != '/'))
```
JakeGinnivan: FYI we are probably going to redo the command line args in 4.0.0, which should help cross plat out as we will probably drop `/` arguments

JakeGinnivan:
Thanks! Just to set expectations, it will likely be another month or so before a 4.0.0 beta is out. I want to get the command line stuff done before I release the beta and I have quite a few speaking engagements coming up.

ermshiperete:
@JakeGinnivan Any idea when a 4.0.0 will be released? IIRC one problem was the need for an updated LibGit2Sharp version which has been released a few weeks now. Would be great if we could get an official 4.0.0 soon... Thanks!

asbjornu:
@ermshiperete I've began tackling that problem in #816, which should be possible to merge into `master` and release as a v3.x. If I've understood @JakeGinnivan correctly, the plan is to merge the `v4.0` branch in its current state into `master` and let the main point of `v4.0` be about breaking changes on the command line.

#816 still needs a bit of work, so I'd highly appreciate any input you might have on that, @ermshiperete, so we can have it merged into `master` as well.

PR #678, opened 2015-10-08
Add support for relative logfile paths

asbjornu:
This changes the handling of `Arguments.LogFilePath` such that it can be relative as well as absolute. It also improves what is logged to the console when the log file can't be created.

### DIFF
--- a/src/GitVersionExe/Program.cs
+++ b/src/GitVersionExe/Program.cs
@@ -113,6 +113,7 @@ namespace GitVersion
             return 0;
         }
 
+
         static void ConfigureLogging(Arguments arguments)
         {
             var writeActions = new List<Action<string>>
@@ -132,11 +133,11 @@ namespace GitVersion
                     var logFileFullPath = Path.GetFullPath(arguments.LogFilePath);
                     var logFile = new FileInfo(logFileFullPath);
 
-                    if (logFile.Directory == null)
-                        throw new DirectoryNotFoundException(String.Format("The directory of {0} does not exist.", logFile));
-
-                    if (!logFile.Directory.Exists)
+                    // NOTE: logFile.Directory will be null if the path is i.e. C:\logfile.log. @asbjornu
+                    if (logFile.Directory != null)
+                    {
                         logFile.Directory.Create();
+                    }
 
                     if (!logFile.Exists)
                     {
--- a/src/GitVersionExe/Program.cs
+++ b/src/GitVersionExe/Program.cs
@@ -160,7 +160,7 @@ namespace GitVersion
         static void WriteLogEntry(Arguments arguments, string s)
         {
             var contents = string.Format("{0}\t\t{1}\r\n", DateTime.Now.ToString("yyyy-MM-dd HH:mm:ss"), s);
-            File.WriteAllText(arguments.LogFilePath, contents);
+            File.AppendAllText(arguments.LogFilePath, contents);
         }
 
         static List<string> GetArgumentsWithoutExeName()
--- a/src/GitVersionExe/Program.cs
+++ b/src/GitVersionExe/Program.cs
@@ -132,11 +132,11 @@ namespace GitVersion
                     var logFileFullPath = Path.GetFullPath(arguments.LogFilePath);
                     var logFile = new FileInfo(logFileFullPath);
 
-                    if (logFile.Directory != null && !logFile.Directory.Exists)
-                    {
-                        // TODO: This should probably be done recursively. @asbjornu
+                    if (logFile.Directory == null)
+                        throw new DirectoryNotFoundException(String.Format("The directory of {0} does not exist.", logFile));
+
+                    if (!logFile.Directory.Exists)
                         logFile.Directory.Create();
-                    }
 
                     if (!logFile.Exists)
                     {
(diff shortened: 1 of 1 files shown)


src/GitVersionExe/Program.cs
```diff
@@ -129,17 +129,27 @@ static void ConfigureLogging(Arguments arguments)
...
+                        {
+                        }
                     }
 
                     writeActions.Add(x => WriteLogEntry(arguments, x));
                 }
                 catch (Exception ex)
                 {
-                    Console.WriteLine("Failed to configure logging: " + ex.Message);
+                    Console.WriteLine("Failed to configure logging for '{0}': {1}", arguments.LogFilePath, ex.Message);
```
orjan: `Logger.WriteWarning` would be better than using `Console.WriteLine`

src/GitVersionExe/Program.cs
```diff
@@ -129,17 +129,27 @@ static void ConfigureLogging(Arguments arguments)
             {
                 try
                 {
-                    Directory.CreateDirectory(Path.GetDirectoryName(arguments.LogFilePath));
-                    if (File.Exists(arguments.LogFilePath))
+                    var logFileFullPath = Path.GetFullPath(arguments.LogFilePath);
+                    var logFile = new FileInfo(logFileFullPath);
+
+                    if (logFile.Directory != null && !logFile.Directory.Exists)
```
orjan: We should probably throw if `logFile.Directory == null` and then just call `logFile.Directory.Create();` no need for recursion in this case.

src/GitVersionExe/Program.cs
asbjornu: I agree, I just didn't want to do anything about that in this (old) part of the code.

src/GitVersionExe/Program.cs
asbjornu: True.

src/GitVersionExe/Program.cs
```diff
@@ -129,17 +129,27 @@ static void ConfigureLogging(Arguments arguments)
             {
                 try
                 {
-                    Directory.CreateDirectory(Path.GetDirectoryName(arguments.LogFilePath));
-                    if (File.Exists(arguments.LogFilePath))
+                    var logFileFullPath = Path.GetFullPath(arguments.LogFilePath);
+                    var logFile = new FileInfo(logFileFullPath);
+
+                    if (logFile.Directory == null)
```
orjan: Would you mind adding {}

src/GitVersionExe/Program.cs
```diff
@@ -129,17 +129,27 @@ static void ConfigureLogging(Arguments arguments)
...
+                        throw new DirectoryNotFoundException(String.Format("The directory of {0} does not exist.", logFile));
+
+                    if (!logFile.Directory.Exists)
```
orjan: I would remove `logFile.Directory.Exists` here since it's not needed and you cannot be sure if the directory is there in the next statement.

JakeGinnivan: `logFile.Directory.Create();` doesn't throw if the directory doesn't exist. Can just remove the if

src/GitVersionExe/Program.cs
JakeGinnivan: I don't mind not having the {}, but is the exception correct? That contradicts the code below.

asbjornu: Thanks for pointing this out. After some investigation, it seems like for paths like `C:\GitVersion.log`, `FileInfo.Directory` will be `null`. In such circumstances, the file should still be possible to create and it is indeed not a failure scenario to place the file on root. I'll rework the code.

src/GitVersionExe/Program.cs
```diff
@@ -129,17 +130,27 @@ static void ConfigureLogging(Arguments arguments)
...
+                    var logFileFullPath = Path.GetFullPath(arguments.LogFilePath);
+                    var logFile = new FileInfo(logFileFullPath);
+
+                    // NOTE: logFile.Directory will be null if the path is i.e. C:\logfile.log. @asbjornu
+                    if (logFile.Directory != null)
+                    {
+                        logFile.Directory.Create();
+                    }
+
+                    if (!logFile.Exists)
```
orjan: @asbjornu we'll possible hit some regression due to this if-statement, since it means that we're appending the log instead of overwriting the file as we're doing in the current `master`

JakeGinnivan: I wouldn't worry about the squash. I don't mind seeing the process people take to get to the end solution. It is just forward merges which I like to see rebased out.

Lets just remove the if statement around this. Then it will either create it or empty the log file?

Or invert the if to put back to the old logic

asbjornu: @JakeGinnivan, @orjan Yea, I can remove the `if`. But given [line 155 in `master`](/GitTools/GitVersion/blob/master/src/GitVersionExe/Program.cs#L155):

```c#
File.AppendAllText(arguments.LogFilePath, contents);
```
I find the old behavior of replacing the file weird. Wouldn't this make the code a bit more intuitive?

```c#
File.WriteAllText(arguments.LogFilePath, contents);
```

orjan:
@asbjornu there're 2 failing log related tests, would you mind fixing them?

asbjornu:
@orjan Yea, I noticed. Not sure why, but I reverted a change from 7aea8ec3d9f9455c57cd8decd697194d42fb5f82 that should make the tests pass again.

PR #682, opened 2015-10-11
Added test and fix for scenario when branch-name contains slashes. Fixes #569.

### LINKED ISSUE #569: Build Metadata for branch includes invalid characters from branch name
When bringing in the "InformationalVersion" number into a class that performs validation on the semver as per semver.org rules, it fails if we have a branch that has anything other than [a-zA-Z0-9\.\-]. 

Often branches are of the form feature/a feature/b where the forward slash is used as a folder indicator. There could be other invalid characters in git branch names.

Can I suggest that the BranchName property should be left untouched but the InformationalVersion number should be cleaned so we can use it directly. All non alphanumeric chars should be changed to be -'s and then it will always be a valid full semver.

Currently I apply this workaround in Msbuild but better if it did this in the tool directly I think?

Spiralis:
This fixes the scenario for issue #569 where multiple slashes in the branch-name (which is perfectly legal for branch-names, but probably should be avoided in the version-name (see issue for discussion).
Note: Added a fix too for the project-references where the the Visualizer.Fody references were off.

### DIFF
--- a/src/GitVersionCore/GitVersionCore.csproj
+++ b/src/GitVersionCore/GitVersionCore.csproj
@@ -52,8 +52,8 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
-    <Reference Include="Visualize, Version=0.4.3.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Visualize.Fody.0.4.3.0\Lib\portable-net4+sl4+wp7+win8+MonoAndroid16+MonoTouch40\Visualize.dll</HintPath>
+    <Reference Include="Visualize, Version=0.4.4.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Visualize.Fody.0.4.4.0\lib\portable-net4+sl4+wp7+win8+MonoAndroid16+MonoTouch40\Visualize.dll</HintPath>
       <Private>False</Private>
     </Reference>
     <Reference Include="YamlDotNet, Version=3.6.0.0, Culture=neutral, processorArchitecture=MSIL">
--- a/src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs
+++ b/src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs
@@ -61,8 +61,12 @@
             if (tagToUse == "useBranchName")
             {
                 Logger.WriteInfo("Using branch name to calculate version tag");
-                var name = branchNameOverride ?? context.CurrentBranch.Name;
-                tagToUse = name.RegexReplace(context.Configuration.BranchPrefixToTrim, string.Empty, RegexOptions.IgnoreCase);
+                tagToUse = branchNameOverride ?? context.CurrentBranch.Name;
+                if (!string.IsNullOrWhiteSpace(context.Configuration.BranchPrefixToTrim))
+                {
+                    tagToUse = tagToUse.RegexReplace(context.Configuration.BranchPrefixToTrim, string.Empty, RegexOptions.IgnoreCase);
+                }
+                tagToUse = tagToUse.RegexReplace("[^a-zA-Z0-9-]", "-");
             }
             int? number = null;
             if (!string.IsNullOrEmpty(context.Configuration.TagNumberPattern))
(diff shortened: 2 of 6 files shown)


pascalberger:
Since several commits only fix others, it might help if you could squash the commits together

src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs
```diff
@@ -61,8 +61,12 @@ void UpdatePreReleaseTag(GitVersionContext context, SemanticVersion semanticVers
             if (tagToUse == "useBranchName")
             {
                 Logger.WriteInfo("Using branch name to calculate version tag");
-                var name = branchNameOverride ?? context.CurrentBranch.Name;
-                tagToUse = name.RegexReplace(context.Configuration.BranchPrefixToTrim, string.Empty, RegexOptions.IgnoreCase);
+                tagToUse = branchNameOverride ?? context.CurrentBranch.Name;
+                if (!string.IsNullOrWhiteSpace(context.Configuration.BranchPrefixToTrim))
```
pascalberger: Why the change to this logic since this shouldn't be related to your issue (only the RegexReplace afterwards should be enough to make it work). Or does this fix another issue?

src/GitVersionCore/GitVersionCore.csproj
```diff
@@ -52,8 +52,8 @@
     <Reference Include="Microsoft.CSharp" />
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
-    <Reference Include="Visualize, Version=0.4.3.0, Culture=neutral, processorArchitecture=MSIL">
-      <HintPath>..\packages\Visualize.Fody.0.4.3.0\Lib\portable-net4+sl4+wp7+win8+MonoAndroid16+MonoTouch40\Visualize.dll</HintPath>
+    <Reference Include="Visualize, Version=0.4.4.0, Culture=neutral, processorArchitecture=MSIL">
+      <HintPath>..\packages\Visualize.Fody.0.4.4.0\lib\portable-net4+sl4+wp7+win8+MonoAndroid16+MonoTouch40\Visualize.dll</HintPath>
```
pascalberger: Don't know if changing casing of lib might lead to problems under Mono. @JakeGinnivan do you know?

Spiralis: This change was a result of the package installation itself. Not me manually changing the line.

src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs
Spiralis: I could have made my change to work on the `name` variable that was used, but since a regex invoke when this often is an empty string I found that this perhaps should be optimized at the same time. So that the regex replace is done only when needed. But, would work without it I guess.

Spiralis:
I can squash them. I just wasn't aware of that it was the preferred way.

src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs
pascalberger: OK, makes sense

Spiralis:
Sorry. I am experiencing some issues while trying to squash. Let me try top clean this...

Spiralis:
Commits squashed. Still kept test and fix as two different commits. Not sure if that is what you want or if I should squash them too?

gep13:
Separate, distinct, commits makes perfect sense. Also, if there is an open issue for this, if you can link back to it in commit message, that would help as well.

Sent from my Windows Phone

src/GitVersionCore/GitVersionCore.csproj
JakeGinnivan: Fody is fine

JakeGinnivan:
@pascalberger I am happy without the squash. It is just the forward merge commits which I tend to ask for a rebase, otherwise I don't worry

Spiralis:
Guys, is there a message-board or a forum somewhere to discuss features in this project? I have been looking for something like that, but can't find any references to something like that?

orjan:
@Spiralis https://gitter.im/GitTools/GitVersion

Spiralis:
@orjan, thanks :)

gep13:
@Spiralis I see you have found the chat room.  It is listed on the readme.md for this project:

https://github.com/GitTools/GitVersion

Is this not discoverable?  Do we need to put it somewhere else?

Spiralis:
Oh. I see it now. It kinda gets lost when it is next to the build-status message. 
I suggest maybe adding it to the quick-links section. Maybe also add a note in the contribution section, possibly in the FAQ.

gep13:
@Spiralis we would be happy to accept a PR to do this :stuck_out_tongue:

PR #685, opened 2015-10-12
Restructure build server documentation

pascalberger:
Put build server documentation together in a single node in the documentation. Extend documentation with a note about build server specific wrappers around GitVersion.exe.


JakeGinnivan:
Can you rebase?

Issue #646, opened 2015-09-16
GitVersionTask 3.1.2 tries to fetch on VsoAgent


onovotny:
On my CI server on VSO, I use both the GitVersionVsoTask and the GitVersionTask for MSBuild. The reason is that I want local builds to have assembly info's updated. 

For now, I get the following warnings when the git version task runs on the VSO agent. I don't think it's getting a nofetch param:

```
3>MSBUILD : warning : WARN [09/16/15 8:31:47:34] Could not determine assembly version: Request failed with status code: 401 [f:\a1\91fa088d\MT\ViewModels\ViewModels.csproj]

348 
2015-09-16T12:31:47.3561791Z          INFO [09/16/15 8:31:47:34] CacheKey: f:\a1\91fa088d\MT\.git:refs/heads/release-3.0.0:be52ac789210c8f72a3943d078bef9726775b1e9

349 
2015-09-16T12:31:47.3561791Z          INFO [09/16/15 8:31:47:34] Version not in cache. Calculating version.

350 
2015-09-16T12:31:47.3561791Z          INFO [09/16/15 8:31:47:34] Applicable build agent found: 'VsoAgent'.

351 
2015-09-16T12:31:47.3561791Z          INFO [09/16/15 8:31:47:34] Branch from build environment: refs/heads/release-3.0.0

352 
2015-09-16T12:31:47.3561791Z          INFO [09/16/15 8:31:47:34] One remote found (origin -> 'https://.....visualstudio.com/DefaultCollection/....').

353 
2015-09-16T12:31:47.3561791Z          INFO [09/16/15 8:31:47:34] Fetching from remote 'origin' using the following refspecs: +refs/heads/*:refs/remotes/origin/*.

354 
2015-09-16T12:31:47.5124075Z ##[warning]WARN [09/16/15 8:31:47:49] Could not determine assembly version: Request failed with status code: 401

355 
2015-09-16T12:31:47.5124075Z      3>MSBUILD : warning : WARN [09/16/15 8:31:47:49] Could not determine assembly version: Request failed with status code: 401 [f:\a1\91fa088d\MT\ViewModels\ViewModels.csproj]
```

PR #687, opened 2015-10-12
Conditional ms build tasks

pi3k14:
Did a retake on this pull request - allows the user to conditionally disable GitVersions MSBuild tasks

Default behavior is to run all task (as before)

### DIFF
--- a/src/GitVersionTask/NugetAssets/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/GitVersionTask.targets
@@ -4,31 +4,39 @@
     <SolutionDir Condition="$(SolutionDir) == '' Or $(SolutionDir) == '*Undefined*'">$(MSBuildProjectDirectory)..\</SolutionDir>
     <IntermediateOutputPath Condition="$(IntermediateOutputPath) == '' Or $(IntermediateOutputPath) == '*Undefined*'">$(MSBuildProjectDirectory)obj\$(Configuration)\</IntermediateOutputPath>
     <GitVersion_NoFetchEnabled Condition="$(GitVersion_NoFetchEnabled) == ''">false</GitVersion_NoFetchEnabled>
+
+    <!-- Property that enables WriteVersionInfoToBuildLog -->
+    <WriteVersionInfoToBuildLog Condition=" '$(WriteVersionInfoToBuildLog)' == '' ">true</WriteVersionInfoToBuildLog>
+
+    <!-- Property that enables UpdateAssemblyInfo -->
+    <UpdateAssemblyInfo Condition=" '$(UpdateAssemblyInfo)' == '' ">true</UpdateAssemblyInfo>
+
+    <!-- Property that enables GetVersion -->
+    <GetVersion Condition=" '$(GetVersion)' == '' ">true</GetVersion>
+
   </PropertyGroup>
 
   <UsingTask
       TaskName="GitVersionTask.UpdateAssemblyInfo"
       AssemblyFile="$(MSBuildThisFileDirectory)..\..\GitVersionTask.dll" />
-  <UsingTask 
-    TaskName="GitVersionTask.GetVersion" 
+  <UsingTask
+    TaskName="GitVersionTask.GetVersion"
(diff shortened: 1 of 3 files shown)


src/GitVersionTask/NugetAssets/GitVersionTask.targets
```diff
@@ -4,31 +4,39 @@
...
-      NoFetch="$(GitVersion_NoFetchEnabled)"
-        />
+  <Target Name="WriteVersionInfoToBuildLog" BeforeTargets="CoreCompile" Condition="$(WriteVersionInfoToBuildLog) == 'true'">
+    <WriteVersionInfoToBuildLog SolutionDirectory="$(SolutionDir)" />
+  </Target>
+  
+  <Target Name="UpdateAssemblyInfo" BeforeTargets="CoreCompile" Condition="$(UpdateAssemblyInfo) == 'true'">
     <UpdateAssemblyInfo
-      SolutionDirectory="$(SolutionDir)"
-      NoFetch="$(GitVersion_NoFetchEnabled)"
```
pascalberger: Have you removed NoFetch intentionally? This was introduced to fix #646

src/GitVersionTask/NugetAssets/GitVersionTask.targets
```diff
@@ -4,31 +4,39 @@
...
     AssemblyFile="$(MSBuildThisFileDirectory)..\..\GitVersionTask.dll"  />
   <UsingTask
       TaskName="GitVersionTask.WriteVersionInfoToBuildLog"
       AssemblyFile="$(MSBuildThisFileDirectory)..\..\GitVersionTask.dll" />
 
-  <Target Name="UpdateAssemblyInfo"
-          BeforeTargets="CoreCompile">
-    <WriteVersionInfoToBuildLog
-      SolutionDirectory="$(SolutionDir)"
-      NoFetch="$(GitVersion_NoFetchEnabled)"
```
pascalberger: Have you removed NoFetch intentionally? This was introduced to fix #646

src/GitVersionTask/NugetAssets/GitVersionTask.targets
pi3k14: No - it's a diff I didn't catch when creating a patch from my "messy" repo.

pi3k14:
Added NoFetch back (and if someone wonder, I am not going to touch git -rebase to make this any "prettier" ;)

src/GitVersionTask/NugetAssets/GitVersionTask.targets
pi3k14: An oversight - fixed

pascalberger:
LGTM. The only thing I'm not sure is, that with now three individual targets, all dependent on the same target, is it ensured that they are always called in the correct order?

pi3k14:
I didn't think there was a dependency between this targets, if so this PR breaks up. The intention was to switch of the tasks you didn't want (UpdateAssemblyInfo in my case, because I want to handle AssemblyVersion, AssemblyFileVersion and AssemblyInformationalVersion individually, and so far I haven't had any problems with it).

This solution only assure that the tasks are run before CoreCompile, in no particular order.

If there is a dependency between them, I can't switch them off at will. Do you have any further information about this?

pascalberger:
@JakeGinnivan Do you know if the execution order is an issue here?

JakeGinnivan:
Not sure.. @SimonCropp ?

pi3k14:
The original order was; WriteVersionInfoToBuildLog, UpdateAssemblyInfo and GetVersion.

WriteVersionInfoToBuildLog cant be dependent on any of the other task. 
I have run a great deal of builds with only GetVersion, and it works without any problems (also tested on new 1.89 build agent on TFS, RC1 version).

The remaining questions then are - is UpdateAssemblyInfo dependent on WriteVersionInfoToBuild Log, would running WriteVersionInfoToBuildLog after the other tasks creating havoc?
To me, that would make no sense.

JakeGinnivan:
This is an area of GitVersion I don't know very well so it all looks fine to me. @pascalberger if you can't see anything, lets merge

PR #691, opened 2015-10-14
Ignore the build folder

asbjornu:
I'm not sure I understand why the `build` folder shouldn't be ignored.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,4 @@ NuGetExeBuild/
 NuGetRefBuild/
 Packages/
 NuGetTaskBuild/
+build/


.gitignore
```diff
@@ -90,3 +90,4 @@ NuGetExeBuild/
 NuGetRefBuild/
 Packages/
 NuGetTaskBuild/
+build/
```
orjan: There're 2 files in the build root so it's a little bit hard to ignore the whole folder.

asbjornu: @orjan Since those files are already committed, it shouldn't matter. But we can be explicit about it:

```
build/*
!build/Build.cmd
!build/Update-GitVersionTfsTaskVersion.ps1
```

orjan: It's still a little bit secret if someone adds a new file, but I don't have a strong opinion about it. One way would be to have all artifacts in a separate folder .e.g. .artifacts/ but that's probably overkill in this case. Another option is just to exclude the new folder new folder.

asbjornu: With the explicit excludes, it at least explains how adding a new one would work. Explicitly adding ignored files still work, too. I thus don't feel this is a bigger issue than always having new, untracked files after a build.

JakeGinnivan: Maybe we should move those two files out of the build folder. Or make everything compile into artifacts or something which can be ignored?

gep13: I personally normally have a `BuildScripts` folder and a `BuildArtifacts` folder.  So I would prefer to see something like that.  Just my 2 cents...

asbjornu: @JakeGinnivan @gep13 I honestly don't see the problem with my proposed solution. The files we want to track inside the `build` folder are already tracked by Git and won't stop being tracked just because we start ignoring the `build` folder. If we want to ensure that it never happens in the future, we can explicitly un-ignore the files with `!`.

gep13: @asbjornu from my point of view, there is nothing wrong with your solution, but rather, there is something wrong with those files being in there in the first place.  If the `build` folder is intended only for items that are generated as part of the build process, then there shouldn't be any other files/folders in there initially.  This just leads to ambiguity.  For instance, normally in my build script, I will `clean` my artifacts folder, but in this case, I would be deleting files that I shouldn't have, with is not a good place to be.

pascalberger: +1 for separate BuildScripts and BuildArtifacts folders

JakeGinnivan:
Gonna close this, opened #719 to track the suggested change

PR #677, opened 2015-10-08
FindCommitBranchWasBranchedFrom should not throw NullReferenceException

asbjornu:
This makes sure #618 and #674 are handled in a more graceful way, by throwing more intuitive exceptions than `NullReferenceException` when a branch has no tip. Before this is merged, it would be good to replace the `example.com` documentation URI with proper documentation explaining how to fix this problem in the repository.

PR #696, opened 2015-10-17
Performance fix

JakeGinnivan:
// @asbjornu @orjan can you guys test this with `commit-message-incrementing: Enabled`. Hopefully it should fix the perf issues we talked about in #677

### DIFF
--- a/src/GitVersionCore/IncrementStrategyFinder.cs
+++ b/src/GitVersionCore/IncrementStrategyFinder.cs
@@ -83,7 +83,7 @@
         {
             if (baseCommit == null) yield break;
 
-            if (intermediateCommitCache == null)
+            if (intermediateCommitCache == null || intermediateCommitCache.Last() != currentBranch.Tip)
             {
                 var filter = new CommitFilter
                 {
--- a/src/GitVersionCore/IncrementStrategyFinder.cs
+++ b/src/GitVersionCore/IncrementStrategyFinder.cs
@@ -54,7 +54,7 @@
                 return null;
             }
             
-            var commits = GetIntermediateCommits(context.Repository, baseVersion.BaseVersionSource, context.CurrentCommit, context.CurrentBranch);
+            var commits = GetIntermediateCommits(context.Repository, baseVersion.BaseVersionSource, context.CurrentCommit);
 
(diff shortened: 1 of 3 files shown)


JakeGinnivan:
Crap, there are not tests around the commit message stuff. I broke it. Need to fix.

JakeGinnivan:
Ok, should be fixed. If either of you can test with one of your repos which caused the perf issues that would be awesome.

orjan:
This weekend is pretty stuffed for me but will give it a go at Monday the latest!

orjan:
It's looking good! From 13s to 10s running latest master against this pull request, see snapshot diff below:
image

src/GitVersionCore/IncrementStrategyFinder.cs
```diff
@@ -80,14 +81,28 @@ public static class IncrementStrategyFinder
...
+            }
+
+            var found = false;
+            foreach (var commit in intermediateCommitCache)
             {
-                Since = headCommit,
-                Until = baseCommit,
-                SortBy = CommitSortStrategies.Topological | CommitSortStrategies.Reverse
-            };
+                if (commit.Sha == baseCommit.Sha)
```
orjan: @JakeGinnivan why are we comparing the `Sha` instead of the commit? We're using `Sha` for comparison over the whole code base so I'll guess it's the way to go, but I'm just curious about why?

JakeGinnivan: No reason, I assume libgit2sharp does implement equal and equality operators properly but I haven't checked.

asbjornu:
@JakeGinnivan Sorry, I haven't had time to test this until now. But yea, this is a major improvement. 

**Before:**

```
2015-10-19 08:14:13 INFO [10/19/15 8:14:13:18] End: Calculating base versions (Took: 24,395.41ms)
````

**After:**

```
2015-10-19 08:15:24 INFO [10/19/15 8:15:24:99] End: Calculating base versions (Took: 5,154.25ms)
```

JakeGinnivan:
Still would be good to make better, but that is a pretty good start.

PR #701, opened 2015-10-20
Add property to always expose commits since last version

onovotny:
There are many places where there's a need for an incrementing number per version. This can be thought of as the "build number", but that term is overloaded. Instead, I just called it CommitsSinceLastVersion to let users do what they need with it.

In our case, I'm using it as the "Patch" version for a continuous deployment scenario where semVer isn't supported -- AppxManifest in the store. I need something like 2.0.101.0. I can't use the continuous deploy mechanism as these are actually beta's. The FullSemVer would be 2.0.0-beta4. If I use CD, then it'd be 2.0.0-beta101 and that's not what I want. I don't really want to update the beta per build. It's similar to wanting 2.0.0-beta4-build101, but I'm ok with the number coming from 2.0.0 and not reseting at "beta 4".


pascalberger:
@onovotny: Could you please also document the new variable in [variables.md](https://github.com/GitTools/GitVersion/blob/master/docs/more-info/variables.md).

Otherwise it LGTM, but I would prefer to have @JakeGinnivan a look at this before merging.

onovotny:
@pascalberger updated

gep13:
Just one comment...

Does CommitsSinceLastVersion correctly described the intention?  Would CommitsSinceLastRelease, CommitsSinceLastReleasedVersion not be more descriptive of the intention?

For instance, in my workflow, I don't tag until I release a version to "somewhere".  I could be barking up the wrong tree, but thought I would put it out there.

asbjornu:
I agree with @gep13. The word `Version` is very conflated and ambiguous. `CommitsSinceLastTag` or `CommitsSinceLastRelease` more accurately describes what it is, imho.

onovotny:
I'm open to changing it, naming is hard!

Given how many touchpoints it is though, I'd rather rename it only once before updating this PR, so we need a decision on the right name.

gep13:
>@onovotny said...
Given how many touchpoints it is though, I'd rather rename it only once before updating this PR, so we need a decision on the right name.

Makes sense to me.

Options:

1. CommitsSinceLastVersion
2. CommitsSinceLastRelease
3. CommitsSinceLastReleasedVersion 
4. CommitsSinceLastTag 

Vote away....

My preference would be 2.

onovotny:
I think of those four, I'd go with 2 as well. GitVersion doesn't require tags, so 4 might be a red herring. 3 is just longer :). Between 1 and 2, I'm indifferent.

gep13:
Yes, although in my workflow, CommitsSinceLastTag makes things crystal clear, the adding of a Tag is just one way to assert the Version Number of your repository,  so it wouldn't work across all the options within GitVersion.

pascalberger:
4 is for me also no option for the mentioned reasons. Otherwise not really sure, not all versions might be released so there might be a difference between Version and Release.

onovotny:
Agree that not all versions might be considered "released," which is why I initially just went with "since last version".

gep13:
The problem I have with Version is... what version are you talking about?  i.e. 2.0.1-beta0001 is a version, but so is 2.0.1.

If my understanding is correct, the counter that you are adding in would give the commits between two "releases" of an application, i.e. 2.0.1 and 2.0.2, or 2.0.2 and 2.0.3.  If that is not the case, then I am definitely barking up the wrong tree.

Really the variable would need to be called something like:

CommitsSinceLastChangeInMajorMinorPatchOCalculatedSemanticVersion

but I think that might be a bit long :-D

onovotny:
In my case, I use 2.0.1-beta1 as the version and have n builds of it. So it'd be 2.0.1-beta1-build001.

That said, I believe the logic to do the commit count ignores pre-release tags. So when I eventually have 2.0.1-beta1-build102 and ship it, then I bump to 2.0.1-beta2. The commit count though would be 102, not back to 0 or 1, so the next build would be 2.0.1-beta2-build103 or some such. Ideally I'd like the commit counter to reset but for my needs, I'm ok with it resetting only between non-pre-releases.

gep13:
>@onovotny said...
then I bump to 2.0.1-beta2

so that I am clear, how exactly are you bumping this?

onovotny:
I'll do it one of two ways: either tag 2.0.1-beta2 or update the next-version in the config to be 2.0.1-beta2

onovotny:
I don't think there's any "automagic" way for GitVersion to know when to bump my beta versions as it depends on whatever I feel like releasing.

gep13:
ok, so not through a commit message or anything like that?

onovotny:
I've never used a commit message for releasing anything :)

gep13:
Hmm, will be interesting to hear @JakeGinnivan thoughts on this one.  I might well be over thinking it :smile:

JakeGinnivan:
The commit counting stuff is referred to internally as the VersionSource, and we take the oldest source to do commit counting (this means that it might not be exactly correct in some scenarios, but it means that the count doesn't reset randomly). This entire workflow is something I have not got my head around yet so feel free to point out things I have missed. 

@onovotny this is to start moving towards your `${semver}-build${commitsSinceVersionSource}` right?

The way I see this would work and the terms I think I would use to describe it. 

When using GitFlow I want to manually control the SemVer in a continuous delivery way but want each publish the NuGet package for each build. The SemVer will be the same for each build so I want to promote the `CommitsSinceVersionSource` from the build metadata to the `PrereleseTag` in the format `${tag}-build${commitsSinceVersionSource}`. Then I will use MyGet or Octopus deploy to rewrite the pre-release tag in my NuGet package to drop the `build${commitsSinceVersionSource}` when I promote a package. When a package is promoted/deployed into an environment I `tag` the built commit with the `SemVer` to cause a SemVer bump. 

Does the terminology/workflow sound right?

onovotny:
So given the naming choices, `CommitsSinceVersionSource` seems to be the best option. I'll update the PR to make it that.

As to the rest of it, yes, I believe exposing this property works towards `${semver}-build${commitsSinceVersionSource}`, but it's not going that far yet. As that will take more time (and has an effect on the versioning mode and `PreReleaseTag` itself, I'd save that for a second PR that uses the `CommitsSinceVersionSource` property.

After renaming, are you good with it being merged in?

docs/more-info/variables.md
```diff
@@ -22,7 +22,10 @@ For the `release/3.0.0` branch of GitVersion it shows:
   "BranchName":"release/3.0.0",
   "Sha":"28c853159a46b5a87e6cc9c4f6e940c59d6bc68a",
   "NuGetVersionV2":"3.0.0-beta0001",
-  "NuGetVersion":"3.0.0-beta0001"
+  "NuGetVersion":"3.0.0-beta0001",
+  "CommitsSinceLastVersion":1,
+  "CommitsSinceLastVersionPadded":"0001",
+  "CommitDate":"2014-03-06"
```
JakeGinnivan: Is this the date of the head commit or the source version commit?

JakeGinnivan: Oh, this isn't new. Hrrmm

onovotny: Yeah, it wasn't new, just fixing it in the docs

JakeGinnivan:
Yeah, happy for it to be merged. As long as the version source thing makes sense to others, just making sure the internal name/concept translates well as an external concept/name

onovotny:
That's been the hard part of this....coming up with a name that conveys its meaning and doesn't overload other words.

asbjornu:
@JakeGinnivan @onovotny I think the term `VersionSource` perfectly explains what it is, without ambiguity. `CommitsSinceVersionSource` is therefore a great name for this variable. :+1:

gep13:
Yip, totally agree, makes sense to me :+1:

pascalberger:
:+1:

PR #711, opened 2015-11-10
Cache version information to disk

asbjornu:
This PR adds support for caching the version information to a YAML file on disk, so successive builds without changes in Git can be sped up. In local tests, the build time is cut in half with this PR in place.

The cache file's name is keyed by the path to the `.git` directory, the canonical name of the repository's `HEAD`, the SHA of the repository's `HEAD` tip and the last modified date of the `.git/refs` directory. This should ensure that the file name changes every time you do something in Git that will affect the version number.

I've had to do quite a bit of refactoring to be able to test this functionality. Most mentionable are the static classes `GitDirFinder`, `RepositoryLoader` and `DirectoryDateFinder` which I've deleted and whose methods I've moved into `IFileSystem` so they can be mocked.

To simplify mocking, I've also added a reference to [NSubstitute](http://nsubstitute.github.io/). I hope this is OK. Mocking LibGit2Sharp's `IRepository` interface would otherwise be a huge [PITA](https://en.wiktionary.org/wiki/pain_in_the_ass).


asbjornu:
Hm, I wonder why that test breaks on AppVeyor, but not on my machine. I'll look into it.

asbjornu:
> GitVersion.WarningException : 0 remote(s) have been detected. When being run on a build server, the Git repository is expected to bear one (and no more than one) remote.

Sigh. A step in the right direction, at least. Adding a remote to the test repository, then!

asbjornu:
This is weird. No other test doing real repository work have any mention of any remotes. How are they working, while my tests fail? @JakeGinnivan do you know?

src/GitVersionCore.Tests/TestFileSystem.cs
```diff
@@ -1,52 +1,188 @@
...
+using System.Linq;
+using System.Text;
+
 using GitVersion.Helpers;
 
+using LibGit2Sharp;
+
+using NSubstitute;
+
 public class TestFileSystem : IFileSystem
```
JakeGinnivan: @GeertvanHorrik fyi. Maybe just make the file system stuff internal in Core. The needed abstractions will be different in each app

src/GitVersionCore.Tests/TestFileSystem.cs
```diff
@@ -1,52 +1,188 @@
...
+    }
+
+
+    public string TreeWalkForDotGitDir(string directory)
+    {
+        return directory;
+    }
+
+
+    public IRepository GetRepository(string gitDirectory)
```
JakeGinnivan: Why not use the Fixture stuff? Mocking out libgit2sharp stuff is annoying in the long run I think

src/GitVersionCore/Helpers/IFileSystem.cs
```diff
@@ -13,5 +15,10 @@ public interface IFileSystem
         void WriteAllText(string file, string fileContents);
         IEnumerable<string> DirectoryGetFiles(string directory, string searchPattern, SearchOption searchOption);
         Stream OpenWrite(string path);
+        Stream OpenRead(string path);
+        void CreateDirectory(string path);
+        long GetLastDirectoryWrite(string path);
+        string TreeWalkForDotGitDir(string directory);
```
JakeGinnivan: I think we should just use the libgit2sharp stuff for this, Repository.Discover. This method was created I think because we didn't realise that existed.

src/GitVersionTask.Tests/VersionAndBranchFinderTests.cs
```diff
@@ -0,0 +1,140 @@
+﻿#region License
```
JakeGinnivan: Remove license info

src/GitVersionTask.Tests/VersionAndBranchFinderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+            fixture.Repository.MakeACommit();
+            var vv = VersionAndBranchFinder.GetVersion(fixture.RepositoryPath, null, false, fileSystem);
+
+            vv.AssemblySemVer.ShouldBe("0.1.0.0");
+
+            vv.FileName.ShouldNotBeNullOrEmpty();
+
+            fileSystem.WriteAllText(vv.FileName, versionCacheFileContent);
+
+            // I would rather see that VersionAndBranchFinder was non-static and could be reinstantiated to
```
JakeGinnivan: I don't mind switching to instance

src/GitVersionTask.Tests/VersionAndBranchFinderTests.cs
```diff
@@ -0,0 +1,140 @@
...
+
+using NUnit.Framework;
+
+using Shouldly;
+
+[TestFixture]
+public class VersionAndBranchFinderTests
+{
+    [Test]
+    public void CacheFileExistsOnDisk()
```
JakeGinnivan: Does that caching only affect the task, or is it core?

src/GitVersionCore.Tests/TestFileSystem.cs
asbjornu: Yea, I hoped the abstraction would be useful, but I ended up not using it. Mocking with `static` classes and methods is nigh impossible.

src/GitVersionCore.Tests/TestFileSystem.cs
asbjornu: @JakeGinnivan Yea, I ended up using the fixture stuff, which is why I ask what is causing the following exception:

> GitVersion.WarningException : 0 remote(s) have been detected. When being run on a build server, the Git repository is expected to bear one (and no more than one) remote.

Why are none of the other tests using the fixture having this problem? I see no mention of any remotes.

src/GitVersionCore/Helpers/IFileSystem.cs
asbjornu: Nice. I've replaced all usage of `TreeWalkForDotGitDir()` with `Repository.Discover()`.

src/GitVersionTask.Tests/VersionAndBranchFinderTests.cs
asbjornu: Doh. ReSharper is a bit too helpful at times. :/

src/GitVersionTask.Tests/VersionAndBranchFinderTests.cs
asbjornu: I'd love to have this PR baked and merged before trying to tackle that, but I'll be sure to look at that sometime. :)

src/GitVersionTask.Tests/VersionAndBranchFinderTests.cs
asbjornu: Unfortunately, it only affect the task, since it's done in the VersionAndBranchFinder, which for some reason aren't used by the EXE.

asbjornu:
@JakeGinnivan I've refactored `VersionAndBranchFinder` to a non-`static` class now. And I'm still dumbfounded by why my tests are failing while other tests using the `RepositoryFixture` stuff don't.

asbjornu:
@JakeGinnivan I've refactored deleted `VersionAndBranchFinder`, moved its caching logic into `ExecuteCore` and replaced all usage of the former with the latter. I've also made `ExecuteCore` non-static. This should make the caching code available to all uses of GitVersion.

However; this, and probably other various changes I've made, are backwards incompatible for someone using GitVersion programmatically. It should not break anything integrating with GitVersionTask or `gitverison.exe`, though.

I'm therefore unsure if this warrants a major version increment or not. If it does, I can create a new pull request that preserves API backwards compatibility, but it will be difficult to make the caching work across the task and exe without breaking, since the code paths are different.

I'm also still unable to figure out why my tests are failing.

JakeGinnivan:
I probably will not get to this one this weekend, hopefully next weekend

If it does need a major version bump I will integrate into the v4 branch, otherwise I will manually merge into the current release branch. Will take me some time to review and get my head around all the changes again.

My speaking commitments are now all done so I have some free time again :)

asbjornu:
@JakeGinnivan Ok. Good to see your comments here again, I hope your talks have gone great! :) 

Besides the regular code review, what I need you to look at is why the tests are failing on AppVeyor and if the internal changes in method signatures and such is enough to warrant a major version bump. I can also bisect and do another PR without backwards incompatible changes, if you want to. But I think having one code path through both the `.exe` and task is so important that I'd rather just have the current state of the code merged one way or another.

Looking forward to reading your comments on the code! :smile:

src/GitVersionCore/ExecuteCore.cs
```diff
@@ -1,12 +1,164 @@
...
+                    cacheKey = BitConverter.ToString(sha1.ComputeHash(Encoding.UTF8.GetBytes(key))).Replace("-", "");
+                }
+
+                var cacheDir = Path.Combine(gitDir, "gitversion_cache");
+                // If the cacheDir already exists, CreateDirectory just won't do anything (it won't fail). @asbjornu
+                this.fileSystem.CreateDirectory(cacheDir);
+
+                var cacheFileName = string.Concat(Path.Combine(cacheDir, cacheKey), ".yml");
+                VersionVariables vv = null;
+                if (this.fileSystem.Exists(cacheFileName))
```
JakeGinnivan: Do we need some sort of cleanup now? Otherwise we will end up with potentially thousands of files in that directory?

Maybe add in a cleanup which deletes everything older than a week?

JakeGinnivan:
I am happy with this. I would like a `/noCache` switch just for debugging purposes and such. Ill merge this first though, then ill try and add it now

JakeGinnivan:
I am just going to play devils advocate on this PR now before I merge it.

What scenarios do you see where gitversion.exe will be used and the refs will not change. I only use gitversion on the build server and it gets triggered when the VCS changes, I can't see I will get a benefit from this PR.

Wondering if maybe caching on exe should be opt in, rather than opt out?

src/GitVersionCore/ExecuteCore.cs
asbjornu: Yes, that sounds like a good idea.

asbjornu:
@JakeGinnivan With `.exe`, I agree with you; the value isn't all that high. In that case, the value is in having less code fragmentation between the task and the exe, which means better test coverage and simplified maintenance. I don't see any harm in having cache on by default, though I agree a `/nocache` switch would be nice.

Caching on the Task should imho not be opt-in, since this speeds up the build quite significantly. I also can't see that this breaks anything other than not being backwards compatible on the code level. The tests of course need to be fixed for AppVeyor, though. I really need help to figure out what's going on there.

src/GitVersionCore/ExecuteCore.cs
JakeGinnivan: We could put the key into the file, rather than the filename?

JakeGinnivan:
``` yml
Branch:
  Key: ...
  SemVer: ..
  ...
```
??

asbjornu: @JakeGinnivan Sure, I can't see why not?

PR #714, opened 2015-11-11
Add version info if it does not exist

FunnyFry:
This pull request addresses issues #473 and #474. Both of these, along with the associated pull request to #473, don't seem to have received any traction since July.

I am using the GitVersion VSO build task and having this issue addressed allows us to not think about what the version info in the AssemblyInfo.cs files look like. Previously all of our projects had been using the GitVersionTask and no longer have the version info in them. I would prefer to not go back and add them and just continue forward using the VSO build task.

@JakeGinnivan

### DIFF
--- a/src/GitVersionExe.Tests/AssemblyInfoFileUpdateTests.cs
+++ b/src/GitVersionExe.Tests/AssemblyInfoFileUpdateTests.cs
@@ -1,10 +1,10 @@
-﻿using System;
-using System.IO;
-using GitVersion;
+﻿using GitVersion;
 using GitVersion.Helpers;
 using GitVersionCore.Tests;
 using NSubstitute;
 using NUnit.Framework;
+using System;
+using System.IO;
 
 [TestFixture]
 public class AssemblyInfoFileUpdateTests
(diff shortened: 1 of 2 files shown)


pascalberger:
@JakeGinnivan Do you still prefer to have this logic in core (as written in #474). Otherwise this LGTM.

asbjornu:
I already think the code is already fragmented enough between Task and exe; I would at least love it if more code was lifted into Core so testing was easier and reuse greater.

JakeGinnivan:
@FunnyFry what do you think, do you think that this could be moved into core so Task and Exe share the same update assembly info logic? This looks like a good step towards that, but if we could move entirely into core that would be an awesome win.

If you don't think it's possible or you don't have time, let us know and we will merge as this is a good first step

JakeGinnivan:
@FunnyFry just wondering on your thoughts on trying to remove this duplication before I continue.

csmager:
This is included in the 3.5.0 release and broke my build on AppVeyor.  I have a couple of legacy projects referenced that reference a 'shared' `AssemblyInfoCommon.cs` file that contains the attributes 

I appreciate that in the previous releases, this meant this files actually never had their version numbers updated by GitVersion... but by adding them into the one it *did* find, it created duplicate attributes.

I'm not sure if this should be logged as an issue (as it didn't actually work as you'd expect before, either), but thought I'd mention!

JakeGinnivan:
@csmager ill leave opening an issue to you, we probably can add some release notes or something about a possible breaking change?

You can now also specify the assembly info filename. Try `/updateassemblyinfo AssemblyInfoCommon.cs`
See http://gitversion.readthedocs.org/en/latest/usage/command-line/#inject-version-metadata-into-the-assembly

If this doesn't resolve your issue then open an issue so we can continue discussion options

csmager:
I think the only solution to have avoided my problem would be that GitVersion checks a project's `.csproj` file and then scans all files for these attributes.  I don't think you want to go there.  I think I'll leave opening an issue.

I think adding something to the release notes would be useful, I had to do a diff between versions, find the code, and look at the blame to find this PR to see it was deliberate!  Perhaps it might save someone else some time.

JakeGinnivan:
Yeah thanks @csmager 

Sometimes breaking changes like that get through, hopefully we can get it sorted soon enough

pkiers:
Hi,
This info may be very useful for you, you won't regret, just take a look at <http://develop.scoremovers.com/e2pkh>

In haste, p.kiers@home.nl

pkiers:
Hi,
I know you were looking for this, I found this stuff for you, take a look at <http://dream.sarahgreim.com/e2eumy>

p.kiers@home.nl

Issue #584, opened 2015-08-18
Documented commandlineswitch /updateassemblyinfofilename not implemented?


vegar:
I'm trying to specify the name of my assemblyinfo. The commandline help states that

> **AssemblyInfo updating**
>    `/updateassemblyinfo`
>                    Will recursively search for all 'AssemblyInfo.cs' files in the git repo and update them
>    `/updateassemblyinfofilename`
>                    Specify name of AssemblyInfo file. Can also /updateAssemblyInfo GlobalAssemblyInfo.cs as a shorthand

So basically, I should be able to do one of the following:

`>GitVersion.exe /updateassemblyinfo true /updateassemblyinfofilename SharedAssemblyInfo.cs`
`>GitVersion.exe /updateassemblyinfo /updateassemblyinfofilename SharedAssemblyInfo.cs`
`>GitVersion.exe /updateassemblyinfo SharedAssemblyInfo.cs`

But whenever I try to include the `/updateassemblyinfofilename` parameter, I get an error: 

> Failed to parse arguments: /updateassemblyinfofilename ....

Looking into the [code for the argument parser](https://github.com/GitTools/GitVersion/blob/7e94062b0b1fce8e19bebdb6b1e3e9670bf85bd5/src/GitVersionExe/ArgumentParser.cs#L154-L174), I can't see that the `updateassemblyinforfilename`-parameter is implemented at all.

That wouldn't be a large problem, except that the VsoTask is implemented in a way that it's impossible to use this form, since it uses a checkbox for specifying the `updateassemblyinfo`-parameter. If I try to override this by not checking the 'Update AssemblyInfo's'-checkbox and then specify `/updateassemblyinfo nameofassemblyinfo.cs` as an extra argument, the task fails since the argument is added twice:

> 2015-08-18T11:42:34.3961389Z Failed to parse arguments: C:\a\4386bae5\Finale /output buildserver /updateAssemblyInfo false /nofetch /updateAssemblyInfo SharedAssemblyInfo.cs
> 2015-08-18T11:42:34.3961389Z Could not parse command line parameter 'SharedAssemblyInfo.cs'.
 

So there are multiple possible bugs/shortages/inconsistencies here, depending how it *should* work:

**For the commandline tool:**
- [ ] `updateassemblyinforfilename`-argument not implemented in ArgumentParser.cs, or
- [ ] non-existing `updateassemblyinforfilename`-argumente documented in HelpWriter.cs

**And for the VsoTask:**
[...]

PR #717, opened 2015-11-15
Converted VSO Task to use PowerShell

johnkattenhorn:
Added option to update specific assembly file

### DIFF
--- a/src/GitVersionTfsTask/task.json
+++ b/src/GitVersionTfsTask/task.json
@@ -10,10 +10,10 @@
     "Build",
     "Release"
   ],
-  "demands": [],
+  "demands": [ ],
   "version": {
-    "Major": "0",
-    "Minor": "0",
+    "Major": "3",
+    "Minor": "4",
     "Patch": "0"
   },
   "minimumAgentVersion": "1.83.0",
@@ -35,6 +35,15 @@
       "helpMarkDown": "Whether to update versions in the AssemblyInfo files"
     },
     {
+      "name": "updateAssemblyInfoFilename",
+      "type": "string",
+      "label": "Update Assembly File",
+      "defaultValue": "",
+      "required": false,
+      "helpMarkDown": "Update versions in specified file",
+      "groupName": "additional"
+    },
+    {
       "name": "additionalArguments",
       "type": "string",
       "label": "Additional GitVersion.exe arguments",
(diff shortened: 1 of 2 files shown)


pascalberger:
Thanks! This was already quickly discussed in #584. Using of a nodejs script is preferred over a PS script for x-plat compatibility.

src/GitVersionTfsTask/task.json
```diff
@@ -10,10 +10,10 @@
     "Build",
     "Release"
   ],
-  "demands": [],
+  "demands": [ ],
   "version": {
-    "Major": "0",
-    "Minor": "0",
+    "Major": "3",
+    "Minor": "4",
```
pascalberger: You don't need to set the version number here. It'll be set at build time

src/GitVersionTfsTask/task.json
```diff
@@ -42,13 +51,22 @@
       "required": false,
       "helpMarkDown": "Additional arguments to send to GitVersion.exe",
       "groupName": "additional"
+    },
+    {
+      "name": "gitVersionPath",
```
pascalberger: What's the use case for this? Since we deliver GitVersion as part of the TFS Task, I don't see the reason to use any other version in a different path.

src/GitVersionTfsTask/task.json
johnkattenhorn: Sorry missed taking this out during testing, I'll fix this up.

src/GitVersionTfsTask/task.json
johnkattenhorn: To be honest I was copying something else I'd done before and thought I'd leave it to allow me to test a future version of GitVersion without having to redeploy VSO as it might disrupt the rest of my teams doing builds.

src/GitVersionTfsTask/task.json
```diff
@@ -35,6 +35,15 @@
       "helpMarkDown": "Whether to update versions in the AssemblyInfo files"
     },
     {
+      "name": "updateAssemblyInfoFilename",
+      "type": "string",
+      "label": "Update Assembly File",
+      "defaultValue": "",
+      "required": false,
+      "helpMarkDown": "Update versions in specified file",
```
pascalberger: A little bit more documentation here would be nice: Are wildcards patterns supported, what is the default value if no value is entered (AssemblyInfo.cs or AssemblyInfo.vb)

johnkattenhorn: Agreed, it doesn't support wildcard right now (but that's a good idea and something that would be easy to do I guess, does the GitVersion parameter support multiple files ?)

pascalberger: Doesn't look like: https://github.com/GitTools/GitVersion/blob/039a43b092134c6a42902fc162fcb4cbc1265c14/src/GitVersionExe/AssemblyInfoFileUpdate.cs#L52. Wildcard support is something that needs first be implemented in GitVersion. But mentioning that only full file names are supported is still useful (since most of the OOTB VSO task support patterns or minimatch).

johnkattenhorn: Agreed, I'll update the help text

JakeGinnivan:
Just to back up @pascalberger, in v4 we are pushing towards mono support for xplat so it would be better if we used nodejs. 

@johnkattenhorn would this be something you would be comfortable giving a shot?

JakeGinnivan:
@johnkattenhorn bump, if not I can give it a crack once I have got through the issue triage and other PR's

johnkattenhorn:
Hey @JakeGinnivan, sorry this has fallen completely off my radar, too be honest node.js isn't my strongest suite so if you're OK having a go then please be my guest.

gep13:
>@JakeGinnivan said...
in v4 we are pushing towards mono support for xplat so it would be better if we used nodejs.

Is this still the plan?  Seems like a lot of work to port what we already have to Node.  Why not CoreCLR?

pascalberger:
@gep13 Node would only be for the build task (instead of the PowerShell script in this PR) not for the whole project.

gep13:
>@pascalberger said...
Node would only be for the build task (instead of the PowerShell script in this PR) not for the whole project.

Ah! That makes a little more sense! :smile:

JakeGinnivan:
I might update it as we get closer to v4, or even after

PR #724, opened 2015-11-27
Remove requirement on master branch existing

gep13:
- As per discussion here:
https://gitter.im/GitTools/GitVersion?at=565781d10d143098620f6250
- Removed redundant code

### DIFF
--- a/src/GitVersionCore.Tests/IntegrationTests/OtherScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/OtherScenarios.cs
@@ -31,6 +31,21 @@
         }
 
         [Test]
+        public void AllowNotHavingMaster()
+        {
+            using (var fixture = new EmptyRepositoryFixture(new Config()))
+            {
+                fixture.Repository.MakeACommit();
+                fixture.Repository.MakeATaggedCommit("1.0.0");
+                fixture.Repository.MakeACommit();
+                fixture.Repository.Checkout(fixture.Repository.CreateBranch("develop"));
+                fixture.Repository.Branches.Remove(fixture.Repository.Branches["master"]);
+
+                fixture.AssertFullSemver("1.1.0-unstable.1");
+            }
+        }
+
+        [Test]
         public void DoNotBlowUpWhenDevelopAndFeatureBranchPointAtSameCommit()
         {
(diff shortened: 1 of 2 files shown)


gep13:
@JakeGinnivan @pascalberger  can you let me know what you think about this?

pascalberger:
LGTM

asbjornu:
@gep13 A test that confirms the new behavior would be nice. :)

gep13:
>@asbjornu said...
@gep13 A test that confirms the new behavior would be nice. :)

@asbjornu  Normally I would agree, but I am not sure in this case.  I was removing redundant code.  If anything, I should have also been removing a redundant test, but there wasn't one.  What exactly would a test do in this case?

pascalberger:
A test for a case without a master branch might indeed be helpful

asbjornu:
@gep13 What @pascalberger wrote. A pretty simple repository fixture test that creates a repository without a master branch and just asserts that everything went okay.

JakeGinnivan:
+1, I assume you also tested this on your local. The only issue I can see is around the branch inheritance stuff, we often fallback to master or develop if we can't figure it out. But majority of the time you should be fine

gep13:
@JakeGinnivan yes, I tested it locally, and it seemed to work, although I got a strange asserted version number for a feature branch taken off the develop branch, but I haven't dug into that yet to see why it was taking the wrong number.

I will see what I can do about creating a test fixture to cover this scenario, but I haven't really looked at that code yet.

Any suggestions as a good one to start with as a basis?

JakeGinnivan:
Maybe https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs#L14

Then just delete master

gep13:
>@JakeGinnivan said...
Then just delete master

What exactly did you mean here?  Thanks!

JakeGinnivan:
The test will always create a master, branch off it to create develop then use libgit2sharp (exposed through the fixture.Repository) to delete master. Then that should be the basis for a test.

gep13:
Thanks to @JakeGinnivan this PR now has a test to go with the changes :+1:

Issue #632, opened 2015-09-03
alpha.branchname, beta, rc tags instead of branchname, unstable, beta?


o-o00o-o:
So, the default config for tags are set as follows (we are using GitFlow)

feature: useBranchName
develop: unstable
release: beta
hotfix: beta
master: 

I may be missing something but I'm not sure why this is as it means that according to SemVer rules your feature branch is now a higher version than the release branch? i.e. v1.0.0-beta < v1.0.0-unstable which is never going to be the case.

This is causing a problem for us as we are checking the version and actively preventing deploys of versions < the current version.

The normal GitFlow is feature->develop->release->master 
or
hotfix->master

So wouldn't it make sense to assign tags that allowed reflecting this?

so, to achieve this I would propose:

feature: alpha.useBranchName
develop: beta
release: rc
hotfix: rc
master: 

This way feature < develop < (release | hotfix) < master

In order to achieve this I think we would need an extension to the way we can specify "useBranchName" to allow it to be specified in-line (I am assuming that this isn't currently the case from what I've seen but I could be wrong). We need this for features so that all features are always have lower precedence to develop regardless of the name of the branch.

The rest we can change now via the config but it would seem to me to be reasonable to also consider changing the defaults here to help people use SemVer correctly?

Interested in thoughts.

JakeGinnivan:
This can be overridden in the config, that said this is intentional. Develop doesn't follow SemVer as the release process is what actually will determine the version number. With that in mind we always make develop (and feature branches targeting develop) a higher version.
The recommended approach is to have develop in a separate NuGet feed then anything which is actually released (from release branches or other) go to the official NuGet feed. 

Without making develop higher there is no way to just upgrade to the latest on develop in an automated way, which breaks far more people who need a continuous integration scenario across multiple packages. nServiceBus is one example of this

JakeGinnivan:
Oops, didn't mean to close. Does that answer the question, and if you have ideas to improve the docs on this suggestions are welcome

o-o00o-o:
Ok so I guess there are couple of things here:

[...]

PR #744, opened 2016-01-17
Bump develop when release branch is created

### LINKED ISSUE #695: Increment "develop" version when branching release
This issue has been broken out of the discussion here: https://github.com/GitTools/GitVersion/issues/632

The current behavior is to use the "-unstable.N" pre-release revision to ensure builds from _develop_ are considered newer than those on release branches that have not yet been merged back to _develop_.

Example:
```
A-B-C (develop)
   \D-E (release/0.1)
```
In the above graph, build A-C would get the versions 0.1.0-unstable.1-3, and D and E would get 0.1.0-rc.1-2. A-C would all be considered *newer* than D and E. Not until _release/0.1_ is merged back into _develop_ does _develop_'s version get bumped to 0.2.0. I find this counter-intuitive. After all, I've branched the release, why is it still being developed (as opposed to just being stabilized), and why are those builds considered newer than the release itself?

One might argue that deleting a release branch, after concluding the release should be cancelled, would then cause _develop_ to drop back(!) to the previous minor version, but I don't think one should ever delete release branches like that. 1) it causes the problem I just mentioned, and 2) if you had at least one build from that release branch before deleting it, and then re-create the same release, there's nothing guarding you from generating the same 0.1.0-rc.1 version again, which would also become confusing. You could of course tag/configure _develop_ to stay at 0.2.0, but that feels like a bit of a kludge, and you still shouldn't re-create releases. There's nothing in SemVer that says you cannot skip a version, AFAIK, so this wouldn't invalidate anything.

If this was to be changed, and _develop_ were to be bumped on branch, then the -unstable tag should be replaced with something that sorts before -rc, but I haven't got a suggestion that is better than the one in https://github.com/GitTools/GitVersion/issues/632.

JakeGinnivan:
- Fixed an commit counting bug
 - Based on discussion at #632
 - Fixes #695
 - Updated related docs

// @GeertvanHorrik @o-o00o-o, @judgeaxl, @asbjornu 

Keen on feedback. I think we can change develops tag to alpha, but I think maybe this should go into v4 and I try get an alpha of that out asap as this is really a breaking change.


JakeGinnivan:
This is going into v4..

GeertvanHorrik:
This is a fantastic feature, great to have!

Not sure if this PR also includes using alpha instead of unstable for the develop branch? If so, I think the images should be updated as well.

The rest looks really good, great work.

asbjornu:
@JakeGinnivan This looks great, but I'm wondering why you've closed it and deleted the branch?

JakeGinnivan:
Closed it out, am forward merging master into the v4 release branch then I will rebase this on top of that.

JakeGinnivan:
Will reopen over the next few days. I forgot that I pulled a bunch of stuff into GitTools.Core in the v4 branch, which now I have to move the changes on master into Core

totht91:
unforgotten bug

asbjornu:
@totht91: If you need this and what @JakeGinnivan writes about the changes having moved to `GitTools.Core` is incorrect, please rebase the changes on the current `HEAD` of `master` and submit a pull request.

PR #757, opened 2016-01-26
Allow next-version to be an integer

asbjornu:
Try to parse the `next-version` configuration value as a `System.Int32` and treat the parsed value as the major version number.

### DIFF
--- a/src/GitVersionCore.Tests/ConfigProviderTests.cs
+++ b/src/GitVersionCore.Tests/ConfigProviderTests.cs
@@ -117,6 +117,36 @@ branches:
     }
 
     [Test]
+    public void NextVersionCanBeInteger()
+    {
+        const string text = "next-version: 2";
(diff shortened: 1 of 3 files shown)


JakeGinnivan:
This is bypassing the yaml deserialisation, can we add a test in `ConfigProviderTests` to just check this :)

asbjornu:
@JakeGinnivan Doesn't the existing [`ConfigProviderTests.NextVersionCanBeInteger`](/okb/GitVersion/commit/ce684c5830c8ced1cb52b3b8158a18d9a163e727#diff-5c240ef90f0aba3d60f74e889bf8fcb5R121) suffice? :)

asbjornu:
@JakeGinnivan I added a couple of more tests to `ConfigProviderTest` so it's in sync with the tests in `ConfigNextVersionBaseVersionStrategyTests`. I've also rebased.

PR #760, opened 2016-01-27
Feature/630 create assembly info

Philo:
#re 630


gep13:
@Philo looks like there are some issues with the CI Build as a result of the changes that you have made.  Are you in a position to look into what is going on?

Thanks!

Philo:
@gep13, I'll take a look, however may not be today

gep13:
@Philo not a problem at all :+1:

Philo:
@gep13 - fixed the build now

gep13:
@Philo thanks for your contribution here! :+1:

JakeGinnivan:
I think this has a bug in it.

It should only add the assembly info attributes in if the filename is specified. I think this is the cause of the assembly info issues since the last release

PR #765, opened 2016-01-30
Perf improve part1

EbenZhangEmbed:
- Cache the PeeledTarget to gain better perf
- We don't have to do anything in `TrackMergeTargetBaseVersionStrategy` if `TrackMergeTarget` is disabled or the current branch name is null or empty
- We don't have to get more than 2 parents because the inherit increment branch configuration fails if there are more than 1 branch.

Part2 #734 is to use Git Command Line to get the branches that contain a commit.


JakeGinnivan:
All look like nice changes, thanks :)

PR #766, opened 2016-01-30
Support branch name variable in tag configuration #664

### LINKED ISSUE #664: Allow {BranchName} in tag config
As per #632 

Currently the prerelease tag configuration is fine except for feature branches where there can be many of them at the same time for the same version. The problem here is that with the current useBranchName config, feature branches starting with "a" are considered the lowest prerelease version and any feature branch starting with "z" will be the latest one. To solve this we should be able to prefix the feature braches "something." so they can be positioned in the desired order consistently no matter what the name of the feaure branch. 

E.g feature/AddForgottenPassword and feature/ZeroAllowedInAmount could be v1.0.0-alpha.AddForgottenPassword and v1.0.0-alpha.ZeroAllowedInAmount which would then both be seen as the same level compared with the other develop/release and master branches.

To achieve this @JakeGinnivan suggested using c# string interpolation so in config something like:

tag: alpha.{BranchName} 

For branch feature/FixLogin
Would produce a semver like 0.0.0-alpha.FixLogin

gusztavvargadr:
Resolves #664.

### DIFF
--- a/src/GitVersionCore.Tests/IntegrationTests/FeatureBranchScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/FeatureBranchScenarios.cs
@@ -170,4 +170,30 @@ public class FeatureBranchScenarios
             fixture.AssertFullSemver("1.2.0-longrunning.2");
         }
     }
+
+    [TestCase("alpha", "JIRA-123", "alpha")]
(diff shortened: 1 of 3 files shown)


JakeGinnivan:
Sweeeet

PR #786, opened 2016-02-17
Fix VSTS known limitation documentation

pascalberger:
Fix VSTS known limitation documentation. Build.Clean needs to be set to all not true


gep13:
:+1:

Issue #575, opened 2015-08-16
Support GitVersion.yml config file


JakeGinnivan:
Because `GitVersionConfig.yaml` has the wrong extension and Config is redundant.

orjan:
I think this should be the default behavior in v4 but with a read fallback on the old file with a warning?

asbjornu:
Fallback with warning sounds like a good plan. We don't even need to wait for v4 if we have backwards compatibility, imho.

JakeGinnivan:
Just support both, shouldn't be too hard

qetza:
Hi, i'll do it:
- read GitVersion.yml if present
- if not read GitVersionConfig.yaml and log a warning for deprecated file name.

qetza:
Done in PR #790

raybooysen:
Possible to close this?

PR #790, opened 2016-02-20
Support GitVersion.yml config file #575

qetza:


src/GitVersionCore/Configuration/ConfigurationProvider.cs
```diff
@@ -144,14 +144,28 @@ public static string GetEffectiveConfigAsString(string gitDirectory, IFileSystem
...
+
+            var yamlPath = Path.Combine(workingDirectory, "GitVersionConfig.yaml");
+            if (fileSystem.Exists(yamlPath))
+            {
+                Logger.WriteWarning("'GitVersionConfig.yaml' is deprecated, use 'GitVersion.yml' instead.");
+
+                return yamlPath;
+            }
+
+            return ymlPath;
```
pascalberger: The two last returns here are redundant. IMHO it would be clearer to return `string.empty` or `null` if none of both files exists, and add checks outside, instead of returning the path to the deprecated file which not even exists. Or otherwise I would suggest to at least return the path to the current file and not the deprecated one.

qetza: Hi Pascal,
The last two returns are not the same, the complete logic is:

- if the new file GitVersion.yml already exists we return it
- if the deprecated file GitVersionConfig.yaml already exists we warn and return it
- if no file exists we return the new file GitVersion.yml

pascalberger: Ah, sorry. missed the additional a in the last return :blush:

asbjornu: @qetza How about renaming the variable `yamlPath` to `deprecatedPath`? Just a single letter separating two variables make them, as obvious by @pascalberger's comment, hard to differentiate. :smiley:

asbjornu:
@qetza If you just rebase this onto the latest `master` to resolve the conflicts, I think this is a welcome addition that @gep13, @pascalberger or @JakeGinnivan can merge in when time allows. :+1:

JakeGinnivan:
Thanks @qetza! Been wanting this for ages :)

pkiers:
Hello,

I just wanted to share some interesting thoughts about amazing people I met yesterday, please read more here <http://delight.unccamps.com/e2rpjljc>

Yours, p.kiers@home.nl

PR #792, opened 2016-02-22
Improved GitVersionTask documentation

asbjornu:
Added more information to how GitVersionTask works, a little TL;DR section and rewrote a few section to make the documentation clearer. Addresses #789.

PR #793, opened 2016-02-23
Fixed problems in MSBuild Task documentation

asbjornu:
This PR tries to fix the problems introduced in #792 which makes the Markdown render badly on ReadTheDocs. I hope this covers everything!


asbjornu:
@pascalberger Thanks!

PR #795, opened 2016-02-24
Additional Test to cover usage of main, instead of master

gep13:


gep13:
@JakeGinnivan ok, so I did some digging into what I thought was an issue, namely this:

https://ci.appveyor.com/project/cakebuild/cake/build/0.9.1-main0001#L232

In the case of Cake, we don't have a `master` branch, we have a `main` branch.  Within this PR https://github.com/GitTools/GitVersion/pull/724 we removed the requirement to have a `master` branch, and I thought there was still an issue, as the wrong asserted version number was coming back.

However, digging into it, I see that I was missing the configuration in the yaml file, to specify the settings for that named branch.

In this PR, I have added another test, to cover the usage of the `main` branch, by providing the necessary branch configuration to make it happy.

Please let me know what you think.

JakeGinnivan:
Looks good.

The only scenario which may have issues is there are hard coded references to master in the inherit branch config for resolving when there are multiple matched branches. This should only affect a really small number of scenarios though, if any?

gep13:
@JakeGinnivan these were the two places that I found hard coded references to master:

https://github.com/GitTools/GitVersion/blob/255a73c69b7e15f09ad926fb63871d2fce09eff9/src/GitVersionCore/Configuration/ConfigurationProvider.cs#L41
https://github.com/GitTools/GitVersion/blob/f64549ac9954a55c3089907796eb64c08ba97dd5/src/GitVersionCore/BuildServers/GitHelper.cs#L68

But I wasn't sure what needed to be done here.

Also, would it make sense to extend the test cases to include running them with save a `dev` and `main` branch?  Making sure all the test cases pass, with a different branch configuration?

PR #801, opened 2016-03-01
Fixed a few things in the index page

dcorriveau-omniscient:
`its` instead of `it's` (possessive of the pronoun it, most likely a typo).
`e.g.` is preferred when citing examples. `i.e.` is used when you want to give further explanation.
`for not just` didn't sound right so I altered it.
Microsoft seems to prefer uppercase `.NET`.

Btw I'm far from being skilled in written English language so if it doesn't make any sense let me know. I'm just reading the docs for the first time since I'm planning on using GitVersion and I thought I'd propose those changes. Hopefully I'll contribute in a more meaningful way in the future once I actually use the app.

Have a good day !

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,13 +13,13 @@ This trips a lot of people up, by default GitVersion *does not* increment the Se
 Read more at [version increments](more-info/version-increments.md)
 
 ### Version sources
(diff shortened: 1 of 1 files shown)


gep13:
@dcorriveau-omniscient these all look like great changes to me! :+1:

PR #802, opened 2016-03-02
Updated doc for a better read

dcorriveau-omniscient:
Looks like the file has carriage returns too, I'm not sure if it could cause problems in the future but I let them there for the moment so that the diff would be easier to see.


gep13:
@dcorriveau-omniscient once again, looks good to me! Thanks!

Yes, whitespace changes should either be submitted as a separate PR, or a separate commit on the same PR.  That way, it makes it easier to see what is going on.

PR #803, opened 2016-03-02
Removed carriage returns

dcorriveau-omniscient:
Like this ?

### DIFF
--- a/docs/usage/msbuild-task.md
+++ b/docs/usage/msbuild-task.md
@@ -1,15 +1,12 @@
 # MSBuild Task
 
-The MSBuild Task for GitVersion — **GitVersionTask** — is a simple solution if
-you want to version your assemblies without writing any command line scripts or
-modifying your build process.
+The MSBuild Task for GitVersion — **GitVersionTask** — is a simple solution if you want to version your assemblies without writing any command line scripts or modifying your build process.
 
 ## TL;DR
 
 ### Install the MSTask targets
 
-Add the [GitVersionTask](https://www.nuget.org/packages/GitVersionTask/) NuGet
-Package into the project you want to be versioned by GitVersion.
+Add the [GitVersionTask](https://www.nuget.org/packages/GitVersionTask/) NuGet Package into the project you want to be versioned by GitVersion.
 
 From the Package Manager Console:
 ```shell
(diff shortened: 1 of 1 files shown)


asbjornu:
What's the point of this? I find that what makes Markdown so brilliant is that it works so well when it is not rendered and just viewed as plain text in a text editor or in the terminal. In that environment, line-wrap at 80 characters is the standard and has been for 40 years. Removing these line breaks makes the document harder to navigate in those environments without adding anything I can think of to a rendered environment.

As for providing some substance for my "cause", you have [this](https://github.com/carwin/markdown-styleguide) and [this](http://www.cirosantilli.com/markdown-style-guide/#option-wrap-inner-sentence) Markdown style guide which both state that lines should break at 80 characters.

I know most Markdown files in this repository don't follow this guideline, but I'd argue they should. :smiley:

dcorriveau-omniscient:
I agree with your points, the main reason I did it was for consistency with the other files in the repo (I thought it was your standard for this repo). But I guess we should do the opposite then.

dcorriveau-omniscient:
Why is there no bullets displayed there:
nobullets

If I test on http://dillinger.io/ it looks fine.
bullets

gep13:
I would suspect that a carriage return is required.  Not sure why it renders ok in GitHub and Dillinger and not on ReadTheDocs.

gep13:
>@asbjornu said...
As for providing some substance for my "cause", you have this and this Markdown style guide which both state that lines should break at 80 characters.
I know most Markdown files in this repository don't follow this guideline, but I'd argue they should.

While part of me agrees with this, I would also argue that most Markdown Editors, or at least the ones that I have played with, wrap the longer lines automatically, so it becomes less of an issue.  I am a big fan of consistency though, so if we are going to do this, we should do it everywhere, and also document how to configure the more popular editors to automatically enforce this.

dcorriveau-omniscient:
Yea that's what I thought too first, that's why I added it sneakily in my last commit (the one that got rejected), but then I double checked in Dillinger to make sure and I saw it was displayed correctly so I thought it was rather odd. Are you aware of a better Markdown test tool that I could use ?

asbjornu:
@gep13 The problem isn't with "Markdown editors", it's with plain text editors or otherwise people reading and editing the files that don't even know the file is anything else than plain text. That scenario is possible because Markdown basically *is* plain text, but if we only set us up to "support Markdown editors", we make the plain text story more difficult without gaining anything, imho.

I agree we should enforce the same policy across all `.md` files in the repository. Let's choose a style guide and stick with it? If we can enforce it through a tool, then even better! :+1:

gep13:
>@dcorriveau-omniscient said...
Are you aware of a better Markdown test tool that I could use ?

None that I am aware of, no.

>@asbjornu said...
I agree we should enforce the same policy across all .md files in the repository. Let's choose a style guide and stick with it? If we can enforce it through a tool, then even better!

I am happy to let you take the lead on this one :smile:

dcorriveau-omniscient:
I think the best thing to do in order to wrap the text to 80 columns would be to use a tool / editor so that the result is consistent.

VIM would be a pretty good standard tool imo.

1. Set the default textwidth to 80 `:set tw=80`.
2. Open the file `:edit <filepath>`.
3. Go to start of document `gg`.
4. Format the document from cursor to end of file `gqG`.

Worked well for me.

asbjornu:
@dcorriveau-omniscient It's easy in [Atom](https://atom.io/) too. Just install the [hard-wrap plugin](https://atom.io/packages/hard-wrap), select the text you want to wrap and press ⌃⌥Y.

gep13:
@JakeGinnivan what are your thoughts on the above?  I am happy to accept a PR is this work is completed.

Thoughts?

JakeGinnivan:
I have no probs with this going in :)

asbjornu:
@JakeGinnivan, awesome. @dcorriveau-omniscient, do you want to do a PR on this or do you want me to tackle it? Either way, we can collaborate.

dcorriveau-omniscient:
@asbjornu I only have converted one page to test the process so I don't really have a PR ready. You can tackle it :P

asbjornu:
@dcorriveau-omniscient Haha, okay. Want to close this PR then?

dcorriveau-omniscient:
Sure, thanks.

PR #805, opened 2016-03-03
Fixed list

dcorriveau-omniscient:
Looks like https://stackedit.io/editor isn't too bad and displays the result correctly using GitHub flavored markdown.


gep13:
@dcorriveau-omniscient noice! stackedit looks very useful, thanks for the link!

Issue #138, opened 2014-03-30
Support generating AssemblyVersion attributes in other languages


SimonCropp:
- [x] VB
  - [x] Command Line
  - [x] MS Build Task
- [ ] F#
  - [ ] Command Line
  - [ ] MS Build Task

pascalberger:
Update state as mentioned in https://github.com/GitTools/GitVersion/issues/633#issuecomment-138137484

pascalberger:
VB.NET support in the MS Build Task was implemented in #811.

pascalberger:
Is F# support still a goal, or can this issue be closed since we have full support for VB.NET reached?

jpierson:
It looks like some work was done here for F# support. Last I can recall there was some specific handling for language specific project extensions where .fsproj was missing but I'm unable to find it at the moment.
https://github.com/GitTools/GitVersion/pull/882/commits/1353214b6a5db6abcadd0c709cc58e372840a950

gep13:
@pascalberger personally, I am still dubious about this.  I would rather see GitVersion as a standalone tool that asserts version numbers.  And leave the updating of various files to other, more capable tools, like magic chunks.

jpierson:
@gep13, magic chunks? 

In my case updating assembly info at build time is the sole reason why I starting using GitVersion. Other tools in the past that I've tried with Subversion, TFS, and Git have required too much manual customization and team conventions with the build pipeline whereas GitVersion msbuild tasks just seem to work.

With respect to this issue and the current support for F# I have forked and done what I believe is the necessary work to put support on par with VB.NET and C# languages. I'll be putting in a pull request for these changes soon.

gep13:
@jpierson sorry, should have linked to it...

https://github.com/sergeyzwezdin/magic-chunks

It is very good at what it does.

jpierson:
@gep13, thanks magic chunks looks interestest. I'm not sure how that relates to updating the actual version numbers on an assembly though unless that tool is also updating source files in addition to config like scripts. BTW, I took both approaches in the [source-version](/365retailmarkets/source-version) package which initially just pushed a version.txt file into the content of a web application but had changed to use a .json file instead in addition to updating the AssemblyVersion stuff as well.

jpierson:
[...]

PR #811, opened 2016-03-24
Implement support for AssemblyInfo.vb in the MSBuild GitVersionTask

apmorton:
Just some general notes about these changes.

- VB.NET generally uses a root namespace at the project level, and everything that doesn't specify a namespace is put there - which is why the `VisualBasicAssemblyInfoBuilder` does not namespace its `GitVersionInformation` class
- VB.NET does not have static classes - the best you can do is make it `NotInheritable`, provide a `Private Sub New()`, and make all your members `Shared`
  - IMO this is better than using a module, as it prevents your members from polluting the namespace
- The majority of changes here are related to unit testing - I basically duplicated all existing tests for C# and VB.NET
- I tried to implement it in a way generic enough to allow future expansion to additional languages without hassle. #138 mentions F# support as a goal.

Lastly, I have this currently working in an existing VB.NET project at work - everything checks out with dotPeak, and it seems to be working as expected.


pascalberger:
LGTM. 

@JakeGinnivan @gep13 Any opinion against merging this PR / VB.NET support?

gep13:
@pascalberger In principal, I have no problem with merging this in, but I haven't as yet, had a chance to review the PR properly.

asbjornu:
I've had a look and I think it's a thorough and good implementation with great test coverage. The only thing I thought while reading through the code was that `AssemblyInfoBuilder` should have been renamed to `CSharpAssemblyInfoBuilder` and then in a new commit, the `AssemblyInfoBuilder` should be added, instead of the other way around.

This is because the contents of `CSharpAssemblyInfoBuilder` is mostly identical to that of the old `AssemblyInfoBuilder` and the new `AssemblyInfoBuilder` is an `abstract` base class for `CSharpAssemblyInfoBuilder` and `VisualBasicAssemblyInfoBuilder`.

Other than that, this PR LGTM. :+1:

gep13:
@apmorton thanks for your work on this, we really appreciate it! :+1:

PR #818, opened 2016-04-01
Fixes #817

### LINKED ISSUE #817: GitVersion command line NuGet package should be marked as dev dependency
When adding this NuGet package to a project, thinking it would be best if it was flagged as a development dependency, so that when someone subsequently does a `nuget pack` on their project, GitVersion.CommandLine won't be added as a dependency to their own package.

dazinator:
#Added developmentDependency flag to nuspec for gitversion commandline, fixes #817


gep13:
Makes sense to me! :+1:

PR #820, opened 2016-04-02
Project Rider changes

asbjornu:
When opening the solution in [Project Rider](https://www.jetbrains.com/rider/), it generates a few files that should be ignored and makes a little change to the `.DotSettings` file.


gep13:
LGTM! :smile_cat:

PR #832, opened 2016-04-16
Bump develop when release branch created

### LINKED ISSUE #695: Increment "develop" version when branching release

JakeGinnivan:
- Fixed an commit counting bug
 - Based on discussion at #632
 - Fixes #695
 - Updated related docs


src/GitVersionCore/EffectiveConfiguration.cs
```diff
@@ -43,8 +44,13 @@ int commitsSinceVersionSourcePadding
             LegacySemVerPadding = legacySemVerPaddding;
             BuildMetaDataPadding = buildMetaDataPadding;
             CommitsSinceVersionSourcePadding = commitsSinceVersionSourcePadding;
+            IsCurrentBranchDevelop = isCurrentBranchDevelop;
+            IsCurrentBranchRelease = isCurrentBranchRelease;
         }
 
+        public bool IsCurrentBranchDevelop { get; private set; }
+        public bool IsCurrentBranchRelease { get; private set; }
```
gep13: @JakeGinnivan very small point, but can we get a space between the properties?

gep13:
@JakeGinnivan apart from one small comment, this looks good to me. :+1:

asbjornu:
I think this looks good! Rebase, let Travis and AppVeyor do their thing and if all green; merge?

DanielRose:
When will this be merged? I could really use this.

JakeGinnivan:
This is a bit of a change of behaviour. What does everyone thing, changing defaults ok for a minor release? develop is going from unstable to alpha with this change

asbjornu:
Since people can always configure `develop` to have the `unstable` tag, I think it's fine doing this in a minor. This configuration option should probably be mentioned in the release notes, though, so it's easily accessible information in case it causes trouble for anyone.

JakeGinnivan:
I need to see the impacts. It changes sorting and bumping rules. Changing back to unstable will cause more problems I think.

I think the sorting is the same, or better in most cases once this is merged. Spose we can release as a pre-release and gather feedback.

JakeGinnivan:
@asbjornu have updated some of the images in the docs.

Will do the mainline docs update in that branch

JakeGinnivan:
This shows the real effect of this branch quite well.

image

asbjornu:
@JakeGinnivan That picture makes perfect sense. LGTM! 😄 👍

DanielRose:
@JakeGinnivan One small thing I noticed: In docs/git-branching-strategies/img/39f9d8b8b007c82f1f80_major-release.png (also shown above), the effect of tagging the master branch is not shown anymore.

JakeGinnivan:
Nice catch @DanielRose, R#s test runner truncates output now and I didn't look properly :P

JakeGinnivan:
Have fixed that image in #950. Thanks @DanielRose

PR #838, opened 2016-04-21
Fix docs to state that Git_Branch is an environment variable for TeamCity

DanielRose:
The "Git_Branch" variable is read via Environment.GetEnvironmentVariable(). Thus, it needs to be an environment variable, not a system variable in TeamCity.


gep13:
@DanielRose thank you very much for making this correction, it is much appreciated!

pkiers:
Hi,

We are planning a short trip this summer and wanted to share some routes with you, what do you think? More info here <http://direct.lalitgolani.com/e2ual>

Best Wishes, p.kiers@home.nl

pkiers:
Hello,

Here is the list of upcoming events you may be interested in, you may find more information here <http://example.brainybrainy.net/e2ejfvij>

Cheers, p.kiers@home.nl

PR #839, opened 2016-04-21
Fix gitlink's broken link in the documentation

KyleOndy:
Noticed that this link was broken when clicking through the documentation.


pascalberger:
Thanks for this fix @KyleOndy, much appreciated!

PR #850, opened 2016-04-29
Working Travis build

asbjornu:
This PR completes the Mono support started in #816 and contains fixes all over the place to achieve a [working build on Travis](https://travis-ci.org/asbjornu/GitVersion). Most tests are executed and succeed, but some tests using the `ShouldMatchApproved()` extension method from Shouldly had to be ignored, since they fail due to missing source information.

If Shouldly gains Mono `.mdb` symbol support, these tests can be un-ignored and the Travis (Linux) build of GitVersion should be on par with Windows and we should be in a great place to ensure Unix, Linux and Mono compatibility going forward.

After merging this, a Travis build badge like this can be added to the README:

[Build Status](https://travis-ci.org/asbjornu/GitVersion)

PR #816, opened 2016-04-01
Mono support

asbjornu:
This PR builds on @BeeWarloc's work in #638 and fixes #264. I've added a [Travis build](https://travis-ci.org/asbjornu/GitVersion) so we can have continuous integration testing on Linux. I don't think I've broken backward compatibility with anything, so it should be safe to merge this into `master` and release as a `3.x`. I think adding Linux support is so important that it can't wait for version 4.

That being said, the tests on Travis fail with the following exception:

```c#
System.TypeInitializationException : The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception.
  ----> System.DllNotFoundException : NativeBinaries/linux/amd64/libgit2-785d8c4.so
  at LibGit2Sharp.Core.Proxy.git_repository_init_ext (LibGit2Sharp.Core.FilePath workdirPath, LibGit2Sharp.Core.FilePath gitdirPath, Boolean isBare) <0x4188cb20 + 0x00058> in <filename unknown>:0 
  at LibGit2Sharp.Repository.Init (System.String path, Boolean isBare) <0x4188b080 + 0x0005a> in <filename unknown>:0 
  at LibGit2Sharp.Repository.Init (System.String path) <0x4188b050 + 0x00016> in <filename unknown>:0 
  at EmptyRepositoryFixture.CreateNewRepository (System.String path) <0x4188afb0 + 0x00017> in <filename unknown>:0 
  at RepositoryFixtureBase..ctor (System.Func`2 repoBuilder, GitVersion.Config configuration) <0x4188acd0 + 0x0002b> in <filename unknown>:0 
  at EmptyRepositoryFixture..ctor (GitVersion.Config configuration) <0x4188ac00 + 0x000ad> in <filename unknown>:0 
  at PullRequestInTeamCityTest.GivenARemoteWithATagOnMaster_AndAPullRequestWithTwoCommits_AndBuildIsRunningInTeamCity_VersionIsCalculatedProperly (System.String pullRequestRef) <0x418931b0 + 0x0009b> in <filename unknown>:0 
  at (wrapper managed-to-native) System.Reflection.MonoMethod:InternalInvoke (System.Reflection.MonoMethod,object,object[],System.Exception&)
  at System.Reflection.MonoMethod.Invoke (System.Object obj, BindingFlags invokeAttr, System.Reflection.Binder binder, System.Object[] parameters, System.Globalization.CultureInfo culture) <0x41733ab0 + 0x000b7> in <filename unknown>:0 
[...]

PR #856, opened 2016-05-10
Working Travis build

asbjornu:
This PR completes the Mono support started in #816, supersedes #850 and contains fixes all over the place to achieve a [working build on Travis](https://travis-ci.org/asbjornu/GitVersion). Most tests are executed and succeed, but some tests using the `ShouldMatchApproved()` extension method from Shouldly had to be ignored, since they fail due to missing source information.

If Shouldly gains Mono `.mdb` symbol support, these tests can be un-ignored and the Travis (Linux) build of GitVersion should be on par with Windows and we should be in a great place to ensure Unix, Linux and Mono compatibility going forward.

After merging this, a Travis build badge like this can be added to the README:

[Build Status](https://travis-ci.org/asbjornu/GitVersion)


asbjornu:
@gep13 There we go. AppVeyor ran this with a new branch name. Anyway, I'd love to have this merged! 😄

Issue #546, opened 2015-07-28
$(GitVersion_NuGetVersion) does seem to get set when Solution is built using TFS Build vNext


johnkattenhorn:
I've already put a comment in the closed issue #530.

I used the article you mentioned in #530 to great effect but 3.0.1 seems to have broken it for me. I changed the variable name to $(GitVersion_NuGetVersion) but it doesn't seem to get set. 

Here is a snippet of the bottom of my project file

    <PropertyGroup>
        <BuildDependsOn>
          $(BuildDependsOn);
          SetOctoPackVersion
        </BuildDependsOn>
      </PropertyGroup>
      
      <Target Name="SetOctoPackVersion">
        <PropertyGroup>
          <OctoPackPackageVersion>$(GitVersion_NuGetVersion)</OctoPackPackageVersion>
        </PropertyGroup>
      </Target>
      <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
      <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
        <PropertyGroup>
          <ErrorText>This project references NuGet package(s) that are missing on this computer. Enable NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
        </PropertyGroup>
        <Error Condition="!Exists('..\packages\GitVersionTask.3.0.1\Build\GitVersionTask.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\GitVersionTask.3.0.1\Build\GitVersionTask.targets'))" />
  </Target>
      <Import Project="..\packages\AutoMapper.3.3.1\tools\AutoMapper.targets" Condition="Exists('..\packages\AutoMapper.3.3.1\tools\AutoMapper.targets')" />
      <Import Project="..\packages\OctoPack.3.0.42\tools\OctoPack.targets" Condition="Exists('..\packages\OctoPack.3.0.42\tools\OctoPack.targets')" />
      <Target Name="EnsureOctoPackImported" BeforeTargets="BeforeBuild" Condition="'$(OctoPackImported)' == ''">
[...]

PR #873, opened 2016-05-23
Add support for TFS 2013.

GaveUp:


src/GitVersionCore/BuildServers/VsoAgent.cs
```diff
@@ -1,51 +1,58 @@
...
+			};
+		}
+
+		public override string GetCurrentBranch(bool usingDynamicRepos)
+		{
+			if (Environment.GetEnvironmentVariable("BUILD_SOURCEBRANCH") != null)
+				return Environment.GetEnvironmentVariable("BUILD_SOURCEBRANCH");
+
+			//TFS2013 Support.  Must parse branch from TF_BUILD_SOURCEGETVERSION.
+			if (Environment.GetEnvironmentVariable("TF_BUILD_SOURCEGETVERSION") != null)
```
pascalberger: IMHO it would be better to use a own `BaseServerBase` implementation for TFS2013, since this is a completely different build system than the one in TFS2015 / VSTS.

src/GitVersionCore/BuildServers/VsoAgent.cs
```diff
@@ -1,51 +1,58 @@
...
+
+			return null;
+		}
+
+		public override bool PreventFetch()
+		{
+			return true;
+		}
+
+		public override string GenerateSetVersionMessage(VersionVariables variables)
```
pascalberger: I don't think this implementation will work for TF2013 / XAML build.

src/GitVersionCore/BuildServers/VsoAgent.cs
GaveUp: To split it out `CanApplyToCurrentContext()` needs to be modified to check for a null BUILD_SOURCEBRANCH in addition to TF_BUILD.  I didn't have a VSO setup readily available, though, to test so this was the safer change.  I'll see about getting something set up to test splitting it out.

src/GitVersionCore/BuildServers/VsoAgent.cs
GaveUp: Can you elaborate on this?  I used a gitflow branching strategy with ContinuousDeployment mode and got the expected versions in all the scenarios I could think of.  It's entirely possible I've missed something, though.

pascalberger: How did you get the versions back to the build? Or how did you have used them further?

This method should set the version of the build. Since it is using a [VSTS logging command](https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md) I don't think it will be supported in TFS2013 / XAML build. Ditto for the GenerateSetParameterMessage method, which should define parameters in the build server containing the version information.

pascalberger: How did you get the versions back to the build? Or how did you have used them further?

This method should set the version of the build. Since it is using a [VSTS logging command](https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md) I don't think it will be supported in TFS2013 / XAML build. Ditto for the GenerateSetParameterMessage method, which should define parameters in the build server containing the version information.

pascalberger: How did you get the versions back to the build? Or how did you have used them further?

This method should set the version of the build. Since it is using a [VSTS logging command](https://github.com/Microsoft/vsts-tasks/blob/master/docs/authoring/commands.md) I don't think it will be supported in TFS2013 / XAML build. Ditto for the GenerateSetParameterMessage method, which should define parameters in the build server containing the version information.

GaveUp: In my setup OctoPack is used to push to Octopus Deploy with the snippet from #546 to set the package version.

pascalberger: OK, in the snippet you call Octopus from MSBuild. The idea behind the `GenerateSetParameterMessage` and `GenerateSetVersionMessage` methods is that they pass the version information back to the build server (TFS in this case), so that it can be used further by the build system (eg. in subsequent build tasks).

GaveUp: Light Bulb has clicked on.  I see what you're getting at now.  I'll go back and refactor this in to its own class since that makes more sense at this point and look at getting the variables available for subsequent tasks, assuming that's possible.

PR #876, opened 2016-05-24
Fix Travis build

asbjornu:
This PR does a few things to fix the build on Travis so it fails when a unit test fails and also fixes a few issues regarding Unix path handling that were uncovered once the build started failing.


asbjornu:
I think it would be wise to merge this and rebase all other pull requests so they are successfully tested on Travis as well (which they aren't now).

PR #890, opened 2016-06-03
Exclude LibGit2Sharp from ILRepack in NuGet.CommandLine

### LINKED ISSUE #854: GitVersion.CommandLine does not work on OS X
Steps to reproduce:

1. mono nuget.exe install GitVersion.CommandLine -Version 3.5.2
2. mono ./GitVersion.CommandLine.3.5.2/tools/GitVersion.exe

Output:

System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. ---> System.DllNotFoundException: git2-381caf5
  at (wrapper managed-to-native) LibGit2Sharp.Core.NativeMethods:git_libgit2_init ()
  at LibGit2Sharp.Core.NativeMethods+LibraryLifetimeObject..ctor () <0x1fa8ae0 + 0x0000b> in <filename unknown>:0
  at LibGit2Sharp.Core.NativeMethods..cctor () <0x1fa8908 + 0x00107> in <filename unknown>:0

Comparing the GitVersion.CommandLine package to the GitVersionTask package (which _does_ work on OS X), the difference is that GitVersionTask includes a GitVersionTask.dll.config file which uses the Mono dllmap capability to map libgit2 for each system; GitVersion.CommandLine does not contain a similar config file. Copying the config file from GitVersionTask to GitVersion.CommandLine (and renaming it to GitVersion.exe.config) causes GitVersion.exe to work correctly.

I will look into creating a PR with a fix to the GitVersion.CommandLine package in the next couple of days, unless someone beats me to it.

Also, since the docs say to open an issue about why I want to use the GitVersion.CommandLine package, let me add that note here:

- I cannot use the Chocolatey package for two reasons:
  - Different projects on my (or any other) machine may be using different versions of GitVersion, making it dangerous to rely on a GitVersion that is installed machine-wide
  - I am building on OS X
- I don't want to use the GitVersionTask package, because it only works in MSBuild. I do not want determining my version number to be tied to MSBuild, because MSBuild is not my primary build tool (although like most .NET projects I am dependent on it for compilation for the time being). For example, there is no easy way to pass the version from GitVersion to my "nuget pack" command.
  - The typical solution of using the "$version$" replacement token in the nuspec to substitute the version from the assembly only works when packing based on a csproj, which doesn't work when running NuGet in mono. I am forced to pack based on the nuspec, and therefore must pass the version directly.
- I don't want to check in GitVersion.exe to my source repository for the same reasons I don't want to check in any external dependencies to my source repository: I lose the ability to easily track and update the versions of my dependencies that NuGet gives me.

For these reasons, the GitVersion.CommandLine package is the best solution for me. Hopefully it can be maintained as a first-class citizen in the future.

eatdrinksleepcode:
Fixes #854 

This is a short-term fix for NuGet.CommandLine on Mono. It excludes LibGit2Sharp from ILRepack, and copies LibGit2Sharp (and its config file) into the package instead. Only NuGet.CommandLine is affected; changing the other packages build by GitVersionExe might (or might not) make sense, but that decision is being deferred to get this change in, in order to unblock other work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ install:
   - mono .nuget/nuget.exe install NUnit.Runners -Version 3.2.1 -OutputDirectory ./src/packages
 script:
   - xbuild ./src/GitVersion.sln /property:Configuration="Debug" /verbosity:detailed
+  - mono ./build/NuGetCommandLineBuild/tools/GitVersion.exe -l -console -output buildserver -updateAssemblyInfo
+  - xbuild ./src/GitVersion.sln /property:Configuration="Debug" /verbosity:detailed
   - mono ./src/packages/NUnit.ConsoleRunner.3.2.1/tools/nunit3-console.exe ./src/GitVersionTask.Tests/bin/Debug/GitVersionTask.Tests.dll ./src/GitVersionCore.Tests/bin/Debug/GitVersionCore.Tests.dll ./src/GitVersionTask.Tests/bin/Debug/GitVersionTask.Tests.dll ./src/GitVersionExe.Tests/bin/Debug/GitVersionExe.Tests.dll --where "cat != NoMono" --noresult
 
 #
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ os:
   - linux
   - osx
 before_install: # We need to download nuget.exe due to: https://github.com/travis-ci/travis-ci/issues/5932
-  - git fetch --unshallow
+  - git fetch --unshallow # Travis always does a shallow clone, but GitVersion needs the full history including branches and tags
   - mkdir -p .nuget
   - wget -O .nuget/nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
   - mono .nuget/nuget.exe
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ os:
   - linux
   - osx
 before_install: # We need to download nuget.exe due to: https://github.com/travis-ci/travis-ci/issues/5932
+  - git fetch --unshallow
   - mkdir -p .nuget
   - wget -O .nuget/nuget.exe https://dist.nuget.org/win-x86-commandline/latest/nuget.exe
   - mono .nuget/nuget.exe
@@ -15,7 +16,7 @@ install:
   - mono .nuget/nuget.exe install NUnit.Runners -Version 3.2.1 -OutputDirectory ./src/packages
 script:
   - xbuild ./src/GitVersion.sln /property:Configuration="Debug" /verbosity:detailed
-  - mono ./build/NuGetCommandLineBuild/tools/GitVersion.exe -l -console -output buildserver -updateAssemblyInfo
+  - mono ./build/NuGetCommandLineBuild/tools/GitVersion.exe -l console -output buildserver -updateAssemblyInfo
   - xbuild ./src/GitVersion.sln /property:Configuration="Debug" /verbosity:detailed
   - mono ./src/packages/NUnit.ConsoleRunner.3.2.1/tools/nunit3-console.exe ./src/GitVersionTask.Tests/bin/Debug/GitVersionTask.Tests.dll ./src/GitVersionCore.Tests/bin/Debug/GitVersionCore.Tests.dll ./src/GitVersionTask.Tests/bin/Debug/GitVersionTask.Tests.dll ./src/GitVersionExe.Tests/bin/Debug/GitVersionExe.Tests.dll --where "cat != NoMono" --noresult
 
--- a/src/GitVersionCore.Tests/ExecuteCoreTests.cs
+++ b/src/GitVersionCore.Tests/ExecuteCoreTests.cs
@@ -141,8 +141,8 @@ CommitDate: 2015-11-10
     string RepositoryScope(ExecuteCore executeCore = null, Action<EmptyRepositoryFixture, VersionVariables> fixtureAction = null)
     {
         // Make sure GitVersion doesn't trigger build server mode when we are running the tests
-        Environment.SetEnvironmentVariable("APPVEYOR", null);
-		Environment.SetEnvironmentVariable("TRAVIS", null);
+        Environment.SetEnvironmentVariable(AppVeyor.EnvironmentVariableName, null);
+        Environment.SetEnvironmentVariable(TravisCI.EnvironmentVariableName, null);
         var infoBuilder = new StringBuilder();
         Action<string> infoLogger = s =>
         {
--- a/src/GitVersionCore.Tests/ExecuteCoreTests.cs
+++ b/src/GitVersionCore.Tests/ExecuteCoreTests.cs
@@ -142,6 +142,7 @@ CommitDate: 2015-11-10
     {
         // Make sure GitVersion doesn't trigger build server mode when we are running the tests
         Environment.SetEnvironmentVariable("APPVEYOR", null);
+		Environment.SetEnvironmentVariable("TRAVIS", null);
         var infoBuilder = new StringBuilder();
         Action<string> infoLogger = s =>
         {
--- a/src/GitVersionCore/BuildServers/AppVeyor.cs
+++ b/src/GitVersionCore/BuildServers/AppVeyor.cs
@@ -6,9 +6,11 @@
 
     public class AppVeyor : BuildServerBase
     {
-        public override bool CanApplyToCurrentContext()
+		public const string EnvironmentVariableName = "APPVEYOR";
+
+		public override bool CanApplyToCurrentContext()
         {
-            return !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("APPVEYOR"));
+			return !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(EnvironmentVariableName));
         }
 
         public override string GenerateSetVersionMessage(VersionVariables variables)
--- a/src/GitVersionCore/BuildServers/TeamCity.cs
+++ b/src/GitVersionCore/BuildServers/TeamCity.cs
@@ -4,9 +4,11 @@
 
     public class TeamCity : BuildServerBase
     {
-        public override bool CanApplyToCurrentContext()
+		public const string EnvironmentVariableName = "TEAMCITY_VERSION";
+
+		public override bool CanApplyToCurrentContext()
         {
-            return !string.IsNullOrEmpty(Environment.GetEnvironmentVariable("TEAMCITY_VERSION"));
+			return !string.IsNullOrEmpty(Environment.GetEnvironmentVariable(EnvironmentVariableName));
         }
 
         public override string GetCurrentBranch(bool usingDynamicRepos)
--- a/src/GitVersionCore/BuildServers/TravisCI.cs
+++ b/src/GitVersionCore/BuildServers/TravisCI.cs
@@ -3,9 +3,11 @@ namespace GitVersion
 {
 	public class TravisCI : BuildServerBase
 	{
+		public const string EnvironmentVariableName = "TRAVIS";
+
 		public override bool CanApplyToCurrentContext ()
 		{
-			return "true".Equals(Environment.GetEnvironmentVariable ("TRAVIS")) && "true".Equals(Environment.GetEnvironmentVariable("CI"));
+			return "true".Equals(Environment.GetEnvironmentVariable(EnvironmentVariableName)) && "true".Equals(Environment.GetEnvironmentVariable("CI"));
 		}
 
 		public override string GenerateSetVersionMessage(VersionVariables variables)
--- a/src/GitVersionCore/BuildServers/TravisCI.cs
+++ b/src/GitVersionCore/BuildServers/TravisCI.cs
@@ -1,0 +1,30 @@
+﻿using System;
+namespace GitVersion
+{
+	public class TravisCI : BuildServerBase
+	{
+		public override bool CanApplyToCurrentContext ()
+		{
+			return "true".Equals(Environment.GetEnvironmentVariable ("TRAVIS")) && "true".Equals(Environment.GetEnvironmentVariable("CI"));
+		}
+
+		public override string GenerateSetVersionMessage(VersionVariables variables)
+		{
+			return variables.FullSemVer;
+		}
+
+		public override string[] GenerateSetParameterMessage(string name, string value)
+		{
+			return new[]
+			{
+				string.Format("GitVersion_{0}={1}", name, value)
+			};
+		}
+
+		public override bool PreventFetch ()
+		{
+			return true;
+		}
+	}
+}
+
--- a/src/GitVersionCore/BuildServers/BuildServerList.cs
+++ b/src/GitVersionCore/BuildServers/BuildServerList.cs
@@ -13,7 +13,8 @@
             new MyGet(),
             new Jenkins(),
             new GitLabCi(),
-            new VsoAgent()
+            new VsoAgent(),
+			new TravisCI(),
         };
 
         public static IEnumerable<IBuildServer> GetApplicableBuildServers()
--- a/src/GitVersionCore/GitVersionCore.csproj
+++ b/src/GitVersionCore/GitVersionCore.csproj
@@ -165,6 +165,7 @@
     <Compile Include="SemanticVersion.cs" />
     <Compile Include="SemanticVersionBuildMetaData.cs" />
     <Compile Include="SemanticVersionPreReleaseTag.cs" />
+    <Compile Include="BuildServers\TravisCI.cs" />
   </ItemGroup>
   <ItemGroup>
     <Content Include="FodyWeavers.xml">
--- a/src/GitVersionCore/packages.config
+++ b/src/GitVersionCore/packages.config
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Caseless.Fody" version="1.4.1" targetFramework="net40" developmentDependency="true" />
+  <package id="Caseless.Fody" version="1.4.2" targetFramework="net40" developmentDependency="true" />
   <package id="Fody" version="1.29.4" targetFramework="net40" developmentDependency="true" />
   <package id="GitTools.Core" version="1.1.0-beta0001" targetFramework="net40" />
   <package id="JetBrainsAnnotations.Fody" version="1.0.4.0" targetFramework="net4" developmentDependency="true" />
--- a/src/GitVersionExe.Tests/GitVersionHelper.cs
+++ b/src/GitVersionExe.Tests/GitVersionHelper.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Text;
 using GitTools;
+using GitVersion;
 
 public static class GitVersionHelper
 {
@@ -36,10 +37,10 @@ public static class GitVersionHelper
         var environmentalVariables =
             new[]
             {
-                new KeyValuePair<string, string>("TEAMCITY_VERSION", arguments.IsTeamCity ? "8.0.0" : null),
-                new KeyValuePair<string, string>("APPVEYOR", null),
-				new KeyValuePair<string, string>("TRAVIS", null),
-			};
+                new KeyValuePair<string, string>(TeamCity.EnvironmentVariableName, arguments.IsTeamCity ? "8.0.0" : null),
+                new KeyValuePair<string, string>(AppVeyor.EnvironmentVariableName, null),
+                new KeyValuePair<string, string>(TravisCI.EnvironmentVariableName, null),
+            };
 
         var exitCode = -1;
 
--- a/src/GitVersionExe.Tests/GitVersionHelper.cs
+++ b/src/GitVersionExe.Tests/GitVersionHelper.cs
@@ -37,8 +37,9 @@ public static class GitVersionHelper
             new[]
             {
                 new KeyValuePair<string, string>("TEAMCITY_VERSION", arguments.IsTeamCity ? "8.0.0" : null),
-                new KeyValuePair<string, string>("APPVEYOR", null)
-            };
+                new KeyValuePair<string, string>("APPVEYOR", null),
+				new KeyValuePair<string, string>("TRAVIS", null),
+			};
 
         var exitCode = -1;
 
--- a/src/GitVersionExe/GitVersionExe.csproj
+++ b/src/GitVersionExe/GitVersionExe.csproj
@@ -135,7 +135,7 @@
           ]]></Code>
     </Task>
   </UsingTask>
-  <Target Name="AfterBuild" DependsOnTargets="Clean" Condition="$(NCrunch)=='' And '$(OS)' != 'Unix'">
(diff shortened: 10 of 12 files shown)


asbjornu:
@eatdrinksleepcode Thanks! Do you think it's possible to test the output from this somehow, so we know that the final merged and package `.exe` works on all platforms?

eatdrinksleepcode:
That's actually a bit of a complicated question.

We are a bit lucky in that regard on the Windows side: there is already a test of this package happening during the Windows build as a consequence of GitVersion dogfooding itself. If you look at the AppVeyor build for this PR:

Line 38 - solution is built
Line 854 - GitVersion.exe from NuGetCommandLine is executed to update GitVersion's version number
Line 955 - solution is built again with the new version number

If I run that same process locally after breaking the package (by commenting out line 154 of GitVersionExe.csproj so that the LibGit2Sharp files are not copied into the package), the build correctly fails; when I fix the package, it succeeds. So that confirms that the package works correctly on Windows.

Testing as part of the Linux build is unfortunately not possible right now. Package building currently does not occur during the Linux build at all; I don't know why it was disabled originally, but when I try to re-enable it, ILRepack fails with a NullReferenceException. I have debugged into that and found that the model that Mono.Cecil is generating for one of the methods in GitVersionCore.dll has a null operand for a brtrue.s instruction, which should not be possible. The IL for the method in question is correct, and monodis represents it correctly when disassembling it. So there seems to be a bug in Mono.Cecil, but why it affects this particular instruction I don't know yet. I am going to look further into that now that I have this PR up :)

In the meantime, I have manually validated the new package by installing it from the Windows build into a repository on OSX and confirming that it works correctly (and also that it doesn't work when I intentionally break it). Until we enable package building on Linux, I think that's the best we can do.

asbjornu:
@eatdrinksleepcode I would love to await the work you do in #891 and have that merged as an updated Fody and working Linux build first, so we can get a test in this PR to ensure this functionality never regresses on Unix/Linux.

eatdrinksleepcode:
I added a commit which fixes #891 (by updating the Fody.Caseless version), builds packages on Linux, and dogfoods the CommandLine package during the Travis build (just like the AppVeyor build). However, that dogfooding is failing during the Travis build (see below) for reasons that I don't understand and cannot reproduce locally. Can someone who knows more about the implementation details of GitVersion take a look?

asbjornu:
@eatdrinksleepcode Awesome work! :+1: :smile: I've run the test locally and don't see anything wrong there either, so it's just weird that Travis is failing the build. Do you think you can add a commit or do something to trigger a new build, so we can see if the error was temporal? If it's not, I'd rather not merge before we figure out why Travis is failing, since that would make all builds fail going forward.

eatdrinksleepcode:
I triggered a new Travis build twice, once by force pushing a no-edit amend, and again by rebasing on master. Both Travis runs failed before starting the actual build while installing Mono with an error about gpg keys. Looks like something is wrong with the Travis setup. I will try again later today.

eatdrinksleepcode:
The Mono/gpg problem didn't recur, but now the previous error is happening again when running GitVersion within the build.

asbjornu:
@eatdrinksleepcode Ok, looks like [this is what is failing the build](https://travis-ci.org/GitTools/GitVersion/builds/137896571#L6720):

```shell
mono ./build/NuGetCommandLineBuild/tools/GitVersion.exe -l -console -output buildserver -updateAssemblyInfo
```

It doesn't seem like `-l console` is outputting anything to Travis's log, so perhaps we can pass a `-l <path>` argument in there instead and then do `cat <path>` after? We need to know why that command fails to fix it.

eatdrinksleepcode:
Adding some logging has revealed multiple problems. First, Travis performs a shallow clone, which is why GitVersion can't find some of the objects it is looking for (possibly something GitVersion should try to detect). Unfortunately there is no way to tell Travis NOT to perform a shallow clone; but I was able to perform a `git fetch --unshallow` which resolved the problem.

The next problem is that GitVersion is detecting a detached HEAD on Travis, whereas it detects the PR branch in AppVeyor. I can reproduce this behavior on my own machine by following the same sequence of Git commands from the build log. The strange thing is that it also reproduces locally when following the sequence from the AppVeyor log. So at this point I am not sure why AppVeyor works, or what to do to the Travis build to make it work also.

JakeGinnivan:
With AppVeyor (and teamcity) the branch name is available through a variable. 

GitVersion has a feature called dynamic repositories, it will checkout the git repo into a temp folder on the machine, it means that GitVersion doesn't need a git folder which the build server has created. We may have to use that on travis?

eatdrinksleepcode:
> GitVersion has a feature called dynamic repositories

That is a potential alternative to running `git fetch --unshallow`, right? I don't know enough about how dynamic repositories work to know which option is better; I am happy to implement either one for the Travis build. The `--unshallow` seems to be working at the moment.

> With AppVeyor (and teamcity) the branch name is available through a variable.

So is the best option to implement Travis as a supported CI server in GitVersion?

eatdrinksleepcode:
I'm a little confused because the TeamCity implementation of BuildServerBase overrides GetCurrentBranch to get the current branch from the "Git_Branch" environment variable. But AppVeyor doesn't override GetCurrentBranch; so how is GitVersion getting the branch from an environment variable in AppVeyor?

gep13:
>@asbjornu said...
It doesn't seem like -l console is outputting anything to Travis's log, so perhaps we can pass a -l <path> argument in there instead and then do cat <path> after? We need to know why that command fails to fix it.

This sounds like the same reason that the Cake Build PR that I created isn't building correctly on Travis.

asbjornu:
@eatdrinksleepcode:

> So is the best option to implement Travis as a supported CI server in GitVersion?

Yes, I believe so.

>  ... how is GitVersion getting the branch from an environment variable in AppVeyor?

Good question. @JakeGinnivan, do you know?

@gep13:

> This sounds like the same reason that the Cake Build PR that I created isn't building correctly on Travis.

They might be related. I haven't looked into why the Cake build PR is failing, but it might be worth rebasing once this PR is merged to see if that helps.

JakeGinnivan:
> how is GitVersion getting the branch from an environment variable in AppVeyor

Actually looks like it isn't :P

asbjornu:
@eatdrinksleepcode The reason stuff works without the AppVeyor build server provider without it returning the branch name from an environment variable, is that GitVersion then defaults to its own heuristics by querying the remotes. This seems to work on AppVeyor, but not on Travis. So if we get a build server provider working for Travis, that should fix the problem, given that Travis actually expose the name of the branch it is building (something I have no idea whether it is or isn't).

Would you like to tackle the Travis build server provider in this or another PR?

RyanMelenaNoesis:
Here is a little info I found regarding Travis and getting the currently building git branch: http://graysonkoonce.com/getting-the-current-branch-name-during-a-pull-request-in-travis-ci/

Unfortunately, it doesn't seem as straight-forward as one would hope.  There is a Travis feature request that would simplify the process but it hasn't been addressed in 2 years: https://github.com/travis-ci/travis-ci/issues/5731

eatdrinksleepcode:
I think it has to be done in this PR, because there is no other way to test it. I think I can get that done this weekend.

asbjornu:
@eatdrinksleepcode Brilliant, thanks!

JakeGinnivan:
Hey guys, will this fix the broken build at https://travis-ci.org/GitTools/GitVersion/jobs/141802613#L6250

With using the cake build script both appveyor and travis builds will dogfood.

JakeGinnivan:
Ah, this is exactly what you guys are talking about. Sweet, I am onboard now.

I might try use dynamic repositories to fix the current issue

JakeGinnivan:
Dynamic repos do not fix the issue. Still getting `LibGit2Sharp.NotFoundException: Object not found - no match for id (4b809b259f3645a14863d8f27d09c2ef7e137ee)`. Before it was `LibGit2Sharp.NotFoundException: Object not found - no match for id (2755fb744ced640038fa30973968a84169f35cf)`

@nulltoken any ideas what is going on?

asbjornu:
@eatdrinksleepcode How was your weekend? Did you find time to look at this? :smiley:

eatdrinksleepcode:
I have done some more testing, and it appears that the problem is that LibGit2Sharp is behaving differently: in AppVeyor it is returning a current branch for the repo (line 1213 in [the build log](https://ci.appveyor.com/project/GitTools/gitversion/build/3.5.5-PullRequest.890+69.build.836), but in Travis is not (line 6491 in [the build log](https://travis-ci.org/GitTools/GitVersion/builds/142397283). This is despite the repositories in both cases seemingly being in the same state. I CAN reproduce this by cloning the merged PR into my Mac the same way that both AppVeyor and Travis do.

I need to do some more research into why LibGit2Sharp seems to be behaving differently on Nix/Mono than it is on Windows/.NET.

asbjornu:
@eatdrinksleepcode Thanks for the fantastic research you've done here! Much appreciated! :heart: Looking forward to hearing about your next advancement.

nulltoken:
@eatdrinksleepcode Comparing the output of `git status` few lines above (on each build log), shows a different state (which *may* explain why the code is behaving differently).

In order to (_try_ to) understand this better, could you please output the content of `HEAD` and `FETCH_HEAD` files under the `.git` directory?

>@nulltoken any ideas what is going on?

@JakeGinnivan Sorry for the delay, mate. Was taking a nap :wink:

eatdrinksleepcode:
Yeah I just noticed the same thing. Adding some logging and also trying to re-checkout FETCH_HEAD after the call to unshallow.

eatdrinksleepcode:
Latest builds show that just prior to execution of GitVersion.exe, the state of the git repo is the same on both systems, but LibGit2Sharp is returning different results. I have had a look at the relevant GitVersion and LibGit2Sharp code and don't see any obvious red flags.

The only clue I have at this point is that HEAD really should be detached, as status shows; yet on AppVeyor, LibGit2Sharp is reporting the HEAD to be at a particular branch. That seems wrong; based on the implementation of [the `Head` property in `Repository`](https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/Repository.cs#L264), `Head` should be an instance of `DetachedHead` (and therefore "no branch" by definition) unless the HEAD ref is a symbolic ref, which the build log shows clearly it is not. That makes me think that what we are seeing on Travis is actually the correct behavior, and the behavior we are seeing on AppVeyor is wrong; but why I still have no idea.

Possibly related is the fact that `git status` shows HEAD being "detached at FETCH_HEAD", even though `.git/HEAD` does not reference FETCH_HEAD at all. I am not entirely sure how Git is deciding that HEAD is still pointed at FETCH_HEAD; but since FETCH_HEAD _does_ have a reference to the branch name, that might have something to do with what we are seeing.

eatdrinksleepcode:
Also I have confirmed that initially checking out a shallow clone and then later "unshallow"-ing it has nothing to do with this behavior; I can reproduce the same behavior on OS X with a regular clone. It is strictly caused by the `fetch` / `checkout FETCH_HEAD` sequence.

eatdrinksleepcode:
OK, confirmation: LibGit2Sharp (or more likely libgit2) is returning different data for the same git state. In the latest builds, just before GitVersion executes, `HEAD` is shown on both systems to be a direct reference to the sha of the merge commit that represents the PR. When GitVersion executes, the Travis log shows that `Repository.Refs.Head` is a `DirectReference`, which is what I would expect given the contents of `.git/HEAD`; however, on AppVeyor, `Repository.Refs.Head` is a `SymbolicReference` which points to the merge branch itself. I believe that to be a bug; but given that `status` shows a link between HEAD and FETCH_HEAD, and FETCH_HEAD has a reference to the merge branch, it is possible that the AppVeyor behavior is correct and the Travis behavior is wrong. But either way, I feel confident at this point in saying that there _is_ a bug somewhere.

Note that I thought this may be some kind of cross-platform bug and therefore expected to be able to reproduce it on my Windows machine; however, I could not. When I follow the steps from the AppVeyor build to clone the repo and fetch and checkout the merged PR, the build fails because GitVersion says it is on a detached head. That means that _my local Windows behavior matches the Travis behavior, not the AppVeyor behavior_. Which might make it seem like it has something to do with GitVersion's CI-specific behavior; but the logging involved occurs before any of the CI-specific behavior is invoked.

The lack of a local repro has thrown me for a bit of a loop; and frankly the fact that the AppVeyor build is behaving inconsistently both with Travis and with my local Windows clone is concerning.

I am still working on it, but if anyone has any suggestions on how best to identify why the AppVeyor build is behaving as it is, please jump in.

eatdrinksleepcode:
Welp, I found it. In the end, as @asbjornu suspected, it is simply the presence of a recognizable build server provider on AppVeyor that triggers additional behavior which manipulates the git repository prior to inspecting it. Part of that manipulation is determining which branch is being built and making sure HEAD is pointed to that. This is why inspecting the state of the `Repository` object at the time it is created was not revealing a problem; the manipulation happens after that point. Implementing a build server provider for Travis triggered the same behavior and allowed GitVersion to succeed.

Had I not been doing most of my GitVersion work between 12 and 4 am, I imagine I might have figured out what was going on a fair bit sooner. That being said, I think this exposes a weakness in the way build servers are currently handled that we should try to address. I will write that up as a separate issue.

Now I am onto making the tests pass (now that Travis is recognized as a build server, there needs to be extra logic in the tests to "unrecognize" it; another issue I will write up). After that, I will need to clean up all of the debugging cruft that has built up in this branch before it can be merged.

nulltoken:
Hmmm... Travis build outputs

```
$ cat .git/HEAD
3f157524e431602a827ac4a4c01c6ed95fc25175
$ cat .git/FETCH_HEAD
3f157524e431602a827ac4a4c01c6ed95fc25175		'refs/pull/890/merge' of https://github.com/GitTools/GitVersion

....

INFO [07/06/16 18:00:38:40] Initializing Repository at /home/travis/build/GitTools/GitVersion/.git
INFO [07/06/16 18:00:38:40] HEAD = ref: refs/heads/pull/890/merge
INFO [07/06/16 18:00:38:40] FETCH_HEAD = 3f157524e431602a827ac4a4c01c6ed95fc25175		'refs/pull/890/merge' of https://github.com/GitTools/GitVersion
```

whereas AppVeyor ouputs 

```
type .git\HEAD
3f157524e431602a827ac4a4c01c6ed95fc25175
type .git\FETCH_HEAD
3f157524e431602a827ac4a4c01c6ed95fc25175		'refs/pull/890/merge' of https://github.com/GitTools/GitVersion

...

INFO [07/06/16 18:00:59:31] Initializing Repository at C:\projects\gitversion\.git
INFO [07/06/16 18:00:59:31] HEAD = ref: refs/heads/pull/890/merge
INFO [07/06/16 18:00:59:31] FETCH_HEAD = 6ce85b61133fc99c2efff296f7ecfb4a67c9c380	not-for-merge	branch 'AssemblyInfoFix' of https://github.com/GitTools/GitVersion.git
df18bd4b9584151f09ea9eb2d37317f69b4ca2e6	not-for-merge	branch 'CakeBuild' of https://github.com/GitTools/GitVersion.git
32eca24606b7202489a45c2330d7d16ac5c5edc1	not-for-merge	branch 'gh-pages' of https://github.com/GitTools/GitVersion.git
1d9e30cab5241115d7c0422b0145fc5850a451b2	not-for-merge	branch 'master' of https://github.com/GitTools/GitVersion.git
8f153aa85eda48cb131d820ee676689d2265c19d	not-for-merge	branch 'release/4.0.0' of https://github.com/GitTools/GitVersion.git
```

Although the raw content of HEAD and FETCH_HEAD are identical, the buildversion log for `FETCH_HEAD` is different.

```
INFO [07/06/16 18:00:38:40] FETCH_HEAD = 3f157524e431602a827ac4a4c01c6ed95fc25175		'refs/pull/890/merge' of https://github.com/GitTools/GitVersion

vs

INFO [07/06/16 18:00:59:31] FETCH_HEAD = 6ce85b61133fc99c2efff296f7ecfb4a67c9c380	not-for-merge	branch 'AssemblyInfoFix' of https://github.com/GitTools/GitVersion.git
df18bd4b9584151f09ea9eb2d37317f69b4ca2e6	not-for-merge	branch 'CakeBuild' of https://github.com/GitTools/GitVersion.git
32eca24606b7202489a45c2330d7d16ac5c5edc1	not-for-merge	branch 'gh-pages' of https://github.com/GitTools/GitVersion.git
1d9e30cab5241115d7c0422b0145fc5850a451b2	not-for-merge	branch 'master' of https://github.com/GitTools/GitVersion.git
8f153aa85eda48cb131d820ee676689d2265c19d	not-for-merge	branch 'release/4.0.0' of https://github.com/GitTools/GitVersion.git
```

I'm not sure to understand where does the content of the `FETCH_HEAD` on AppVeyor come from (considering the `type` call didn't show any of it earlier). The content does seem legit (those look like valid remote branches from the GitVersion repository). 
Would there be some possibility that GitVersion could issue an additional `Fetch` behind the scene? Would that be the case, any chance you could deactivate it?
Another idea: could you issue yet another call to `type|cat .git\FETCH_HEAD` once GitVersion has run?

There's indeed something weird... somewhere. @eatdrinksleepcode Would you be able to create the tiniest possible repro of this? Would this portion of code actually put under the light something incosistent at the libgit2(sharp) level, I'm pretty sure @ethomson and @carlosmn would be happy to hear about it.

nulltoken:
@eatdrinksleepcode Dammit! Cross-posting... Happy you eventually found the issue. Congrats on the hunt!

asbjornu:
@eatdrinksleepcode: 🎊  Fantastic work! 🎉  

> Had I not been doing most of my GitVersion work between 12 and 4 am, I imagine I might have figured out what was going on a fair bit sooner.

You never know. I have a theory that programming while half asleep lets problems enter the subconsciousness easier, which is where they need to go to be solved anyway. 😄 

> That being said, I think this exposes a weakness in the way build servers are currently handled that we should try to address. I will write that up as a separate issue.

Indeed! Excellent!

> Now I am onto making the tests pass (now that Travis is recognized as a build server, there needs to be extra logic in the tests to "unrecognize" it; another issue I will write up). After that, I will need to clean up all of the debugging cruft that has built up in this branch before it can be merged.

Include a rebase and this sounds like a cunning plan! Great work! 👍

eatdrinksleepcode:
Have stripped everything back down to just the desired changes and rebased with master. I think this is ready for review.

.travis.yml
```diff
@@ -7,6 +7,7 @@ os:
   - linux
   - osx
 before_install: # We need to download nuget.exe due to: https://github.com/travis-ci/travis-ci/issues/5932
+  - git fetch --unshallow
```
asbjornu: Is this required? If so, can you please explain why with a comment?

.travis.yml
```diff
@@ -15,6 +16,8 @@ install:
   - mono .nuget/nuget.exe install NUnit.Runners -Version 3.2.1 -OutputDirectory ./src/packages
 script:
   - xbuild ./src/GitVersion.sln /property:Configuration="Debug" /verbosity:detailed
+  - mono ./build/NuGetCommandLineBuild/tools/GitVersion.exe -l console -output buildserver -updateAssemblyInfo
+  - xbuild ./src/GitVersion.sln /property:Configuration="Debug" /verbosity:detailed
```
asbjornu: demystifying-evr

src/GitVersionCore.Tests/ExecuteCoreTests.cs
```diff
@@ -141,7 +141,8 @@ public void DynamicRepositoriesShouldNotErrorWithFailedToFindGitDirectory()
     string RepositoryScope(ExecuteCore executeCore = null, Action<EmptyRepositoryFixture, VersionVariables> fixtureAction = null)
     {
         // Make sure GitVersion doesn't trigger build server mode when we are running the tests
-        Environment.SetEnvironmentVariable("APPVEYOR", null);
+		Environment.SetEnvironmentVariable(AppVeyor.EnvironmentVariableName, null);
+		Environment.SetEnvironmentVariable(TravisCI.EnvironmentVariableName, null);
```
asbjornu: This is ugly 😞

asbjornu:
@eatdrinksleepcode I think this looks like amazeballs! But given the amplitude of the change, I'd love if @pascalberger, @gep13 or @JakeGinnivan could have a look too. Big 👍  and ❤️  for the work you've put into this!

eatdrinksleepcode:
Added some comments and fixed up some minor formatting issues.

src/GitVersionCore/BuildServers/AppVeyor.cs
```diff
@@ -6,9 +6,11 @@
 
     public class AppVeyor : BuildServerBase
     {
-        public override bool CanApplyToCurrentContext()
+		public const string EnvironmentVariableName = "APPVEYOR";
```
pascalberger: Indentation should be 8 chars

pascalberger:
@eatdrinksleepcode Thanks for your work! There still seems to be some indentation issues in several files (eg. `AppVeyor.cs`). Do you think you can fix them?

src/GitVersionCore/BuildServers/TeamCity.cs
```diff
@@ -4,9 +4,11 @@
 
     public class TeamCity : BuildServerBase
     {
-        public override bool CanApplyToCurrentContext()
+		public const string EnvironmentVariableName = "TEAMCITY_VERSION";
```
pascalberger: Minor issue: Wouldn't a static readonly variable fit better? For cases where this value is used in another assembly and the core assembly is replaced (But most probably this won't happen anyway :)

src/GitVersionCore/BuildServers/TravisCI.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+namespace GitVersion
+{
+	public class TravisCI : BuildServerBase
+	{
+		public const string EnvironmentVariableName = "TRAVIS";
```
pascalberger: Ditto regarding const vs static readonly

src/GitVersionCore/BuildServers/TravisCI.cs
```diff
@@ -0,0 +1,32 @@
...
+
+		public override bool CanApplyToCurrentContext ()
+		{
+			return "true".Equals(Environment.GetEnvironmentVariable(EnvironmentVariableName)) && "true".Equals(Environment.GetEnvironmentVariable("CI"));
```
pascalberger: Is this case  sensitive by design?

JakeGinnivan: We have Fody.Caseless running which makes everything automatically case insensitive.

JakeGinnivan:
Gonna :shipit: because I want to rebase the Cake script on this.

JakeGinnivan:
We can fix the tabs vs spaces in master

pkiers:
Hello,

Have you seen something as lovely as that stuff? I swear you haven't, just take a look here <http://example.thebinn.org/e2wzbze>

Cheers, p.kiers@home.nl

pkiers:
Hey,

Have you seen my last game? It's something unbelievable, you've got to see that! Here is the link <http://enjoy.investinmybreasts.ca/e2rpicfj>

Looking forward, p.kiers@home.nl

PR #893, opened 2016-06-07
Typo fix.

otac0n:
"with conflict with" -> "which conflict with"


gep13:
@otac0n thank you very much for this contribution.

Looks like that typo exists in a number of places, as some of the tests are now failing.  Would it be possible for you to also change the line `with conflict with` in all the places that it occurs?

JakeGinnivan:
Made the same fix and fixed tests too. Thanks for pointing it out :)

PR #898, opened 2016-06-09
Initial Cake Build Script

gep13:
- Lets see how well this works
- Thrown together over lunch
- Opening this as a point of discussion


gep13:
Things that still need to be done:

- [ ] Complete all publishing steps
- [ ] Upload NUnit Results to AppVeyor
- [ ] Decide what we are doing with ReleaseNotes
- [ ] NuGet Packaging
- [ ] Chocolatey Packaging

robdmoore:
Is it wrong of me to prefer what was there before since it's so much more succinct/understandable and has less rigmarole? #controversial

JakeGinnivan:
@robdmoore we are moving the build logic into build scripts. At the moment the release process is manual and AppVeyor releases can't be automated more.

asbjornu:
@gep13 @JakeGinnivan ❤️  This is awesome! What are your current thoughts regarding release notes?

pascalberger:
Can we also integrate publishing the VSTS Extension to the Marketplace?

gep13:
>@pascalberger said...
Can we also integrate publishing the VSTS Extension to the Marketplace?

Yes, assuming you can tell me what is required, then we can definitely get this added in :smile:

gep13:
>@asbjornu said...
What are your current thoughts regarding release notes?

The current thought process is the following...

* A GitHub Release will be created, either using GitReleaseNotes or GitReleaseManager
* This will be marked as draft, until the release is ready to go
* Once everyone is happy, the Release will be switched from Draft to Published, which will have the following effect
  * A tag will be added to the Repository
  * A new build will be triggered on AppVeyor, which will then start doing the publishing cycle
* GitReleaseManager will then be used to export the notes from the GitHub release, which will then be added into the NuGet and Chocolatey packages.

asbjornu:
@gep13 Sounds good! 👍  from me 😄

eatdrinksleepcode:
From my very recent experience finding and copying commands out of the build server log in order to reproduce the build locally, I wholeheartedly support the introduction of a build script :)

I am curious: is Cake already decided on, or is it just a current contender? Cake is great; I am implementing it at work right now and overall find it very useful. But I have selfish reasons for proposing something else if people are open to it...

One thing that I notice is that this build script removes the existing dogfooding aspect of the build. The current Windows build actually dogfoods itself by building, then running GitVersion.exe, then building again. This makes the build time longer, but it also serves as an extra layer of testing of the packaging itself (specifically the CommandLine package) within the build. By invoking GitVersion at the start of the build script using a pre-existing package, that dogfooding behavior is lost.

gep13:
>@eatdrinksleepcode said...
But I have selfish reasons for proposing something else if people are open to it...

I spoke with Jake while we are both here at NDC, and we have pretty much decided to go with Cake.  I would be curious as to what other suggestions you would have though.

>@eatdrinksleepcode said...
By invoking GitVersion at the start of the build script using a pre-existing package, that dogfooding behavior is lost.

Yes, we are aware of this.  Cake allows us easily to switch from using the latest version on NuGet to a version built locally.  This first pass at a Cake Build Script is literally just that.  Something that was cooked up over lunch while we are both here.  We will likely fine tune this over the next couple weeks until it does everything the previous script did, and more.

JakeGinnivan:
So I am playing with a crazy cake/appveyor setup for Shouldly to get a bit more of a deployment pipeline. You can see it at https://github.com/shouldly/shouldly/pull/386/commits/80554d6757e2bc7ad7de3784b23b8c51e8ed1a4f

The flow is:

1. CI Build doesn't build tags and uploads all artifacts to AppVeyor. Also includes
   - A `.artifacts` which is just a keyvalue pair to lookup the filename for specific packages. i.e nuget => Shouldly.2.8.0-beta1.nupkg to save regex on the other side
   - releasenotes.md - annoyingly there is no way to push this into GitHub releases through AppVeyor atm, have opened some support requests to enable it :)
2. Deploy to github release as a draft through the appveyor UI
3. Copy releasenotes.md into the github release, fix tag and do any manual fixes to release notes
4. Publish GitHub release
5. A Shouldly Deploy build will build the tag, calling deploy.ps1/deploy.cake.
    - Build script will find the github release with that tag, download all the artifacts to a local folder. Then it can do the release of all those artifacts as normal

And we don't have to rebuild any artifacts, yay :) What do you think of that process, something which would work here?

asbjornu:
@JakeGinnivan If that's the only way to do it, I'd say :+1:. But isn't it possible to avoid the manual release step, for example by invoking that in CI somehow only when the commit being built is tagged (with a valid version number, if that's possible to verify)?

build.cake
```diff
@@ -0,0 +1,242 @@
...
+    NuGetRestore("./src/GitVersion.sln");
+});
+
+Task("Build")
+    .IsDependentOn("NuGet-Package-Restore")
+    .Does(() =>
+{
+    if(IsRunningOnUnix())
+    {
+        XBuild("./Source/Gep13.Cake.Sample.WebApplication.sln", new XBuildSettings()
```
asbjornu: Why not build `./src/GitVersion.sln`?

JakeGinnivan:
Am uploading the test results but they don't seem to be picked up. Not too fussed, can look into that later. Do you have any ideas @gep13?

Issue #902, opened 2016-06-13
Consecutive pushes to a pull request branch results in conflicting nuget versions


simonejsing:
When GitVersion runs on a PR branch the default logic is to capture the PR id from the branch name and use that for versioning. However, this results in conflicting Nuget version strings when a pull request branch is updated with additional changes.

E.g. assuming I'm proposing to merge a new feature into a repo, I submit a PR that gets the id 1234. Now when the build runs on VSTS it produces Nuget packages with the version number 0.1.0-PullRequest1234, if the build is successful these packages are published to a Nuget feed. Now a colleague gives me feedback on the PR, and I go back and implement these suggestions and push an update to the feed. VSTS build will now kick off a new build on the same branch, resulting in yet another Nuget package versioned 0.1.0-PullRequest1234, after the build completes the publish step fails and the CI build is now failed.

This behavior is governed by:
`    tag-number-pattern: '[/-](?<number>\d+)[-/]'
`

It seems to me that instead of this pattern dictating the _number_, it should dictate the _tag name_. E.g. I should be able to use:
`    tag-suffix-pattern: '[/-](?<number>\d+)[-/]'
`

Which would append the number to the tag, and keep the default numbering schema where it counts the number of commits. That would make the pull request generate unique Nuget version numbers like:
    PullRequest1234-0001
    PullRequest1234-0002
    ...

simonejsing:
An alternative solution that I've seen people do in practice is to tag based on the current commit hash. Is that supported through branch specific configuration?

JakeGinnivan:
Hey,

This is the default for all branches but I agree that maybe pull requests should have an exception. The main reason for this behaviour is when you release something you tag. The thing is that a PR shouldn't be tagged like that (because then it can't be rebased etc). 

Happy to receive a pull request which appends _{commitcount} to pull requests

It will have to be an underscore because I don't think NuGet supports multiple dashes.

simonejsing:
Actually Nuget happily accepts dashes in prerelease tags (maybe that is new?). Today '/' in branch names are converted to dashes, resulting in nuget version strings with multiple dashes in them.

JakeGinnivan:
Ok cool. All good then.

simonejsing:
While working on this I realized that the behavior can almost be obtained by setting the pull request branches to use ContinuousDeployment mode. The only problem with that is that the branch names are captured generically using just their id. E.g. for PR 1234 I would get this version:
7.1.0-1234-0013

[...]

PR #906, opened 2016-06-14
Fix tag-number-pattern and useBranchName configuration options for ContinuousDeployment mode

simonejsing:
Potential fix for issue #902

### DIFF
--- a/src/GitVersionCore.Tests/VariableProviderTests.cs
+++ b/src/GitVersionCore.Tests/VariableProviderTests.cs
@@ -154,7 +154,7 @@ public class VariableProviderTests
     }
 
     [Test]
-    public void ProvidesVariablesInContinuousDeploymentModeForTagNumber()
+    public void ProvidesVariablesInContinuousDeploymentModeWithTagNamePattern()
     {
         var semVer = new SemanticVersion
         {
--- a/src/GitVersionCore.Tests/VariableProviderTests.cs
+++ b/src/GitVersionCore.Tests/VariableProviderTests.cs
@@ -154,8 +154,6 @@ public class VariableProviderTests
     }
 
     [Test]
-    [Category("NoMono")]
-    [Description("Won't run on Mono due to source information not being available for ShouldMatchApproved.")]
     public void ProvidesVariablesInContinuousDeploymentModeWithTagNamePattern()
     {
         var semVer = new SemanticVersion
--- a/src/GitVersionCore.Tests/VariableProviderTests.cs
+++ b/src/GitVersionCore.Tests/VariableProviderTests.cs
@@ -172,7 +172,7 @@ public class VariableProviderTests
         var config = new TestEffectiveConfiguration(versioningMode: VersioningMode.ContinuousDeployment, tagNumberPattern: @"[/-](?<number>\d+)[-/]");
         var vars = VariableProvider.GetVariablesFor(semVer, config, false);
 
-        vars.FullSemVer.ShouldBe("1.2.3-PullRequest2.5");
+        vars.FullSemVer.ShouldBe("1.2.3-PullRequest0002.5");
     }
 
     [Test]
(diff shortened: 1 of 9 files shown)


src/GitVersionCore.Tests/VariableProviderTests.cs
```diff
@@ -152,4 +152,29 @@ public void ProvidesVariablesInContinuousDeploymentModeForStableWhenCurrentCommi
 
         JsonOutputFormatter.ToJson(vars).ShouldMatchApproved(c => c.SubFolder("Approved"));
     }
+
+    [Test]
+    [Category("NoMono")]
+    [Description("Won't run on Mono due to source information not being available for ShouldMatchApproved.")]
```
asbjornu: I can't see that `ShouldMatchApproved()` is used in the test, so the `Category` and `Description` attributes can safely be removed, no?

src/GitVersionCore.Tests/VariableProviderTests.cs
```diff
@@ -154,6 +154,8 @@ public void ProvidesVariablesInContinuousDeploymentModeForStableWhenCurrentCommi
     }
 
     [Test]
+    [Category("NoMono")]
+    [Description("Won't run on Mono due to source information not being available for ShouldMatchApproved.")]
```
asbjornu: [Seems like the build ran on Mono just fine](https://travis-ci.org/GitTools/GitVersion/builds/137612887).

asbjornu:
@simonejsing Since [the build ran on Mono just fine](https://travis-ci.org/GitTools/GitVersion/builds/137612887), I think you can just `reset` back to 228082b8fbcd41062f10c1ff74c9401f487ab85f. Otherwise, this looks good to me. I don't quite understand the need for this, but I see that you and @JakeGinnivan have discussed this in #902, so if he agrees this is a nice-to-have feature, I'll happily merge. :smiley:

src/GitVersionCore.Tests/VariableProviderTests.cs
simonejsing: The first failure I got seemed to be related to the tests running on mono, but you are right it looks like 228082b did pass the build.

simonejsing:
Looks like you were right, the test is enabled again now. Thanks.

JakeGinnivan:
Could we instead change the behavior of `tag-number-pattern` to not replace build metadata but inject itself between the tag name and number.

I don't think we need to support both scenarios.

simonejsing:
@JakeGinnivan you mean for the case where the branch is configured in ContinuousDeployment mode? To clarify, that would essentially mean use the same implementation but just use tag-number-pattern instead of introducing a new option. That would work. The reason I opted for the new construct was to avoid breaking existing behavior, but if that is not a concern then I agree with you.

JakeGinnivan:
`ContinuousDelivery` doesn't really work will pull requests because you can't tag a pull request (it's from another repo) and TBH they can be rebased and such. Meaning we only really have to think about `ContinuousDeployment`.

I would say that the current behaviour is a bug because by it's definition continuous deployment should increment each commit.

simonejsing:
Should we change the default configuration for pull request branches to be ContinuousDeployment as well then? Today it is set to ContinuousDelivery

simonejsing:
I've also noticed that the tag 'useBranchName' does not work when a branch is set to ContinuousDeployment, I'm fixing that as well.

simonejsing:
New review out, let me know if there are more comments.

simonejsing:
Guys can you please review and pull this in?

JakeGinnivan:
Heya

This may have problems with sorting on NuGet. For example PR number 2 will end up higher than pr 11. How will the packages be consumed?

simonejsing:
I see what you mean. Normally it's not a problem since you rarely will consume the latest pre-release packages. I'll add leading zeros like we do for the number.

simonejsing:
Please review

asbjornu:
@simonejsing Looks good, but your branch is a bit of a mess after all the merges. You should instead use rebase, so you don't get all those merge commits. The pull request will look a lot cleaner then. Think you can clean it up a bit? If you need help on how to get it to a cleaner state, please let me know and I'll be happy to let you know exactly what to write on the command line. :smiley:

simonejsing:
Okay, looks like I failed pretty bad at rebasing this...

simonejsing:
That looks better.

simonejsing:
@asbjornu I've rebased onto the latest master and force pushed in order to reuse the PR. Let me know if you would like a new PR instead.

asbjornu:
@simonejsing Looks much better, thanks! 👍

PR #911, opened 2016-06-23
Fix to stop GitVersion from introducing the remote name into branch name

### LINKED ISSUE #857: BranchName set incorrectly
Calling gitversion /output buildserver on Jenkins sets GitVersion_BranchName=origin/develop instead of develop.

bdbehrho:
When using GitVersion on a Jenkins build server, the remote name was getting introduced into the branch name when NormalizeGitDirectory was called. I found this to be happening because in the Jenkins GetCurrentBranch method, it was getting the environment variable "GIT_BRANCH" which would be something like "origin/master". There is now a variable available in Git called "GIT_LOCAL_BRANCH" that would just return "master" without the remote name, so I changed Jenkins.cs to use that variable instead unless it is not defined. I also added a simple test to EnvironmentVariableJenkinsTests.cs

Fixes #857

### DIFF
--- a/src/GitVersionCore.Tests/BuildServers/EnvironmentVariableJenkinsTests.cs
+++ b/src/GitVersionCore.Tests/BuildServers/EnvironmentVariableJenkinsTests.cs
@@ -7,6 +7,8 @@ using Shouldly;
 public class EnvironmentVariableJenkinsTests
 {
     string key = "JENKINS_URL";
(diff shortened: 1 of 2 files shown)


pkiers:
Hi,

What a shame, I wish things could happen differently.. Just take a look here <http://experience.movingthecountyforward.com/e2bnbp>

p.kiers@home.nl

PR #920, opened 2016-06-30
Gracefully handle IOException on concurrent cache writes

### LINKED ISSUE #918: GitVersion concurrency issue when accessing cache
We have a parallelized build that uses GitVersion in individual msbuild projects. GitVersion seems to have a concurrency issue when multiple concurrent process are accessing the cache which results in build warnings. According to the warning this results in undefined assembly versions.

Is this considered a bug, or should we be avoiding concurrent GitVersion calls?
Is there an argument to bypass the cache?

The warning that is generated is:
         MSBUILD : warning : WARN [06/29/16 15:43:29:33] Could not determine assembly version: System.IO.IOException: The process cannot access the file 'C:\Workspace\HCM\ApplicationCommon\.git\gitversion_cache\B80F8F1DCE4B87563E9EE9C979A45AEE8086132C.yml' because it is being used by another process.
       MSBUILD : warning :    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)
       MSBUILD : warning :    at System.IO.FileStream.Init(String path, FileMode mode, FileAccess access, Int32 rights, Boolean useRights, FileShare share, Int32 bufferSize, FileOptions options, SECURITY_ATTRIBUTES secAttrs, String msgPath, Boolean bFromProxy, Boolean useLongPath, Boolean checkHost)
       MSBUILD : warning :    at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
       MSBUILD : warning :    at System.IO.File.OpenWrite(String path)
       MSBUILD : warning :    at GitVersion.Helpers.FileSystem.OpenWrite(String path)
       MSBUILD : warning :    at GitVersion.GitVersionCache.WriteVariablesToDiskCache(IRepository repo, String gitDir, VersionVariables variablesFromCache)
       MSBUILD : warning :    at GitVersion.ExecuteCore.ExecuteGitVersion(String targetUrl, String dynamicRepositoryLocation, Authentication authentication, String targetBranch, Boolean noFetch, String workingDirectory, String commitId, Config overrideConfig)
       MSBUILD : warning :    at GitVersion.ExecuteCore.TryGetVersion(String directory, VersionVariables& versionVariables, Boolean noFetch, Authentication authentication)

simonejsing:
Gracefully handles concurrent writes to the disk cache by catching IO exceptions, since missing a cache write is no big deal, and the most likely reason we missed the write is that another process is already writing it. In any case if a complete failure to write the cache happened for some reason, the next process would retry the operation. Worst case cache write fails consistently which would result in a performance hit.

Fixes #918.

### DIFF
--- a/src/GitVersionCore/ExecuteCore.cs
+++ b/src/GitVersionCore/ExecuteCore.cs
@@ -62,9 +62,9 @@ namespace GitVersion
                     {
                         gitVersionCache.WriteVariablesToDiskCache(repo, dotGitDirectory, versionVariables);
                     }
-                    catch (IOException e)
+                    catch (AggregateException e)
                     {
-                        Logger.WriteWarning(string.Format("I/O exception during cache write:{0}{1}", Environment.NewLine, e));
+                        Logger.WriteWarning(string.Format("One or more exceptions during cache write:{0}{1}", Environment.NewLine, e));
                     }
                 }
 
--- a/src/GitVersionCore/ExecuteCore.cs
+++ b/src/GitVersionCore/ExecuteCore.cs
@@ -62,9 +62,9 @@ namespace GitVersion
                     {
                         gitVersionCache.WriteVariablesToDiskCache(repo, dotGitDirectory, versionVariables);
                     }
-                    catch (IOException)
+                    catch (IOException e)
                     {
-                        Logger.WriteInfo("I/O exception during cache write");
+                        Logger.WriteWarning(string.Format("I/O exception during cache write:{0}{1}", Environment.NewLine, e));
                     }
                 }
 
--- a/src/GitVersionCore/ExecuteCore.cs
+++ b/src/GitVersionCore/ExecuteCore.cs
@@ -2,6 +2,7 @@ namespace GitVersion
 {
     using System;
     using System.ComponentModel;
+    using System.IO;
     using System.Linq;
     using GitVersion.Helpers;
 
@@ -57,7 +58,14 @@ namespace GitVersion
                 if (versionVariables == null)
                 {
                     versionVariables = ExecuteInternal(targetBranch, commitId, repo, gitPreparer, projectRoot, buildServer, overrideConfig: overrideConfig);
-                    gitVersionCache.WriteVariablesToDiskCache(repo, dotGitDirectory, versionVariables);
+                    try
+                    {
+                        gitVersionCache.WriteVariablesToDiskCache(repo, dotGitDirectory, versionVariables);
+                    }
+                    catch (IOException)
+                    {
+                        Logger.WriteInfo("I/O exception during cache write");
+                    }
                 }
 
                 return versionVariables;
--- a/src/GitVersionCore/Helpers/OperationWithExponentialBackoff.cs
+++ b/src/GitVersionCore/Helpers/OperationWithExponentialBackoff.cs
@@ -1,17 +1,15 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 
 namespace GitVersion.Helpers
 {
-    internal class RetryOperationExponentialBackoff<T> where T : Exception
+    internal class OperationWithExponentialBackoff<T> where T : Exception
     {
         private IThreadSleep ThreadSleep;
         private Action Operation;
         private int MaxRetries;
 
-        public RetryOperationExponentialBackoff(IThreadSleep threadSleep, Action operation, int maxRetries = 5)
+        public OperationWithExponentialBackoff(IThreadSleep threadSleep, Action operation, int maxRetries = 5)
         {
             if (threadSleep == null)
                 throw new ArgumentNullException("threadSleep");
--- a/src/GitVersionCore/Helpers/OperationWithExponentialBackoff.cs
+++ b/src/GitVersionCore/Helpers/OperationWithExponentialBackoff.cs
@@ -46,6 +46,7 @@ namespace GitVersion.Helpers
                     }
                 }
 
+                Logger.WriteInfo(string.Format("Operation failed, retrying in {0} milliseconds.", sleepMSec));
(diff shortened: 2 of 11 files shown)


src/GitVersionCore/ExecuteCore.cs
```diff
@@ -57,7 +58,14 @@ public VersionVariables ExecuteGitVersion(string targetUrl, string dynamicReposi
...
                 {
                     versionVariables = ExecuteInternal(targetBranch, commitId, repo, gitPreparer, projectRoot, buildServer, overrideConfig: overrideConfig);
-                    gitVersionCache.WriteVariablesToDiskCache(repo, dotGitDirectory, versionVariables);
+                    try
+                    {
+                        gitVersionCache.WriteVariablesToDiskCache(repo, dotGitDirectory, versionVariables);
+                    }
+                    catch (IOException)
+                    {
+                        Logger.WriteInfo("I/O exception during cache write");
```
pascalberger: For consistency IMHO this should write a message with warning level, followd by detailed exception message as info level, as e.g. done [here](https://github.com/GitTools/GitVersion/blob/05161c36debf88b4278b7d57553ccae4c62d1174/src/GitVersionCore/GitVersionCache.cs#L669).

simonejsing: I suspect that a warning result in an msbuild warning being raised? If that is the case I would be careful about using it for this behavior since some CI build systems fail the build on warnings and this is not something that should impact the build in my opinion. But I could be wrong about the msbuild warning.

pascalberger: Good point. @asbjornu @JakeGinnivan Do you know the impact of this in CI build systems?

asbjornu:
@simonejsing Thanks for this! Do you think it's possible to handle this via some sort of locking instead? Also, do you think it's possible to write a test that verifies the intended and wanted behavior?

src/GitVersionCore/ExecuteCore.cs
asbjornu: @pascalberger I've suggested handling this with locking instead, so we avoid the exceptions altogether. I have not experienced CI failing on warnings, but that's possible to configure, I guess.

pascalberger:
Yes, implementing locking makes sense. But also in this case exceptions which are still thrown while trying to write to the cache for any reason, should be handled and logged and not lead to a failure IMHO.

asbjornu:
@pascalberger I agree that exceptions should be handled, but since in a locking scenario we're not really expecting anything to fail, I think the exceptions that do occur should be written out as a warning.

simonejsing:
@asbjornu @pascalberger - I initially thought about locking as well, but for multi-process synchronization I would need to use a named mutex. Besides requiring a fairly tricky implementation to get it right, it also adds a performance overhead on the happy path. It also leads to all sorts of considerations to take into account, e.g. how do we scope the locking to be at the repo level to avoid locking GitVersion across all repos. After discussing with a coworker, he brought up the point that a widely used practice for concurrent file system operations is to use exponential backoff. This is for example used by the C# compiler with a 30 second max wait. But at this point we agreed that for a caching scenario where the consequence of write failure is that another process will handle the retry then we should simply ignore a cache write.

Now that being said, I would highly suggest the exponential backoff over named mutex locking if you disagree with out train of thoughts. Let me know what you think.

asbjornu:
@simonejsing Excellent points. I agree that a named mutex is too much of an overhead, on execution time, code complexity and how long it will take to implement. I like the idea of an exponential backoff, though, so if you can implement something like that, it would be neat. Even neater if you could adjust the code wrt. logging as @pascalberger suggests and add at least one test that provokes the race condition.

simonejsing:
Implemented exponential backoff with unit tests. Please review.

simonejsing:
I get the following build error. That error message is not very helpful, so any insights would be nice...

Error : /home/travis/build/GitTools/GitVersion/src/GitVersionTask.Tests/bin/Debug/GitVersionTask.Tests.dll

7850The object with ID 2 implements the IObjectReference interface for which all dependencies cannot be resolved. The likely cause is two instances of IObjectReference that have a mutual dependency on each other.

7851  at (wrapper managed-to-native) System.Object:__icall_wrapper_mono_remoting_wrapper (intptr,intptr)

7852  at (wrapper remoting-invoke) NUnit.Engine.Agents.RemoteTestAgent:Run (NUnit.Engine.ITestEventListener,NUnit.Engine.TestFilter)

7853  at NUnit.Engine.Runners.ProcessRunner.RunTests (ITestEventListener listener, NUnit.Engine.TestFilter filter) <0x413be5d0 + 0x00042> in <filename unknown>:0

asbjornu:
@simonejsing I restarted the build. Everything is green now. 👍

simonejsing:
Thanks!

src/GitVersionCore/Helpers/OperationWithExponentialBackoff.cs
```diff
@@ -0,0 +1,54 @@
...
+                catch (T e)
+                {
+                    exceptions.Add(e);
+                    if (tries > MaxRetries)
+                    {
+                        throw new AggregateException("Operation failed after maximum number of retries were exceeded.", exceptions);
+                    }
+                }
+
+                ThreadSleep.Sleep(sleepMSec);
```
JakeGinnivan: It would probably be good to log here how long it is going to sleep for, that way if it is failing multiple times the user can see why it's taking so long.

simonejsing:
@JakeGinnivan addressed your feedback and rebase onto latest master to avoid the merge commits.

JakeGinnivan:
Awesome, thanks for the PR and making that change so quickly

pkiers:
Hello,
I've recently read some useful information, I guess it might be useful for you too, please read at <http://desire.ssg-corp.com/e2izvju>

p.kiers@home.nl

PR #927, opened 2016-07-07
Add config wizard changes to support private repositories on Appveyor

### LINKED ISSUE #869: GitVersion init settings for private github repos with Appveyor
Appveyor uses ssh to clone/fetch a repo as you can configure the public key for this operation in Github.

I wanted to enable Gitversion for my private repo on Appveyor. The suggested workarounds to use https and set environment variables weren't working for me as appveyor does not include encrypted environment variables (GITVERSION_REMOTE_PASSWORD) on PR builds for security reasons.

@JakeGinnivan suggested that I change the process to avoid the extra fetch instead of trying to use ssl.

This worked out and here's what I ended up with in my appveyor.yml:

```
before_build:
  - nuget restore
  - ps: gitversion $env:APPVEYOR_BUILD_FOLDER /l console /output buildserver /updateAssemblyInfo /nofetch /b $env:APPVEYOR_REPO_BRANCH
```

I think it would be a great idea to add a prompt in `gitversion init` asking if your repo is private and configuring appveyor output to match this when a repo is private.

I'll leave this open several days to see what people think.  If it seems a good idea, I should be able to make the changes this weekend and open a PR.

bigjonroberts:
This is to resolve #869 
I checked [the docs](https://github.com/GitTools/GitVersion/blob/master/docs/build-server-support/build-server/appveyor.md) for this.  They look OK as-is for this change.

@Philo also expressed some interest in documenting this use case of `/nofetch` command-line option under this PR

### DIFF
--- a/src/GitVersionCore/Configuration/Init/BuildServer/AppVeyorSetup.cs
+++ b/src/GitVersionCore/Configuration/Init/BuildServer/AppVeyorSetup.cs
@@ -7,10 +7,19 @@ namespace GitVersion.Configuration.Init.BuildServer
     using GitVersion.Configuration.Init.Wizard;
     using GitVersion.Helpers;
 
+    enum ProjectVisibility
+    {
+        Public = 0,
+        Private = 1
+    }
+
     class AppVeyorSetup : ConfigInitWizardStep
     {
-        public AppVeyorSetup(IConsole console, IFileSystem fileSystem) : base(console, fileSystem)
+        private ProjectVisibility _projectVisibility;
+
+        public AppVeyorSetup(IConsole console, IFileSystem fileSystem, ProjectVisibility visibility) : base(console, fileSystem)
         {
+            _projectVisibility = visibility;
         }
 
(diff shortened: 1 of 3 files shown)


src/GitVersionCore/Configuration/Init/BuildServer/AppVeyorSetup.cs
```diff
@@ -32,38 +41,55 @@ protected override StepResult HandleResult(string result, Queue<ConfigInitWizard
...
-  project: <your sln file>");
+  project: <your sln file>",
+                GetGVCommand(_projectVisibility)
+            ));
         }
 
         void GenerateNuGetConfig(string workingDirectory)
         {
-            WriteConfig(workingDirectory, FileSystem, @"install:
+            WriteConfig(workingDirectory, FileSystem, String.Format(@"install:
```
JakeGinnivan: Can use string interpolation here instead just as an FYI.

JakeGinnivan:
Nice, thanks :)

PR #943, opened 2016-07-14
Fix tfs task build

### LINKED ISSUE #941: Tasks.json not being updated in build
It seems the task.json for the VSTS Task is not being updated properly in the build.

The major, minor, patch fields were empty. Putting this here as I don't have time right now to investigate and track it down. If someone wants to look into it and sort it out that would be awesome :smile:

pascalberger:
Fix versioning of TFS task during build.

Fixes #941


JakeGinnivan:
Nice, thanks mate :)

PR #958, opened 2016-07-19
Update documentation to use alpha prefix

pascalberger:
Update documentation to use new alpha prefix instead of old unstable one.


JakeGinnivan:
Ah nice, I just updated images. Thanks!

PR #957, opened 2016-07-19
Fix build server doc links

pascalberger:
The link to the GitLabl build server documentation is currently wrong. Trying to fix this with full paths.

PR #959, opened 2016-07-19
Revert "Fix build server doc links"

pascalberger:
Reverts GitTools/GitVersion#957

Unfortunately this broke all links. Still no clue why the gitlab link in the original way didn't work...


gep13:
Que?!?

Issue #942, opened 2016-07-14
Cache broken with GitVersionTask 3.6.0


tpluscode:
I've just updated the MSBuild task package and apparently the cache is broken. Each target (`WriteVersionInfoToBuildLog`, `WriteVersionInfoToBuildLog` and `GitVersion`) in each built project recalculates the version number from scratch. Every time I see 

```
Cache file C:\TeamCity\buildAgent\work\aaa244402599d927\.git\gitversion_cache\XYZ123.yml not found
```

And every time (during one build) the `XYZ123` part is different. Worked good in version 3.5.4

Martin-Andersen:
@tpluscode glad I a not alone and not crazy. I am getting the exact same error
`[WriteVersionInfoToBuildLog]   INFO [07/14/16 13:55:32:72] Cache file C:\TeamCity\buildAgent\work\e21746ec702e5b5d\.git\gitversion_cache\3012A3203E0BBA59F13642534D34033F6E301B69.yml not found.`

It takes forever when the cache is not working.

DanielRose:
I have the same problem here. My build time went from `MSBuild (3m:08s)` to `MSBuild (2h:04m:52s)`!

Even while building an assembly, the name of the cache file keeps on changing for every step:
`[WriteVersionInfoToBuildLog]   INFO [07/13/16 10:42:19:40] Cache file C:\TeamCity_BuildAgent_2\work\60f2387b524bacc3\.git\gitversion_cache\B979AEEFD2C8836D2DD487256DFC4D03DFE097D1.yml not found.`
`[UpdateAssemblyInfo]   INFO [07/13/16 10:42:51:60] Cache file C:\TeamCity_BuildAgent_2\work\60f2387b524bacc3\.git\gitversion_cache\4641D2244564FA2AAB7EE360234E2E4AD9544A41.yml not found.`
`[GetVersion]   INFO [07/13/16 10:43:17:57] Cache file C:\TeamCity_BuildAgent_2\work\60f2387b524bacc3\.git\gitversion_cache\4F5D3967671A88C07685ED01CF8CCF4AC536CD7C.yml not found.`

JakeGinnivan:
Hey, thanks for posting the issue.

Not sure why this is happening, the code to calculate the hash is:

```
var ticks = fileSystem.GetLastDirectoryWrite(Path.Combine(gitDir, "refs"));
[...]

PR #967, opened 2016-07-22
add a test to ensure cache key is same even if renormalized

tofutim:
Re: https://github.com/GitTools/GitVersion/issues/942

If the gitPreparer normalize (in Initialize) is called multiple times, the cachekey will change each time, even though .git/refs has exactly the same content. This is because the cachekey is based on the timestamp rather than the contents. This PR changes the cachekey calculation so that it computes a hash based on the filenames and directory names in .git/refs.

### DIFF
--- a/src/GitVersionCore.Tests/ExecuteCoreTests.cs
+++ b/src/GitVersionCore.Tests/ExecuteCoreTests.cs
@@ -28,13 +28,13 @@ public class ExecuteCoreTests
             var targetUrl = "https://github.com/GitTools/GitVersion.git";
             var targetBranch = "refs/head/master";
             var gitPreparer = new GitPreparer(targetUrl, null, new Authentication(), false, fixture.RepositoryPath);
+            //var cacheKey0 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
+            gitPreparer.Initialise(true, targetBranch);
             var cacheKey1 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
             gitPreparer.Initialise(true, targetBranch);
             var cacheKey2 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
-            gitPreparer.Initialise(true, targetBranch);
-            var cacheKey3 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
 
-            cacheKey3.Value.ShouldBe(cacheKey2.Value);
+            cacheKey2.Value.ShouldBe(cacheKey1.Value);
         });        
     }
 
--- a/src/GitVersionCore.Tests/ExecuteCoreTests.cs
+++ b/src/GitVersionCore.Tests/ExecuteCoreTests.cs
@@ -19,6 +19,26 @@ public class ExecuteCoreTests
     }
 
     [Test]
+    public void CacheKeySameAfterReNormalizing()
+    {
+        var versionAndBranchFinder = new ExecuteCore(fileSystem);
+
+        RepositoryScope(versionAndBranchFinder, (fixture, vv) =>
+        {
+            var targetUrl = "https://github.com/GitTools/GitVersion.git";
+            var targetBranch = "refs/head/master";
+            var gitPreparer = new GitPreparer(targetUrl, null, new Authentication(), false, fixture.RepositoryPath);
+            var cacheKey1 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
+            gitPreparer.Initialise(true, targetBranch);
+            var cacheKey2 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
+            gitPreparer.Initialise(true, targetBranch);
+            var cacheKey3 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
+
+            cacheKey3.Value.ShouldBe(cacheKey2.Value);
+        });        
+    }
+
+    [Test]
     public void CacheFileExistsOnDisk()
     {
         const string versionCacheFileContent = @"
--- a/src/GitVersionCore/GitVersionCacheKeyFactory.cs
+++ b/src/GitVersionCore/GitVersionCacheKeyFactory.cs
@@ -104,7 +104,11 @@
                         // Perform whatever action is required in your scenario.
                         System.IO.FileInfo fi = new System.IO.FileInfo(file);
                         result.Add(fi.Name);
-                        Logger.WriteInfo(string.Format("{0}: {1}, {2}", fi.Name, fi.Length, fi.CreationTime));
+
+                        var reader = fi.OpenText();
+                        result.Add(reader.ReadToEnd());
+
+                        //Logger.WriteInfo(string.Format("{0}: {1}, {2}", fi.Name, fi.Length, fi.CreationTime));
                     }
                     catch (System.IO.FileNotFoundException e)
                     {
--- a/src/GitVersionCore/GitVersionCacheKeyFactory.cs
+++ b/src/GitVersionCore/GitVersionCacheKeyFactory.cs
@@ -2,6 +2,7 @@
 {
     using GitVersion.Helpers;
     using System;
+    using System.Collections.Generic;
     using System.IO;
     using System.Security.Cryptography;
     using System.Text;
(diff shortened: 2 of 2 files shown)


src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,108 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                }
+                // Perform the required action on each file here.
+                // Modify this block to perform your required task.
+                foreach (string file in files)
+                {
+                    try
+                    {
+                        // Perform whatever action is required in your scenario.
+                        System.IO.FileInfo fi = new System.IO.FileInfo(file);
+                        result.Add(fi.Name);
```
DanielRose: You are just using the name of the file. Even if it is modified by normalization, this shouldn't change.

tofutim: You are right. I'm just using the name - and we could pull in the contents as well to be more precise. Though I think normalization is just bringing in the branches from origin. I could go with using the contents though.

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,112 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+
+                        var reader = fi.OpenText();
```
DanielRose: You can simplify this by using `results.Add(File.ReadAllText(file))`

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,112 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                        continue;
+                    }
+                }
+
+                // Push the subdirectories onto the stack for traversal.
+                // This could also be done before handing the files.
+                // push in reverse order
+                for (int i = subDirs.Length - 1; i >= 0; i--)
+                {
+                    dirs.Push(subDirs[i]);
```
DanielRose: This pushes the name of the subdirectory only. Replace this with `dirs.Push(Path.Combine(currentDir, subDirs[i]));`

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,112 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                        // Perform whatever action is required in your scenario.
+                        System.IO.FileInfo fi = new System.IO.FileInfo(file);
+                        result.Add(fi.Name);
+
+                        var reader = fi.OpenText();
+                        result.Add(reader.ReadToEnd());
+
+                        //Logger.WriteInfo(string.Format("{0}: {1}, {2}", fi.Name, fi.Length, fi.CreationTime));
+                    }
+                    catch (System.IO.FileNotFoundException e)
```
DanielRose: Catch IOException as well.

src/GitVersionCore/GitVersionCacheKeyFactory.cs
tofutim: right, except the first line already has the whole directory

tofutim: or are you saying there will be a traversal error?

tofutim: you mean just in case?

DanielRose: Yes, there will be a traversal error. The current directory does not change anywhere, so it will never go into the subdirectories.

src/GitVersionCore/GitVersionCacheKeyFactory.cs
tofutim: It does traverse.

src/GitVersionCore/GitVersionCacheKeyFactory.cs
DanielRose: Okay, very strange. I tried it earlier and I'm sure the subdirectories were not rooted, i.e. it was "refs", etc. But now I tried it again, and they were "C:\...\refs", etc. So you can ignore what I wrote here :)

src/GitVersionCore.Tests/ExecuteCoreTests.cs
```diff
@@ -19,6 +19,26 @@ public void SetUp()
...
+        RepositoryScope(versionAndBranchFinder, (fixture, vv) =>
+        {
+            var targetUrl = "https://github.com/GitTools/GitVersion.git";
+            var targetBranch = "refs/head/master";
+            var gitPreparer = new GitPreparer(targetUrl, null, new Authentication(), false, fixture.RepositoryPath);
+            var cacheKey1 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
+            gitPreparer.Initialise(true, targetBranch);
+            var cacheKey2 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
+            gitPreparer.Initialise(true, targetBranch);
+            var cacheKey3 = GitVersionCacheKeyFactory.Create(fileSystem, gitPreparer, null);
```
pascalberger: Why get cache key 3 times, and only compare 2 and 3? Shouldn't the first one also be the same?

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+        }
+
+        // lifted from https://msdn.microsoft.com/en-us/library/bb513869.aspx
+        public static List<string> TraverseTree(string root)
+        {
+            var result = new List<string>();
+            result.Add(root);
+            // Data structure to hold names of subfolders to be
+            // examined for files.
+            Stack<string> dirs = new Stack<string>(20);
```
pascalberger: Use var

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
 
-            return GetHash(dotGitDirectory, lastGitRefsChangedTicks.ToString());
+            return GetHash(traverse.ToArray());
+        }
+
+        // lifted from https://msdn.microsoft.com/en-us/library/bb513869.aspx
+        public static List<string> TraverseTree(string root)
+        {
+            var result = new List<string>();
+            result.Add(root);
```
pascalberger: You can use collection initialization here

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+            Stack<string> dirs = new Stack<string>(20);
+
+            if (!System.IO.Directory.Exists(root))
+            {
+                throw new ArgumentException();
+            }
+
+            dirs.Push(root);
+
+            while (dirs.Count > 0)
```
pascalberger: Use Any() for better readability and slightly better performance

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+        // lifted from https://msdn.microsoft.com/en-us/library/bb513869.aspx
+        public static List<string> TraverseTree(string root)
+        {
+            var result = new List<string>();
+            result.Add(root);
+            // Data structure to hold names of subfolders to be
+            // examined for files.
+            Stack<string> dirs = new Stack<string>(20);
+
+            if (!System.IO.Directory.Exists(root))
```
pascalberger: System.IO is not required since the namespace already is included

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+            {
+                string currentDir = dirs.Pop();
+
+                var di = new DirectoryInfo(currentDir);
+                result.Add(di.Name);
+
+                string[] subDirs;
+                try
+                {
+                    subDirs = System.IO.Directory.GetDirectories(currentDir);
```
pascalberger: System.IO is not required since the namespace already is included

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                // another application or thread after our call to Directory.Exists. The 
+                // choice of which exceptions to catch depends entirely on the specific task 
+                // you are intending to perform and also on how much you know with certainty 
+                // about the systems on which this code will run.
+                catch (UnauthorizedAccessException e)
+                {
+                    Logger.WriteError(e.Message);
+                    continue;
+                }
+                catch (System.IO.DirectoryNotFoundException e)
```
pascalberger: System.IO is not required since the namespace already is included

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                {
+                    Logger.WriteError(e.Message);
+                    continue;
+                }
+
+                string[] files = null;
+                try
+                {
+                    files = System.IO.Directory.GetFiles(currentDir);
```
pascalberger: System.IO is not required since the namespace already is included

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                }
+
+                catch (UnauthorizedAccessException e)
+                {
+                    Logger.WriteError(e.Message);
+                    continue;
+                }
+
+                catch (System.IO.DirectoryNotFoundException e)
```
pascalberger: System.IO is not required since the namespace already is included

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                    continue;
+                }
+                // Perform the required action on each file here.
+                // Modify this block to perform your required task.
+                foreach (string file in files)
+                {
+                    try
+                    {
+                        // Perform whatever action is required in your scenario.
+                        System.IO.FileInfo fi = new System.IO.FileInfo(file);
```
pascalberger: System.IO is not required since the namespace already is included

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                    Logger.WriteError(e.Message);
+                    continue;
+                }
+                // Perform the required action on each file here.
+                // Modify this block to perform your required task.
+                foreach (string file in files)
+                {
+                    try
+                    {
+                        // Perform whatever action is required in your scenario.
```
pascalberger: You should remove this comment, or update it to the concrete implementation

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -23,10 +24,106 @@ private static string GetGitSystemHash(GitPreparer gitPreparer, IFileSystem file
...
+                    continue;
+                }
+
+                catch (System.IO.DirectoryNotFoundException e)
+                {
+                    Logger.WriteError(e.Message);
+                    continue;
+                }
+                // Perform the required action on each file here.
+                // Modify this block to perform your required task.
```
pascalberger: You should remove this comment, or update it to the concrete implementation

src/GitVersionCore.Tests/ExecuteCoreTests.cs
tofutim: actually, no - the first one is the un-normalized cache key - the problem we are trying address is a different cache key after re-normalizing - probably don't even need to lookup cachekey1 though

src/GitVersionCore/GitVersionCacheKeyFactory.cs
tofutim: TraverseTree comes from https://msdn.microsoft.com/en-us/library/bb513869.aspx - I didn't to change it too much - but if you'd like me to...

tofutim:
@rubenmamo Hi Ruben, can you confirm that these changes work in your scenario? Thanks so much.

JakeGinnivan:
We should probably have a 3.6.2 release right?

I have created support/3.6.0 to allow us to fix issues in the 3.6.x version. Can you re-target this PR to that branch?

JakeGinnivan:
`master` is currently v4. But this is a pretty decent bug which I think should be patched in the last release

tofutim:
@JakeGinnivan can i re-target the PR from this PR or do I need to make a new PR?

tofutim:
https://github.com/isaacs/github/issues/18 sigh

rubenmamo:
@totufim, apologies for not replying earlier. I wasn't in the office and couldn't check with the latest version. Dynamic repositories are not working in 3.6.1 and I suspect this would be the same for 3.6.2. The new issue is not caused by this change. I have created issue #981 which details the new issue.

Issue #971, opened 2016-07-25
Add strategy for incrementing the version by looking for commits tagged with version bumps


MikeLarah:
Similar to incrementing based off of commit messages containing `+semver:minor` or `+semver:major` - if we could tag commits with these messages and increment the version accordingly.

The reason being is that we want to do a binary comparison between the current build and the last master build to detect API changes (which obviously happens after the commit), then retrospectively tag the commit to state if there was a major or minor change.

mwadams:
(Prior to a merge back into e.g. Develop or Master)

JakeGinnivan:
I suggest the following for this scenario.

1. Upgrade to v4 beta
2. Switch to mainline mode
3. Add the `+semver: minor` or `+semver: breaking` to the merge commit when merging the PR.

If this works for you, it would be awesome to get this sort of flow documented somewhere in the docs!

JakeGinnivan:
Going to close, I do not think we want to encourage abusing tags in this way. Would love a follow up from your side @MikeLarah if the above suggestion works for you

PR #972, opened 2016-07-26
Updated IncrementStrategyFinder to support version bumping from tags

MikeLarah:
Updated IncrementStrategyFinder to support version bumping from tagged commits.

Updated tests and docs.

#971

### DIFF
--- a/src/GitVersionCore/Configuration/Config.cs
+++ b/src/GitVersionCore/Configuration/Config.cs
@@ -55,6 +55,19 @@
         [YamlMember(Alias = "no-bump-message")]
         public string NoBumpMessage { get; set; }
 
+        [YamlMember(Alias = "major-version-bump-tag")]
+        public string MajorVersionBumpTag { get; set; }
+
+        [YamlMember(Alias = "minor-version-bump-tag")]
+        public string MinorVersionBumpTag { get; set; }
+
+        [YamlMember(Alias = "patch-version-bump-tag")]
+        public string PatchVersionBumpTag { get; set; }
+
+        [YamlMember(Alias = "no-bump-tag")]
+        public string NoBumpTag { get; set; }
+
+
         [YamlMember(Alias = "legacy-semver-padding")]
         public int? LegacySemVerPadding { get; set; }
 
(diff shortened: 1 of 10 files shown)


src/GitVersionCore/Configuration/Config.cs
```diff
@@ -55,6 +55,19 @@ public string NextVersion
...
+        [YamlMember(Alias = "minor-version-bump-tag")]
+        public string MinorVersionBumpTag { get; set; }
+
+        [YamlMember(Alias = "patch-version-bump-tag")]
+        public string PatchVersionBumpTag { get; set; }
+
+        [YamlMember(Alias = "no-bump-tag")]
+        public string NoBumpTag { get; set; }
+
+
```
pascalberger: Remove double linebreak

pascalberger:
Thanks for your contribution! LGTM

@asbjornu @gep13 @JakeGinnivan OK for you too merge this?

JakeGinnivan:
Would you be able to walk me through a typical workflow on how you would use this feature @MikeLarah?

JakeGinnivan:
@MikeLarah am keen to hear more about the requirements for this change?

MikeLarah:
@JakeGinnivan Sorry for the delayed response. Our intention was to use this in combination with a process in the CI build that determined whether the public API has been changed.
e.g. 1) developer commits and pushes up changes to a feature branch, 2) CI build does comparison between feature branch binaries and latest master binaries to determine if API has changed, 3) we tag the commit with relevant major/minor increment tag if necessary, 4) future builds are automatically version incremented using GitVersion (and will be correctly incremented when the feature is merged to develop branch)

JakeGinnivan:
Hey Mike,

I see a few issues with this approach

1. More config :(
2. `git tag` will have a heap of `+semver-minor` tags, which is noisy

Alternatives:

When you merge pull requests, you can alter the merge message in every tool I know. At that point simply append `+semver: minor` to the merge PR commit message.

Then you can use the new Mainline development mode in v4 to aggregate the versions from each merged PR. I think your process is working around the fact mainline was missing and I think it should be able to solve your problem. Let me know how you go, going to close this out for now (but we can keep the convo open)

Issue #974, opened 2016-07-26
Version is incremented twice if the base version commit also has an explicit semver message


MikeLarah:
Test to repro:

```
[Test]
    public void CanUseCommitMessagesToBumpVersionBaseVersionTagIsAppliedToSameCommit()
    {
        using (var fixture = new EmptyRepositoryFixture())
        {
            fixture.Repository.MakeACommit();
            fixture.MakeATaggedCommit("1.0.0");
            fixture.Repository.MakeACommit("+semver:minor");
            fixture.AssertFullSemver("1.1.0+1");

            fixture.ApplyTag("2.0.0");

            fixture.Repository.MakeACommit("Hello");

            // Default bump is patch

            fixture.AssertFullSemver("2.0.1+1");
        }
    }
```

pascalberger:
Fix available in #972

PR #975, opened 2016-07-26
Exclude base commit when getting intermediate commits.

MikeLarah:
Fix for #974 (if #972 is rejected)


JakeGinnivan:
Gonna merge this one for now. Want to discuss #972 a bit more before we go down that road.

JakeGinnivan:
Also, if you want this in 3.6.2 then resubmit this to `support/3.6.0` otherwise this will only land in v4 betas.

JakeGinnivan:
Gonna cherry pick this across.

Issue #961, opened 2016-07-20
Only push VSTS Extension on stable build


JakeGinnivan:
Otherwise v4 beta will get pushed to everyone automatically =/

pascalberger:
PR available: #977

PR #977, opened 2016-07-27
Throw if for some reason we are missing any artifacts. Do not upload …

JakeGinnivan:
…pre-release to VSTS

@gep13 is the random missing artifacts from AppVeyor in the v15 fix? :)


gep13:
>@JakeGinnivan said...
@gep13 is the random missing artifacts from AppVeyor in the v15 fix? :)

Yes, should be.  I think the check that you are adding here is still a good fallback strategy though.

pascalberger:
LGTM

asbjornu:
Fixes #961.

Issue #862, opened 2016-05-16
Incorrect version calculation


lerthe61:
We have faced some strange behavior, when fork in repository is interpreted as merge:
![canvas](https://cloud.githubusercontent.com/assets/6067323/15296475/388c3a50-1b9e-11e6-8746-9fe4d525fc0c.png)

We also found workaround by changing _track-merge-target_ to false for _develop_ branch.
Original configuration is
[GitVersionConfig.zip](https://github.com/GitTools/GitVersion/files/266536/GitVersionConfig.zip)

We use version 3.4.1 (portable). But issue is still presented in latest codebase.

lerthe61:
Prepare test repository for illustrating problem. For branch develop GitVersion calculate SemVer "1.0.1-feature.2"
[TestRepo.zip](https://github.com/GitTools/GitVersion/files/267466/TestRepo.zip)

lerthe61:
Attach unit test for issue. It seems that problem in TrackMergeTargetBaseVersionStrategy.IsDirectMergeFromCommit method.

pascalberger:
@lerthe61 Can you please test with latest [4.0.0-beta release](https://github.com/GitTools/GitVersion/releases/tag/v4.0.0-beta.1) as it changes some behavior how versioning on develop is handled.

lerthe61:
@pascalberger Thanks! Everything look ok now. I will create pull request with my test to prevent regression in future

PR #978, opened 2016-07-27
Version calculation should not use tags from not merged branch

lerthe61:
Create unit test which cover scenario described in #862 issue. Short description:
Repositories consist from three branch - feature, develop and master. There is only fork (NO merge commits) in repository. Feature and Master branches grow up from the same commit. Tags in feature branch should not taking in count for version calculation for develop branch.


lerthe61:
This test failed for version 3.4.1 (portable). And I would like to introduce this unit test to be sure that not facing the same problem in future.

asbjornu:
@lerthe61 LGTM, thanks! :+1:

PR #980, opened 2016-07-28
Update version in project.json files

### LINKED ISSUE #647: Support for dotnet (prev. DNX) / .NET Core projects
In DNX / .NET Core projects the version is set via the `project.json` file [described here](https://github.com/aspnet/Home/wiki/Project.json-file) in the Metadata section. `AssemblyInfo.cs` on the other hand is gone. I'd like to use GitVersion with such projects.

droyad:
Added the `/updateprojectjson` switch to GitVersion.exe which replaces the version in Project.json files. Closes #647.

### DIFF
--- a/docs/usage/command-line.md
+++ b/docs/usage/command-line.md
@@ -57,12 +57,11 @@ Will result in command line argument error
 Will iterate through each file and update known attributes (`AssemblyVersion`, `AssemblyFileVersion`, `AssemblyInformationalVersion`).
 
 ## Replace version in project.json
-`GitVersion.exe /updateprojectjson` will recursively search for all `project.json` files that are
-in the same directory as a `.xproj` file. For each found file it will update the value of the `version` element in the
-root of the document with the `InformationalVersion`.
+`GitVersion.exe /updateprojectjson` will recursively search for all `project.json` files and set the `InformationalVersion` as
+the value of the `version` element in the root of the JSON document.
 
 The `dotnet build` command uses the `major.minor.patch` version from the `project.json` as the file and assembly version if the corresponding 
-attributes do not exist. It also automatically adds the `AssemblyInformationalVersion` attribute with the full version excluding metadata.
+attributes do not exist and also automatically adds the `AssemblyInformationalVersion` attribute with the full version excluding metadata.
 
 ### Example:
 `GitVersion.exe /updateprojectjson`
--- a/docs/usage/command-line.md
+++ b/docs/usage/command-line.md
@@ -61,7 +61,7 @@ Will iterate through each file and update known attributes (`AssemblyVersion`, `
 the value of the `version` element in the root of the JSON document.
 
 The `dotnet build` command uses the `major.minor.patch` version from the `project.json` as the file and assembly version if the corresponding 
-attributes do not exist and also automatically adds the `AssemblyInformationalVersion` attribute with the full version excluding metadata.
+attributes do not exist and also automatically adds the `AssemblyInformationalVersion` attribute with the full version excluding metadata. 
 
 ### Example:
 `GitVersion.exe /updateprojectjson`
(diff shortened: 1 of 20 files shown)


dazinator:
This capability would be really nice to have. I would love to see this in GitVersion because project.json's are currently a fact of life, even though they are expected to go away at some point in the future.. What are thoughts opinions around this?

gep13:
@dazinator I have started using this: https://github.com/sergeyzwezdin/magic-chunks/ As it integrates really nicely with Cake, which I am now using for all my builds.

Here is an example of updating a project.json file with it:

https://github.com/OctopusDeploy/CSProjToXProj/blob/master/build.cake#L110

dazinator:
@gep13 - thanks for that, I haven't looked at Cake yet, but I like the fact you can use c# syntax! Unfortunately switching to a Cake build isn't an option for us at present, so I am looking for alternatives like this PR..

@droyad I have noticed a bit of a problem with this PR in that it can break the build as it doesn't deal with updating version numbers in project references. For example, if you have ProjectA.json which references ProjectB.json, then ProjectA.json will have this dependency entry:

```
{
"dependencies": {
 "ProjectB": "1.0.0-*"
}
```

This functionality updates the `version`'s of both projects, for example to `1.1.0-alpha1` and now the solution won't build because the dependencies section are invalid. So for using this to build a multi-project solution, I think this would need to be solved..

gep13:
@dazinator what are you currently using to run GitVersion?

droyad:
@dazinator I use just the wildcard to reference other projects, ie:
```
{
"dependencies": {
 "ProjectB": "*"
}
```

Are there any drawbacks to that approach?

dazinator:
@gep13 - a team city command line step!

@droyad - ah yeah so I tried that myself before I posted but it didn't work.. i'll try again - if it works for you then i probably just need to clear the project.lock files or something.

gep13:
>@dazinator said,,,
@gep13 - a team city command line step!

Gotcha.  I was going to say that if you were using any other type of build script, i.e. psake, or something like that, then Magic Chunks has some plugins for there too.  It isn't just limited to Cake usage.

JakeGinnivan:
Fyi to reference other projects you can do:

```
{
"dependencies": {
 "ProjectB": { "target": "project" },
}
```

JakeGinnivan:
Thoughts on https://github.com/ah-/dotnet-gitversion

We could move this into a new project as done in that repo?

gep13:
Makes sense to me.

asbjornu:
@JakeGinnivan: That looks like a great effort from @ah-. Can't we just move his repository into the GitTools organization and make it official?

JakeGinnivan:
I think his repo is a port to dotnet core. I am happy for someone else to take the lead on this feature :)

JakeGinnivan:
If you need something specific from me let me know

onovotny:
I'm not sure this is needed anymore. Instead, cross-compile the existing GitVersionTask for net4 and NetCoreApp. Then it can be used directly by csproj everywhere as I describe in #1118

droyad:
Agreed. With the move back to msbuild system, the classic ;) approach can be used.

asbjornu:
Can this be closed, then?

dazinator:
Does that require an upgrade from VS2015 to VS2017 in order to migrate from project.json back to msbuild? Just concerned that project.json files might remain for those stuck on vs2015 is so?

onovotny:
`project.json` will only work with vs 2015 and the new tooling will only work with vs 2017. The tooling are preview tooling with vs 2015 and will not get any more updates. Moving to VS 2017 is really the only option.

dazinator:
hmm ok.. I'm going to enjoy explaining that cost to my manager haha. How about the cross plat story, for example Linux.. will their be new tooling on Linux to replace project.json as well?

onovotny:
All of the tooling is going to be csproj only, including cross platform. The current preview4 CLI tooling already supports the csproj.

dazinator:
Ok cool, thanks for explaining @onovotny - project.json shall be consigned to the dustbin of history, and anyone stuck on it will have a migration path... sounds very sensible to close this issue :-)

dazinator:
@gep13 - I ended up moving from team city to appveyor and cake in the end and using Magic Chunks. I don't regret it. Team City is now dead to me.

PR #1000, opened 2016-08-11
Develop version not bump if no new commit since release branch

qetza:
Updated develop version strategy so that develop version is not bumped until a commit is done after  release branch merge base.


JakeGinnivan:
Nice change

Issue #834, opened 2016-04-19
Add NuGetPreReleaseTag


roji:
project.json builds allow you to leave the prerelease tag component of the version as a wildcard (`"version": "3.1.0-*"`) and then pass it in via a command-line parameter (`dotnet pack --version-suffix`).

Ironically with all the variables GitVersion generates, there doesn't seem to be a simple nuget-compatible prerelease tag (i.e. unstable0456 and not unstable.456). Maybe you guys would consider adding NuGetPreReleaseTag?

Related to #647

JakeGinnivan:
Sounds fine to me.

qetza:
Done in PR #1002

PR #1002, opened 2016-08-11
Add NuGetPreReleaseTag variable

qetza:
Issue #834


JakeGinnivan:
The only comment is maybe we should have a v2 variable as well. The idea for the v2 variable is that when NuGet adds proper SemVer support, the NuGet variable will stop padding but the v2 variable won't.

qetza:
Will do when i'm back home :smile:

qetza:
Done

JakeGinnivan:
Was a good move asking for the v2 var ask the change to SemVer 2 is underway for NuGet. https://emgarten.com/2016/08/09/semver-2-0-0-with-nuget-3-5-0-rc1/

Issue #988, opened 2016-08-08
How to configure GitVersion to fit our needs


maxraab:
Suppose you have a big community and many contributions. Every contribution starts with a feature branch from `master` and after it's finished gets merged back to `master`. (Normal GitHub Flow). The main difference is the release stragegy:

If a release is sheduled - a new release branch (e.g. 8-1-stable, 8-2-stable, ...) is created from current `master`. From this time only some contributions gets cherry picked from `master` into the `stable branch`. All release candidates (e.g. 8.1-rc1, 8.1-rc2, ...) where tagged on this `stable branch`.

If the Version is ready to ship, the Version gets tagged but the release branch will not merged back into `master`. (To support up to 10 versions at the same time).
From this time, the commits on `master` should have a version similar to: "next-version"-pre.

asbjornu:
@maxraab Could you please provide a more visual example of how your Git graph looks like? A simple [Web Sequence Diagram](https://www.websequencediagrams.com/) would do.

maxraab:
![untitled](https://cloud.githubusercontent.com/assets/12211887/17480985/f81f8354-5d7a-11e6-9c63-85a02a7302f8.png)

As you can see, we have no tags directly in `master`.

/cc @asbjornu

DanielRose:
You will need to use the feature #832 for this to work (currently only in the preview releases).

For configuration (see http://gitversion.readthedocs.io/en/latest/configuration/): Mark the master branch as `is-develop`. You will also need a custom branch name and `tag-number-pattern`, or change the name of your stable branches (i.e. release branches) to a standard `release-1.1` or `release/1.1`

maxraab:
Thank you for your response.

I don't get it.

Is it possible to get [one of these szenarios](https://github.com/maxraab/GitVersion/tree/ticket-988) working?
I've added two files (szenario1.cs, szenario2.cs) with our testcases.

DanielRose:
I couldn't test your scenario, but try a `GitVersion.yml` with the following contents:
```
branches:
  master:
    tag: pre
    is-develop: true
  releases?[/-]:
    tag: rc
```

gep13:
@maxraab @DanielRose **NOTE:** In order to verify any changes that you make in the configuration file, you will need to clear the GitVersion cache, which you can find here:

![image](https://cloud.githubusercontent.com/assets/1271146/17554949/688e3848-5f05-11e6-91c0-09ce92f8b466.png)

[...]

PR #1004, opened 2016-08-11
Feature branch does not pick up version

maxraab:
based on a bug discovered in #988

### DIFF
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -16,7 +16,7 @@ namespace GitVersion
             if (matchingBranches.Length == 0)
             {
                 var branchConfig = new BranchConfig();
-                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig, isDevelop: true);
+                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig, tracksReleaseBranches: true);
                 return new KeyValuePair<string, BranchConfig>(string.Empty, branchConfig);
             }
             if (matchingBranches.Length == 1)
@@ -114,7 +114,7 @@ namespace GitVersion
                             Increment = branchConfig.Increment,
                             PreventIncrementOfMergedBranchVersion = branchConfig.PreventIncrementOfMergedBranchVersion,
                             // If we are inheriting from develop then we should behave like develop
-                            IsDevelop = branchConfig.IsDevelop
+                            TracksReleaseBranches = branchConfig.TracksReleaseBranches
                         });
                 }
 
@@ -142,7 +142,7 @@ namespace GitVersion
                         Increment = inheritingBranchConfig.Increment,
                         PreventIncrementOfMergedBranchVersion = inheritingBranchConfig.PreventIncrementOfMergedBranchVersion,
                         // If we are inheriting from develop then we should behave like develop
-                        IsDevelop = inheritingBranchConfig.IsDevelop
+                        TracksReleaseBranches = inheritingBranchConfig.TracksReleaseBranches
                     });
             }
         }
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -1,13 +1,11 @@
 namespace GitVersion
 {
+    using JetBrains.Annotations;
+    using LibGit2Sharp;
     using System;
     using System.Collections.Generic;
     using System.Linq;
     using System.Text.RegularExpressions;
-
-    using JetBrains.Annotations;
-
-    using LibGit2Sharp;
 
     public class BranchConfigurationCalculator
     {
@@ -18,7 +16,7 @@ namespace GitVersion
             if (matchingBranches.Length == 0)
             {
                 var branchConfig = new BranchConfig();
-                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig);
+                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig, isDevelop: true);
                 return new KeyValuePair<string, BranchConfig>(string.Empty, branchConfig);
             }
             if (matchingBranches.Length == 1)
@@ -44,12 +42,12 @@ namespace GitVersion
             {
                 throw new ArgumentNullException("config");
             }
-            
+
             if (currentBranch == null)
             {
                 throw new ArgumentNullException("currentBranch");
             }
-            
+
             return config.Branches.Where(b => Regex.IsMatch(currentBranch.FriendlyName, "^" + b.Key, RegexOptions.IgnoreCase)).ToArray();
         }
 
--- a/src/GitVersionCore/Configuration/BranchConfig.cs
+++ b/src/GitVersionCore/Configuration/BranchConfig.cs
@@ -17,7 +17,7 @@
             TagNumberPattern = branchConfiguration.TagNumberPattern;
             TrackMergeTarget = branchConfiguration.TrackMergeTarget;
             CommitMessageIncrementing = branchConfiguration.CommitMessageIncrementing;
-            IsDevelop = branchConfiguration.IsDevelop;
+            TracksReleaseBranches = branchConfiguration.TracksReleaseBranches;
             IsReleaseBranch = branchConfiguration.IsReleaseBranch;
             IsMainline = branchConfiguration.IsMainline;
         }
@@ -42,12 +42,12 @@
 
         [YamlMember(Alias = "track-merge-target")]
         public bool? TrackMergeTarget { get; set; }
-        
+
         [YamlMember(Alias = "commit-message-incrementing")]
         public CommitMessageIncrementMode? CommitMessageIncrementing { get; set; }
 
-        [YamlMember(Alias = "is-develop")]
-        public bool? IsDevelop { get; set; }
+        [YamlMember(Alias = "tracks-release-branches")]
+        public bool? TracksReleaseBranches { get; set; }
 
         [YamlMember(Alias = "is-release-branch")]
         public bool? IsReleaseBranch { get; set; }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -144,7 +144,7 @@ branches:
     increment: Patch
     prevent-increment-of-merged-branch-version: true
     track-merge-target: false
-    is-develop: false
+    tracks-release-branches: false
     is-release-branch: false
   releases?[/-]:
     mode: ContinuousDelivery
@@ -152,7 +152,7 @@ branches:
     increment: Patch
     prevent-increment-of-merged-branch-version: true
     track-merge-target: false
-    is-develop: false
+    tracks-release-branches: false
     is-release-branch: true
   features?[/-]:
     mode: ContinuousDelivery
@@ -160,7 +160,7 @@ branches:
     increment: Inherit
     prevent-increment-of-merged-branch-version: false
     track-merge-target: false
-    is-develop: false
+    tracks-release-branches: false
     is-release-branch: false
   (pull|pull\-requests|pr)[/-]:
     mode: ContinuousDelivery
@@ -169,7 +169,7 @@ branches:
     prevent-increment-of-merged-branch-version: false
     tag-number-pattern: '[/-](?<number>\d+)[-/]'
     track-merge-target: false
-    is-develop: false
+    tracks-release-branches: false
     is-release-branch: false
   hotfix(es)?[/-]:
     mode: ContinuousDelivery
@@ -177,7 +177,7 @@ branches:
     increment: Patch
     prevent-increment-of-merged-branch-version: false
     track-merge-target: false
-    is-develop: false
+    tracks-release-branches: false
     is-release-branch: false
   support[/-]:
     mode: ContinuousDelivery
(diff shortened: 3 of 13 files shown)


asbjornu:
@maxraab Could you rebase this PR on the latest master, please? :smile:

maxraab:
@asbjornu done :smile:

JakeGinnivan:
I wonder if we should rename `is-develop` to `tracks-release-branches` which is more clear about what it is doing?

maxraab:
Sounds good, should I rename all occurences?

JakeGinnivan:
Yeah, also you will need to add validation in the configuration provider to tell people it has renamed.

maxraab:
@JakeGinnivan I've rebased to the current master and renamed `is-develop` to `tracks-release-branches`.

src/GitVersionCore/Configuration/BranchConfig.cs
```diff
@@ -17,42 +17,72 @@ public BranchConfig(BranchConfig branchConfiguration)
...
             CommitMessageIncrementing = branchConfiguration.CommitMessageIncrementing;
-            IsDevelop = branchConfiguration.IsDevelop;
+            TracksReleaseBranches = branchConfiguration.TracksReleaseBranches;
             IsReleaseBranch = branchConfiguration.IsReleaseBranch;
             IsMainline = branchConfiguration.IsMainline;
         }
 
         [YamlMember(Alias = "mode")]
-        public VersioningMode? VersioningMode { get; set; }
+        public VersioningMode? VersioningMode
```
asbjornu: Can you please not reformat code that is not a part of your PR? :smiley:

maxraab: I've restored the original format but we should improve that in another PR.

maxraab:
@asbjornu @JakeGinnivan for some reasons the AppVeyor build fails.. How can I get around with this?

src/GitVersionCore/Configuration/BranchConfig.cs
asbjornu: @maxraab If the current ReSharper settings made the properties multiline, we should indeed fix that in its own PR.

asbjornu:
@maxraab It's the `Create-Release-Notes` task that fails. I think @JakeGinnivan has a theory that it might be due to a missing GitHub API key, making the calls to GitHub throttled and thus failing. Not sure, but just giving it a bit of time and then rerunning the test should do.

DanielRose:
I've been thinking about this a bit, and I'm not so sure if this PR (while fixing the stated issue) actually makes sense.

The way I understand GitVersion _logically/conceptionally_, i.e. not the actual implementation (to simplify, I'll ignore `next-version` and when/how the obtained version number is later incremented):

1. First it checks in the current branch for a version number (ex. via tag, branch name, merged branch name; stated in the documentation as [Version Sources](http://gitversion.readthedocs.io/en/stable/more-info/version-sources/))
  * branch is `release/1.1` => 1.1
  * commit with tag `1.2` => 1.2
  * branch `release/1.3` was merged into current branch => 1.3
2. If no version number could be found, it checks the parent branch for its version number at the point where it branched
  * branch `feature/special` branched from `release/1.1` => no version number in `feature/special` => asks `release/1.1` => 1.1

The change with `is-develop` was supposed to be: _For a branch marked with it, creating a branch off it (with a version number) also picks up the version-number._ So in step 1:
* branch is `develop`, branch `release/1.4` was created from it => 1.4

This works. But the bug is that step 2 does not work properly:
* situation as above, then branch `feature/awesome` is branched from `develop` => 0.1 (simple case, no tags, merges, etc.); _should be 1.4_

The bug is that in step 2, the develop branch, when asked for its version number, acts as if `is-develop` was not set, and thus returns 0.1.

Now the OP noticed that if the branch `feature/awesome` is marked as `is-develop`, then it suddenly works. So the solution is basically to mark every branch from `develop` with that setting (and give it a different name).

I think this solution tries to fix the bug at the wrong place. It essentially means that all branch configurations (except perhaps master) need to be marked that way. Also logically it makes no sense: Why do I need to mark a branch with `is-develop`, so that when asking the parent branch, the parent branch acts as if `is-develop` was set there? Shouldn't that be a setting of the parent branch?

I see two more sensible solutions:

1. Add "Version of child branch name" as an additional Version Source, which always applies. In that case, `is-develop` can be dropped.
  * Solution: replace all checks for `is-develop` in the code with true, and drop dead code.
2. Fix the bug, without needing to apply `is-develop` everywhere (and perhaps give it a better name).
  * My _guess_ at what actually happens in the code, and why this bug occurs: Step 2 is not implemented as _ask the parent branch what its version is at the branch point_ (i.e. use the config of the parent branch), but rather as _go into parent branch and continue version calculation_ (i.e. use the config of the current branch). However, the config for version calculation can be different between the current branch and the parent branch. By continuing the use the original config, the bug occurs.
  * Solution: When going into a parent branch as part of version calculation, change to use the config of that parent branch.

Note that if my guess is correct, then there should be more bugs in the version calculation when version-calculation-relevant configs are different between two branches, such as `increment` or `track-merge-target`.

src/GitVersionCore/BranchConfigurationCalculator.cs
```diff
@@ -18,7 +16,7 @@ public class BranchConfigurationCalculator
             if (matchingBranches.Length == 0)
             {
                 var branchConfig = new BranchConfig();
-                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig);
+                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig, tracksReleaseBranches: true);
```
JakeGinnivan: Not sure this is right actually as @DanielRose says. The default shouldn't be to track release branches because there could be feature branches off support branches or any number of other scenarios.

JakeGinnivan:
`is-develop` and `is-release-branch` work together. Basically if a branch is marked `is-develop` then gitversion will try and find any release branches and use them as a base version. Track merge target I found really confusing and it kept having random bugs, this was quite simple.

It is an odd one, because by branching from the develop branch it automatically bumps. `DevelopVersionStrategy` should also be renamed to `TrackReleaseBranch` and instead of just checking `context.Configuration.IsCurrentBranchDevelop` that should be renamed to `context.Configuration.DoesCurrentBranchInheritFromDevelop` and we should check if any of the parent branches are develop. @DanielRose does that sound closer to fixing the actual issue?

maxraab:
@JakeGinnivan Looks good to me. It would be handy if all branches inherit settings of their parent branches (except the branch has specific settings).

Anyway.. this bug prevents us using GitVersion. I hope we could fix it soon :)

DanielRose:
I created a number of tests, as #1009. When using "normally named" feature branches, ex. `feature/test`, it works as expected (both with a normal git flow and branching directly off master).

However, when using these "misnamed" feature branches, i.e. those without a config (as in this issue), it does not work. The reason seems to be the value of the increment strategy: The default value (in `ConfigurationProvider`) is `IncrementStrategy.Patch`, while for feature-branches it is `IncrementStrategy.Inherit`. With Inherit, the version numbers get picked up correctly.

The documentation of increment is very short: _The part of the SemVer to increment when GitVersion detects it needs to be (i.e commit after a tag)_.

In the code, Inherit is additionally documented as _Uses the increment strategy from the branch the current branch was branched from_. But looking at the test results, and a quick look at the code in `BranchConfigurationCalculator.GetBranchConfiguration()`, Inherit actually means _Use the version number as calculated by the parent branch at the branch-point_ (which can be overridden by using tags, etc.). For all other values, the parent branch config is apparently not used.

So what I previously described as step 2 is actually only done when the branch uses the Inherit strategy. So either the description or the code should be updated. Also, it would be useful to have a way for the user to set the default config values (i.e. those which are used when no branch config exists).

_Aside:_

In the tests, when an increment != Inherit is used, the calculated version number in the feature branch is different when a release branch is only created off develop/master, or the release branch is merged back.

In the first case, no version number is found => uses 0.1.0 as base value.

In the second case, the merged number (1.0) in the parent branch is then incremented according to the strategy of the feature branch (=> 2.0, 1.1, 1.0.1, ...), ignoring how it was incremented in the parent branch.

DanielRose:
@JakeGinnivan Can you comment on what should be done?

JakeGinnivan:
I am not sure right now. Have DDD Perth tomorrow and this week has been pretty busy. Will try to catch up with this thread on Sunday

JakeGinnivan:
@DanielRose is it just that the default should be inherit rather than patch? Or is there something else?

DanielRose:
@JakeGinnivan Well, the bug here would (more or less - the commit counts might be unexpected) be fixed if the default was inherit.

However, this bug exposes that the behaviour of "increment" is different than what is documented (and expected - at least to me), as well as the behaviour when switching branches.

Expected:
major, minor, patch => Increase that part of the version number
inherit => Use the value of the parent branch
All these values are unique per branch, according to that branch's config. So during version calculation, when changing to the parent branch, that branch's config is then (always) used subsequently.

Actual:
major, minor, patch => Increase that part of the version number
inherit => _Use the config of the parent branch, once branching into it_
These values belong to the current branch.  So during version calculation, when changing to the parent branch, _that branch's config is only used if inherit was specified._

An example (numbers don't make sense, just to illustrate):
master branch, increment major, tagged 1.0.0, some more commits
child_minor branch, increment minor, later tagged 3.2.0, some more commits
child_inherit branch, increment inherit, later tagged 5.0.0, some more commits

Point in time | expected | actual
----|---|---
master (after tag) | 2.0.0 | 2.0.0
child_minor (before tag) | 2.0.0 | **_1.1.0_**
child_minor (after tag) | 3.3.0 | 3.3.0
child_inherit (before tag) | 2.0.0 | 2.0.0
child_inherit (after tag) | 6.0.0 | 6.0.0

JakeGinnivan:
> More precisely: Take the graph along the commits (including following any parent branches). Then act as if they were all on one branch, and use the single value of increment to determine how increases in the version number are to be handled.

In reality does this end up with a different version than us figuring out parent, calculating version then taking that version instead? I know the description doesn't exactly match what we end up doing, but if we switched the default over to inherit would that fix the incorrect versions. Or do we need to change the way we calculate the version when inheriting?

JakeGinnivan:
I am happy for you to propose fixes, either to the docs or to the code so they line up. Also happy for default to change to inherit in v4.

DanielRose:
>> More precisely: Take the graph along the commits (including following any parent branches). Then act as if they were all on one branch, and use the single value of increment to determine how increases in the version number are to be handled.

> In reality does this end up with a different version than us figuring out parent, calculating version then taking that version instead? I know the description doesn't exactly match what we end up doing, but if we switched the default over to inherit would that fix the incorrect versions. Or do we need to change the way we calculate the version when inheriting?

You can see the difference in the table I posted. On master, we are on version 2.0. Then branch to "child_minor", and suddenly the version is 1.1. Changing the default to inherit wouldn't help, since it has an explicit value "minor".

> I am happy for you to propose fixes, either to the docs or to the code so they line up. Also happy for default to change to inherit in v4.

Ok, I'll try to find some time for that.

maxraab:
@JakeGinnivan @DanielRose: What is the current state of play?

DanielRose:
@maxraab @JakeGinnivan I was on vacation. I'll take a closer look at the code in the next couple of days.

DanielRose:
I was able to take a closer look at the code and started looking how to fix this. Mainly so far I've added many comments to better understand what is being done.

The main changes need to be done in the version calculators themselves. Those that work with the commits (ex. MergeMessageBaseVersionStrategy) use the LibGit Branch.Commits directly. However, a branch (and especially its commits) are defined by LibGit to include the parent branches as well. So I'll need to change this to have a property with the commits only of the current branch (i.e. only up to the point where the branch was created).

So that is my next step, which I wasn't able to do so far.

maxraab:
@DanielRose Are there any news concerning the version calculation?

DanielRose:
@maxraab I was able to get a list of branch configs and the commits belonging to them. Next I'll need to modify the version calculators to use that list instead. I'll try to find some time to continue working on it.

DanielRose:
@maxraab @JakeGinnivan I tried a bunch of approaches, but my idea fundamentally does not seem to work with Git. The problem is that Git has no concept of a "parent branch", since in its heart is a DAG with labels (see also http://stackoverflow.com/questions/3161204/find-the-parent-branch-of-a-git-branch).

Lets say I have the following repository:
```
     A1 --- A2 (branch A)
    /
---X
    \
     B1 --- B2 (branch B)
```

What branch does commit `X` belong to? The correct answer (in Git) is: A _and_ B.

So when running GitVersion at commit `B2`, should I use the config of branch B for the commits B2, B1, X? Or should I use the config of branch B for the commits B2, B1, and then the config of branch A for the commit X? I could use some heuristics and/or additional config (ex. "master" always wins), but then this will get complicated quickly.

BTW, the reason this was not a problem until now, is that looking for a "parent" branch only happened when the configuration said to inherit from the parent branch. That way we know there is a parent branch (or at least there is supposed to be one), and in the previous case it is clear that the commit `X` belongs to branch A.

So how can I salvage at least part of what I did? My suggestion would be:
- Keep the additional comments I put in the code.
- Clarify the documentation of the inherit increment.
- Have the default value for branches (i.e. those without any config) be _inherit_.

JakeGinnivan:
Thanks for your patience @maxraab. I have rebased this.

@DanielRose I don't think this and your branch are incompatible, I like the rename of is-develop to tracks-release-branch so want to get that part in. Going to pull your inherit changes in next

PR #1020, opened 2016-08-25
Add nocache argument to bypass file cache

qetza:


JakeGinnivan:
Is this to work around issues where cache should be invalidated but is not? I guess it's a good thing to have just in case

qetza:
I don't remember the exact repro steps but i was playing with a test repo to get the configuration which would suite my needs and some times i had to manually delete the cache after updating the config or adding a tag.
Maybe it was a bug on the cache key computation but i think that having a explicit argument can always be handy :smile:

PR #1037, opened 2016-09-06
Add diag switch

JakeGinnivan:


pascalberger:
LGTM. Maybe it would make sense to call /diag in the VSTS task if system.verbose variable is set (at least if it runs under the new agent where Git should be available)

asbjornu:
Perhaps something similar can be done in GitVersionTask?

JakeGinnivan:
@pascalberger @asbjornu am unsure about how to add this into GitVersionTask. I think having the CLI to diagnose issues is ok?

gep13:
@JakeGinnivan agreed

asbjornu:
Hm. Since [Visual Studio has a setting for MSBuild verbosity](http://stackoverflow.com/questions/1211841/how-can-i-make-visual-studios-build-be-very-verbose#1211886) and MSBuild itself has the `/verbosity:level` switch, you'd think this was something that could be detected somehow. Propagating that into GitVersionTask would be the ideal thing to do, imho.

PR #1042, opened 2016-09-19
Switch to NUnit3TestAdapter

maxraab:
Switch to NUnit3TestAdapter to show tests in the VS testrunner window.

### DIFF
--- a/src/GitVersionCore.Tests/packages.config
+++ b/src/GitVersionCore.Tests/packages.config
@@ -9,7 +9,7 @@
   <package id="ModuleInit.Fody" version="1.5.9.0" targetFramework="net45" developmentDependency="true" />
   <package id="NSubstitute" version="1.10.0.0" targetFramework="net45" />
   <package id="NUnit" version="3.2.1" targetFramework="net45" />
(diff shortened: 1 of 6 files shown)


src/GitVersionCore.Tests/packages.config
```diff
@@ -8,8 +8,8 @@
   <package id="LibGit2Sharp.NativeBinaries" version="1.0.137" targetFramework="net45" />
   <package id="ModuleInit.Fody" version="1.5.9.0" targetFramework="net45" developmentDependency="true" />
   <package id="NSubstitute" version="1.10.0.0" targetFramework="net45" />
-  <package id="NUnit" version="3.2.1" targetFramework="net45" />
-  <package id="NUnitTestAdapter" version="2.0.0" targetFramework="net45" />
+  <package id="NUnit" version="3.4.1" targetFramework="net45" />
```
JakeGinnivan: Looks like you updated NUnit but didn't save the .csproj file before committing, which is why build is failing.

maxraab:
@JakeGinnivan You're right. I've updated the commits.

PR #1047, opened 2016-09-26
Update version-increments.md

Wesley-Lomax:
Added space to examples for minor, patch and fix as required for the regex

### DIFF
--- a/docs/more-info/version-increments.md
+++ b/docs/more-info/version-increments.md
@@ -23,7 +23,7 @@ See [Octopus deploy](../build-server-support/build-server/octopus-deploy.md)
 With v3 there are multiple approaches.
 
 ### Commit messages
-Adding `+semver: breaking` or `+semver: major` will cause the major version to be increased, `+semver: feature` or `+semver:minor` will bump minor and `+semver:patch` or `+semver:fix` will bump the patch.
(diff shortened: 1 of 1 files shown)


pascalberger:
Thanks for the PR!

I never used this feature, but shouldn't this also work without the whitespaces according to [this test case](https://github.com/GitTools/GitVersion/blob/da0cca830b9c56b3572d2c8ed5a58fd7a74871e3/src/GitVersionCore.Tests/IntegrationTests/VersionBumpingScenarios.cs#L36)? Anyone knows how this was intended to work?

Otherwise I would suggest to fix the code to allow whitespaces instead.

asbjornu:
Since `+semver: breaking`, `+semver: major` and `+semver: feature` are with a space, it makes sense for consistency to add a space to `+semver:minor`, `+semver:patch` and `+semver:fix` as well. I do agree the code should allow both, though.

Wesley-Lomax:
They are show with a space here as well  - https://gitversion.readthedocs.io/en/latest/configuration/

JakeGinnivan:
It works for both looking at the regex: https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/IncrementStrategyFinder.cs#L19

JakeGinnivan:
Thanks!

Issue #720, opened 2015-11-24
StrictFilePathMarshaler must be used on a FilePath


michaelnoonan:
One of our developers was consistently getting this error when building in Visual Studio 2015: `Could not determine assembly version: StrictFilePathMarshaler must be used on a FilePath`

I can see some references to this in `libgit2sharp`, and one in https://github.com/GitTools/GitVersion/pull/652.

The only remedy we could find under pressure was to re-clone the repository.

Unfortunately we needed to keep moving and don't have the repository to reproduce it any more. Hopefully if anyone else gets this error message we can narrow down the root cause.

JakeGinnivan:
Gathering this is on the GitVersionTask only? I wonder if it is another bug/error which is being triggered which stops the task from generating the version number and then causes this. Hopefully the underlying bug will get found via the exe and fixed then this goes away.

I might close this issue for now, if anyone hits this then post and we can reopen to investigate.

andrejohansson:
I have this error now on a cleanly installed Windows 10 machine with latest stable git and Visual Studio 2015 enterprise.

Suggestions how I can narrow it down?

asbjornu:
@andrejohansson It would be nice to know if you use GitVersionTask or `gitversion.exe`. If you use the former, output from MSBuild would be nice. Run `msbuild.exe` from the command line if you don't get any valuable build output in Visual Studio. If you use the latter, it should give you build output directly that will be valuable. Any build output you can get your hands on, post it here in a comment.

andrejohansson:
I'm using the gitversiontask, this is from my csproj file:

```
  <Error Condition="!Exists('..\packages\GitVersionTask.3.1.2\Build\portable-net+sl+win+wpa+wp\GitVersionTask.targets')" Text="$([System.String]::Format('$(ErrorText)', '..\packages\GitVersionTask.3.1.2\Build\portable-net+sl+win+wpa+wp\GitVersionTask.targets'))" />
  </Target>
  <Import Project="$(SolutionDir)\.nuget\NuGet.targets" Condition="Exists('$(SolutionDir)\.nuget\NuGet.targets')" />
  <Import Project="..\packages\GitVersionTask.3.1.2\Build\portable-net+sl+win+wpa+wp\GitVersionTask.targets" Condition="Exists('..\packages\GitVersionTask.3.1.2\Build\portable-net+sl+win+wpa+wp\GitVersionTask.targets')" />
[...]

PR #1070, opened 2016-10-21
Use the first matched branch configuration

### LINKED ISSUE #881: Explicit branch match should override default config
As mentioned in #868, it's a bit surprising when adding custom branch config to `GitVersion.yml` and then have GitVersion implicitly apply its default config to match the same branch (via regex). It would be less surprising if explicit config did override the default/implicit config. If we get a branch match on the explicit config, we shouldn't apply the default one to the same branch.

cvpcs:
This change causes GitVersion to allow multiple branch configurations to match the current branch, and instead of throwing an exception it will use the first matching configuration that was found.

Fixes #881

### DIFF
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -20,6 +20,14 @@ namespace GitVersion
             if (matchingBranches.Length > 0)
             {
                 branchConfiguration = matchingBranches[0];
+
+                if (matchingBranches.Length > 1) {
+                    Logger.WriteInfo(string.Format(
+                        "Multiple branch configurations match the current branch branchName of '{0}'. Using the first matching configuration, '{1}'. Matching configurations include: '{2}'",
+                        currentBranch.FriendlyName,
+                        branchConfiguration.Name,
+                        string.Join("', '", matchingBranches.Select(b => b.Name))));
+                }
             }
             else
             {
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -21,8 +21,9 @@ namespace GitVersion
             {
                 branchConfiguration = matchingBranches[0];
 
-                if (matchingBranches.Length > 1) {
-                    Logger.WriteInfo(string.Format(
+                if (matchingBranches.Length > 1)
+                {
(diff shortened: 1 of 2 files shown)


src/GitVersionCore/BranchConfigurationCalculator.cs
```diff
@@ -37,9 +37,6 @@ public class BranchConfigurationCalculator
 
                 return keyValuePair;
             }
-
-            const string format = "Multiple branch configurations match the current branch branchName of '{0}'. Matching configurations: '{1}'";
-            throw new Exception(string.Format(format, currentBranch.FriendlyName, string.Join(", ", matchingBranches.Select(b => b.Key))));
```
asbjornu: It would probably be nice to know that there were multiple matches, even though it now doesn't fail.

cvpcs:
I added a simple test to the GitVersionContextTests that creates two release configurations with non-mutually exclusive regex values and test which gets picked among two branches, one of which matches both.

cvpcs:
I'm not sure what's causing the error on the Travis CI build. It claims its something wrong with DocumentationSamples.GitHubFlowMajorRelease but there doesn't appear to be much output from the log on that test run. Also the error is Win32Exception: Success?

asbjornu:
@cvpcs: I think I have identified the problem in GitTools/GitTools.Core#38. As soon as that's merged, released and upgraded in GitVersion, I hope we can get rid of this and the problem @mkoertgen is having in #720.

src/GitVersionCore/BranchConfigurationCalculator.cs
```diff
@@ -25,21 +25,27 @@ public class BranchConfigurationCalculator
...
-            if (matchingBranches.Length == 1)
+            else
             {
                 var keyValuePair = matchingBranches[0];
                 var branchConfiguration = keyValuePair.Value;
 
+                if (matchingBranches.Length > 1)
+                {
+                    Logger.WriteInfo(string.Format(
+                        "Multiple branch configurations match the current branch branchName of '{0}'. Using the first matching configuration, '{1}'. Matching configurations include: '{2}'",
```
JakeGinnivan: I think this should probably be a warning..

asbjornu:
@cvpcs: Can you please have a look at the feedback you've gotten on this PR and rebase it on `master` so it hopefully turns green? 😄

PR #1072, opened 2016-10-27
Remove no break space character that messes up header

serra:
See http://gitversion.readthedocs.io/en/latest/configuration/ where header "Configuration" is not properly formatted.

Removing \ufeff (nbsp) fixes this.


serra:
\ufeff is the byte order mark and generally should not be present in files. I did not do any other encoding checks on the file.

pascalberger:
Thanks @serra!

Issue #953, opened 2016-07-18
Cannot get dynamic repository to work


tpluscode:
Dynamic repository doesn't work for me with version 3.6.1 (installed GitVersion.Portable from Choco). Below is my command and output. As you see I'm running TC on Windows

```
[Step 1/4] Starting: C:\Windows\system32\cmd.exe /c gitversion /url https://bitbucket.org/*** /u buildsvr /p ******* /b refs/heads/XYZ /c 5a078f3ab356914a9ae9ff5c69146e6663c1c9f3 /output buildserver /updateassemblyinfo
[08:31:08][Step 1/4] in directory: C:\TeamCity\buildAgent_2\work\f18fd9d2d9018dcf
INFO [07/18/16 8:31:09:11] Working directory: C:\TeamCity\buildAgent_2\work\f18fd9d2d9018dcf
INFO [07/18/16 8:31:09:11] IsDynamicGitRepository: False
ERROR [07/18/16 8:31:09:16] An unexpected error occurred:
System.IO.DirectoryNotFoundException: Can't find the .git directory in C:\TeamCity\buildAgent_2\work\f18fd9d2d9018dcf
   at GitVersion.GitPreparer.GetDotGitDirectory()
   at GitVersion.GitPreparer.GetProjectRootDirectory()
   at GitVersion.ConfigurationProvider.Verify(GitPreparer gitPreparer, IFileSystem fileSystem)
   at GitVersion.Program.VerifyArgumentsAndRun()
INFO [07/18/16 8:31:09:16] 
INFO [07/18/16 8:31:09:16] Attempting to show the current git graph (please include in issue): 
INFO [07/18/16 8:31:09:16] Showing max of 100 commits
INFO [07/18/16 8:31:09:26] ERROR: fatal: Not a git repository (or any of the parent directories): .git

INFO [07/18/16 8:31:09:11] Working directory: C:\TeamCity\buildAgent_2\work\f18fd9d2d9018dcf
INFO [07/18/16 8:31:09:11] IsDynamicGitRepository: False
ERROR [07/18/16 8:31:09:16] An unexpected error occurred:
System.IO.DirectoryNotFoundException: Can't find the .git directory in C:\TeamCity\buildAgent_2\work\f18fd9d2d9018dcf
   at GitVersion.GitPreparer.GetDotGitDirectory()
   at GitVersion.GitPreparer.GetProjectRootDirectory()
[...]

PR #1073, opened 2016-10-27
#953 Fix dynamic repositories

GeertvanHorrik:


GeertvanHorrik:
Fixes #953

asbjornu:
@GeertvanHorrik: Now that `master` builds are green again, can you please rebase this so we can see if this turns green as well?

GeertvanHorrik:
I just hit "Update branch", I hope that was the right button :-)

aateeque:
Not sure if this fixes the issue with dynamic repositories. As reported by others in #953 it seems an uninitialized `GitPreparer` may have something to do here. Maybe its my setup and I am looking at this completely the wrong way, but the issue I have found is that the constructor parameter `targetPath` for `GitPreparer` needs to be a path that contains a _.git_ directory - in other words the target path has to be within a git repository. 
If the `targetPath` is not set it uses the current directory which will be the download path for GitVersion.exe (from nuget for example while using Cake). In this instance, the call to `GetDotGitDirectory()` from `ConfigurationProvider.Verify()` throws an exception as `Repository.Discover(targetPath)` returns null for a non-git `targetPath`. Following this GitVersion will exit with 1 after failing an attempt to print a graph for a non-git directory.

GeertvanHorrik:
This requires an update of GitTools.Core, checking the status now.

GeertvanHorrik:
Just verified this stuff locally on my machine, all seems to work, I think we should merge this.

GeertvanHorrik:
But it should be merged into 4.0.0 beta. @JakeGinnivan which branch this I merge this into?

asbjornu:
@GeertvanHorrik: I've restarted all failing builds since it seems like the Mono source server was down when they were run. Stuff targeting v4 should be merged into master.

GeertvanHorrik:
Looks like the builds are not really doing anything (or at least I cannot check the details of the AppVeyor builds).

GeertvanHorrik:
But since GitTools.Core was already merged, this PR only:

1. Fixes possible NullReferenceException
2. Adds more unit tests

asbjornu:
@GeertvanHorrik: True. Since Windows and OS X builds are green, this can safely be merged. It's only Linux builds that fail with:

```
Failed to fetch http://download.mono-project.com/repo/debian/pool/main/m/mono/mono-jay_4.6.2.16-0xamarin1_amd64.deb  503  Service Temporarily Unavailable [IP: 52.85.142.49 80]
```

GeertvanHorrik:
@asbjornu Thanks for the fast replies, really appreciate this. Can you merge it, not sure if I should use "rebase & merge" or "squash & merge".

asbjornu:
@GeertvanHorrik: No problem. I've rebased 3badcd2d0f65201ead5c209ceadc5a3d33a1307b on top of the `HEAD` of `master` to make the merge a bit less messy. When it's done testing, I think it's good to go. :shipit:.

GeertvanHorrik:
Thanks, looks like everything succeeded (at least the AppVeyor builds). I'll merge this now.

JakeGinnivan:
@GeertvanHorrik just running these tests, they are pretty slow because they talk to GitHub and actually close another repo (they take about 5 seconds each..). Can you submit a PR which changes these tests to use the fixture approach. We have other tests which simulate the dynamic repo builds and run super fast.

JakeGinnivan:
I also get

```
Message: System.UnauthorizedAccessException : Access to the path 'pack-101bf845e769c8eee6dfff4cefaaf619d573f0b4.idx' is denied.
```

The second time I run the tests..

PR #1081, opened 2016-11-08
Support setting Verbosity of GitVersion

### LINKED ISSUE #955: Introduce Verbose/Debug logging level
We are adding more and more logging in and often I want less info. We should create another logging level and enable it with a --debug switch or a config setting?

tomap:
Hi,

I added support to set verbosity of GitVersion when launching command line

GitVersion.exe -Verbosity Error allows to have less info/debug messages in your logs

fixes #955 

Thomas P.

### DIFF
--- a/src/GitVersionTfsTask/GitVersion.ps1
+++ b/src/GitVersionTfsTask/GitVersion.ps1
@@ -54,6 +54,13 @@ if($additionalArguments)
   $argsGitVersion = $argsGitVersion + " $additionalArguments"
 }
 
+if(!$additionalArguments -or !$additionalArguments.Contains("-debug"))
+{
+  if([string]::IsNullOrEmpty($env:SYSTEM_DEBUG) -or $env:SYSTEM_DEBUG -ne "true"){
+    $argsGitVersion = $argsGitVersion + " -verbosity Error"
+  }
+}
+
 Write-Host (Get-LocalizedString -Key "Invoking GitVersion with {0}" -ArgumentList $argsGitVersion)
 
 Invoke-Tool -Path $GitVersionPath -Arguments "$argsGitVersion"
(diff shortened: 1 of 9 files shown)


src/GitVersionTfsTask/GitVersion.ps1
```diff
@@ -54,6 +54,13 @@ if($additionalArguments)
   $argsGitVersion = $argsGitVersion + " $additionalArguments"
 }
 
+if(!$additionalArguments -or !$additionalArguments.Contains("-debug"))
+{
+  if([string]::IsNullOrEmpty($env:SYSTEM_DEBUG) -or $env:SYSTEM_DEBUG -ne "true"){
```
JakeGinnivan: This is too generic for a variable name. We should name it `GITVERSION_LOG_LEVEL`

JakeGinnivan:
The build is fixed now. If you could change the variable above and rebase this branch that would be great. You should be able to just `git pull --rebase` to rebase instead of merge

src/GitVersionTfsTask/GitVersion.ps1
JakeGinnivan: I see where you were going with this now after looking through the commits. I am going to remove for now. I think we should extend the VSTS task to expose the log level.

JakeGinnivan:
This is in now through #1107. Thanks!

src/GitVersionTfsTask/GitVersion.ps1
JakeGinnivan: I removed this for the moment. I figured it can be another PR which we can look at separately

pascalberger: Makes sense

PR #1082, opened 2016-11-09
Fixes #1053: exclude NCrunch ItemGroup when NCrunch not present

### LINKED ISSUE #1053: "Path is not relative" when GitVersionTask is used in ccproj files
Adding the Import

`<Import Project="..\packages\GitVersionTask.3.6.3\build\portable-net+sl+win+wpa+wp\GitVersionTask.targets" Condition="Exists('..\packages\GitVersionTask.3.6.3\build\portable-net+sl+win+wpa+wp\GitVersionTask.targets')" />`

into a ccproj (Azure cloud service) project results in the project not loading in Visual Studio with the error:

`path is not relative`

By a process of elimination, it is caused by [this block of ncrunch-related `ItemGroup` `None` definitions](https://github.com/GitTools/GitVersion/blob/v3.6.3/src/GitVersionTask/NugetAssets/GitVersionTask.targets#L82-L92). If those are commented out then the error does not occur. If the first one is left in, then the error comes back and the project will not load.

Workarounds:
- Comment out [this entire `ItemGroup`](https://github.com/GitTools/GitVersion/blob/v3.6.3/src/GitVersionTask/NugetAssets/GitVersionTask.targets#L81-L92)
- Add a `Condition=" $(NCrunch) != '' "` to the item group on [this line](https://github.com/GitTools/GitVersion/blob/v3.6.3/src/GitVersionTask/NugetAssets/GitVersionTask.targets#L82).

Are these ncrunch definitions required in the targets file once it is deployed to the nupkg? If not could they be removed, and if so could we just add the condition to them (I'm not sure if that condition is correct, but it "Works On My MachineTM")?

AGBrown:
This adds a condition to the ItemGroup that is required for NCrunch so that it is excluded when NCrunch is not installed. This is similar to the conditions [here](https://github.com/GitTools/GitVersion/blob/v3.6.3/src/GitVersionTask/GitVersionTask.csproj#L110) and [here](https://github.com/GitTools/GitVersion/blob/v3.6.3/src/GitVersionExe/GitVersionExe.csproj#L83) which also affect a section depending on the presence of NCrunch

### DIFF
--- a/src/GitVersionTask/NugetAssets/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/GitVersionTask.targets
@@ -79,7 +79,7 @@
   </Target>
 
   <!--Support for ncrunch-->
-  <ItemGroup>
+  <ItemGroup Condition=" $(NCrunch) != '' ">
     <None Include="$(MSBuildThisFileDirectory)..\..\GitVersionTask.dll">
       <Visible>False</Visible>
     </None>


AGBrown:
appveyor error is seemingly unrelated:

```
Cake.exe : An error occured when executing task 'Create-Release-Notes'.
At line:1 char:1
    + CategoryInfo          : NotSpecified: (An error occure...Release-Notes'.:String) [], RemoteException
 
Error: 
Could not find file 'C:\projects\gitversion\build\releasenotes.md'.
Command executed with exception: 
```

pascalberger:
@gep13 Any idea why AppVeyor tried to create release notes? I think we removed this for PR builds

pascalberger:
@AGBrown You have created the PR to merge into GitTools:support/3.6.0 which tries to build release notes also for PRs (and fails with that). Can you please change the PR to merge into master instead?

AGBrown:
@pascalberger yes, I can do that. Does that mean that (if accepted) it will still get applied to the support/v3.6.0 branch as well?

Sorry for the confusion - I was looking at the support branch and it looked like previous PRs (e.g. 9846b201) had been merged to the support branch first (e.g. 5ad31739), and then back-merged into master (e.g. 08f882e1).

I'll change the PR tonight/tomorrow for you.

AGBrown:
Closing to redo this based on `master` instead

AGBrown:
Re-opened as per @JakeGinnivan [request](https://gitter.im/GitTools/GitVersion?at=5839b3fb89d179bf4dc98154)

> Would you be able to submit a pr against the 3.6 branch?

JakeGinnivan:
Thanks @AGBrown

PR #1085, opened 2016-11-17
Use the Inherit strategy by default

DanielRose:
Alternative implementation for PR #1004:

By default, branches without an explicit value for `increment` use the value `Inherit` (this was `Patch` before). If the default value should remain Patch, this can be easily changed in ConfigurationProvider.cs.

The default value for `increment` can be set "globally" in the config, just as can be done for `mode`.

Extended and clarified the documentation for increment and the version number sources.

Added a bunch of comments in the code and did some refactoring/cleanup.


JakeGinnivan:
Rebased then reopened at #1103

PR #1087, opened 2016-11-18
Fixes #1053: exclude NCrunch ItemGroup when NCrunch not present

### LINKED ISSUE #1053: "Path is not relative" when GitVersionTask is used in ccproj files

AGBrown:
This adds a condition to the ItemGroup that is required for NCrunch so that it is excluded when NCrunch is not installed. This is similar to the conditions here and here which also affect a section depending on the presence of NCrunch

This replaces #1082 (which was a PR into the support/3.6.0 branch) at the request of @pascalberger

### DIFF
--- a/src/GitVersionTask/NugetAssets/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/GitVersionTask.targets
@@ -81,7 +81,7 @@
   </Target>
 
   <!--Support for ncrunch-->
-  <ItemGroup>
+  <ItemGroup Condition=" $(NCrunch) != '' ">
     <None Include="$(GitVersionTaskDir)GitVersionTask.dll">
       <Visible>False</Visible>
     </None>


AGBrown:
appveyor error:

```
File not found: C:/projects/gitversion/build/releasenotes.md
Cake.exe : An error occurred when executing task 'Upload-AppVeyor-Artifacts'.
At line:1 char:1
    + CategoryInfo          : NotSpecified: (An error occurr...yor-Artifacts'.:String) [], RemoteException
 
Error: 
UploadArtifact failed (2).
Command executed with exception: UploadArtifact failed (2).
```

pascalberger:
@AGBrown Thanks! 

We are aware of the build issue and need to fix build on master first, afterwards you can rebase the branch and we can merge. I'll let you know as soon as master is fixed.

src/GitVersionTask/NugetAssets/GitVersionTask.targets
```diff
@@ -81,7 +81,7 @@
   </Target>
 
   <!--Support for ncrunch-->
-  <ItemGroup>
+  <ItemGroup Condition=" $(NCrunch) != '' ">
```
JakeGinnivan: Oops, looks like that goe removed at some point. Thanks for fixing

PR #1102, opened 2016-11-26
Feature branch does not pick up version

JakeGinnivan:
Rebased #1004 and some minor fixes

### DIFF
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -1,13 +1,11 @@
 namespace GitVersion
 {
+    using JetBrains.Annotations;
+    using LibGit2Sharp;
     using System;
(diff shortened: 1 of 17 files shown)


src/GitVersionCore/BranchConfigurationCalculator.cs
```diff
@@ -22,7 +20,7 @@ public class BranchConfigurationCalculator
                     currentBranch.FriendlyName));
 
                 var branchConfig = new BranchConfig();
-                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig, "");
+                ConfigurationProvider.ApplyBranchDefaults(config, branchConfig, "", isDevelop: true);
```
JakeGinnivan: I removed that with my cleanup. This branch doesn't functionally change anything, it is more about the rename from is-develop to tracks-release-branches

PR #1107, opened 2016-11-27
Log verbosity

JakeGinnivan:
Rebased and builds on the PR by @tomap. #1081


JakeGinnivan:
Thanks @pascalberger

asbjornu:
Great! Just a quick question: Can we propagate MSBuild's verbosity down into GitVersionTask somehow?

JakeGinnivan:
Yeah probably, would just be a case of updating the msbuild tasks. Maybe open a up for grabs

PR #1112, opened 2016-12-05
More caching of merge bases

DanielRose:
Merge bases are symmetric, i.e. the merge base of 'branchA' and 'branchB' is the same as 'branchB' and 'branchA'.
Consider a local and its remote (tracked) branch the same, with the state of the local branch preferred.

### DIFF
--- a/src/GitVersionCore/GitRepoMetadataProvider.cs
+++ b/src/GitVersionCore/GitRepoMetadataProvider.cs
@@ -2,6 +2,7 @@
 using LibGit2Sharp;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 
 namespace GitVersion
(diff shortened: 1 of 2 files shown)


src/GitVersionCore/GitRepoMetadataProvider.cs
```diff
@@ -105,17 +106,26 @@ public IEnumerable<Branch> GetBranchesContainingCommit([NotNull] Commit commit,
 
         /// <summary>
         /// Find the merge base of the two branches, i.e. the best common ancestor of the two branches' tips.
+        /// Note that a local branch and a remote branch tracked by it (or when untracked, has the same name) are considered the same.
         /// </summary>
         public Commit FindMergeBase(Branch branch, Branch otherBranch)
         {
-            var key = Tuple.Create(branch, otherBranch);
+            if (branch.IsSameBranch(otherBranch))
```
JakeGinnivan: This should ensure the tips are also pointing at the same place.

DanielRose: @JakeGinnivan Actually here it might not be the case.

> Consider a local and its remote (tracked) branch the same, with the state of the local branch preferred.

Normally for the merge base it will make no difference if the local or remote branch is not up to date. That is only the case if the local and/or the remote branch was rebased. With this change (actually the algorithm in `IsSameBranch()`), the state (and thus merge base) of the local branch is always chosen.

JakeGinnivan:
After trying to rebase this I have a failing test and thought I would check to see if you can diagnose easier than I can. I am out of time for my mornings catch up session @DanielRose

DanielRose:
@JakeGinnivan I was (and am) pretty busy the last few weeks. I'll take a closer look at this as soon as I can.

JakeGinnivan:
Yeah no worries @DanielRose, understand that too well.

JakeGinnivan:
I know this is a while back now, should we close this or did you want to take another run at it?

JakeGinnivan:
@DanielRose

JakeGinnivan:
I am having a play with a completely new approach at #1243, I have high hopes its a better way forward than internal caching. Going to close this one down as it will likely have to be redone manually due to too much drift. Sorry about that. In fact I am now using the sha for the cache keys instead of branches so that may actually have the same effects as this PR anyways.

DanielRose:
@JakeGinnivan I was on a long vacation with limited Internet access. Once I find some free time I can take a look at your PR and see if I find anything :)

PR #1116, opened 2016-12-09
More logging when preparing the git repository

DanielRose:
Adds some more logging in GitPreparer:

- Use Logger.IndentLog() to show where time is spent.
- Enable the logging of GitTools.Core.


JakeGinnivan:
Nice improvements

Issue #1118, opened 2016-12-14
Support for VS 2017 cross-targeting builds


onovotny:
VS 2017 now has support for cross-targeting builds that can generate a NuGet package. This operates with an inner and outer loop. The inner loop builds each TFM and works with the current `GitVersionTask` package as-is. The outer-loop (where things like Pack/Publish) targets run needs one update to the packaging of the targets: a copy of the targets also needs to go into `\buildCrossTargeting`. This name may change as per https://github.com/dotnet/sdk/issues/486 to `\buildMultiTargeting`, so for now, easiest just to put a copy of the targets in each.

Instead of being a copy, it can be a new set of targets that have this:

` <Import Project="..\build\GitVersionTask.targets"/>`

There may be some other updates too. Instead of/in addition to doing a `CoreCompile` hook for the inner loop, it would be good to use `CoreGenerateAssemblyInfo` since that task using the new SDK build system is what writes the assembly info.

Here's some more info:
`C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\MSBuild\Sdks\NuGet.Build.Tasks.Pack\buildCrossTargeting` look there and you can see how you can distinguish between MSBuild running on CoreCLR and Desktop as to load the right version of a task...

In the outer loop, you'd want to inject the logic before the `GenerateNuspec` target

onovotny:
Example of how to load different build tasks for desktop vs coreclr:

```csproj
  <PropertyGroup Condition="$(NugetTaskAssemblyFile) == ''">
    <NugetTaskAssemblyFile Condition="'$(MSBuildRuntimeType)' == 'Core'">..\CoreCLR\NuGet.Build.Tasks.Pack.dll</NugetTaskAssemblyFile>
    <NugetTaskAssemblyFile Condition="'$(MSBuildRuntimeType)' != 'Core'">..\Desktop\NuGet.Build.Tasks.Pack.dll</NugetTaskAssemblyFile>
  </PropertyGroup>
```

kzu:
Or you could just switch to GitInfo which supports this already (see https://github.com/kzu/GitInfo/pull/53)

PR #1119, opened 2016-12-15
Inject into correct point in build pipeline for VS 2017

onovotny:
This is an initial attempt at implementing the .NET Desktop side of #1118.

Notably, it's a bit "whack a mole" in terms of ensuring the `GetVersion` target is executed early enough. Then, without trying to overwrite what the user wants, the user will have to add something like this into their project file in order to copy the generated values into the correct version variables:

```csproj
  <Target Name="UpdateVersionVars" AfterTargets="GetVersion">
    <PropertyGroup>
      <Version>$(GitVersion_SemVer)</Version>
      <PackageVersion>$(GitVersion_SemVer)</PackageVersion>
    </PropertyGroup>
  </Target>
```

Copying to `Version` at that stage will ensure that the `AssemblyFileVersion` and `AssemblyVersion` are set based on that `Version` turns into the informational version. I'm not sure if you can use the `FullSemVer` there or if that'll cause the parsing to barf.... either way, the user is in control of what variables they assign to what there.

If there's a better way to do this than to have the user supply an `AfterTargets`, then I'm all for it. I tested this locally and it did correctly generate the file/assembly versions and the package version when using the pack command.

### DIFF
--- a/src/GitVersionTask/GitVersionTask.csproj
+++ b/src/GitVersionTask/GitVersionTask.csproj
@@ -128,7 +128,6 @@
     <Copy SourceFiles="$(TargetDir)GitVersionTask.dll.mdb" DestinationFolder="$(BuildDir)NuGetTaskBuild\build" Condition="Exists('$(TargetDir)GitVersionTask.dll.mdb')" />
     <Copy SourceFiles="$(ProjectDir)NugetAssets\build\GitVersionTask.targets" DestinationFolder="$(BuildDir)NuGetTaskBuild\build" />
     <Copy SourceFiles="$(ProjectDir)NugetAssets\buildMultiTargeting\GitVersionTask.targets" DestinationFolder="$(BuildDir)NuGetTaskBuild\buildMultiTargeting" />
-    <Copy SourceFiles="$(ProjectDir)NugetAssets\buildMultiTargeting\GitVersionTask.targets" DestinationFolder="$(BuildDir)NuGetTaskBuild\buildCrossTargeting" />
     <Copy SourceFiles="$(ProjectDir)NugetAssets\GitVersionTask.nuspec" DestinationFolder="$(BuildDir)NuGetTaskBuild" />
     <PepitaPackage.CreatePackageTask NuGetBuildDirectory="$(BuildDir)NuGetTaskBuild" MetadataAssembly="$(ILMergeTemp)GitVersionTask.dll" Version="$(GitVersion_NuGetVersion)" />
     <Delete Files="@(TempFiles)" />
--- a/src/GitVersionTask/GitVersionTask.csproj
+++ b/src/GitVersionTask/GitVersionTask.csproj
@@ -79,9 +79,8 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
-    <None Include="NugetAssets\GitVersionTask.targets">
-      <SubType>Designer</SubType>
-    </None>
+    <None Include="NugetAssets\buildMultiTargeting\GitVersionTask.targets" />
+    <None Include="NugetAssets\build\GitVersionTask.targets" />
     <None Include="NugetAssets\GitVersionTask.nuspec">
       <SubType>Designer</SubType>
     </None>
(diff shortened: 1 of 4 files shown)


onovotny:
Related to this in terms of making it easier to extend the `Version` tag https://github.com/dotnet/sdk/issues/504

onovotny:
I've made a few updates:

1. Just works now by default in all projects. In "SDK" projects, `UpdateAssemblyInfo` gets defaulted to off and instead it'll set the appropriate MSBuild properties that other tasks use to generate the assembly info and nuget packaging. Of course, this can be turned off.
2. It can optionally use the full semver 2.0 in the nuget packaging. This is off by default since many servers don't yet support it (I believe the symbolsource stuff breaks currently). You can set `<UseFullSemVerForNuGet>true</UseFullSemVerForNuGet>` in your project file to opt-in to using SemVer 2.0 package versions.

asbjornu:
@onovotny: Is this backwards compatible with VS2015?

onovotny:
It should be, it adds a few properties that weren’t in or used before and takes an additional “or” dependency on a couple of targets that weren’t there before either.

Sent from my Windows 10 phone

thoemmi:
Any news on this? I have a new VS2017 RC project cross-targeting `net40` and `net45`: 
```xml
  <PropertyGroup>
    <TargetFrameworks>net40;net45</TargetFrameworks>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0009" />
  </ItemGroup>
  <!-- ... -->
```
The compile failes because of duplicate `Assembly*VersionAttribute`s.
```
1>obj\Debug\net40\StickyWindows.AssemblyInfo.cs(17,12,17,58): error CS0579: Duplicate 'System.Reflection.AssemblyFileVersionAttribute' attribute
1>obj\Debug\net40\StickyWindows.AssemblyInfo.cs(18,12,18,67): error CS0579: Duplicate 'System.Reflection.AssemblyInformationalVersionAttribute' attribute
1>obj\Debug\net40\StickyWindows.AssemblyInfo.cs(21,12,21,54): error CS0579: Duplicate 'System.Reflection.AssemblyVersionAttribute' attribute
```

onovotny:
@thoemmi The fix to that is to remove those attributes from your AssemblyInfo file. That file is ideally completely generated now anyway by attributes in the csproj.

thoemmi:
@onovotny  Wisecracker 😉 Of course I already would have done that if there were an AssemblyInfo file.
It's a MSBuild "15" project, and there's no AssemblyInfo.cs.

Anyway, I think I can manage to create the correct version information. My current .csproj looks like this
```xml
  <PropertyGroup>
    <TargetFrameworks>net40;net45</TargetFrameworks>

    <UpdateAssemblyInfo>false</UpdateAssemblyInfo>
    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
  </PropertyGroup>
  <ItemGroup>
    <PackageReference Include="GitVersionTask" Version="4.0.0-beta0009" />
  </ItemGroup>
  <Target Name="UpdateVersionVars" AfterTargets="GetVersion">
    <Message Text="$(GitVersion_SemVer)" Importance="High" />
    <PropertyGroup>
      <FileVersion>$(GitVersion_AssemblySemVer)</FileVersion>
      <AssemblyVersion>$(GitVersion_AssemblySemVer)</AssemblyVersion>
      <PackageVersion>$(GitVersion_SemVer)</PackageVersion>
      <InformationalVersion>$(GitVersion_SemVer)</InformationalVersion>
    </PropertyGroup>
  </Target>
  <!-- ... -->
```
The `<UpdateAssemblyInfo>false</UpdateAssemblyInfo>` says GitVersion ot to create an AssemblyInfo on its own. Instead, MSBuild uses a common set of properties: `FileVersion`, `AssemblyVersion`, `PackageVersion`, and `InformationalVersion`.
This works **except** `PackageVersion`: The generated Nupgk still uses the AssemblySemVer. I'll continue exploring MSBuild's .targets files, and post an update when I have found a solution.

src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets
```diff
@@ -0,0 +1,80 @@
...
+    <!-- Property that enables WriteVersionInfoToBuildLog -->
+    <WriteVersionInfoToBuildLog Condition=" '$(WriteVersionInfoToBuildLog)' == '' ">true</WriteVersionInfoToBuildLog>
+
+    <!-- Property that enables UpdateAssemblyInfo. Default to off for SDK builds -->
+    <UpdateAssemblyInfo Condition=" '$(UpdateAssemblyInfo)' == '' and '$(TargetFramework)' != '' ">false</UpdateAssemblyInfo>
+    <UpdateAssemblyInfo Condition=" '$(UpdateAssemblyInfo)' == '' ">true</UpdateAssemblyInfo>
+
+    <GetVersion Condition=" '$(GetVersion)' == '' ">true</GetVersion>
+    
+
```
pascalberger: Remove multiple empty lines

src/GitVersionTask/GitVersionTask.csproj
```diff
@@ -79,9 +79,8 @@
   </ItemGroup>
   <ItemGroup>
     <None Include="app.config" />
-    <None Include="NugetAssets\GitVersionTask.targets">
```
pascalberger: Shouldn't the file NugetAssets\GitVersionTask.targets be deleted?

src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
```diff
@@ -8,8 +8,13 @@
     <!-- Property that enables WriteVersionInfoToBuildLog -->
     <WriteVersionInfoToBuildLog Condition=" '$(WriteVersionInfoToBuildLog)' == '' ">true</WriteVersionInfoToBuildLog>
 
-    <!-- Property that enables UpdateAssemblyInfo -->
+    <!-- Property that enables UpdateAssemblyInfo. Default to off for SDK builds -->
+    <UpdateAssemblyInfo Condition=" '$(UpdateAssemblyInfo)' == '' and '$(TargetFramework)' != '' ">false</UpdateAssemblyInfo>
     <UpdateAssemblyInfo Condition=" '$(UpdateAssemblyInfo)' == '' ">true</UpdateAssemblyInfo>
+    
+    <!-- Property that enables setting of Version -->
+    <UpdateVersionProperties Condition=" '$(UpdateVersionProperties)' == '' ">true</UpdateVersionProperties>
```
pascalberger: Would be nice if these properties can be documented [here](https://github.com/GitTools/GitVersion/blob/master/docs/usage/msbuild-task.md)

src/GitVersionTask/GitVersionTask.csproj
onovotny: it was moved into the build folder

src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
onovotny: will update the docs now

src/GitVersionTask/GitVersionTask.csproj
pascalberger: Ah, missed this :)

onovotny:
I've added docs - I think I've covered it?

pascalberger:
@onovotny Thanks LGTM. I'll merge it as soon as the build has passed.

onovotny:
Once GitVersion supports CoreCLR and linux/mac, we can include a .net core version of these as well. I think LibGit2Sharp has that support now? I think NerdBank.GitVersioning is using that and he has a coreclr task.

pascalberger:
@onovotny It's merged now. Thanks for your contribution!

Issue #544, opened 2015-07-28
MajorMinorPatchTag - AssemblyVersion vs AssemblyFileVersion


veccie:
I don't know if this is a bug or how it's suppose to be, so this is maybe more a question about the logic and thought behind MajorMinorPatchTag.

When using the MajorMinorPatchTag scheme the applied assembly versions look like this:

```
[assembly: AssemblyVersion("0.4.0.12")]
[assembly: AssemblyFileVersion("0.4.0.0")]
[assembly: AssemblyInformationalVersion("0.4.0-alpha.12+Branch...")]
```

The AssemblyVersion has number of commits(which is correct) but AssemblyFileVersion doesn't. When I look in the code I can see that AssemblyVersion is set according to the scheme(via AssemblySemVer). AssemblyFileVersion however is always set to MajorMinorPatch + ".0", which means that the MajorMinorPatchTag scheme actually is ignored?

Why isn't AssemblyFileVersion set to AssemblySemVer also? Does it course problems elsewhere?

The AssemblyVersion is hidden from file properties, so it can be confusing to see 0.4.0.0 when deep down it's actually 0.4.0.12. I know that AssemblyInformationalVersion for example shows 0.4.0-alpha.12+Branch... but that doesn't necessarily tell me the truth about the AssemblyVersion, unless MajorMinorPatch scheme is used.

If these version numbers really are suppose to be different then shouldn't it be the other way around, if you consider what the CLR cares about: AssemblyVersion?

```
[assembly: AssemblyVersion("0.4.0.0")]
[assembly: AssemblyFileVersion("0.4.0.12")]
```

In any case, I think both version properties should be set to the same version: 0.4.0.12 when using the MajorMinorPatchTag scheme. In fact couldn't they always be set to the same version?

JakeGinnivan:
Is this in the Task or the exe?

veccie:
Both:

Exe:
https://github.com/GitTools/GitVersion/blob/master/src/GitVersionExe/AssemblyInfoFileUpdate.cs#L38

Task:
https://github.com/GitTools/GitVersion/blob/master/src/GitVersionTask/AssemblyInfoBuilder/AssemblyInfoBuilder.cs#L42

JakeGinnivan:
@SimonCropp I think both should listen to the config?

JakeGinnivan:
@veccie feel free to submit a PR for this, I agree. We should respect that setting.

cjam:
Hey Guys, 

[...]

PR #1130, opened 2017-01-06
Created new configuration-option assembly-file-versioning-scheme:

jensbrand:
Created new configuration-option assembly-file-versioning-scheme: It has the same possible options as assembly-versioning-scheme and it is used for generating the AssamblyFileVersion just in the same matter as the AssemblyVersion.
Since the AssemblyVersion is used for referencing and AssemblyLoading and AssemblyFileVersion is just a representation of the version / build of the assembly, we can now configure both independent.
Default setting is MajorMinorPatch for both options.

This change is discussed in issue: #544


JakeGinnivan:
Looks good, two small comments.

1. None of the tests actually verify MajorMinorPatchTag, could you change an existing test or just add a new one
2. It would be great to replace the tabs with spaces to be consistent. We probably should have http://editorconfig.org/ configured to prevent this issue (feel free to add)

jensbrand:
I've done the requested changes. One additional test checks the configuration-option (ConfigProviderTests.CanUpdateAssemblyFileVersioningScheme()) and another test checks the result of the possible settings of AssemblyFileVersioningScheme (AssemblyFileVersionTests.ValidateAssemblyFileVersionBuilder()).

JakeGinnivan:
All merged @jensbrand, thanks. Sorry it took so long

PR #1138, opened 2017-01-16
Commit message counted twice?

LarsAndreasEk:
Hi,

git init

add some file
commit to master
tag the commit "1.0"

make a new commit to master, with message "+semver:major"

create release branch (release/2.0) from master

run GitVersion from release branch.

For the given scenario, GitVersion thinks semver should be 3.0.0-beta.1+1. I think it should be 2.0.0-beta.1+0.

I might be mistaken, but I assume this is not the intended behaviour?

What do you think?

### DIFF
--- a/src/GitVersionCore.Tests/IntegrationTests/ReleaseBranchScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/ReleaseBranchScenarios.cs
@@ -360,4 +360,18 @@ public class ReleaseBranchScenarios
             fixture.AssertFullSemver(config, "2.0.0-beta.2");
         }
     }
+
+    [Test]
+    public void ReleaseBranchShouldUseBranchNameVersionDespiteBumpInPreviousCommit()
+    {
+        using (var fixture = new EmptyRepositoryFixture())
+        {
+            fixture.Repository.MakeATaggedCommit("1.0");
+            fixture.Repository.MakeACommit("+semver:major");
+
+            Commands.Checkout(fixture.Repository, fixture.Repository.CreateBranch("release/2.0"));
+
+            fixture.AssertFullSemver("2.0.0-beta.1+0");
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


LarsAndreasEk:
What do you think, @JakeGinnivan and @asbjornu? :)

JakeGinnivan:
Does it do it if you put a commit on master before taking the release branch?

Technically the commit is included in both branches and because git does not track what branches were made off what other branches there is no way to know which branch was created second.

At first look, I would not know how to approach fixing this.

LarsAndreasEk:
Hi,

I've committed a change to the test with the commit you asked. The result is the same, though. 

I don't know how to approach fixing this either, but I suspect it has something to do with the base version source selected by BaseVersionCalculator at line 73-75. If I change

`                var calculatedBase = new BaseVersion(
                    context, maxVersion.Version.Source, maxVersion.Version.ShouldIncrement, maxVersion.Version.SemanticVersion,
                    baseVersionWithOldestSource.BaseVersionSource, maxVersion.Version.BranchNameOverride);
`
 to

` var calculatedBase = new BaseVersion(
                    context, maxVersion.Version.Source, maxVersion.Version.ShouldIncrement, maxVersion.Version.SemanticVersion,
                    maxVersion.Version.BaseVersionSource, maxVersion.Version.BranchNameOverride);
`

my test goes green, but 12 existing tests goes red :)

As far as I can understand, the base version source is used both to count commits and to iterate these commits to look for +semver commit messages.

JakeGinnivan:
Yeah, the thing is the last release is the last tag as far as gitversion is concerned. Branch points have less to do with it. 

A fix for the way the code is now is to possibly remove any commit message bumps from any commits later than the merge base. Unsure how difficult that would be

LarsAndreasEk:
I made a commit that seems to solve the issue at hand. All tests are green at least, but I'm not sure whether or not I've introduced some unwanted behavior to the code.

LarsAndreasEk:
Any thoughs, @JakeGinnivan? :)

JakeGinnivan:
I guess we trust our tests :) if they say all is good then it is. These sort of changes always seem to break someones workflow in some way but it makes sense to me..

To be clear, this won't fix your original issue though I don't think (if you removed the additional commit between the first message before you branch)

LarsAndreasEk:
I've committed a few tests that verify some variants. One of them is the test from the original issue. It's still green :)

LarsAndreasEk:
Any hope getting this merged to master? :)

JakeGinnivan:
Thanks for this @LarsAndreasEk, good improvement

PR #1144, opened 2017-01-20
Add check to see if a logger is enabled before logging

tomap:
See commit comments

Thomas

### DIFF
--- a/src/GitVersionCore.Tests/ModuleInitializer.cs
+++ b/src/GitVersionCore.Tests/ModuleInitializer.cs
@@ -11,7 +11,7 @@ public static class ModuleInitializer
     /// </summary>
     public static void Initialize()
     {
-        Logger.SetLoggers(
+        Logger.SetLoggers(VerbosityLevel.Debug,
             s => Console.WriteLine(s),
             s => Console.WriteLine(s),
             s => Console.WriteLine(s),
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -32,9 +32,9 @@ namespace GitVersion
             }
             else
             {
-                Logger.WriteInfo(string.Format(
-                    "No branch configuration found for branch {0}, falling back to default configuration",
-                    targetBranch.FriendlyName));
+                if (Logger.IsInfoEnabled) Logger.WriteInfo(string.Format(
+                                                    "No branch configuration found for branch {0}, falling back to default configuration",
+                                                    targetBranch.FriendlyName));
 
                 branchConfiguration = new BranchConfig { Name = string.Empty };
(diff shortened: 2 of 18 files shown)


asbjornu:
I like the idea of this, but wouldn't it be better to move the check for `Logger.IsInfoEnabled` and similar inside the `LogMessage` method, so it doesn't spread all over the code base? I like being able to write out log statements without having to remember checking a boolean first.

DanielRose:
@asbjornu The logger itself already does that. The reason for this commit is so that a call to string.Format() is avoided if the logger is not enabled.

Personally, I don't see the need for this. The memory and CPU requirements for string.Format() are vanishingly small. Last time I profiled (2-3 months ago), more than 99% of the time was spent waiting for git, specifically finding the merge-bases.

tomap:
Hi, the check is only here to avoid unncessary treatments (access property, evaluate lists, ...)
A good example is 
https://github.com/GitTools/GitVersion/pull/1144/commits/22f5093847ad13f8bc0817b81de7880c56a518b7#diff-240091ccd756d009e1677290734085acR122

A less legitimate example is 
https://github.com/GitTools/GitVersion/pull/1144/commits/22f5093847ad13f8bc0817b81de7880c56a518b7#diff-dc5fcc418fdfb7dd363dd7eeec68adfaL65
where there is no formating of any kind, so the test is unnecessary.

But in some cases, you have to be aware that your debug code is taking precious time for nothnig because it will never be displayed.
The aim of this dev is performance.


As for adding the test inside the debug function, I already did that by pointing the Debug/Warn/Error/... to an empty function when the debug level is disabled.
See 
https://github.com/GitTools/GitVersion/pull/1144/commits/af70c489ba8bceffc406ef821abdf0d97c80247e#diff-4abc10f1875195711870f3564a5d5526R59

DanielRose:
I appreciate what you want to do. But as part of what GitVersion does, it hits the local git repository (i.e. the disk) very heavily. Depending on the circumstances, it even accesses the network quite a bit. In contrast to that, the time spent on logging (which will be enabled most of the time anyway) is almost immeasurably small. So the upside is very small. The downside is that for every future logging call, you have to remember to check if it is enabled first, plus it clutters the code even more than the call to the logger already does.

JakeGinnivan:
I think there are better ways to resolve removing the string format. The immediate one which comes to mind is change our logging signature from:

`public static Action<string> WriteDebug { get; private set; }`

to

`public static Action<Func<string>> WriteDebug { get; private set; }`

Which would change usage to `Logger.WriteDebug(() => `Use ${stringInterp} here`)`. Then it can be a cross cutting change?

Going to close this as it conflicts, @tomap if you like the idea of the above change another PR is welcome. Another side effect is we can put heavy debug information into the lambda (git operations).

PR #1147, opened 2017-01-27
Fix typo in gitflow-examples

rasa:


gep13:
@rasa thanks for fixing that!

PR #1149, opened 2017-02-01
Preserve existing file encoding if it can be easily determined otherwise use UTF8 with BOM.

### LINKED ISSUE #1074: UpdateAssemblyInfo strips byte order mark
The AssemblyInfo files get overwritten using `File.WriteAllText(file, fileContents)` which writes in UTF8 without a byte-order mark. All of my source files include the BOM to comply with default StyleCop rules and this strips it out causing the files to get flagged.

The simple fix for me would be to use the WriteAllText overload that takes an encoding and to pass in Encoding.UTF8, but then people that specifically don't want byte order marks would get them. The encoding of the existing file could be detected and re-used to write the new contents of the file, but based on my initial research into that it is a bit tricky to get right and may be overkill for this. Perhaps a configuration setting to set the encoding to use?

I can make the change and submit a PR but wanted to get opinions before deciding which way to go.

kll:
This fixes #1074 and will no longer strip the BOM or change the encoding when a file has a BOM. It does not handle cases where a BOM was not included or the existing encoding was not Unicode based. In those cases the files will be written using UTF8 with BOM as that is the most common and complies with [StyleCop SA1412](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1412.md)

### DIFF
--- a/src/GitVersionCore.Tests/GitVersionCore.Tests.csproj
+++ b/src/GitVersionCore.Tests/GitVersionCore.Tests.csproj
@@ -125,6 +125,7 @@
     <Compile Include="ConfigProviderTests.cs" />
     <Compile Include="GitVersionContextTests.cs" />
     <Compile Include="Helpers\DirectoryHelper.cs" />
+    <Compile Include="IntegrationTests\FileSystemTests.cs" />
     <Compile Include="IntegrationTests\MainlineDevelopmentMode.cs" />
     <Compile Include="LogMessages.cs" />
     <Compile Include="Mocks\MockThreadSleep.cs" />
--- a/src/GitVersionCore.Tests/IntegrationTests/FileSystemTests.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/FileSystemTests.cs
@@ -1,0 +1,68 @@
+﻿using System.IO;
+using System.Text;
+
+using GitVersion.Helpers;
+
+using NUnit.Framework;
+
+using Shouldly;
+
+[TestFixture]
+public class FileSystemTests
+{
+    public string TempFilePath { get; set; }
+
+    [SetUp]
+    public void CreateTempFile()
+    {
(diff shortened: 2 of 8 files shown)


kll:
At the moment it looks like the travis-ci MacOS queue is having some [issues](https://www.traviscistatus.com/).

kll:
Yesterday when I submitted the PR the appveyor build passed but the travis-ci build failed due to their MacOS queue issues. Today I nudged the PR to cause a rebuild since travis-ci was back to normal but this time the inverse happened and appveyor failed while travis-ci passed. No code changed and looking at the log everything built and it shows 431 tests passed, 0 failed, 0 warnings, 0 inconclusive, 0 skipped. It just looks like NUnit3 crashed.

asbjornu:
Seems like the problem is this:

```
Error : C:\projects\gitversion\src\GitVersionTask.Tests\bin\Release\GitVersionTask.Tests.dll
Object reference not set to an instance of an object.
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at NUnit.Engine.Services.TestAgency.LaunchAgentProcess(TestPackage package)
   at NUnit.Engine.Services.TestAgency.CreateRemoteAgent(TestPackage package, Int32 waitTime)
   at NUnit.Engine.Runners.ProcessRunner.CreateAgentAndRunner()
   at NUnit.Engine.Runners.ProcessRunner.RunTests(ITestEventListener listener, TestFilter filter)
```

Perhaps upgrading NUnit can help?

kll:
I don't know, maybe it will, but why did it run and pass the day before? It also runs and passes fine when I run the build locally. I suspect something transient on the appveyor server and if it was just run again it would probably pass.

asbjornu:
@kll: Ok, I've restarted the build. Let's see how it goes.

PR #1163, opened 2017-02-23
GitVersionTask should use .git relative to project directory

### LINKED ISSUE #1162: GitVersionTask uses .git relative to SolutionDir
The targets and coding for GitVersionTask seem to use the .git repository relative to the SolutionDir instead of the project folder. This has started to cause versioning issues for projects in sub repositories. I am not sure when this change occurred but it used to not be this way. Ideally it should use the .git repository closest to the project being built and only fall back to the solution repository via option (i.e. when explicitly wanted.)

For now I worked around this by manually editing the targets file to use MSBuildProjectDirectory instead of SolutionDir.

mrswain:
This should fix the issue with GitVersionTask using the solution
directory .git folder instead of a project level .git folder, this is
more in line with how the command line works. It also adds property
support so the current use of the solution dir can be used or another
location can be specified. Also adds the ability to define a project
level or solution level import for global configuration.

Fixes #1162.

### DIFF
--- a/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
@@ -1,7 +1,14 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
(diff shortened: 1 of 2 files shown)


asbjornu:
@mrswain: Nice, thanks. Can you please associate your user with the e-mail address you committed 3355263 with? Just click the commit and then the question mark alongside your name and then jot in your e-mail address in your account.

mrswain:
Sorry about that, forgot that my local client has my actual work email address. Corrected.

PR #1172, opened 2017-02-26
Skip initial verification in case of dynamic repositories

GeertvanHorrik:
The initial verification breaks dynamic repositories because it requires a .git directory (but we don't have it yet). The change in this PR skips the initial verification in case of dynamic repositories.


GeertvanHorrik:
Merging this one immediately because it's required for dynamic repositories to run in addition to what was released in beta10.

PR #1177, opened 2017-02-28
Fix GitVersionTask on Linux

ermshiperete:
The additional import statement introduced in #1163 causes failure
on Linux when a project tries to use the GitVersionTask. It seems
that Mono can't cope with an empty `Project` attribute in the
`Import` element. Additionally there seems to be a bug at least in
Mono 3.x in that it requires the argument in `Exists` to be passed
in quotes.

This change fixes these two problems and allows to use the
GitVersionTask again on Linux.

I'd appreciate if a Beta-12 with this fix could be released soon.

### DIFF
--- a/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
@@ -9,6 +9,7 @@
     <GitVersionPath Condition="'$(GitVersionPath)' == ''">$(MSBuildProjectDirectory)</GitVersionPath>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And Exists('$(MSBuildProjectDirectory)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And '$(SolutionDir)' != '' And Exists('$(SolutionDir)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
+    <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == ''"></GitVersionCustomizeTargetFile>
     <IntermediateOutputPath Condition="$(IntermediateOutputPath) == '' Or $(IntermediateOutputPath) == '*Undefined*'">$(MSBuildProjectDirectory)\obj\$(Configuration)\</IntermediateOutputPath>
     <GitVersion_NoFetchEnabled Condition="$(GitVersion_NoFetchEnabled) == ''">false</GitVersion_NoFetchEnabled>
 
@@ -39,7 +40,7 @@
       TaskName="GitVersionTask.WriteVersionInfoToBuildLog"
       AssemblyFile="$(GitVersionTaskLibrary)GitVersionTask.dll" />
 
-  <Import Project="$(GitVersionCustomizeTargetFile)" Condition="'$(GitVersionCustomizeTargetFile)' != '' and Exists($(GitVersionCustomizeTargetFile))" />
+  <Import Project="$(GitVersionCustomizeTargetFile)" Condition="'$(GitVersionCustomizeTargetFile)' != '' and Exists('$(GitVersionCustomizeTargetFile)')" />
 
   <Target Name="WriteVersionInfoToBuildLog" BeforeTargets="CoreCompile;GetAssemblyVersion;GenerateNuspec" Condition="$(WriteVersionInfoToBuildLog) == 'true'">
     <WriteVersionInfoToBuildLog SolutionDirectory="$(GitVersionPath)" NoFetch="$(GitVersion_NoFetchEnabled)"/>
--- a/src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets
@@ -9,6 +9,7 @@
     <GitVersionPath Condition="'$(GitVersionPath)' == ''">$(MSBuildProjectDirectory)</GitVersionPath>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And Exists('$(MSBuildProjectDirectory)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And '$(SolutionDir)' != '' And Exists('$(SolutionDir)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
+    <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == ''"></GitVersionCustomizeTargetFile>
     <GitVersion_NoFetchEnabled Condition="$(GitVersion_NoFetchEnabled) == ''">false</GitVersion_NoFetchEnabled>
 
     <!-- Property that enables WriteVersionInfoToBuildLog -->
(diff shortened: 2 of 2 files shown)


src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
```diff
@@ -9,6 +9,7 @@
     <GitVersionPath Condition="'$(GitVersionPath)' == ''">$(MSBuildProjectDirectory)</GitVersionPath>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And Exists('$(MSBuildProjectDirectory)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And '$(SolutionDir)' != '' And Exists('$(SolutionDir)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
+    <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == ''">*Undefined*</GitVersionCustomizeTargetFile>
```
asbjornu: I've never seen `*Undefined*` used as a placeholder in MSBuild before. Won't an empty element initialise the property to an empty string?

src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
```diff
@@ -9,6 +9,7 @@
     <GitVersionPath Condition="'$(GitVersionPath)' == ''">$(MSBuildProjectDirectory)</GitVersionPath>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And Exists('$(MSBuildProjectDirectory)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And '$(SolutionDir)' != '' And Exists('$(SolutionDir)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
+    <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == ''">*Undefined*</GitVersionCustomizeTargetFile>
     <IntermediateOutputPath Condition="$(IntermediateOutputPath) == '' Or $(IntermediateOutputPath) == '*Undefined*'">$(MSBuildProjectDirectory)\obj\$(Configuration)\</IntermediateOutputPath>
```
asbjornu: I can see `*Undefined*` is used here too and I honestly haven't noticed it before. Seems like @SimonCropp added it in ec6e3697802d7dba297d90cb009220777d7685c7. Is this normal MSBuild nomenclature?

src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets
```diff
@@ -9,6 +9,7 @@
     <GitVersionPath Condition="'$(GitVersionPath)' == ''">$(MSBuildProjectDirectory)</GitVersionPath>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And Exists('$(MSBuildProjectDirectory)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
     <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == '' And '$(SolutionDir)' != '' And Exists('$(SolutionDir)\GitVersionTask.targets')">$(SolutionDir)\GitVersionTask.targets</GitVersionCustomizeTargetFile>
+    <GitVersionCustomizeTargetFile Condition="'$(GitVersionCustomizeTargetFile)' == ''">*Undefined*</GitVersionCustomizeTargetFile>
```
asbjornu: Same here.

src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
```diff
@@ -39,7 +40,7 @@
       TaskName="GitVersionTask.WriteVersionInfoToBuildLog"
       AssemblyFile="$(GitVersionTaskLibrary)GitVersionTask.dll" />
 
-  <Import Project="$(GitVersionCustomizeTargetFile)" Condition="'$(GitVersionCustomizeTargetFile)' != '' and Exists($(GitVersionCustomizeTargetFile))" />
+  <Import Project="$(GitVersionCustomizeTargetFile)" Condition="'$(GitVersionCustomizeTargetFile)' != '' and Exists('$(GitVersionCustomizeTargetFile)')" />
```
asbjornu: Don't you have to check for `*Undefined*` here as well?

src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets
```diff
@@ -37,7 +38,7 @@
       TaskName="GitVersionTask.WriteVersionInfoToBuildLog"
       AssemblyFile="$(GitVersionTaskLibrary)GitVersionTask.dll" />
 
-  <Import Project="$(GitVersionCustomizeTargetFile)" Condition="'$(GitVersionCustomizeTargetFile)' != '' and Exists($(GitVersionCustomizeTargetFile))" />
+  <Import Project="$(GitVersionCustomizeTargetFile)" Condition="'$(GitVersionCustomizeTargetFile)' != '' and Exists('$(GitVersionCustomizeTargetFile)')" />
```
asbjornu: Don't you have to check for `*Undefined*` here as well?

ermshiperete:
The travis build failures don't look like they are related to this change.

---

Review status: 0 of 2 files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/GitVersionTask/NugetAssets/build/GitVersionTask.targets, line 12 at r1](https://reviewable.io:443/reviews/gittools/gitversion/1177#-Ke4O5p8xonnm3dLRZF5-r1-12:-Ke9qMyL0F7yX2-nYNub:b-iumld5) ([raw file](https://github.com/gittools/gitversion/blob/a2ba8a8f86a267e2d3be0a49395fa1c1572bf936/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets#L12)):*
<details><summary><i>Previously, asbjornu (Asbjørn Ulsberg) wrote…</i></summary><blockquote>

I've never seen `*Undefined*` used as a placeholder in MSBuild before. Won't an empty element initialise the property to an empty string?
</blockquote></details>

You're right. I didn't think that it would work, but it does! Done.

---

*[src/GitVersionTask/NugetAssets/build/GitVersionTask.targets, line 13 at r1](https://reviewable.io:443/reviews/gittools/gitversion/1177#-Ke4O5p8xonnm3dLRZF5-r1-13:-Ke9qYQgpEO5FCiHATpl:b-ms6szs) ([raw file](https://github.com/gittools/gitversion/blob/a2ba8a8f86a267e2d3be0a49395fa1c1572bf936/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets#L13)):*
<details><summary><i>Previously, asbjornu (Asbjørn Ulsberg) wrote…</i></summary><blockquote>

I can see `*Undefined*` is used here too and I honestly haven't noticed it before. Seems like @SimonCropp added it in ec6e3697802d7dba297d90cb009220777d7685c7. Is this normal MSBuild nomenclature? 
</blockquote></details>

I've occasionally seen it in Conditions, but I don't know if it will actually ever have this value (unless explicitly set to it :smiley: )

---

*[src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets, line 12 at r1](https://reviewable.io:443/reviews/gittools/gitversion/1177#-Ke4O5p8xonnm3dLRZF6-r1-12:-Ke9qsFlxGzjJeK0xKa0:b-896fix) ([raw file](https://github.com/gittools/gitversion/blob/a2ba8a8f86a267e2d3be0a49395fa1c1572bf936/src/GitVersionTask/NugetAssets/buildMultiTargeting/GitVersionTask.targets#L12)):*
<details><summary><i>Previously, asbjornu (Asbjørn Ulsberg) wrote…</i></summary><blockquote>

Same here.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/gittools/gitversion/1177#-:-Ke9rcEYlpS2Qx54_2aP:bl4rmie)*

PR #1193, opened 2017-03-18
Better default support for NCrunch

onovotny:
This PR will disable writing to the build log, updating assembly info's and calling get version if running in NCrunch by default.


gep13:
@onovotny thank you for your contribution! 👍

PR #1194, opened 2017-03-21
ensure version is calc'd before restore

onovotny:
this is needed to ensure multi-project dependencies get packed correctly

### DIFF
--- a/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
@@ -66,7 +66,7 @@
     </ItemGroup>
   </Target>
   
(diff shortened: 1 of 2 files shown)


JakeGinnivan:
Thanks @onovotny, will get this released over the weekend

gep13:
>@JakeGinnivan said...
will get this released over the weekend

Would someone else be free (@pascalberger @asbjornu) to do a pairing session with Jake while he does this so that more people are familiar (and more importantly happy about doing) with the release process?  I would do it, but I am already max'ed out this weekend.

asbjornu:
@gep13: I'd love to, but my weekend is packed. I do feel, however, that if the process was described in detail in a Markdown file within this repository, I would be comfortable with **attempting** to follow it, if nothing else. Can't guarantee success, that's all. 😄

PR #1201, opened 2017-04-07
Jenkins Pipeline compatibility

erik-am:
GitVersion includes support for Jenkins as a build system, but this only supports traditional Jenkins jobs, not the new way of using [Jenkins Pipeline](https://jenkins.io/doc/book/pipeline/).

Currently, you can find all kinds of workarounds on the web to make GitVersion play nicely with Jenkins Pipeline-as-code. This pull request aims to fix it within GitVersion so that one doesn't need workarounds in his pipeline.

Current problems:
- GitVersion expects the branch name in the environment variable `GIT_BRANCH` while using Jenkins. This is true for the old Git plugin, but when using Jenkins pipeline, the branch name is stored in `BRANCH_NAME` instead.
A workaround that people use is to add `env.GIT_BRANCH = env.BRANCH_NAME` in the Jenkinsfile, but this pollutes the pipeline.
- Somehow Jenkins will add two remotes `origin` and `origin1` when executing the Pipeline.
[GitTools](https://github.com/GitTools/GitTools.Core/blob/master/src/GitTools.Core/GitTools.Core.Shared/Git/Helpers/GitRepositoryHelper.cs) will therefore fail with the error "2 remote(s) have been detected"
A workaround to remove the duplicate origin has been posted at http://stackoverflow.com/questions/39475135/gitversion-in-a-jenkins-multibranch-pipeline-job
- Within a pull request, both the pull request number and the number of commit should be taken into account, e.g., `1.0.4-PullRequest0005.3`. However, the actual result was `1.0.4-PullRequest.3`, missing the PR number and therefore leading to duplicates. This can be fixed by changing the default tag number pattern from `[/-](?<number>\d+)[/-]` to `[/-](?<number>\d+)`.

### DIFF
--- a/src/GitVersionCore.Tests/BuildServers/EnvironmentVariableJenkinsTests.cs
+++ b/src/GitVersionCore.Tests/BuildServers/EnvironmentVariableJenkinsTests.cs
@@ -9,6 +9,7 @@ public class EnvironmentVariableJenkinsTests
     string key = "JENKINS_URL";
     string branch = "GIT_BRANCH";
     string localBranch = "GIT_LOCAL_BRANCH";
+    string pipelineBranch = "BRANCH_NAME";
 
     private void SetEnvironmentVariableForDetection()
     {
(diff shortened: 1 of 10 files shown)


erik-am:
The build and tests pass in AppVeyor, but fail in Travis CI.
I see that's the case for other PR's as well, so I guess that's normal?

JakeGinnivan:
Nice work

JakeGinnivan:
And yeah, travis is stuffed...

PR #1204, opened 2017-04-19
Invalid release version when some branch created from it

czesiu:
I added test for problem with version calculating on release branch. When you create some branch from release branch then version calculating will take version from this other branch. I think that it is invalid result, but I don't know how to fix this.

Line 453:

    fixture.AssertFullSemver(config, "2.0.0-beta.10");

 throws exception:

    Shouldly.ShouldAssertException : variables.FullSemVer should be "2.0.0-beta.10" but was "2.0.0-beta.2"


JakeGinnivan:
This is actually a pretty hard problem to solve, but am having another crack at making it better. It may actually help GitVersion perform better.

Take this scenario: `master -> release/ -> feature`

If gitversion runs, it can't know if feature was branched from release or the other way around.. This is what is happening in your test, when you create the feature branch GitVersion thinks that release is now branched off the feature branch.

JakeGinnivan:
This issue should be fixed in #1225, I want to get feedback on the approach before I merge/release it though. If you can have a look at even the docs in that PR and check to see if the problem and the solution make sense to you that would be handy @czesiu

Issue #883, opened 2016-05-26
Consolidate AssemblyInfo update logic


asbjornu:
I added a few TODOs in cb4698d72b8d7aca6087769dfa402472abc3e9bc for consolidation of the AssemblyInfo update logic in `GitVersionTask` and `GitVersionExe` that should be tackled. There's quite a bit of duplicated functionality that leads to diverging support for different features and bugs in the task that might be well tested in the `.exe` and vice versa. Consolidating all of this into `GitVersionCore` with accompanying tests would be neat.

bvirkler:
Came here from #1319, bumping to "un-stale".

stunney:
@asbjornu what is left with this issue?

asbjornu:
@stunney, pretty much everything I wrote in the issue desription is still applicable.

arturcic:
@asbjornu I think we can close this one, as we moved the logic for AssemblyInfo update into GitVersion.Core, and both the `GitVersionTask` and `GitVersionExe` are using the same class

asbjornu:
Is 55d066462c21acf02fd6a1a15ff5707317850ce5 the commit that did the consolidation, @arturcic?

arturcic:
yes

asbjornu:
Ok. Closing as resolved in 55d066462c21acf02fd6a1a15ff5707317850ce5.

PR #1216, opened 2017-05-16
Match RegEx for AssemblyInfo attributes without quotations in version

### LINKED ISSUE #1212: Duplicate Assembly attribute error when using constants
Related to #871 and #760

Using a shared assembly info I am setting my version number in a constant string 
```
[assembly: AssemblyVersion(AssemblyInfo.Version)]
[assembly: AssemblyFileVersion(AssemblyInfo.FileVersion)]
[assembly: AssemblyInformationalVersion(AssemblyInfo.InformationalVersion)]
```
Looking at the RegEx to replace the attributes, this is not going to work as it requires quotation marks in the between the parentheses of the attribute.

This of course results in duplicate Assembly attributes, as it will try to append new attributes, assuming there are none.

I don't understand why the RegEx in [AssemblyInfoFileUpdate.cs](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionExe/AssemblyInfoFileUpdate.cs#L28) requires the quotation marks
RegEx:
`AssemblyVersion(Attribute)?\s*\(\s*"[^""]*"\s*\)`

Would it not be better to just replace any content (including whitespace) between the parentheses of the Attribute?
RegEx:
`AssemblyVersion(Attribute)?\s*\(\s*\S*\s*\)`

Of course, the string that's replacing the original should then have the quotation marks again. 
Probably this would require some refactoring on AssemblyInfoFileUpdate.cs and maybe some other places. I'm glad to make a PR, does anyone see any issues with this?

Freakazoid182:
Fixes #1212 

This allows for the replacing of AssemblyInfo attributes containing variables instead of literal strings
Regarding #883 this should not add any duplicate work. I think this would actually make it easier to also replace values in AssemblyInfo template strings.

### DIFF
--- a/src/GitVersionExe.Tests/AssemblyInfoFileUpdateTests.cs
+++ b/src/GitVersionExe.Tests/AssemblyInfoFileUpdateTests.cs
@@ -445,6 +445,61 @@ public class AssemblyInfoFileUpdateTests
         });
     }
 
+    [TestCase("cs", "[assembly: AssemblyVersion (AssemblyInfo.Version) ]\r\n[assembly: AssemblyInformationalVersion(AssemblyInfo.InformationalVersion)]\r\n[assembly: AssemblyFileVersion(AssemblyInfo.FileVersion)]")]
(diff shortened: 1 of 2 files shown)


stunney:
@asbjornu should this be closed?

asbjornu:
@stunney, the pull request you're commenting on was merged 3 years ago, so I don't quite understand your question.

Issue #1218, opened 2017-05-17
Docker improvements


JakeGinnivan:
I am pretty new with Docker, what we have works but I wonder if anyone can make what we have better.

We have a base image at https://github.com/GitTools/GitVersion/tree/master/src/DockerBase, the mono image doesn't run libgit2 because wheezy is super old and glib2.15 doesn't run on it without adding experimental repos and updating.

So I went with an Ubuntu base image and put mono on it, this seems pretty heavy but seems to work.

Then https://github.com/GitTools/GitVersion/blob/master/Dockerfile builds off the base. Then we publish through our deploy script at https://github.com/GitTools/GitVersion/blob/master/deploy.cake#L196

If you can improve it, send a pull request, it would be greatly appreciated.

mkoertgen:
Hi @JakeGinnivan. I did a quick evaluation on how to wrap GitVersion inside docker.
First i tried dotnetcore & mono-slim base images but finally arrived at adding `gitversion` (latest github release) and `git` to the latest mono image.

You can review my evaluation experiment here: https://github.com/mkoertgen/docker.gitversion/blob/master/Dockerfile

asbjornu:
I believe this can be closed due to the merging of #1215, #1269, #1344, #1422, #1432 and #1485 which all improve our Docker story. I've also created #1572 to discuss which further, concrete, improvements we can make to the current naming scheme.

PR #1219, opened 2017-05-17
Improvements to Dockerfile

BenHall:
Related to #1218

Main fixes:
 - Set a version for Ubuntu. Currently using :latest which will change in future
 - Move commands to a single instruction, this reduces the total image size as cache can be cleared

### DIFF
--- a/src/DockerBase/Dockerfile
+++ b/src/DockerBase/Dockerfile
@@ -5,9 +5,7 @@ MAINTAINER GitTools Maintainers <jake@ginnivan.net>
 # Wheezy doesn't support glibc 2.15 which libgit2sharp requires
 # So we are going to install mono on ubuntu instead
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-RUN apt-get update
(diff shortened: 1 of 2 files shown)


src/DockerBase/Dockerfile
```diff
@@ -1,13 +1,11 @@
...
 # Wheezy doesn't support glibc 2.15 which libgit2sharp requires
 # So we are going to install mono on ubuntu instead
 RUN apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF
-RUN apt-get update
-RUN apt-get install libcurl3 tzdata unzip curl git-all -y
-RUN echo "deb http://download.mono-project.com/repo/debian wheezy main" | tee /etc/apt/sources.list.d/mono-xamarin.list
-RUN apt-get update
-RUN apt-get install mono-complete -y
+RUN echo "deb http://download.mono-project.com/repo/debian wheezy main" | tee /etc/apt/sources.list
+.d/mono-xamarin.list
```
pascalberger: Should this be on a new line?

BenHall: Good spot. No, it shouldn't. I copied and pasted from my terminal window

PR #1220, opened 2017-05-17
Added 'EnsureWixToolsetInstalled' BeforeTarget to GetVersionTask.targets

forestb:
The [Wix toolset](http://wixtoolset.org/) does not trigger any of the following events/targets: `CoreCompile;GetAssemblyVersion;GenerateNuspec;_GenerateRestoreProjectSpec`, and as a result the `GetVersion` target never gets triggered, which means none of the environment variables are populated as a result of using MSBuild.

The `EnsureWixToolsetInstalled` is set as the default `InitialTargets="EnsureWixToolsetInstalled"` value for all Wix projects.

It would be hugely helpful to have the GitVersion variables seamlessly available as part of the build process for Wix projects!

RFR

### DIFF
--- a/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
@@ -67,7 +67,7 @@
     </ItemGroup>
   </Target>
   
-  <Target Name="GetVersion" BeforeTargets="CoreCompile;GetAssemblyVersion;GenerateNuspec;_GenerateRestoreProjectSpec" Condition="$(GetVersion) == 'true'">
+  <Target Name="GetVersion" BeforeTargets="CoreCompile;GetAssemblyVersion;GenerateNuspec;_GenerateRestoreProjectSpec;EnsureWixToolsetInstalled" Condition="$(GetVersion) == 'true'">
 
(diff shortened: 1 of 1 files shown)


asbjornu:
This won't fail in non-Wix projects that don't have `EnsureWixToolsetInstalled` defined?

forestb:
> This won't fail in non-Wix projects that don't have EnsureWixToolsetInstalled defined?

Correct.  

It used to be that GitVersion was more-incompatible with Wix until a fix was added in GitVersionTask v4.

[After adding a reference to a Wix project using GitVersionTask v3.6.5](https://github.com/forestb/GitVersionTesting/compare/master...adding-gvt-v3) v.s. 
[After adding a reference to a Wix project using GitVersionTask v4.Beta](https://github.com/forestb/GitVersionTesting/compare/master...adding-gvt-v4)

Now that a reference to the reference to `GitVersionTask.targets` is being added automatically, we just need to trigger the `GetVersion` target in order to populate the environment variables.

The addition I'm proposing only adds the condition, "Hey, I noticed the `EnsureWixToolsetInstalled` target is about to be executed; Let's execute `GetVersion` first."

It does not break anything; only subscribes to that additional event if it's found.

asbjornu:
Aha, I'm such an MSBuild noob that I misunderstood `BeforeTargets` to mean "run these targets before the current one", when it seems to mean exactly the opposite, which I understand won't break anything. Excellent! 👍

PR #1224, opened 2017-05-21
Feature/fix mainline feature branch bump

JakeGinnivan:
Fixes #1221 by using the test in #1223

@mabead

PR #1225, opened 2017-05-21
Pick better merge base

JakeGinnivan:
Builds on #1224, includes the test from #1204 and gets it passing.

This change allows GitVersion to better decide which branch the current branch was likely branched from. Have added some new configuration options and updated the related docs.

Would love any other ideas on how to solve this problem better.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -249,15 +249,15 @@ Because git is a directed graph GitVersion sometimes cannot tell which branch th
 Take this graph
 
 ```
-  *   feature/foo
-* |   release/1.0.0
+* release/1.0.0   * feature/foo
+| ________________/
 |/
 *
 *
 * (master)
 ```
 
-By looking at this graph, you cannot tell which of these scenarios happened
+By looking at this graph, you cannot tell which of these scenarios happened:
 
 1. feature/foo branches off release/1.0.0
    - Branch release/1.0.0 from master
@@ -270,6 +270,8 @@ By looking at this graph, you cannot tell which of these scenarios happened
    - Branch release/1.0.0 from feature/foo 
    - Add a commit to both release/1.0.0 and feature/foo
    - feature/foo is the base for release/1.0.0
+
+Or put more simply, what branch was created first, `release/1.0.0` or `feature/foo`.
 
 To resolve this issue we give GitVersion a hint to how we normally do our branching, feature branches have a value of:
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -273,11 +273,11 @@ By looking at this graph, you cannot tell which of these scenarios happened:
 
 Or put more simply, what branch was created first, `release/1.0.0` or `feature/foo`.
 
-To resolve this issue we give GitVersion a hint to how we normally do our branching, feature branches have a value of:
+To resolve this issue, we give GitVersion a hint to our branching workflows by telling it what types of branches it can be created from. For example feature branches are by default configured to have the following source branches:
 
 `sourceBranches: ['master', 'develop', 'feature', 'hotfix', 'support']`
 
-This means that we will never bother to evaluate pull requests as merge base options, being explicit like this helps GitVersion be much faster too.
+This means that we will never bother to evaluate pull request branches as merge base options, being explicit like this helps GitVersion be much faster too.
 
 ### is-source-branch-for
 The reverse of the above setting. This property was introduced to keep it easy to extend GitVersion's config.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -243,6 +243,41 @@ values, but here they are if you need to:
 ### regex
 This is the regex which is used to match the current branch to the correct branch configuration.
 
+### source-branches
+Because git is a directed graph GitVersion sometimes cannot tell which branch the current branch was branched from.
+
+Take this graph
+
+```
+  *   feature/foo
+* |   release/1.0.0
+|/
+*
+*
+* (master)
+```
+
+By looking at this graph, you cannot tell which of these scenarios happened
+
+1. feature/foo branches off release/1.0.0
+   - Branch release/1.0.0 from master
+   - Branch feature/foo from release/1.0.0
+   - Add a commit to both release/1.0.0 and feature/foo
+   - release/1.0.0 is the base for feature/foo
+
+2. release/1.0.0 branches off feature/foo
+   - Branch feature/foo from master
+   - Branch release/1.0.0 from feature/foo 
+   - Add a commit to both release/1.0.0 and feature/foo
+   - feature/foo is the base for release/1.0.0
+
+To resolve this issue we give GitVersion a hint to how we normally do our branching, feature branches have a value of:
+
+`sourceBranches: ['master', 'develop', 'feature', 'hotfix', 'support']`
+
+This means that we will never bother to evaluate pull requests as merge base options, being explicit like this helps GitVersion be much faster too.
+
+
 ### branches
 The header for all the individual branch configuration.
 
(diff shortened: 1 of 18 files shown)


docs/configuration.md
```diff
@@ -243,6 +243,45 @@ values, but here they are if you need to:
...
+
+Take this graph
+
+```
+  *   feature/foo
+* |   release/1.0.0
+|/
+*
+*
+* (master)
```
adamralph: It's not obvious to me what each branch label refers to in this diagram. I think the branch labels should really be shown next to the head commit of the branch.

Specifically, I'm not sure if you mean this:

```
             *   feature/foo
* master     |   release/1.0.0      // head of release branch is behind head of feature branch
| ___________/
|/
|
|
```

or this:

```
             *   feature/foo, release/1.0.0      // head of both branches is the same
* master     |
| ___________/
|/
|
|
|
```

JakeGinnivan: Nice idea, have fixed

JakeGinnivan: It was

```
* release/1.0.0   * feature/foo
| ________________/
|/
*
*
* (master)
```

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
...
+
+2. release/1.0.0 branches off feature/foo
+   - Branch feature/foo from master
+   - Branch release/1.0.0 from feature/foo 
+   - Add a commit to both release/1.0.0 and feature/foo
+   - feature/foo is the base for release/1.0.0
+
+Or put more simply, what branch was created first, `release/1.0.0` or `feature/foo`.
+
+To resolve this issue we give GitVersion a hint to how we normally do our branching, feature branches have a value of:
```
adamralph: Perhaps the fault is on me, but this sentence, and the following list, do not make any sense to me. I've honestly no idea what you mean by  "feature branches have a value of", and what information the list is designed to convey.

asbjornu: I think what's meant is what the default configuration for feature branches is. So doing `GitVersion init` should give you a `GitVersion.yml` containing the list below. Perhaps a better description would be:

> To resolve this issue, we give GitVersion a hint to how we normally do our branching. Feature branches are by default configured to have the following source branches:

JakeGinnivan: Rewrote it, hopefully makes it a bit clearer

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
 ### regex
 This is the regex which is used to match the current branch to the correct branch configuration.
 
+### source-branches
+Because git is a directed graph GitVersion sometimes cannot tell which branch the current branch was branched from.
```
adamralph: > Because git is a directed graph...

I'm guessing that a significant number of people won't know what this means, and it seems redundant. Perhaps remove it?

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
...
+   - Add a commit to both release/1.0.0 and feature/foo
+   - release/1.0.0 is the base for feature/foo
+
+2. release/1.0.0 branches off feature/foo
+   - Branch feature/foo from master
+   - Branch release/1.0.0 from feature/foo 
+   - Add a commit to both release/1.0.0 and feature/foo
+   - feature/foo is the base for release/1.0.0
+
+Or put more simply, what branch was created first, `release/1.0.0` or `feature/foo`.
```
adamralph: Suggest:

> Or put more simply, ~what~ *you cannot tell which* branch was created first, `release/1.0.0` or `feature/foo`.

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
...
+
+To resolve this issue, we give GitVersion a hint to our branching workflows by telling it what types of branches it can be created from. For example feature branches are by default configured to have the following source branches:
```
adamralph: Suggest:

> To resolve this issue, we give GitVersion a hint ~to~ *about* our branching workflows by telling it what types of branches ~it~ *a branch* can be created from. For example*,* feature branches are*,* by default*,* configured to have the following source branches:

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
...
+
+`sourceBranches: ['master', 'develop', 'feature', 'hotfix', 'support']`
```
adamralph: Is this list correct? Under GitFlow I was under the impression that feature branches should only be created from, and merged to, `develop`. And for GitHubFlow, `master`.

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
...
+
+This means that we will never bother to evaluate pull request branches as merge base options, being explicit like this helps GitVersion be much faster too.
```
adamralph: Suggest:

> This means that we will never bother to evaluate pull request branches as merge base options~,~ *and* being explicit ~like this helps GitVersion be much faster too~ *in this way also improves the performance of GitVersion*.

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
...
+
+### is-source-branch-for
```
adamralph: Why is this kebab-cased, whereas sourceBranches is camelCased above (but kebab-cased below)?

docs/configuration.md
```diff
@@ -243,6 +243,47 @@ values, but here they are if you need to:
...
+The reverse of the above setting. This property was introduced to keep it easy to extend GitVersion's config.
+
+When you add a new branch type this allows you to specify both source-branches for the new branch, and also add the new branch to existing branch configurations. For example if you create a new branch called `unstable` you could set a value of `['release', 'master', 'feature']` etc.
```
adamralph: Again, it might be me, but I don't get this at all.

docs/configuration.md
asbjornu: I think it's valuable to point out that sometimes GitVersion can't figure out the origin of the current branch. Whether that's because it's a DAG or not is perhaps not that important.

docs/configuration.md
asbjornu: Kebab case would be more consistent, so I believe the camel casing is a typo?

docs/configuration.md
asbjornu: @adamralph: It's not just you. I don't get how the branch `unstable` can be the source branch for `release`, `master` or `feature`. 😄

zvirja: One more point is to not use position depending things like "above setting". It's always better to refer to the setting by name, so that later description will not be broken due to changed sorting.

zvirja:
@JakeGinnivan Do you currently have any time to follow up? :flushed: I have yet another thing that I need improve in this direction, so I would like to base my PR on your one. Given that the documentation stuff is under the discussion only, it looks not too much left 😉

JakeGinnivan:
@zvirja sorry, will get the feedback addressed and just merge

docs/configuration.md
JakeGinnivan: This configuration param is for all flows, GitVersion actually doesn't know about the different workflows, it just has branch configuration which works for both

JakeGinnivan:
Have addressed all the feedback, thanks all. Going to merge once builds are green.

adamralph:
@JakeGinnivan the final revision LGTM.

The only thing I still don't understand is how `unstable` can ever be the source branch for branches such as `master`, `develop`, `hotfix`, `release` or `support` as demonstrated in the `is-source-branch-for` example. I realise it's only an example, but is it demonstrating something that almost no-one would ever want to do? Or am I misunderstanding something?

PR #1243, opened 2017-06-24
Feature/introduce in memory git metadata

JakeGinnivan:
My goal here is to move GitVersion into more distinct stages.

1. Load all the information it needs from git, this would be branches, tags, merge bases, parent branches etc.
   - Ideally this would source from both local and remote branches so we don't need to normalise
   - Afaik we introduced normalisation in to simplify the codebase and remove the concept of remote and local branches, it just makes everything local.
2. Calculate version, this would now be entirely in memory and not touch libgit2
3. We could then switch out the provider which actually loads the information (for instance, libgit2 or github api or git command line)

After this happens the slow part should be step 1, which should be way easier to speed up because all git operations will be in one place and we can make better decisions how to load it all.

Keen for feedback @gep13 @asbjornu


JakeGinnivan:
In terms of performance, this should have reduced the number of times the current branch commits are scanned by at least one pass.

src/GitVersionCore/GitVersionContext.cs
```diff
@@ -80,6 +82,13 @@ public GitVersionContext(IRepository repository, Branch currentBranch, Config co
         public Commit CurrentCommit { get; private set; }
         public bool IsCurrentCommitTagged { get; private set; }
         public GitRepoMetadataProvider RepositoryMetadataProvider { get; private set; }
+        /// <summary>
+        /// This is a new concept which is static metadata
+        /// It has no runtime dependencies on libgit2 and all information is pre-calculated.
+        /// We should move to use this, that way libgit2 is used when bootstrapping
+        /// this will be far easier to optimise then all the logic for gitversion will be entirely in memory
+        /// </summary>
+        public GitRepoMetadata RepositoryMetadata { get; private set; }
```
asbjornu: If we agree that `GitRepoMetadata` should be called just `Repository`, the same goes for this property.

src/GitVersionCore/GitRepoInformation/GitRepoMetadataProvider.cs
```diff
@@ -110,22 +110,36 @@ public IEnumerable<Branch> GetBranchesContainingCommit(Commit commit, IList<Bran
...
-            var key = Tuple.Create(branch, otherBranch);
+            return FindMergeBase(branch.Tip.Sha, otherBranch.Tip.Sha, branch.FriendlyName, otherBranch.FriendlyName);
+        }
+
+        /// <summary>
+        /// Find the merge base of the two branches, i.e. the best common ancestor of the two branches' tips.
+        /// </summary>
+        public Commit FindMergeBase(string sha1, string sha2, string sha1Desc = null, string sha2Desc = null)
+        {
+            var key = Tuple.Create(sha1, sha2);
```
asbjornu: A `MergeKey` class could also fix symmetry problems as mentioned in #1112.

src/GitVersionCore/GitRepoInformation/GitRepoMetadata.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System.Collections.Generic;
+
+namespace GitVersion.GitRepoInformation
+{
+    public class GitRepoMetadata
```
JakeGinnivan: I am unsure at the moment to be honest. I don't want to re-implement git data structures, this really is information about the commit graph with a bunch of pre-calculated stuff. Lets see where it goes then fix naming next

src/GitVersionCore/GitRepoInformation/GitRepoMetadata.cs
```diff
@@ -0,0 +1,76 @@
+﻿using System.Collections.Generic;
+
+namespace GitVersion.GitRepoInformation
```
JakeGinnivan: I like GitVersion.Graph! Will keep that in mind as a progress

src/GitVersionCore/GitRepoInformation/GitRepoMetadata.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        // Wonder if this can be 'mainline'
+        public MBranch MasterBranch { get; }
+        public List<MBranch> ReleaseBranches { get; }
+        public MBranch CurrentBranch { get; }
+        public List<MTag> Tags { get; private set; }
+    }
+
+    public class MTag
```
JakeGinnivan: Rather than aliasing in a bunch of files which reference both this namespace and libgit2 the metadata related ones are prefixed with M. I expect this to go at some point

src/GitVersionCore/GitRepoInformation/GitRepoMetadata.cs
```diff
@@ -0,0 +1,76 @@
...
+            TipSha = tipSha;
+            Parent = parent;
+            MergeMessages = mergeMessages;
+            Tags = tags;
+        }
+
+        public string Name { get; }
+        public string TipSha { get; }
+        public MParent Parent { get; }
+        public List<MergeMessage> MergeMessages { get; }
```
JakeGinnivan: 👍

src/GitVersionCore/GitRepoInformation/GitRepoMetadataProvider.cs
```diff
@@ -9,15 +9,15 @@ namespace GitVersion
...
+        private Dictionary<Tuple<string, string>, MergeBaseData> mergeBaseCache;
         private Dictionary<Branch, List<SemanticVersion>> semanticVersionTagsOnBranchCache;
         private IRepository Repository { get; set; }
         const string missingTipFormat = "{0} has no tip. Please see http://example.com/docs for information on how to fix this.";
         private Config configuration;
 
         public GitRepoMetadataProvider(IRepository repository, Config configuration)
         {
-            mergeBaseCache = new Dictionary<Tuple<Branch, Branch>, MergeBaseData>();
+            mergeBaseCache = new Dictionary<Tuple<string, string>, MergeBaseData>();
```
JakeGinnivan: This cache will disappear soon, it won't be needed once everything is querying the in memory representations :)

src/GitVersionCore/GitRepoInformation/GitRepoMetadataProvider.cs
JakeGinnivan: Ditto here

src/GitVersionCore/GitRepoInformation/GitRepoMetadataProvider.cs
```diff
@@ -259,17 +273,15 @@ List<BranchCommit> GetMergeCommitsForBranch(Branch branch)
...
         private class MergeBaseData
         {
-            public Branch Branch { get; private set; }
-            public Branch OtherBranch { get; private set; }
-            public IRepository Repository { get; private set; }
+            public string Sha1 { get; private set; }
+            public string Sha2 { get; private set; }
 
-            public Commit MergeBase { get; private set; }
+            public string MergeBase { get; private set; }
```
JakeGinnivan: Just trying to reduce the number of libgit2 references

src/GitVersionCore/GitRepoInformation/Libgit2RepoMetadataProvider.cs
```diff
@@ -0,0 +1,87 @@
+﻿using LibGit2Sharp;
+using System.Collections.Generic;
+using System.Linq;
+using System;
+using System.Text.RegularExpressions;
+
+namespace GitVersion.GitRepoInformation
+{
+    public class Libgit2RepoMetadataProvider
```
JakeGinnivan: Good idea, will do once I get the structure correct

src/GitVersionCore/GitRepoInformation/Libgit2RepoMetadataProvider.cs
```diff
@@ -0,0 +1,87 @@
...
+    {
+        public static GitRepoMetadata ReadMetadata(GitVersionContext context)
```
JakeGinnivan: Yeah I think so, just wasn't sure the shape at the time, this allowed me to just hook it in.

I am trying to make seams at the moment which will simplify code elsewhere, then can put the right abstraction in place

src/GitVersionCore/GitRepoInformation/GitRepoMetadataProvider.cs
asbjornu: Ah, of course! 👍

src/GitVersionCore/GitRepoInformation/GitRepoMetadataProvider.cs
asbjornu: Right. Good idea.

erikbra:
@JakeGinnivan sorry to post on a closed PR, but I'm really interested in speeding up this stuff as well. Would you care to share some thoughts on why this PR was closed? And similar with #1244, Was it due to lack of time, or do you think they are a step in the wrong direction? I've only just started looking at GitVersion, so I haven't looked at this in detail. I just wanted to know if this was a good place to start if wanting to work on the perf wrt the integration with LibGit2.

JakeGinnivan:
Hey, all good.

Was a few things, I kinda lost context around where I was and how to finish it off mainly though from memory.

Generally I think the approach should be:

1) Create a single interface which all interactions with libgit2 is put behind. The methods on this should be pretty specific like `getPossibleParentBranches()` or something (and could be memoized like you suggested in the other thread).
2) Once this has been refactored, try to implement a in memory version of this interface which the in memory structure is populated in the most efficient way possible ahead of time.

Doing this would also allow you to implement say a `github api` implementation, then gitversion as a service could be created :P

JakeGinnivan:
It's been a while, but if there are other questions I'll try to answer best I can @erikbra

erikbra:
Thanks a lot for your replies, @JakeGinnivan. I'm only just starting to look at this, so I don't have a good understanding of the structures required yet. I'll read more of the original source code, and fork off this PR to get an understanding of the thoughts behind it. 

I really appreciate if you would be able to answer some questions when the time comes. I'll get back to you on that one.

I might be jumping ahead here, but C# has gotten a lot of perf-focused new features lately, so we might be able to do some of the reading in pure C# with Memory spans, etc. in the future. But let's work on the separation first :)

Really like the idea of GitVersion-as-a-service ;)

PR #1247, opened 2017-06-30
Sync FileVersion

### LINKED ISSUE #1241: FileVersion inconsistency between UpdateAssemblyInfo and UpdateVersionProperties
The new `UpdateVersionProperties` section in the 4.0 `GitVersionTask.targets` file is using a different value for  `FileVersion` than the `UpdateAssemblyInfo` task.

`UpdateAssemblyInfo` is using [AssemblySemFileVer](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionTask/AssemblyInfoBuilder/CSharpAssemblyInfoBuilder.cs#L45).

`UpdateVersionProperties` is using [$(GitVersion_MajorMinorPatch).$(GitVersion_CommitsSinceVersionSource)](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets#L108) in both versions of the targets file.

To maintain consistency with previous behavior, I think `UpdateVersionProperties` should be changed to match `UpdateAssemblyInfo`.

I'd be happy to submit a PR if you agree that this should be changed.

bording:
This ensures that `AssemblySemFileVer` is used consistently for the `FileVersion` attribute throughout GitVersionTask.

Fixes #1241

### DIFF
--- a/src/GitVersionTask/AssemblyInfoBuilder/FSharpAssemblyInfoBuilder.cs
+++ b/src/GitVersionTask/AssemblyInfoBuilder/FSharpAssemblyInfoBuilder.cs
@@ -41,7 +41,7 @@ module GitVersionInformation =
 {3}
 ",
         vars.AssemblySemVer,
-        vars.MajorMinorPatch + ".0",
+        vars.AssemblySemFileVer,
         vars.InformationalVersion,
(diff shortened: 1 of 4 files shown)


asbjornu:
LGTM!

PR #1249, opened 2017-07-01
Clean up AssemblyInfo file updating

### LINKED ISSUE #1242: GitVersionInformation in SDK-style projects
In 4.0, `UpdateAssemblyInfo` is false for the new SDK-style projects, which makes sense since the appropriate values can be injected via `UpdateVersionProperties`.

However, this means that the static `GitVersionInformation` class is also no longer generated by default, which means SDK-style projects lose the ability to inspect version properties at run-time.

Setting `UpdateAssemblyInfo` to true works, but it's a pain because you then end up with assembly attribute conflicts, which means you have to add something like this to your project file also:
```
<GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
<GenerateAssemblyInformationalVersionAttribute>false</GenerateAssemblyInformationalVersionAttribute>
<GenerateAssemblyVersionAttribute>false</GenerateAssemblyVersionAttribute>
```

I'd like to see the generation of `GitVersionInformation` be controlled separately and be enabled by default even for SDK-style projects.

Is this something that sounds reasonable? I'd be willing to work on this if it sounds like something you'd accept.

bording:
This is still WIP, but here's what I have so far:

- `AssemblyInfoFileUpdate` has been moved into GitVersionCore and renamed `AssemblyInfoFileUpdater`
-  The `UpdateAssemblyInfo` task has been rewritten to use `AssemblyInfoFileUpdater`

Things to do still:
- [x] Add a new way to generate the `GitVersionInformation` class
- [x] Remove the `AssemblyInfoBuilder` stuff from GitVersionTask
- [x] Look at the existing templates and see what changes might be needed
- [ ] Figure out which of the other issues linked [here](https://github.com/GitTools/GitVersion/issues/1242#issuecomment-311648842) should be fixed as part of this PR and which should be handled in subsequent PRs

Fixes #1242


bording:
I see there there are some test failures, but none of them look related to my changes...

bording:
Made a bit more progress. I've turned `AssemblyVersionInfoTemplates` into a more general purpose `TemplateManager` class, and I'm getting the add formats from embedded resources as well.

I've also started the `GitVersionInformationGenerator` implementation, which uses the `TemplateManager` to load its templates from embedded resources as well.

src/GitVersionCore.Tests/GitVersionCore.Tests.csproj
```diff
@@ -62,9 +62,8 @@
       <HintPath>..\packages\NUnit.3.6.0\lib\net45\nunit.framework.dll</HintPath>
       <Private>True</Private>
     </Reference>
-    <Reference Include="Shouldly, Version=2.7.0.0, Culture=neutral, PublicKeyToken=6042cbcb05cbc941, processorArchitecture=MSIL">
-      <HintPath>..\packages\Shouldly.2.7.0\lib\net40\Shouldly.dll</HintPath>
-      <Private>True</Private>
+    <Reference Include="Shouldly, Version=2.8.3.0, Culture=neutral, PublicKeyToken=6042cbcb05cbc941, processorArchitecture=MSIL">
```
asbjornu: Since some of the build failures seems related to the upgrade of Shoudly, can you perhaps avoid that upgrade in this PR? On one of the Travis builds, this is the output:

>  The type initializer for `Shouldly.ShouldlyConfiguration` threw an exception.

bording: I can back this out, but the reason I upgraded it was that the older version was throwing an error about it not having a diff tool, so I was blocked. Upgrading it allowed it to use the VS diff tool.

asbjornu: I would appreciate if you could install an external diff tool so we could have this PR free of any package upgrades. I hope that should take care of the build failures.

bording: @asbjornu I do have an external diff tool, SourceGear DiffMerge, but Shouldly doesn't appear to know about it. Are there some docs I can follow to temporarily wire that up so I can actually run the tests but not need to bump the version?

asbjornu: Seems like [this might be a good place to start](http://shouldly.readthedocs.io/en/latest/assertions/shouldMatchApproved.html#configuration).

bording: Thanks, I'll see what I can do to get it working and I'll remove the package upgrade.

bording:
@asbjornu I have cleaned up the project files and removed the Shouldly update, but it looks like AppVeyor is still failing. I've run all of those failing tests locally, and they seem to work fine.

Regarding the rest of the changes to finish this PR up, I hope to have it wrapped up this week.

asbjornu:
@bording: Those build failures are indeed very strange. If someone else from @GitTools/developers should come by, here's the failure:

```c#
Error : ConfigProviderTests.CanWriteOutEffectiveConfiguration
System.Exception : Cannot find method in call stack with attribute NUnit.Framework.TestCaseAttribute
   at Shouldly.Configuration.FindMethodUsingAttribute`1.GetTestMethodInfo(StackTrace stackTrace, Int32 startAt) in C:\projects\shouldly\src\Shouldly.Shared\Configuration\FindMethodUsingAttribute.cs:line 20
   at Shouldly.ShouldMatchApprovedTestExtensions.ShouldMatchApproved(String actual, Func`1 customMessage, Action`1 configureOptions) in C:\projects\shouldly\src\Shouldly.Shared\ShouldlyExtensionMethods\ShouldMatchApprovedTestExtensions.cs:line 49
   at Shouldly.ShouldMatchApprovedTestExtensions.ShouldMatchApproved(String actual) in C:\projects\shouldly\src\Shouldly.Shared\ShouldlyExtensionMethods\ShouldMatchApprovedTestExtensions.cs:line 17
   at ConfigProviderTests.CanWriteOutEffectiveConfiguration() in C:\projects\gitversion\src\GitVersionCore.Tests\ConfigProviderTests.cs:line 177
```

bording:
Pushed up a few more changes, but didn't get as much done as I'd hoped I would. 😞 

@asbjornu I started looking into turning the existing `AssemblyInfoBuilder` tests in GitVersionTask.Tests  tests into new `GitVersionInformationGenerator` tests, but they're a lot more involved than I was expecting. In addition to using Shouldly to write the file output to disk, it's also compiling the code with Rosyln and loading the assembly into memory to test it.

Is it really important to keep this aspect in the new tests? It seems a bit excessive to me. Also, it only tests the C# and VB output, and not F#. I'm thinking that having the Shouldly output to verify the output looks correct would be enough here, especially considering that's all the `AssemblyInfoFileUpdater` tests do.

bording:
Hmm, still seeing those weird test failures. I see that other PRs have been merged recently, and the failures didn't happen on those branches, but I'm not sure what I've done here that would cause them. I guess I'll need to spend some time figuring that out since it doesn't look like it's going away on it's own...

asbjornu:
Compiling the generated files with Roslyn is a bit excessive, I agree. It's good to know that the generated content actually compiles (especially since it's just basic and unsafe string replacements), but we can probably move that to the build pipeline instead. @gep13, do you have Ideas on how we can have Cake do the instrumentation there instead of doing it within a test?

It's unfortunate that the tests are failing and that I don't have time assisting you. I'm back from holiday in a couple of weeks and might be able to take a closer look then. The best of luck meanwhile, though and thanks for this work, @bording! It's highly appreciated!

bording:
Just wanted to chime in and say that I'm still planning on getting back to finishing this. Lack of spare time and then vacation has conspired to prevent me from working on it recently.

Hope to get back to it soon!

bording:
@asbjornu With the changes I've just pushed up, I'm removing the WIP and think this is ready for a review.

I haven't added a way to create the `GitVersionInformation` file from GitVersionExe yet, but that wasn't something it could do before anyway, so it could be added as a followup PR since this one is large enough as-is.

asbjornu:
@bording: Thanks for your work on this! ❤️  Those failing tests are bizarre. Especially because they're only failing on AppVeyor. Perhaps an upgrade of Shouldly can help?

bording:
> @bording: Thanks for your work on this! ❤️ Those failing tests are bizarre. Especially because they're only failing on AppVeyor. Perhaps an upgrade of Shouldly can help?

I can try it and see if it makes a difference.

dazinator:
Quick look at the stack trace a few comments above, nunit attribute is mentioned - so could also check if updating nunit packages helps at all.

bording:
Dug into the Shouldly code a bit and I realized what was going on! The tests I added were calling `ShouldlyConfiguration.ShouldMatchApprovedDefaults.LocateTestMethodUsingAttribute<TestCaseAttribute>();`, but the failing tests were all using `Test`, not `TestCase`.

The configuration appears to be global, so I added `ShouldlyConfiguration.ShouldMatchApprovedDefaults.LocateTestMethodUsingAttribute<TestAttribute>();` to the tests that were failing, and now everything is passing!

And, since those tests were all marked to not run on mono, that's why Travis wasn't failing.

@asbjornu Should I revert the Shouldly updates? Maybe go back to 2.8.3 instead of the beta?

asbjornu:
Excellent! 2.8.3 sounds fine.

bording:
> Excellent! 2.8.3 sounds fine.

Done!

bording:
Sigh, now it looks like 2.8.3 is causing Travis to fail because it added a Visual Studio difftool option, and looking for that is throwing.

I guess back to 2.7.0 it is!

asbjornu:
@bording: Fantastic work, thank you so much for this! ❤️

PR #1264, opened 2017-07-25
Nitpicky typo correction

cbuice:


gep13:
@cbuice thank you for taking the time to make this change, every bit helps 😄

Issue #1175, opened 2017-02-28
Building using dotnet build fails for .NET Standard library project


jpierson:
I'm using the latest pre-release of GitVersionTask (4.0.0-beta0011) on a .NET standard library project and it seems to work from Visual Studio 2017 but not from the command line using `dotnet build`. I'm targeting netstandard1.2.

I wasn't sure whether this would work at all. I started off looking around for an implementation that installs a GitVersion package as a `DotnetCliToolReference` in an SDK csproj file. With nothing conclusive from that or from other .NET Core related issues (ex. https://github.com/GitTools/GitVersion/issues/647) I figured I should just try adding the package reference and see what it does. I was happy at first to see it was working when building the solution from Visual Studio but building using `dotnet build` gives me the following error.

> Microsoft (R) Build Engine version 15.1.545.13942
> Copyright (C) Microsoft Corporation. All rights reserved.
> 
> C:\Users\me\.nuget\packages\gitversiontask\4.0.0-beta0011\build\GitVersionTask.targets(45,5): error MSB4062: The "GitVersionTask.WriteVersionInfoToBuildLog" task could not be loaded from the assembly C:\Users\me\.nuget\packages\gitversiontask\4.0.0-beta0011\build\GitVersionTask.dll. Could not load file or assembly 'Microsoft.Build.Utilities.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL'. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask. [C:\Users\me\source\myproj\src\MyProj\MyProj.csproj]
> C:\Program Files\dotnet\sdk\1.0.0-rc4-004771\Sdks\Microsoft.NET.Sdk\build\Microsoft.NET.Sdk.targets(92,5): error : Cannot find project info for 'C:\Users\me\source\myproj\src\MyProj\MyProj.csproj'. This can indicate a missing project reference. [C:\Users\me\source\myproj\src\TestCore\TestCore.csproj]
> 
> Build FAILED.
> 
[...]

Issue #1178, opened 2017-02-28
Question: GitVersion Configuration


Tankatronic:
I have spent a lot of time trying to wrap my head around GitVersion and what it is doing. Just when I think I understand the workflow, I realize I don't! I wanted to give an example workflow with how we would like GitVersion to work, and if it is possible to configure it to behave this way or not.

Our master branch is our release branch. Development happens on develop, where we branch off for features/bugs. We really like the idea of using the commit bump messages to control when and what versions increment. I have not been able to get them to work, at least in the way that I would expect. 

Say develop is currently version 2.0.0. I branch off featureA, and finish it and it's ready to merge back into develop. Another team member reviews featureA, determines that it is just a minor version difference, and merges it back in with +semver: minor in the merge message. bugA was also branched off and when it is merged back into develop a +semver: fix is put into the merge message. 

Now what I would _expect_ to happen would be that when develop merges in featureA, it is incremented to version 2.1.0. When it merges in bugA it is incremented to version 2.1.1. I understand that this isn't really what is going to happen because it has not been released yet. develop stays at version 2.0.0 through this entire process. Please correct me if I am wrong here.

After doing some research and going through issues, what I think should happen is that once development is done on develop and we merge it into master, that is when the version will be incremented. Since there was a +semver: minor and a +semver: fix, only the minor version will change and the patch version will remain at zero with a resulting version of 2.1.0 on master. When development resumes on develop, it should see that master has had the version incremented and then increment its own version to 2.1.0.

The problem is, I have not been able to get master to increment its version. It remains at 2.0.0. What mode should I be using to get this to function?

asbjornu:
Does @JakeGinnivan's answer https://github.com/GitTools/GitVersion/issues/1137#issuecomment-282482205 help in any way?

Tankatronic:
Sorry for the delayed response. This question can be closed. The comment did help slightly but I think we figured out a configuration that worked for us.

PR #1269, opened 2017-07-29
#1175 - NetStandard

dazinator:
## Current Status

`GitVersionTask` is working on `dotnet core`.
Also there is a new nuget package called: `GitVersion.CommandLine.DotNetCore` which contains the GitVersion commandline, except built for `dotnet core`, so you can run `GitVersion` via `dotnet GitVersion.dll`.
And there is a new dotnetcore based docker image.

## To try out the build task:

1. Install the `dotnet core sdk`. (Need a minimum of version `1.1` installed).
2. Add the appveyor nuget feed: https://ci.appveyor.com/nuget/gitversion-8nigugxjftrw
3. Add the nuget package:

```xml
 <PackageReference Include="GitVersionTask" Version="4.0.0-pullrequest1269-1535">
      <PrivateAssets>all</PrivateAssets>
    </PackageReference>
```

[Here is a sample consumer project building on appveyor with dotnet sdk](https://ci.appveyor.com/project/dazinator/gitversion-pr1269/history)

Be aware when building or restoring using `dotnet`  commands you need to use the --disable-parallel flag (https://github.com/GitTools/GitVersion/issues/1381)

## To try GitVersion DotNetCore CLI

1. Install the `dotnet core runtime`. (Need a minimum of version `2.0` installed).
2. Add the appveyor nuget feed: https://ci.appveyor.com/nuget/gitversion-8nigugxjftrw
3. Install the `GitVersion.CommandLine.DotNetCore` nuget package.
4. CD into the `/tools` folder of the nuget package and run: `dotnet gitversion.dll`

Also there are a couple of artifacts in this PR not yet being built / uploaded to appveyor:
- Tfs Build Task
- Gem Package

This is because since converting the projects to sdk style, the `AfterTargets` logic in some of the csproj files doesn't seem to run. I have moved most of the logic (for all of the other artifacts) out of the csproj files and into the cake script instead - but haven't yet done the TFS or the Gem package. If someone has some time to contribute, please start there..

# Remaining Tasks

- Get the build to produce TFS task and Gem package again.

# Completed Tasks

- Updated all of the projects to SDK style (vs2017) format.
- Upgraded various dependencies, including `GitTools.Core`, `GitTools.Testing` and a few others to newer versions that that support `netstandard`. 
- `GitVersion` command line now builds for `netcoreapp20` in addition to just `net40` and the additional nuget package is uploaded to appveyor.
- Removed Pepita (now use standard /t:Pack mechanism)
- Removed `Fody`
- GitVersion.sln now builds using `dotnet core` sdk, so this might pave the way to building under a 
 linux/dotnet core image, as opposed to linux/mono image.
- Removed ILMerge and ILRepack from the GitVersionTask project. The DESKTOP version of the assembly has all of its dependencies included as separate dll's inside the nuget package. The .NETCORE version of the assembly, has all of it's dependencies in it's nuspec, and they are restored and resolved at runtime thanks to `UtilPack` and therefore doesn't need to worry about embedding them (which is fortunate because ILMerge and Costura do not work with netcore assemblies so I know no way of embedding assemblies with netcore anyway).
- Removed dependency that `ContinuaCI` build server support had on registry (I fixed this separately in master and merged that into this pr)
- Removed dependency on System.Web (I fixed this separately #1274 in master and merged that into this pr)

# Other Notes
-  `Encoding.GetEncodings()` isn't available on NETStandard1.3. ([it's in netstandard2]( https://stackoverflow.com/questions/44351507/what-is-net-core-equivalent-of-encoding-getencodings
) I am manually populating a collection of common encodings for netstandard.
- Web Requests look different in netstandard api. (had to use compilation directives and alternative methods in many places to get to compile)
- I am using AsyncBridge nuget package for async / await support in net40 of GitVersionCore. I have replaced Thread.Sleep with Task.Delay - as the former not available under netstandard and the latter is better practice anyway.
-System.ComponentModel.WarningException --> GitTools.WarningException

- I have moved a lot of the packaging logic out of csproj files, and into the cake script.
- I added a `Directory.Build.props` file to control some common dependency versions at a global level.


asbjornu:
@dazinator: I would love to review this, but I have no experience with .NET Core and can thus provide very little guidance. Sorry. 😕

GeertvanHorrik:
> I had to remove ContinuaCI support from NETStandard build - as lack of registry API's.

Why does it need access to the registry? It should just check if environment variables are registered, same as on Cake:

https://github.com/cake-build/cake/blob/08907d1a5d97b66f58c01ae82506280882dcfacc/src/Cake.Common/Build/ContinuaCI/ContinuaCIProvider.cs#L62

agc93:
Given .NET Core 2.0 has now been released, is the plan to update to 2.0, or to continue with this work?

As I've mentioned to @JakeGinnivan the current non-Core GitVersion doesn't work on a huge number of Linux platforms with no workaround available..

dazinator:
My own personal view is that:

1. Yes netstandard2 might make this easier. (Would  help to verify using api portability analysis tool)
2. We shouldn't really require netstandard2 and should aim for the lowest netstandard thats reasonable, longer term.

I think that relying on `system.web` for example, is madness and its not unreasonable to want to eliminate this as a dependency.

gitversion core should atleast be able to match the netstandard level targeted by libgit2sharp, if we tidy up the implementation.

If anyone wants to aim for netstandard2 as a first step though - nothing stopping you and its better than nothing. Many of the dependencies are now all compatible so its just waiting for someone to do this work!

dazinator:
@geertvanhortik - have raised #1178 for the continua ci registry issue you pointed out

dazinator:
Going to replace Thread.Sleep with Task.Delay. Thread.Sleep isn't available in desired netstandard version.

GeertvanHorrik:
And Task.Delay is much better (doesn't block CPU).

dazinator:
Yup exactly. Sadly `Task.Delay` isn't available for net40. So I have had to do this:

```
 public static class TaskHelper
    {

#if NETDESKTOP
        public static Task Delay(int milliseconds)
        {
            var tcs = new TaskCompletionSource<bool>();
            System.Timers.Timer timer = new System.Timers.Timer();
            timer.Elapsed += (obj, args) =>
            {
                tcs.TrySetResult(true);
            };
            timer.Interval = (double)milliseconds;
            timer.AutoReset = false;
            timer.Start();
            return tcs.Task;
        }
#else
        public static Task Delay(int milliseconds)
        {
            return Task.Delay(milliseconds);
        }
#endif
    }

```

So we call `await TaskHelper.Delay(x)` now instead.

I have taken another look at the `GetEncodings` api (or lack thereof), I've decided to workaround this for netstandard by explivitly supporting an easily accessible set of encodings:

```
#if NETDESKTOP
            var encodings = (Encoding.GetEncodings());
           
#else
            // GetEncodings not available on netstandard, so just manually adding some common ones.
            var encodings = new List<Encoding>() { Encoding.ASCII, Encoding.Unicode, Encoding.UTF32, Encoding.UTF7, Encoding.UTF8 };

#endif
```

Do you think that would be acceptable?

dazinator:
Hmm so I am struggling to get `Tasks` to compile in net40.

>   GitVersionCache.cs(49,13,49,48): error CS1061: 'Task' does not contain a definition for 'GetAwaiter' and no extension method 'GetAwaiter' accepting a first argument of type 'Task' could be found (are you missing a using directive or an assembly reference?)
1>        Helpers\ThreadSleep.cs(9,13,9,49): error CS1061: 'Task' does not contain a definition for 'GetAwaiter' and no extension method 'GetAwaiter' accepting a first argument of type 'Task' could be found (are you missing a using directive or an assembly reference?)

I thought adding the `https://www.nuget.org/packages/Microsoft.Bcl.Async` nugget package to net4 projects was enough. Perhaps its different with the new csproj file format (i.e PackageReference). I'm checking in what I have done so far, if anyone can resolve this please do. Could we even frop support for net40 and make net45 the new minimum?

dazinator:
Ok so down to 2 last issues:

1. Need to replace this for netstandard:

```
informationalVersion = config.AssemblyInformationalFormat.FormatWith(semverFormatValues);
```

2. When building the tasks project it currently fails complaining about ILRepack - @JakeGinnivan I think this is your neck of the woods. We need to build and package the msbuild task for both desktop and netcore so we need to make changes to the build anyway - can we get rid of ILMerge / ILRepack and PepitaPackage now?

dazinator:
@JakeGinnivan I'm planning to remove ILMerge in favour of Fody weaving of the msbuild task - think that will simplify things. I'll upgrade the project to vs2017 multitargeting, and remove pepita as well as that should no longer be required.

dazinator:
Note to self: following needs to removed and replaced by newer msbuild techniques (+ fody)
```
<Target Name="AfterBuild" DependsOnTargets="Clean" Condition="$(NCrunch)=='' And '$(OS)' != 'Unix'">
    <MakeDir Directories="$(TargetDir)ILMergeTemp\" />
    <PropertyGroup>
      <Runtime Condition="'$(OS)'=='Unix'">mono</Runtime>
    </PropertyGroup>
    <!-- Copy target file so that our temporary file gets the correct permissions -->
    <Copy SourceFiles="$(TargetPath)" DestinationFolder="$(TargetDir)ILMergeTemp\" />
    <Exec Command="$(Runtime) &quot;$(SolutionDir)packages\ILRepack.2.0.13\tools\ILRepack.exe&quot; /allowDup /keyfile:&quot;$(SolutionDir)key.snk&quot; /out:&quot;$(TargetDir)ILMergeTemp\$(TargetFileName)&quot; &quot;$(TargetPath)&quot; &quot;$(TargetDir)GitVersionCore.dll&quot;  &quot;$(TargetDir)GitTools.Core.dll&quot; &quot;$(TargetDir)LibGit2Sharp.dll&quot;  &quot;$(TargetDir)YamlDotNet.dll&quot; /target:library /targetplatform:&quot;v4,$(FrameworkPathOverride)&quot; /ndebug /internalize " />
    <ItemGroup>
      <TempFiles Include="$(TargetDir)ILMergeTemp\*.*" />
    </ItemGroup>
    <MakeDir Directories="$(BuildDir)NuGetTaskBuild" />
    <ItemGroup>
      <NativeBinaries Include="$(TargetDir)lib\**\*.*" />
    </ItemGroup>
    <Copy SourceFiles="@(NativeBinaries)" DestinationFiles="@(NativeBinaries->'$(BuildDir)NuGetTaskBuild\build\lib\%(RecursiveDir)%(Filename)%(Extension)')" />
    <Copy SourceFiles="$(TargetDir)LibGit2Sharp.dll.config" DestinationFiles="$(BuildDir)NuGetTaskBuild\build\GitVersionTask.dll.config" />
    <Copy SourceFiles="$(TargetDir)ILMergeTemp\GitVersionTask.dll" DestinationFolder="$(BuildDir)NuGetTaskBuild\build" Condition="Exists('$(TargetDir)ILMergeTemp\GitVersionTask.dll')" />
    <Copy SourceFiles="$(TargetDir)GitVersionTask.pdb" DestinationFolder="$(BuildDir)NuGetTaskBuild\build" Condition="Exists('$(TargetDir)GitVersionTask.pdb')" />
    <Copy SourceFiles="$(TargetDir)GitVersionTask.dll.mdb" DestinationFolder="$(BuildDir)NuGetTaskBuild\build" Condition="Exists('$(TargetDir)GitVersionTask.dll.mdb')" />
    <Copy SourceFiles="$(ProjectDir)NugetAssets\build\GitVersionTask.targets" DestinationFolder="$(BuildDir)NuGetTaskBuild\build" />
    <Copy SourceFiles="$(ProjectDir)NugetAssets\buildMultiTargeting\GitVersionTask.targets" DestinationFolder="$(BuildDir)NuGetTaskBuild\buildMultiTargeting" />
    <Copy SourceFiles="$(ProjectDir)NugetAssets\GitVersionTask.nuspec" DestinationFolder="$(BuildDir)NuGetTaskBuild" />
    <PepitaPackage.CreatePackageTask NuGetBuildDirectory="$(BuildDir)NuGetTaskBuild" MetadataAssembly="$(ILMergeTemp)GitVersionTask.dll" Version="$(GitVersion_NuGetVersion)" />
    <Delete Files="@(TempFiles)" />
    <RemoveDir Directories="$(TargetDir)ILMergeTemp\" />
  </Target>

```

dazinator:
Seperate PR submitted that removes System.Web!

dazinator:
This is almost there! Would appreciate if anyone could give the three failing approval tests a quick check..

agc93:
Is there any movement or plans for this PR? This would be an awesome improvement (and I could finally go back to using GitVersion!)..

gep13:
@JakeGinnivan @asbjornu @pascalberger are any of you in a position to take a look at this?

asbjornu:
@gep13: I outlined my prerequisites for reviewing this in https://github.com/GitTools/GitVersion/pull/1269#issuecomment-319917991. Sorry. 😕

gep13:
@onovotny any chance you could cast your eyes over this and give us your thoughts? Thanks!

onovotny:
Will take a look, but a few thoughts:

Do you *really* care about `net40` support? It's dead, not supported by Microsoft, and probably has 0-day exploits in the wild. `net452` should be the minimum IMO. Even there, as a build-time tool, I don't think it's unreasonable to assume at least `net461` to exist on a build box. If that's the case, then you can use `netstandard2.0` directly. Keep in mind this is a build-time tool and doesn't limit what people target their libraries to.

dazinator:
@onovotny - thanks for the feedback. 
I also think dropping net40 is a good idea, but I didn't want to conflate this PR anymore than necessary, so I have kept existing support.
I don't think there is any need to upgrade this PR to netstandard2.0 as I have got everything compiling for netstandard 1.3 now - dropping unnecessary API surface area in the process (removed system.web etc). This "trimming the fat" is an improvement that we shouldn't be too quick to throw away imho..

> I should think that you need two sets of MSBuild Tasks, one for full MSBuild and one for Core. There can be some subtlety here around loading dependencies, and you can look at how AssemblyLoadContext can help. Check out the Refit build tasks to see how it works for both full .NET and .NET Core builds.

Yep this is currently outstanding work that I have not had time to do. It's listed as such in the PR description - it just needs someone to get this done really. I am not sure when I will next have any time to continue this work.

gep13:
> @onovotny said...
Will take a look,

Thank you! Appreciate it!

dazinator:
Waiting on: https://github.com/GitTools/GitTools.Core/pull/47 to be merged. Once thats done, I should be able to finally produce the nuget package for the gitversion task that supports dotnet cli and msbuild desktop.

dazinator:
FYI my idea to use Fody / Costura to embed dependencies for the netcoreapp gitversion msbuild task was a no go, as Fody / Costura don't support net core, only desktop .net.

manastalukdar:
Looks like GitTools/GitTools.Core#47 has been merged. Would appreciate this being merged in soon, and a fresh nuget being provided. Thanks.

dazinator:
@manastalukdar - oh you would appreciate it? Ok I wont go out tonight, i'll get right on this for you..

manastalukdar:
@dazinator Thank you for your contributions and civility.

dazinator:
Haha you are welcome. Thank you for yours!

dazinator:
Just an update on the status of this!

The gitversion msbuild task itself is compiling for netcore and desktop msbuild - so its ready to be used for both scenarios... but..

It has dependencies! And those dependencies live along side it in the nuget package. The way dependencies are resolved for an executing msbuild task, under each platform (desktop and netcoreapp) is different. Also it has changed over time in the version of netcoreapp used. netcoreapp1.0 looked different from netcoreapp2.0. For desktop, we were merging all the dependency assemblies into the msbuild task assembly itself - easy. This doesn't work for netcore. 

Luckily, someone out there has already started working on this problem - i.e allow you to write an msbuild task as a nuget package and be executed under the different flavours of msbuild without having to handle all dependency resolution plumbing. I have been involved with testing this out: 
 (see here: https://github.com/CometaSolutions/UtilPack/issues/7)

Once that is ready, I will implement this here in GitVersion - and that should be the completion to this PR :-)

bording:
>Luckily, someone out there has already started working on this problem - i.e allow you to write an msbuild task as a nuget package and be executed under the different flavours of msbuild without having to handle all dependency resolution plumbing. I have been involved with testing this out:
(see here: CometaSolutions/UtilPack#7)

Are you sure something so complicated is necessary? Take a look at how [Nerdbank.GitVersioning](https://www.nuget.org/packages/Nerdbank.GitVersioning/) is bundling everything.

dazinator:
@bording - I had already taken a look at nerdbank gitversioning. I tried to use the same approach and hit casting problems due to the approach it uses. 

The problem I hit was due to the fact it relies on writing a task that acts as a proxy to itself - which caused casting issues.
Example:

1. User does dotnet cli command like  `dotnet build`
2. Latest version of dotnet cli runs. For me this is 2.0.2 preview. It could be 1.0.0.or 1.1.0 or 2.0.0 or whichever version is installed on users machine (we dont enforce a global.json so we cant control this) 
3. Dotnet cli loads an appropriate version of msbuild. For me this was 15.5 preview. It could be 15.3 or 15.1 for older dotnet cli versions.
4. Msbuild runs. Loads the custom task from the task assembly, and executes it.
5. This is where the Task cant directly execute anything that would require a dependency to be loaded because there is nothing that knows how to find the dependency in nuget package where the task assembly itself is. So nerdbank gitversion handles this by using reflection / assembly load context to dynamically load another instance if the task type and its dependency assemblies. It then copies its input property values (task inputs) to the proxied instance input property values (this is where I saw the casting exceptions). Executes the proxied task instance. Copies back the output properties to.its own output properties. Returns result to msbuild.

The casting issues I saw were things like "cannot cast `ITaskItem` to `ITaskItem`. I.am assuming it is to do with the version of msbuild the task is compiled against versus version loaded by dotnet cli at runtime. I wasnt able to find a solution but didnt spend too.much time on it.

So problems are with this approach:

1. Have to include all dependency assemblies in nuget package (not great)
2. Casting issues as explained above - seemingly to do with the version of msbuild compiled against versus version loaded by dotnet cli (which we cant control)
3. Have to compile the task for netcoreapp and desktop - as we have to handle dependency resolution per platform. 

With the utilpack task factory approach:

1. No need to include dependency assemblies in nuget package.
2. No need to have a weird proxying task - code for the task is more standardised and looks ordinary.
3. No need to compile task per platform. Can just compile task once for netstandard.
4. No need to maintain any dependency resolution logic. Can allow dependencies to be restored and  resolved for us.

After some careful consideration I decided to stick with util pack approach as I think it will be a cleaner solution.

bording:
@dazinator Thanks for the detailed explanation. I knew getting dependencies to work inside MSBuild was problematic because of all the different versions of MSBuild, but I hadn't looked in depth into what  Nerdbank.GitVersioning was doing internally to make it look like it was just able to drop the assemblies into the folder.

I'm beginning to think that instead of all of the complexity around getting an MSBuild task working for all the different scenarios, it would be better to look at the model adopted by [SourceLink](https://github.com/ctaggart/SourceLink). The task has no dependencies itself, but calls out to a DotnetCliTool that does all the heavy lifting with LibGit2Sharp.

You do end up needing to add both a package and a tool reference to your project, but that doesn't seem like a problem considering how much simpler it makes everything else.

dazinator:
> I'm beginning to think that instead of all of the complexity around getting an MSBuild task working for all the different scenarios, it would be better to look at the model adopted by SourceLink. The task has no dependencies itself, but calls out to a DotnetCliTool that does all the heavy lifting with LibGit2Sharp.

Sounds interesting. I don't think calling out to a `DotnetCliTool` is a safe assumption because the task might be executing under desktop msbuild and the `dotnetcli` may not even be installed.. but we could call out to our existing `gitversion.exe` which we could bundle inside the same nuget package? But then you would probably have to include multiple `gitversion.exe` in the nuget package, compiled for different targets (netcoreapp, net45 etc). Because if you are running under dotnet cli you can't call the gitversion.exe compiled for `net45` and vice versa? But once you do that, you are back to having a non trivial msbuild task that is concerned with how its being run and which exe to call. What do you think? 

My aim is to keep a single nuget package, with a single msbuild task, that works accross all scenarios:-
1. Legacy projects (old csproj format, project.json projects) 
2. New "sdk" style project format. 
3. Running under desktop msbuild
4. Running on netcoreapp msbuild (all stable releases)

I am open to any approach that satisfies the above. Utilpack task factory is one approach I am looking at currently, but if you have some time, you could clone this branch and spike up an alternative - if it works then great we could roll with that instead!

dazinator:
[UtilPack.NuGet.MSBuild](https://github.com/CometaSolutions/UtilPack/tree/develop/Source/UtilPack.NuGet.MSBuild) is ready for me to try with gitversion msbuild task now. I will try it out, and hopefully, finally, get a nuget package for `gitversion` task published by the end of the week - so people can try it out. Looks like there are some conflicts to resolve too. Hopefully this won't be anything too difficult to merge.

dazinator:
@bording - I have hit a bit of a wall with the `UtilPack` approach at the moment. The task runs, but there is a problem locating the `libgit2sharp` native binaries. I assumed they have to be included in the nuget package with the msbuild task assembly itself, but that didn't appear to work. If you have any other approaches you feel might work, please feel free to contribute!

dazinator:
Might seek some help over on the libgit2sharp repo.

dazinator:
BTW - The task is running under desktop as before, and that works. The issue of libgit2sharp not being able to locate its binaries is only happening when running `dotnet build`

dazinator:
Blocked by: https://github.com/libgit2/libgit2sharp/issues/1498

bording:
> Sounds interesting. I don't think calling out to a DotnetCliTool is a safe assumption because the task might be executing under desktop msbuild and the dotnetcli may not even be installed.. but we could call out to our existing gitversion.exe which we could bundle inside the same nuget package? But then you would probably have to include multiple gitversion.exe in the nuget package, compiled for different targets (netcoreapp, net45 etc). Because if you are running under dotnet cli you can't call the gitversion.exe compiled for net45 and vice versa? But once you do that, you are back to having a non trivial msbuild task that is concerned with how its being run and which exe to call. What do you think?

I suppose it's too radical of an option for now to consider the dotnet SDK and SDK-style csproj projects a requirement for this work!

However, the existing targets/tasks in GitVersionTask work fine for desktop msbuild scenarios, so perhaps those could continue to be used unaltered for that case.

Then if wanting to invoke from `dotnet build` etc, then the CLI tool could be required? The .NET Core version of the task would be the only one to have that dependency.

Obviously, there would be a lot details to work out, but it seems like something that might have some merit.

dazinator:
I have got the gitversiontask running under desktop and netcore msbuild now. But it required a fix to libgit2sharp which I submitted a PR for and may not be merged due to ongoing changes in the same area. Worst comes to worst, I may have to switch to a temporary fork of libgit2sharp containing my fix.

@bording other option is we pursue your idea further. I think I see what you are suggesting.. So we could keep a single `gitversiontask` nuget package. Keep the old task in it that behaves as before. Then add a new task to it, that will be called only under netcore msbuild? The new task would call out to the dotnetcli tool of gitversion exe. We would create that as an additional package. Those building under dotnetcore would need to add that additional dotnetclitool package. 

If we went down that route we would have:

1. Duplicate task class with different implementations (legacy vs netcore)
2. Another .exe in the form of a dotnetclitool. Unclear whether or how much that overlaps with the existing gitversion.exe or whether it would proxy calls to it?

dazinator:
I could use some help figuring out why the appveyor build is failing: https://ci.appveyor.com/project/GitTools/gitversion/build/0.0.1440

I won't have much more time to work on this for a while, and fixing the appveyor / travis builds is the only thing holding back getting a nuget package out there for testing with netcore.

dazinator:
Sorry just to be super clear on status of this PR. You can build this locally and produce GitVersionTask nuget package that works with both NETCore and NET Desktop. So basically it's looking pretty good. However becasue the AppVeyor / Travis builds are broken (due to upgrading the build environment to vs2017) we are unable to publish any official nuget packages anywhere. 

If anyone has anytime to look at the AppVeyor or Travis builds, please do and it will help speed this along.

bording:
@dazinator I'm hoping to have some time this weekend to take a look at the changes here, so while I'm doing that, I'll see if I can figure out the problem with the CI builds.

dazinator:
@bording - thank you. I have a suspicion its probably the support for appveyor within gitversion itself that is breaking the appveyor build.. I had to make a few changes to work around lack of api's in netstandard for making a web call. Not looked at all at travis build. Appreciate your help!

jnm2:
> Then add a new task to it, that will be called only under netcore msbuild?

No way. I need this for modern-SDK csproj targeting .NET Framework and regular too.

jnm2:
This was super helpful to me in creating cross platform MSBuild and dotnet build packages: http://www.natemcmaster.com/blog/2017/07/05/msbuild-task-in-nuget/

src/GitVersionCore/StringComparerUtils.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public static class StringComparerUtils
+    {
+#if NETDESKTOP
+        public static readonly System.StringComparer IngoreCaseComparer = StringComparer.InvariantCultureIgnoreCase;
+        public static readonly StringComparison IngoreCaseComparison = StringComparison.InvariantCultureIgnoreCase;
+        public static readonly StringComparison CaseSensitiveComparison = StringComparison.InvariantCulture;
+#else
+        public static readonly System.StringComparer IngoreCaseComparer = StringComparer.OrdinalIgnoreCase;
+        public static readonly StringComparison IngoreCaseComparison = StringComparison.OrdinalIgnoreCase;
```
jnm2: `Ingore` is a typo (4x)

dazinator:
@jnm2 - don't worry there is only a single msbuild task and it runs under netcore and desktop msbuild. You should be able to check this branch out and give it a try. This will be easier to try out once we fix the appveyor / travis builds and get nuget packages published somewhere.

src/GitVersionCore.Tests/AssemblyInfoFileUpdaterTests.cs
```diff
@@ -4,482 +4,486 @@
...
 using GitVersion.Helpers;
-using GitVersionCore.Tests;
 using NSubstitute;
 using NUnit.Framework;
 using Shouldly;
+using GitVersionCore.Tests;
 
-[TestFixture]
-public class AssemblyInfoFileUpdaterTests
+namespace GitVersionExe.Tests
```
bording: The changes here seem to indicate the forward integration master missed some things, particularly around the changes I made in #1249.

bording: The fact that the approve files above all are showing changes seems to be related to this as well.

bording:
As mentioned earlier in the PR, I think that targeting netstandard2.0 is a good idea. By the time the final version of LibGit2Sharp 0.25.0 arrives, it will also be targeting netstandard2.0: https://github.com/libgit2/libgit2sharp/pull/1483

dazinator:
> As mentioned earlier in the PR, I think that targeting netstandard2.0 is a good idea. By the time the final version of LibGit2Sharp 0.25.0 arrives, it will also be targeting netstandard2.0

We have no choice but to drop support for anything below netstandard 2.0 if thats what libgit2sharp next version will require. Hopefully that should be fine with gitversion consumers. Not much choice if its not other than to remain on libgit2sharp preview which is obviously not good long term.

dazinator:
@bording  - I take it you didn't get any time for this on the weekend, so I am going to have a quick look at the build issue now!

gep13:
@dazinator @bording regarding the AppVeyor Build Failure...

I wonder if this:

https://ci.appveyor.com/project/GitTools/gitversion/build/0.0.1440#L1349

Is part of the problem?

gep13:
@dazinator @bording also, there were some known GitVersion issues in Cake, specifically around the handling of a GitVersion property which is returned.  I would suggest, as a minimum, moving to Cake 0.22.2.

See release notes for more information: https://github.com/cake-build/cake/releases

dazinator:
@gep13  - oh thanks I missed that line. Yes that looks to be the issue. I'll have a go at fixing that up in the AppVeyor build support.

I will also investigate cake update!

src/GitVersionCore.Tests/AssemblyInfoFileUpdaterTests.cs
dazinator: > The changes here seem to indicate the forward integration master missed some things, particularly around the changes I made in #1249.

Yeah.. Looks like I had a merge conflict which I resolved in Commit 42d253e7 by adding a namespace and some other minor changes that you see. GitHub shows massive changes to that file, but a VS compare shows only serveral lines were changed. However I have no idea why I made those changes so perhaps it was unintentional. I have just reverted the file to match master version anyhow.

> The fact that the approve files above all are showing changes seems to be related to this as well.

They are all listed as "No changes." though. Perhaps it shows them just because they were renamed / relocated in master.

dazinator:
@gep13 - Made a bit of progress on fixing the build (I hope).

Now getting:

> INFO [11/07/17 6:29:54:58] Working directory: C:\Windows\system32
INFO [11/07/17 6:29:54:59] IsDynamicGitRepository: False
ERROR [11/07/17 6:29:54:61] An unexpected error occurred:
System.IO.DirectoryNotFoundException: Can't find the .git directory in C:\Windows\system32
   at GitVersion.GitPreparer.GetDotGitDirectory()
   at GitVersion.GitPreparer.GetProjectRootDirectory()
   at GitVersion.ConfigurationProvider.Verify(GitPreparer gitPreparer, IFileSystem fileSystem)
   at GitVersion.Program.VerifyArgumentsAndRun()
INFO [11/07/17 6:29:54:61] 
INFO [11/07/17 6:29:54:61] Attempting to show the current git graph (please include in issue): 
INFO [11/07/17 6:29:54:61] Showing max of 100 commits

 https://ci.appveyor.com/project/GitTools/gitversion/build/4.0.0-PullRequest1269.1443.build.1442#L1905

gep13:
@dazinator we are probably going to want to use the `DotNetCoreTest` alias, rather than the `NUnit3` alias.

dazinator:
@gep13 - I will give that a go. Worth noting we can't switch to `DotNetCoreBuild` yet due to some incompatible msbuild tasks in use like Fody weaver. I'll try just switching the tests out to `DotNetCoreTest` though - hopefully there is a no-build argument still.

dazinator:
@gep. I tried switching to `DotNetCoreTest`:

```
Task("Run-NUnit-Tests")
    .IsDependentOn("Build")
    .Does(() =>
{

     var settings = new DotNetCoreTestSettings
     {
         Configuration = "Release",
		 NoBuild = true
     };

         DotNetCoreTest("./src/GitVersionCore.Tests/GitVersionCore.Tests.csproj", settings);
	 DotNetCoreTest("./src/GitVersionExe.Tests/GitVersionExe.Tests.csproj", settings);
	 DotNetCoreTest("./src/GitVersionTask.Tests/GitVersionTask.Tests.csproj", settings);
    
});

```

NUnit had the benefit that it output an xml file which we uploaded to appveyor. Not clear on how to do this with `DotNetTest`. 

It did clear those tests that were failing in NUnit Console runner. But now all the tests with ApprovalTests are failing (approx 30) becuase [ApprovalTests doesn't appear to support netcore yet](https://github.com/approvals/ApprovalTests.Net/issues/162)


There is also the other issue that we build under XBuild on Unix:

```
if(IsRunningOnUnix())
    {
        XBuild("./src/GitVersion.sln",  new XBuildSettings()
            .SetConfiguration(configuration)
            .WithProperty("POSIX", "True")
            .SetVerbosity(Verbosity.Minimal));

```

I'm a windows guy - will `DotNetCoreTest` run on Unix?

dazinator:
Also running the cake build script locally it seems to get far enough to the failed tests. AppVeyor build is failing straight away:

```
Executing task: DogfoodBuild
Cake.exe : An error occurred when executing task 'DogfoodBuild'.
At line:1 char:1
    + CategoryInfo          : NotSpecified: (An error occurr...'DogfoodBuild'.:String) [], RemoteException
 
Error: 
One or more errors occurred.
```

jnm2:
> NUnit console runner had the benefit that it output an xml file which we uploaded to appveyor. Not clear on how to do this with DotNetTest.

Yeah... I just ran into this myself at https://github.com/AMVSoftware/Cake.SqlServer/commit/4a3747672dec0320203f977090e2933d50fdacf2#diff-4d3634747586b964df9f7712bc3ec669.

NUnit has not yet implemented support for outputting test result XML via the `dotnet test` adapter, and so far that's the only way to run .NET Core tests. You can tell `dotnet test` to output in MSTest format via `Logger = "trx"`; the problem is that the MSTest format can only contain one test run per file, and dotnet test will only test one target framework per test run. So you'll end up with one file per target framework and no way to merge them, and AppVeyor ignores all but the first.

https://github.com/Microsoft/vstest/issues/880#issuecomment-341912021

dazinator:
@JakeGinnivan - any chance you could lend your considerable talents against the failing builds? I am not really sure how to fix them.. The builds work fine locally for me :-(

JakeGinnivan:
@dazinator will take a look over the next few days :)

gabrielweyer:
@dazinator I fixed the issue, it was due to a `NullReferenceException` in `Cake`

In [`build.cake`](https://github.com/dazinator/GitVersion/blob/feature/netstandard/build.cake#L42-L46) replace:

```csharp
msBuildSettings = msBuildSettings
	.WithProperty("GitVersion_NuGetVersion", nugetVersion)
	.WithProperty("GitVersion_SemVer", semVersion)
	.WithProperty("GitVersion_MajorMinorPatch", version)
	.WithProperty("GitVersion_PreReleaseTag", **preReleaseTag);**
```

By:

```csharp
if (!string.IsNullOrWhiteSpace(nugetVersion))
	msBuildSettings.WithProperty("GitVersion_NuGetVersion", nugetVersion);

if (!string.IsNullOrWhiteSpace(semVersion))
	msBuildSettings.WithProperty("GitVersion_SemVer", semVersion);

if (!string.IsNullOrWhiteSpace(version))
	msBuildSettings.WithProperty("GitVersion_MajorMinorPatch", version);

if (!string.IsNullOrWhiteSpace(preReleaseTag))
	msBuildSettings.WithProperty("GitVersion_PreReleaseTag", preReleaseTag);
```

Those lines are protected by this guard: `if (BuildSystem.AppVeyor.IsRunningOnAppVeyor)`, that's why it worked on your machine but failed in `AppVeyor`.

dazinator:
@gabrielweyer thank you so much! I will give that a try.

dazinator:
@gabrielweyer - Ok I made that change to the cake script.. I think that solved that one, but the build still fails, probably now as a result of something else. If I run `.\build.ps1 > output.txt` locally here is a [gist of the result](https://gist.github.com/dazinator/bd915cf728d10d8f1860aaea6cf593d5)

Basically, I see this i my output:

```
GitVersionExe.Tests -> C:\Users\Darrell\Source\Repos\GitVersion\src\GitVersionExe.Tests\bin\Release\net452\GitVersionExe.Tests.dll
  GitVersionTask -> C:\Users\Darrell\Source\Repos\GitVersion\src\GitVersionTask\bin\Release\netstandard1.5\GitVersionTask.dll
Finished executing task: DogfoodBuild
```

Where as the AppVeyor build gives this:

```
GitVersionExe.Tests -> C:\projects\gitversion\src\GitVersionExe.Tests\bin\Release\net452\GitVersionExe.Tests.dll
Cake.exe : An error occurred when executing task 'DogfoodBuild'.
At line:1 char:1
    + CategoryInfo          : NotSpecified: (An error occurr...'DogfoodBuild'.:String) [], RemoteException
 
Error: 
One or more errors occurred.
Command executed with exception: 

```

I think perhaps we need some better output from cake (via msbuild) to work out what is going wrong there!

gabrielweyer:
@dazinator Interesting, I forked your repo and applied the fix. The build on `AppVeyor` is successful but afterwards I've got failing tests: https://ci.appveyor.com/project/GabrielWeyer/gitversion

This is the error on your latest `AppVeyor` build:

```
"C:\projects\gitversion\src\GitVersion.sln" (Build target) (1) ->
"C:\projects\gitversion\src\GitVersionCore\GitVersionCore.csproj" (default target) (3:17) ->
"C:\projects\gitversion\src\GitVersionCore\GitVersionCore.csproj" (Build target) (3:19) ->
(CoreCompile target) -> 
  GitPreparer.cs(86,108): error CS0117: 'StringComparison' does not contain a definition for 'InvariantCultureIgnoreCase' [C:\projects\gitversion\src\GitVersionCore\GitVersionCore.csproj]
  GitPreparer.cs(93,106): error CS0117: 'StringComparison' does not contain a definition for 'InvariantCultureIgnoreCase' [C:\projects\gitversion\src\GitVersionCore\GitVersionCore.csproj]
```

I'm not sure yet why they behave differently. Will have a look now.

dazinator:
Sorry I managed to miss this somehow! Thank you for investigating. I will make those changes and lets see if it fixes the build :-)

dazinator:
@gabrielweyer  - thanks I think thats solved it. We are down to some tests now that fail due to using ApprovalTests under dotnet core (ApprovalTests doesn't support net core).

I am not sure what the best answer is for that. Perhaps we still need to run this subset of the tests under full net framework? @JakeGinnivan  any suggestions?

dazinator:
I'll try and get the tests running nunit console again, instead of dotnet cli for now.

dazinator:
It looks like that many tests fail when running under a Release build configuration due to the following Shouldly exception:

https://github.com/shouldly/shouldly/blob/cb58830180c1c3f87db9b0eb74ef9d7c446ec71e/src/Shouldly/ShouldlyExtensionMethods/ShouldMatchApprovedTestExtensions.cs#L53

dazinator:
Turns out Shouldy requires full pdb debug symbols when building tests projects in release mode. One of the tests projects was missing:

```
 <DebugSymbols>true</DebugSymbols>
```

dazinator:
Bug announcement!

GitVersionTask should now run under normal desktop .net msbuild and also.. DotNetCore ms build!

Here are the steps to try it:

Steps:

1.  Add AppVeyor nuget feed: https://ci.appveyor.com/nuget/gitversion-8nigugxjftrw
2. Create a project:

```

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.0</TargetFramework>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="GitVersionTask" Version="4.0.0-pullrequest1269-1500" />
  </ItemGroup>

</Project>
```

Note: There are lots of other pull request packages - be sure to pick the one shown above for PR 1269.

3. Make sure you have initialiased a Git repository and made atleast one commit. Otherwise when you build you will get the following error:

> Error	MSB4044	The "GetAssemblyVersion" task was not given a value for the required parameter "NuGetVersion".

It seems GitVersionTask has always has an error if you attempt to build without having a git repo with atleast one commit. We can fix this as a seperate issue.

Let me know how it goes..

manastalukdar:
@dazinator Thanks for your work on this. I tried out the nuget from AppVeyor, and it works great from within Visual Studio. However, running dotnet build fails:

```
MSBUILD : Task factory error error NMSBT000: Failed to load actual task factory assembly System.IO.FileNotFoundException: Could not load file or assembly 'C:\Users\manas\.nuget\packages\utilpack.nuget.msbuild\2.2.0\build\netcoreapp1.1\UtilPack.NuGet.MSBuild.NuGet.4.5.0.dll'. The system cannot find the file specified. [D:\dev\work\Source\some.server\some.server.csproj]
MSBUILD : Task factory error error NMSBT000: File name: 'C:\Users\manas\.nuget\packages\utilpack.nuget.msbuild\2.2.0\build\netcoreapp1.1\UtilPack.NuGet.MSBuild.NuGet.4.5.0.dll' [D:\dev\work\Source\some.server\some.server.csproj]
MSBUILD : Task factory error error NMSBT000:    at System.Runtime.Loader.AssemblyLoadContext.LoadFromPath(IntPtr ptrNativeAssemblyLoadContext, String ilPath, String niPath, ObjectHandleOnStack retAssembly) [D:\dev\work\Source\some.server\some.server.csproj]
MSBUILD : Task factory error error NMSBT000:    at System.Runtime.Loader.AssemblyLoadContext.LoadFromAssemblyPath(String assemblyPath) [D:\dev\work\Source\some.server\some.server.csproj]
MSBUILD : Task factory error error NMSBT000:    at UtilPack.NuGet.MSBuild.NuGetTaskRunnerFactory..ctor() [D:\dev\work\Source\some.server\some.server.csproj]
MSBUILD : Task factory error error NMSBT000:  [D:\dev\work\Source\some.server\some.server.csproj]
MSBUILD : Task factory error error NMSBT000: . [D:\dev\work\Source\some.server\some.server.csproj]
C:\Users\manas\.nuget\packages\gitversiontask\4.0.0-pullrequest1269-1500\build\functionality\GitVersionBuild.targets(6,5): error MSB4175: The task factory "UtilPack.NuGet.MSBuild.NuGetTaskRunnerFactory" could not be loaded from the assembly "C:\Users\manas\.nuget\packages\utilpack.nuget.msbuild\2.2.0\build\\netcoreapp1.1\UtilPack.NuGet.MSBuild.dll". Object reference not set to an instance of an object. [D:\dev\work\Source\some.server\some.server.csproj]
```

stebueh:
According to the release notes, version 2.3.0 of UtilPack.NuGet.MSBuild  might fix this; see https://github.com/CometaSolutions/UtilPack/issues/17

dazinator:
@manastalukdar @stebueh  - thanks - I will update utilpack dependency. I managed to build from VS and dotnet cli successfully, I wonder what the difference was for me!

dazinator:
@manastalukdar - please try `<PackageReference Include="GitVersionTask" Version="4.0.0-pullrequest1269-1501" />`

stebueh:
@dazinator I think you missed to update the UtilPack.NuGet.MSBuild version in the .nuspec; using 4.0.0-pullrequest1269-1501 and manually modifying the version in the .nuspec locally made it work for me.

dazinator:
@stebueh - thanks - well spotted, I'll correct that :-)

dazinator:
I am too used to having the nuspec generate automatically via `dotnet pack` :-)

manastalukdar:
"dotnet build" now works with  4.0.0-pullrequest1269-1502. However, the versioning is broken. The data in GitVersionInformation.g.cs looks correct, But the *.AssemblyInfo.cs file does not reflect that. It has picked up default version numbers. Just to be sure, I went back to 1500 and 1501, and in both cases the versioning is correct and the project builds fine from VS. I manually edited the .nuspec in 1501 to use 2.3.0 of utilpack, and that got dotnet build to work but the versioning broke. So, looks like unitpack 2.3.0 somehow breaks the versioning.

Also, in case anyone else runs into this, I had to delete the 2.2.0 folder for utilpack.nuget.msbuild in the nuget package cache to get rid of the previous dotnet build errors even with the newer 1502 (and manually edited 1501 nuspec) GitVersionTask nuget package. Somehow it was picking up the older utilpack package.

dazinator:
@manastalukdar  -  Do you have an assemblyinfo.cs file your project or are you allowing the project system to generate one for you? If you have one in the project, try deleting the assembly version attributes. Failing that are you able to upload the build logs? Do you get the expected result in a Visual Studio build still or does this happen with both VS build and a dotnet core build? or just dotnet core?

I will do some more testing myself using net core 2.0

manastalukdar:
@dazinator 

I do not have an assemblyinfo.cs in my project. The project system generates one.

- With 1502 of GitVersionTask (which has 2.3.0 of utilpack), I do not get the expected result with both VS and dotnet core build.
- With 1500 and 1501, I get the expected result with VS. "dotnet build" fails with the errors I had reported earlier.
- With 1501 (and manually editing its .nuspec to use 2.3.0 of utilpack), the build succeeds with both VS and "dotnet build". However the versioning is incorrect in both cases.

The build log I got with the (work) project where this happens is massive and will probably be a pain to parse. I'll try to reproduce this with a simpler sample project within a day or two.

dazinator:
@manastalukdar thanks a repro / sample project would really help here.

manastalukdar:
@dazinator 

Project in [this repo](https://github.com/manastalukdar/gitversion_pr1269) reproduces the issue. Some points to note:

With 1502, if you have 2.2.0 of utiltask in your nuget cache, then 
- building using VS will provide the correctly versioned binary.
- building using "dotnet build" will fail with the errors I have previously mentioned.

With 1502, after having removed 2.2.0 of utiltask from nuget cache, building using VS and "dotnet build" both succeed. However, the versioning of the binary is incorrect in both cases.

ermshiperete:
With 1502 I'm getting:

    error MSB4184: The expression ""$HOME/.nuget/packages/gitversiontask/4.0.0-pullrequest1269-1502/build/".Substring(-1)" cannot be evaluated. StartIndex cannot be less than zero.

ermshiperete:
A possible fix for the problem with the differing path separators would be:

  	<ItemGroup>
		<PathSeparators Include="/"/>
		<PathSeparators Include="\"/>
	</ItemGroup>
	<PropertyGroup>
		<GitVersionTaskBuildTools_IsMultiTargeting>false</GitVersionTaskBuildTools_IsMultiTargeting>
		<GitVersionTaskBuildTools_IsMultiTargeting Condition="$(MSBuildThisFileDirectory.Substring($(MSBuildThisFileDirectory.LastIndexOfAny(@(PathSeparators)))).Substring(1).Contains('buildMultiTargeting'))">true</GitVersionTaskBuildTools_IsMultiTargeting>
	</PropertyGroup>
`

dazinator:
@ermshiperete - thanks I have applied your suggestion - would you be able to give it another test on Linux and let me know if it works? I'd be interested to know if you actually get correctly version stamped assembies and what version of the dotnet cli you are using to build with :-)

dazinator:
@asbjornu @JakeGinnivan - I have managed migrate much of the build logic that was in the csproj files over to cake. Including ILREPACK and nuget packaging etc.

A couple of things still elude me - how to produce the VSIX for the tfs task, and how to produce the gem. I'd like to add those steps to the cake script too but for now I have had to comment out the upload of those two artifacts.

dazinator:
@manastalukdar I think I have solved the problem - here is a working repro: https://github.com/dazinator/gitversion_pr1269

Note: If you have installed VS 15.6 your dotnet sdk has probably been bumped to version `2.1.100`. In that case you will need a `global.json` in your repo that pins the sdk version to 2.0.0 like in the sample repo above. This won't be necessary if / when the following gets fixed: https://github.com/CometaSolutions/UtilPack/issues/19

dazinator:
This is now ready for some further testing, please try it out and let me know if you encounter any issues. I have updated the issue description but here is how you test:

1. Add the appveyor nuget feed: https://ci.appveyor.com/nuget/gitversion-8nigugxjftrw
2. In your project folder: `dotnet add package gitversiontask -v 4.0.0-pullrequest1269-1530`

dazinator:
It seems that GitVersionTask has an issue when building from VS2017 and using multi-targeting to target more than one framework. In that scenario, Costura fails to resolve the embedded dependencies on the second build. Most likely something to do with the way costura registers itself to resolve assemblies, and it not registering on the second run due to something getting cached. Very speculative but for now, I think the easiest solution is to remove embedding the GitVersionTask assemblies completely - as I don't think it is necessary..

dazinator:
Have published Version `4.0.0-pullrequest1269-1531` of GitVersionTask, which fixes the vs - multi-targeting scenario mentioned above.

dazinator:
Be aware when building or restoring using `dotnet` commands you need to use the `--disable-parallel` flag (#1381)

dazinator:
Found the following issue: https://github.com/CometaSolutions/UtilPack/issues/24

ermshiperete:
@dazinator I tried on Linux. With 1534 the build of my test project fails with

    Error /usr/share/dotnet/sdk/2.1.4/Sdks/Microsoft.NET.Sdk/build
    /Microsoft.NET.GenerateAssemblyInfo.targets (146,5)
    MSB4044: The "GetAssemblyVersion" task was not given a value for the required 
    parameter "NuGetVersion".

dazinator:
Do you have an initialised git repo with atleast one commit?

ermshiperete:
Ahh - thanks! Yes, that was the problem now. It now works and it seems I get correctly version stamped assemblies.

dazinator:
Excellent news :-)

dazinator:
Just be careful of: CometaSolutions/UtilPack#24

manastalukdar:
@dazinator Happy to report that with 1534, both dotnet build and VS2017 based build now works fine with the correct version being stamped onto the assemblies.

dazinator:
@manastalukdar excellent... finally getting somewhere haha.
I have just fixed a remaining issue hopefully which you would only notice if your project is targeting a platform that doesn't support a minumum of `netstandard1.5`. This should be fixed in the next build of this PR but I'll test and confirm.

dazinator:
Have also tested that this works when building with `dotnetcore sdk version 1.1.`. I suspect there may be some production environments that never shifted to 2.0.0 so I though this was worthwhile to make sure this was supported.

dazinator:
Might as well keep going on this PR and get GitVersion.exe running on dotnet core too.
I think:

1. Remove Fody
2. Use `Microsoft.Extensions.CommandlineUtils` for command line argument parsing.
3. Get it building for dotnet core (in addition to the existing net40 exe)

Once that is done, appveyor can publish the dotnet core version,  and we can make a dotnetcore docker image that uses it.

stephenwilliams:
@dazinator might I suggest an alternative to `Microsoft.Extensions.CommandlineUtils` as they are [no longer actively developing it](https://github.com/aspnet/Common/issues/257). I'd suggest using [McMaster.Extensions.CommandLineUtils](https://github.com/natemcmaster/CommandLineUtils) as it is an actively develop (and, IMHO, improved) fork.

dazinator:
@stephenwilliams  ah yes I had read that somewhere also. Yes, that sounds sensible.
Although actually I think now that I should probably just get the existing stuff working on dotnet core, and then the agrument parser can be changed in future as a seperate PR - so for now forget I mentioned it :-)

dazinator:
@adamralph - that's a good point, and to be honest I am a bit confused when it comes to netstandard and executables.

I can build the `GitVersion` exe project locally targeting netstandard20. The output is some dll's and a `deps` file. It obviously won't output an `exe` as that is windows only, and I get that. But how is this output going to be beneficial to anyone? If I try and run `dotnet gitversion.dll` in the output folder it fails:

> dotnet gitversion.dll
A fatal error was encountered. The library 'hostpolicy.dll' required to execute the application was not found in 'F:\Repos\GitVersion\src\GitVersionExe\bin\Release\netstandard2\'.

So rather than `dotnet build` I tried:

> dotnet publish -c release -f netstandard2

I then changed into the publish folder and tried `dotnet gitversion.dll` and got the same error:
 >dotnet gitversion.dll
A fatal error was encountered. The library 'hostpolicy.dll' required to execute the application was not found in 'F:\Repos\GitVersion\src\GitVersionExe\bin\Release\netstandard2\publish\'.

So this has me confused. How can I make this runnable!? :-)

dazinator:
Fyi its now also possible to `dotnet build` the GitVersion.sln itself which means we can potentially move our own build environments to dotnet core based ones.

adamralph:
@dazinator I didn't realise that the final output is an executable. In that case, you do indeed need to build it for .NET Core, and not .NET Standard.

.NET Standard is not a runtime platform, it's more like a contract that libraries should conform to in order to be able to run on the platforms that support that version of .NET Standard. You may find it useful to watch [my talk](https://vimeo.com/254635878) on the subject.

In a nutshell, you should be aiming to build any libraries (e.g. GitVersionCore.dll) for .NET Standard and then any executables for specific platforms such .NET Framework, .NET Core, etc.

bording:
> So current thinking is rather than publish native executables for all of the various platforms, we would just publish for netcoreapp20. Then if you are on linux, you would install dotnet core and the you can run gitversion with it. This means that we would support all the platforms that can install dotnet core, similar to how we currently support other platforms via mono,

It would probably be a better model to also have some sort of .NET Framework version of the exe as well because if someone isn't using .NET Core at all, they shouldn't have to install it just to use GitVersion.

dazinator:
Gitversion running nicely on dotnet core:

image

dazinator:
`GitVersion.exe` is currently distributed as:
1. A nuget package called `GitVersion.CommandLine`.
2. A choco package named `GitVersion.Portable` (`portable`?) 
3. A ruby Gem package.

For the dotnet core version
1. I'll create a new `tool` nuget package: `GitVersion.CommandLine.DotNet`. If you install this nuget package you would then have to CD into its directory and run `dotnet GitVersion.dll`
2. I am unsure what support chocolatey has for dotnet core. For example, how would chocolatey automatically make it available on the command line given that you have to invoke using `dotnet` 
3. I have no experience with Ruby so I won;t be looking to create a ruby package for this.

dazinator:
It might be worth waiting a bit and distributing the dotnet version of GitVersion as a global tool: https://www.natemcmaster.com/blog/2018/02/02/dotnet-global-tool/

dazinator:
There is not an additional nuget package avaialble, which contains `GitVersion` command line, but build for `dotnet core`. Have updated this issue with instructions on how to use it.

ermshiperete:
Could this PR be merged even if we don't yet have a GitVersion executable and don't have a global tool? Having the GitVersion task available would greatly help in creating nuget packages for dotnet core. The remaining to-do items could then be put in a new PR.

dazinator:
Well in this pr everything except for the tfs and gem packages artifacts are building. So its just those two holding this up. If somebody wants to merge it prematurely and address those two later then thats fine by me.

dazinator:
I think the travis build is using mono. I am going to look into changing this to run on dotnet core instead.

dazinator:
Seems this branch is a bit behind recent changes in master with [StringFormatWith](https://github.com/GitTools/GitVersion/commits/master/src/GitVersionCore/StringFormatWith.cs)
I'm going to merge those in now

dazinator:
Ok so code that was recently added in master to do with using environment variables in assembly version format strings, doesn't compile for netstandard.

image

I have had to change it to this to get it to compile:

image


I _think_ that both flavours of the method call behave the same from their descriptions, the second one says it looks at the process level so hopefully this should be fine.

asbjornu:
As long as the tests are run for all supported platforms and they are all green, it should be fine.

dazinator:
Travis is giving:

> All packages listed in /home/travis/build/GitTools/GitVersion/tools/packages.config are already installed.
Analyzing build script...
Processing build script...
Installing tools...
Installing addins...
Compiling build script...
Error: Error occurred when compiling build script: (2399,40): error CS0012: The type 'Object' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'.

I have no idea what that's about.

dazinator:
Bit confused about travis and I have run out of time. Hopefully someone will have more luck figuring out the problem than I!

asbjornu:
How about trying `latest` for Mono instead of version `5.4.2`?

asbjornu:
@gep13: Being the Cake man here, any ideas what https://github.com/GitTools/GitVersion/pull/1269#issuecomment-378983670 might be about?

asbjornu:
Related: cake/issues#2063. Sorry I can't help out here. I'm actually surprised that Mono is required to execute `dotnet` on Linux. I thought Mono was only for .NET Framework and that `dotnet` was going to be cross platform. Does that only apply to the runtime and not the build environment?

dazinator:
No I converted the build environment to dotnet core too as part of this PR so mono isnt required to do a dotnet restore, dotnet build etc.. however the build calls a couple of tools like ILMerge and NUnit console which I believe still require mono as those tools use full net.

asbjornu:
It seems like [NUnit v3.9 has .NET Core support](https://github.com/nunit/docs/wiki/.NET-Core-and-.NET-Standard) and I thought we decided to throw ILMerge in the bin?

bording:
@asbjornu NUnit does work with .NET Core, but you need to switch to using `dotnet test` along with adding the test adapter package to your test projects to make it work.

asbjornu:
@bording: Sounds like we should be doing that, then? Would be very nice to not having to rely on Mono for anything.

dazinator:
I did switch to running nunit tests using dotnet cli and the nunit adaptor, and that cake task is actually still in the cake script just not called, because the issue was (and was discussed somewhere above at the time), its not easy to get the results out in a file for upload to appveyor as an artifact. Therefore I kept using nunit console. If someone can figure that bit out then we could drop nunit console.

dazinator:
ILMerge is still needed to produce the single GitVersion.exe on NET Desktop. We dropped ILMerge from gitversiontask.

dazinator:
@asbjornu another problem with switching to dotnet test (also mentioned up above) was that some tests are `approvaltests` and `approvaltests` don't run under dot net core (because its not supported.)

bording:
@dazinator Which dependencies are being merged into GitVersion.exe?

Which library is being used for the approval tests? I know [ApprovalTests.Net](https://github.com/approvals/ApprovalTests.Net) can be made to work with .NET Core. [Assent](https://github.com/droyad/Assent) is one that also works. It even has a .NET Core assembly in the package.

dazinator:
@bording I agree, if there is a way to get the approval tests running or an easy migration path we should do it. 

Regarding ILMerge, ILRepack is used to merge all assemblies in the bin folder into the .exe. This is for the distribution of Gitversion.exe in a couple of the nuget packages (targeting full .Net). In this PR I moved this logic out of csproj files AfterBuild steps and into the cake script so if you need more detail check the cake script.

bording:
> Regarding ILMerge, ILRepack is used to merge all assemblies in the bin folder into the .exe. This is for the distribution of Gitversion.exe in a couple of the nuget packages (targeting full .Net). In this PR I moved this logic out of csproj files AfterBuild steps and into the cake script so if you need more detail check the cake script.

I'll take a look, but overall I'm wondering what the benefit of all the assemblies being combined into a single exe really is. If it was a library that would be referenced, then yeah having a private copy of a dependency can be useful, but for an exe, why not just let the dlls sit in the same folder?

dazinator:
Its debateable.. Some people think providing a single exe makes consumption easier. Nuget.exe as an example, its a self contained exe, a consumer can download it from the nuget website and use it immediately, copy deploy it etc. Its self contained end the end user doesnt see anything they dont care directly about (dependencies). Imho it doesnt make a great deal of difference really. Best person to provide an opinion / context would be @jakeginnivan 

Also no one downloads gitversion.exe as a file download like they may do for nuget.exe. They typically obtain it via nuget or choco so having multiple files rather than a single exe surely doesnt matter at all in that scenario as they dont manage it.

asbjornu:
I don't think merging is worth the trouble. When merging can be done in a cross platform way with .NET Core, I'm all for it, but until it is I say we ditch it. The value it provides over the pain it causes just isn't worth it in my humble opinion.

bigjonroberts:
If you decide to keep single-file merge for the new .net standard build, it looks like CoreRT is the way to do that:
https://github.com/dotnet/corert/blob/master/Documentation/intro-to-corert.md 

As referenced here:
https://github.com/dotnet/corefx/issues/13329
and here:
https://github.com/dotnet/cli/issues/7737

bording:
@bigjonroberts CoreRT is more about creating native applications with an ahead-of-time compiler instead of the normal JIT compliation with IL assemblies. It's not really ready for use yet.

Once it is, that would be an interesting way of shipping a binary that has no dependencies. However, it would be platform-specific, so you'd need to ship a separate exe for Windows, linux, etc.

bording:
@dazinator FYI, with https://github.com/libgit2/libgit2sharp/pull/1571, the story around getting the native library loaded from .NET Core inside an MSBuild task will be drastically simplified. As long as the runtimes folder is in the right place, it should just work!

baurzhan:
@asbjornu running "dotnet gitversion.dll" doesn't update .csproj file with version information. How to achieve same result like "gitversion /updateassemblyinfo" command?

dazinator:
Add the gitversiontask nuget package to your project.

dazinator:
@bording good to hear! :-)

baurzhan:
@dazinator after adding GitVersionTask and running "dotnet build" I get the error "Could not load file or assembly 'Microsoft.Build.Utilities.v4.0, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL'".

dazinator:
What version if gitversiontask nuget package did you add? Was it the same version as described above in the issue description?

dazinator:
@baurzhan - any news?

dazinator:
FYI - I believe its only the failing travis build that is preventing this from being merged.. So if anyone finds value in the PR and wants to see it merged, and has any time to investigate the failing travis build - your help would be very appreciated :-)

Farthom:
I took a quick look at the build scripts for this beast under a Linux environment. (In an attempt to wrestle this failing TravisCI build.) There seems to be two issues at play here. The first issue is choice of build SDK. The cake script calls "DotNetBuild" which on a Linux environment appears to trickle down into XBuild. From what I can tell XBuild is fundamentally incapable of building the new .netcore style projects. Cake also provides "DotNetCoreMSBuild" which uses the official .Netcore SDK to build. I'm no expert on the history of the build systems for this project and I'll admit I never used cake before 1 hour ago but I would expect a build using the official SDK to be desirable. Attempting to use "DotNetCoreMSBuild" works on Windows but leads into issue #2.

Issue 2: Official .NetCore SDK does not support targeting net40 or other "full framework" targets when building on a Linux platform! (Doh!)


So our options appear to be as follows:
1) Drop net40 and transition everything to .netstandardX.X or .netcoreappX.X
2) Give up Linux compilation.
3) Find some other Linux based build system which might work. 

So my real question is: 
Why is Linux build so important? As long as the finished library can RUN on Linux who cares how it was compiled? Perhaps some of the more seasoned contributers may be able to shed more light on why things are done this way. However my "outsider" opinion would be to choose option #2. 

Cheers!

bording:
Is there a way to tell DotNetCoreMSBuild to only build the targets on Travis that can actually be built?

If not, it should be possible to adjust the `TargetFrameworks` when building on linux to exclude the `net40` target and only build the targets that can be built on linux.

One thing to point out is that the not being able to build .NET Framework on linux is a temporary problem. Microsoft is working on making the .NET Framework targeting packs something that can be pulled down via NuGet, so the .NET Core SDK will able to use those to compile assemblies targeting the .NET Framework as well.

So, the `TargetFrameworks` modification could be reversed once that happens.

devlead:
It is possible to build legacy framework on .NET Core

Example below project targets netstandard 2.0, net45, netstandard 1.5, net 40, net 35 and net20. 

https://github.com/LitJSON/litjson/blob/develop/build.cake#L57

dazinator:
@Farthom thank you for sparing some time to investigate.

> Why is Linux build so important? As long as the finished library can RUN on Linux who cares how it was compiled? 

This is a good question.
The only thing I can think of in favour of a linux build is that once we run on linux, there will inevitably be some issue that occurs whilst running on Linux, and some unlucky linux developer with netcore installed will be tasked with fixing it. They will want to clone our code base and build it locally and debug / test any contributions before hopefully submitting a PR. If we don't have a CI build on linux we are basically admitting we don't know how to build our code base on linux and that we don't care about this scenario. I think it would be nice to support this scenario given that we have added support for running on this environment it would show we have thought about it more fully end to end?

gjonespf:
> Why is Linux build so important? As long as the finished library can RUN on Linux who cares how it was compiled?
Agree with the points you've made @dazinator and I guess the importance of dogfooding shouldn't really be understated.
By not building on Linux, I would suggest you're limiting:
1. The CI build setups you can use to build the project
2. The input from developers that can help with code changes (as they'll want to be able to build locally)

If you've decided on CI tools (which I expect you have), and aren't concerned about them disappearing future, I guess 1 is less of an issue. I'd suggest 2 is probably worth considering though.
Can recommend looking at Cake to help with smoothing cross plat build, though that'd be better approached as a separate piece of work.  Not sure how well Cake plays on Travis and Appveyor tho (my experience is with Jenkins).

devlead:
Cake plays just fine with Travis, one consideration AppVeyor now supports Linux. Another option for Linux/Mac builds is Bitrise which has an free tier.

Regardless even if you don't build on Linux, you'll need to test on multiple distributions/Mac. Especially since libgit2 native bits behaves slightly different and has different dependencies on each platform. 

One way to achieve automated integration tests, could be to push artifacts to MyGet and use its web hook functionality to trigger CI tests.

pauldotknopf:
Is there a myget feed that I can get this, as is?

csMACnz:
@pauldotknopf there is the appveyor ci nuget feed that I have used against this pr to test it out: https://ci.appveyor.com/nuget/gitversion-8nigugxjftrw  using the pull request as the version (currently `4.0.0-pullrequest1269-1576`)

arturcic:
@dazinator there is a pull request that fixes the travis build

dazinator:
@arturcic - many thanks will take a look!

dazinator:
@arturcic have merged! Thanks for the PR. Hopefully the travis build show now turn green...

dazinator:
Tada! [travis is fixed](https://travis-ci.org/GitTools/GitVersion/builds/384323143?utm_source=github_status&utm_medium=notification) - cheers @arturcic

dazinator:
@asbjornu @JakeGinnivan 
I think we should consider merging this PR now - I have merge powers - but also some questions:

1. What's the strategy - do we just merge straight to master? Any manual tagging necessary?
2. This PR is missing the automated build / publish steps for the TFS task and the ruby gem packages. After a merge we would have to relase updates to those packages manually until we can follow up with a PR that adds this logic to the cake script.

dazinator:
Also we should consider what docs need to be updated.

thoemmi:
I have changed our projects to use version `4.0.0-pullrequest1269-1580` of `GitVersionTask`, and it works pretty fine. 👍 

However, for each project I get this warning:

> MSBUILD : Task factory warning NMSBT010: There is a mismatch between SDK NuGet version (4.7.0) and the NuGet version the task factory was compiled against (4.6.0). There might occur some exotic errors.

I guess it's related to the issue https://github.com/CometaSolutions/UtilPack/issues/19 of `UtilPack`.

dazinator:
@jakeginnivan
Can you follow these steps on mac to make sure your setup:
https://www.microsoft.com/net/learn/get-started/macos

JakeGinnivan:
Should I just ignore the errors about the main framework missing @dazinator ?

JakeGinnivan:
How feasible is it to completely switch over? Tests and all, currently tests are still full framework so it's not really possible to develop GitVersion on a mac.

I am super out of touch with all this so should I even be bothering trying to get up and running on mac?

kll:
I've been fighting for days trying to get a version of gitversion that can run on the official microsoft/dotnet:2.1-sdk docker image. It still needs a newer version of libgit2sharp / libgit2sharp.nativebinaries than this PR is targeting in order to work. I tried rolling my own based on this PR but hit on the fact that the newer versions of libgit2sharp targets netstandard2.0 (and has for almost a year now) which means they can't be used without also converting GitTools.Core and GitVersion over to netstandard2.0. Based on that fact I'd like the decision not to just target netstandard2.0 now to be reconsidered.

If the decision is made to target netstandard2.0 I'll attempt to do the work and submit a PR. I don't have any experience in the GitTools code bases though so that might be overly optimistic on my part.

dazinator:
@kll Sorry to hear of your docker struggles. I am eager to see this PR merged really, without expanding it any further. It delivers what it set out to do. I think it would be good to get this merged and then add support for the 2.1-sdk docker image as a separate issue.

dazinator:
@JakeGinnivan I am the wrong person to ask about Mac. I thought in theory it should work - can anyone offer any advice to Jake on this?

kll:
I'll start on a new PR based on this one but will keep it distinct so this one can be merged first.

In theory at least re-targeting everything including the test projects to just netstandard2.0/netcoreapp2.0 should allow it to work on a mac.

dazinator:
@kll ok sounds good. Is there anyway you could add netstandard2.0 as an additional target rather than retarget everything to 2.0 as a new baseline? I'd love to keep support for 1.1 if possible

kll:
@dazinator Unfortunately not since LibGit2Sharp only targets netstandard2.0. We are limited by what the libraries we consume target. I got a good start yesterday and I think its a positive change because a lot of the extra complexity that was added in to work around the missing APIs in netstandard1.3 can go back away now since they are present in netstandard2.0.

dazinator:
@kll 

> LibGit2Sharp only targets netstandard2.0

Yeah.. the official release does you are right. This PR though uses a beta package which I think supports `netstandard1.5`. So to add netstandard2.0 support I was thinking something like this:


```

<ItemGroup Condition="'$(TargetFramework)'=='netstandard1.5'">
    // current packages as per this pr.
</ItemGroup>

<ItemGroup Condition="'$(TargetFramework)'=='netstandard2.0'">
    // newer libgit2sharp package targeting netstandard20
</ItemGroup>

```

However that is a lot of build complexity and just an idea. If this PR gets merged I think most people (inlcuding me) would be perfectly fine with moving the baseline to netstandard20 as of a following release :-)

JakeGinnivan:
Thoughts on just bringing GitTools.Core into this repo, I don't think it's being used by anything else properly? May simplify things going forward

petriashev:
Any plans for release?

dazinator:
@jakeginnivan - probs a good idea to bring gittools in if its not being consumed independently anywhere else - but can we get this PR merged first and track that as a seperate issue if necessary?

Also what are your thoughts about this PR - can it be merged now? Or are there any specific things that need to be addressed?

dazinator:
I prefer not to merge my own PRs (just a policy of mine) but in order to get this moving:

Unless there are any objections / concerns raised then in two weeks I will merge this myself. That is unless someone else wants to do the honours :-)

pauldotknopf:
2 weeks? Geesh!

I say, by the power granted to me by Mr. Internet himself, you have until Monday of next week to raise your concerns. Or forever hold your peace!

JeremyCade:
@dazinator just merge, release a alpha. The world shall go forth and break from there.

JakeGinnivan:
I am fine with you merging. Lets just merge, release, and the community can help us straighten it out

Lets get back on the releasing train.

lecoque:
So if I understand correctly in order to use `GitVersionTask` with `dotnet core` projects I have to use appveyor nuget feed and version `4.0.0-pullrequest1269-1535`?

dazinator:
@lecoque that's presently correct, however I have just merged this PR, hopefully an official nuget.org release will occur soon - will update back here once that has happened.

pauldotknopf:
Any update on this?

dazinator:
@pauldotknopf - This needs sorting: #1445

pauldotknopf:
Is this the package we should be using?

https://www.nuget.org/packages/GitVersionCore

asbjornu:
@pauldotknopf: That depends on how you're using GitVersion. If you use it as a library, that's the right reference. If you use it as a command line executable, [GitVersion.CommandLine.DotNetCore](https://www.nuget.org/packages/GitVersion.CommandLine.DotNetCore) is the right one. If you use it as an MSBuild Task, [GitVersionTask](https://www.nuget.org/packages/GitVersionTask) is the right one.

dazinator:
There is also now the dotnet cli global tool which is probably easier to consume if you want the command line usage (on any platform):

install instructions here: https://www.nuget.org/packages/GitVersion.Tool

Issue #1271, opened 2017-07-31
Allow for brach names to start with reserved names


tgbarnett:
The current implementation of GitVer has a specific behavior for branches called _master_
Unfortunately the way that this is implemented it is for branches that start with _master_ which means that _masterfix_ has the same rules applied (though _fixmaster_ would be treated differently.

I think that _masterfix_ and _fixmaster_ should be treated the same (i.e. not the same as _master_)

asbjornu:
Closing as #1273 is merged.

PR #1273, opened 2017-07-31
 Allow branch names starting with master

tgbarnett:
#1271 
_masterfix_ and _fixmaster_ now behave the same way (i.e. in a different
way to _master_, but in the same way as other branches)


asbjornu:
LGTM, thanks!

PR #1279, opened 2017-08-16
Fixes #1278 - ContinuaCI use environment variable not registry.

### LINKED ISSUE #1278: Continua Ci - uses registry - needs fix
Originated from #1269:

> I had to remove ContinuaCI support from NETStandard build - as lack of registry API's.

> Why does it need access to the registry? It should just check if environment variables are registered, same as on Cake:

https://github.com/cake-build/cake/blob/08907d1a5d97b66f58c01ae82506280882dcfacc/src/Cake.Common/Build/ContinuaCI/ContinuaCIProvider.cs#L62

dazinator:
Fixes #1278.


dazinator:
@GeertvanHorrik - does this look better?

GeertvanHorrik:
Yup, looks good to me 👍

dazinator:
Ok merging, cheers

PR #1280, opened 2017-08-17
Closes #1274 - Removed system.web (replaced databinder.eval)

### LINKED ISSUE #1274: Remove System.Web dependency
System.Web is used for it's databinder, in order to format version strings to a template. 

We need to get rid of it. It's hampering our ability (#1175) to build for netstandard.

dazinator:
Replaced databinder.eval and dropped system.web as a dependency. System.Web should never have been adopted imho :-)
This will unblock my open PR to add netstandard support.
Minor correction in docs.


dazinator:
All of the following tests cover this already: 
image

dazinator:
Merging..

PR #1287, opened 2017-09-05
Improve shallow clone exception

### LINKED ISSUE #1043: LibGit2Sharp.NotFoundException: Object not found - no match for id
Any idea why I would get this error and is there anyway to ignore the error? The error is LibGit2Sharp.NotFoundException: Object not found - no match for id

asbjornu:
Improve upon the `LibGit2Sharp.NotFoundException` thrown when a commit ID can't be found due to a shallow clone by adding instructions to do `git fetch --unshallow`. Fixes #1043.

### DIFF
--- a/src/GitVersionCore/GitVersionException.cs
+++ b/src/GitVersionCore/GitVersionException.cs
@@ -1,0 +1,31 @@
+﻿namespace GitVersion
+{
+    using System;
+    using System.Runtime.Serialization;
+
+    [Serializable]
+    public class GitVersionException : ApplicationException
+    {
+        public GitVersionException()
+        {
+        }
+
+
+        public GitVersionException(string message)
+            : base(message)
+        {
+        }
+
+
+        public GitVersionException(string message, Exception innerException)
+            : base(message, innerException)
+        {
+        }
+
+
+        protected GitVersionException(SerializationInfo info, StreamingContext context)
+            : base(info, context)
+        {
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


src/GitVersionCore/GitVersionException.cs
```diff
@@ -0,0 +1,42 @@
...
+
+#endregion
+
+namespace GitVersion
+{
+    using System;
+    using System.Runtime.Serialization;
+
+    [Serializable]
+    public class GitVersionException : ApplicationException
```
dazinator: Would you consider deriving from `GitToolsException` instead : `https://github.com/GitTools/GitTools.Core/blob/11fa325b9f7afa0f9d8abf12d260210f82a29e15/src/GitTools.Core/Exceptions/GitToolsException.cs` - It will make it a bit easier to get this building under netstandard1.3 where `ApplicationException` isn't avialable, and it will also make exceptions accross the tools more consistent which I think was the purpose behind GitToolsException.

asbjornu: @dazinator: Sure. Done.

src/GitVersionCore/GitVersionException.cs
```diff
@@ -0,0 +1,32 @@
+﻿#region License
+
+// --------------------------------------------------
```
gep13: @asbjornu where is this license coming from?

asbjornu: Good question. It's from ReSharper, but I thought GitVersion's DotSettings should override it? Either way, it's gone.

PR #1291, opened 2017-09-07
Improve ExecuteCore.ExecuteInternal when HEAD is not pointing at the desired branch

### LINKED ISSUE #1289: Incorrect calculated version on new branch without any commits
Here's an issue I am experiencing when building on AppVeyor but I am unable to reproduce locally.

Let's say the current build number is `0.4.2`, let's also say that every push triggers a build on AppVeyor and finally let's say that a step in the build script uses GitVersion to calculate the build version.

**Repro steps**
- Create a new branch, say `release/0.4.3`
- Push to GitHub
- Check your AppVeyor build. The calculated version is 0.5.0 which is incorrect
- Here's an example that demonstrates this [incorrect result](https://ci.appveyor.com/project/cakecontrib/cake-email/build/0.5.0-alpha.10.build.61#L118)

**Steps to restore expected behavior**
- Commit a change to the new branch
- Push to GitHub
- Check your AppVeyor build. The calculated version is 0.4.3 which is correct
- Here's an example that demonstrates this [result](https://ci.appveyor.com/project/cakecontrib/cake-email/build/0.4.3-rc.1.build.62#L126).

I'm guessing GitVersion is confused because the latest commit on the new branch is the same as the latest on the develop branch, but that's just a guess on my part.

I have confirmed this incorrect behavior with GitVersion 3.6.2 I'll double check and confirm if 4.0 beta also suffers from this issue.

Jericho:
This resolves #1289

### DIFF
--- a/src/GitVersionCore/GitVersionContext.cs
+++ b/src/GitVersionCore/GitVersionContext.cs
@@ -9,11 +9,6 @@
     /// </summary>
     public class GitVersionContext
     {
-        public GitVersionContext(IRepository repository, Config configuration, bool isForTrackingBranchOnly = true, string commitId = null)
-            : this(repository, repository.Head, configuration, isForTrackingBranchOnly, commitId)
-        {
-        }
-
         public GitVersionContext(IRepository repository, string targetBranch, Config configuration, bool onlyEvaluateTrackedBranches = true, string commitId = null)
              : this(repository, repository.Branches.SingleOrDefault(b => b.CanonicalName == targetBranch || b.FriendlyName == targetBranch) ?? repository.Head, configuration, onlyEvaluateTrackedBranches, commitId)
         {
--- a/src/GitVersionCore/GitVersionContext.cs
+++ b/src/GitVersionCore/GitVersionContext.cs
@@ -10,7 +10,9 @@
     public class GitVersionContext
     {
         public GitVersionContext(IRepository repository, string targetBranch, Config configuration, bool onlyEvaluateTrackedBranches = true, string commitId = null)
-             : this(repository, repository.Branches.SingleOrDefault(b => b.CanonicalName == targetBranch || b.FriendlyName == targetBranch) ?? repository.Head, configuration, onlyEvaluateTrackedBranches, commitId)
+             : this(repository,
+                   repository.Branches.SingleOrDefault(b => b.CanonicalName == targetBranch || b.FriendlyName == targetBranch) ?? repository.Head,
+                   configuration, onlyEvaluateTrackedBranches, commitId)
         {
         }
 
--- a/src/GitVersionCore/GitVersionContext.cs
+++ b/src/GitVersionCore/GitVersionContext.cs
@@ -10,9 +10,7 @@
     public class GitVersionContext
     {
         public GitVersionContext(IRepository repository, string targetBranch, Config configuration, bool onlyEvaluateTrackedBranches = true, string commitId = null)
-             : this(repository,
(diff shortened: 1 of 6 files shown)


pascalberger:
@Jericho Thanks for the PR! Would it be possible that you can also create a test case similar to the scenario with your demo repo (branch without a commit on it) and repo checked out to specific commit?

src/GitVersionCore/GitVersionContext.cs
```diff
@@ -9,8 +9,8 @@
     /// </summary>
     public class GitVersionContext
     {
-        public GitVersionContext(IRepository repository, Config configuration, bool isForTrackingBranchOnly = true, string commitId = null)
-            : this(repository, repository.Head, configuration, isForTrackingBranchOnly, commitId)
+        public GitVersionContext(IRepository repository, string targetBranch, Config configuration, bool onlyEvaluateTrackedBranches = true, string commitId = null)
+             : this(repository, repository.Branches.SingleOrDefault(b => b.CanonicalName == targetBranch || b.FriendlyName == targetBranch) ?? repository.Head, configuration, onlyEvaluateTrackedBranches, commitId)
```
asbjornu: Can you please reduce the number of statements per line here? It's a bit hard to read with so much going on in one line of code. 😃

Jericho:
@pascalberger I already created a file called `BranchWithoutCommitScenario.cs` where I have a test with this exact scenario: 

- create a commit
- create a new branch (the commit is now on two branches)
- checkout the specific commit
- assert that version has been properly calculated

I was under the impression this was the standard way of testing scenarios in GitVersion. For instance I see that's the way other scenarios are tested such as: develop scenarios, mainline scenarios, pull request scenarios, etc. But if I failed to follow some other standard, please give some guidelines and I'll be happy to adjust how I test my scenario.

Speaking of this scenario, I mentioned it before but I want to repeat it because it's been bothering me: if you remove the fix I wrote and run the unit test for my scenario, you would hope to see the version being miscalculated like I observed when using GitVersion on AppVeyor. But that's not the case. The testing fixture throws an exception because it can't handle a detached head in a git repo. I don't know why the behavior is different and, to be honest, I don't have the appetite to investigate but I want everyone to be aware.

src/GitVersionCore/GitVersionContext.cs
```diff
@@ -10,7 +10,9 @@
     public class GitVersionContext
     {
         public GitVersionContext(IRepository repository, string targetBranch, Config configuration, bool onlyEvaluateTrackedBranches = true, string commitId = null)
-             : this(repository, repository.Branches.SingleOrDefault(b => b.CanonicalName == targetBranch || b.FriendlyName == targetBranch) ?? repository.Head, configuration, onlyEvaluateTrackedBranches, commitId)
+             : this(repository,
+                   repository.Branches.SingleOrDefault(b => b.CanonicalName == targetBranch || b.FriendlyName == targetBranch) ?? repository.Head,
+                   configuration, onlyEvaluateTrackedBranches, commitId)
```
asbjornu: I only now discovered that this all happens within a constructor. Sorry for not catching that earlier. Could you please add a static method that wraps `repository.Branches.SingleOrDefault(b => b.CanonicalName == targetBranch || b.FriendlyName == targetBranch) ?? repository.Head` and invoke that method from the constructor?

Jericho:
@asbjornu done. Also, instead of just a one-liner, I broke it up into multiple line which allowed me to add comments to explain each step in the logic.

asbjornu:
@Jericho: Beautiful! As soon as the tests are green, this will be merged. Thank you so much for your persistence and tireless effort! 😄 👍 ❤️

pascalberger:
@Jericho Thanks for getting this done and your persistence 👍

PR #1293, opened 2017-09-13
Ensure that PreReleaseNumber is always a number

pitermarx:
Because of the [JsonOutputFormatter](https://github.com/GitTools/GitVersion/blob/c5ae891d56a8d6d52ac4f756a6727a41071332b1/src/GitVersionCore/OutputFormatters/JsonOutputFormatter.cs), the json output can either be "PreReleaseNumber": number or "PreReleaseNumber": null
this can have impact on desserialization as found on [Cake](https://github.com/cake-build/cake/issues/1798)

### DIFF
--- a/src/GitVersionCore.Tests/Approved/VariableProviderTests.ProvidesVariablesInContinuousDeliveryModeForStable.approved.txt
+++ b/src/GitVersionCore.Tests/Approved/VariableProviderTests.ProvidesVariablesInContinuousDeliveryModeForStable.approved.txt
@@ -5,7 +5,7 @@
   "PreReleaseTag":"",
   "PreReleaseTagWithDash":"",
   "PreReleaseLabel":"",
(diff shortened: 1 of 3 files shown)


matkoch:
@pitermarx  I don't think this should be done. There is a clear distinction between _null_ and _0_.

asbjornu:
@pitermarx: I agree with @matkoch. I interpret a `PreReleaseNumber` set to `0` to mean that the release is not stable and at version `0`. A `PreReleaseNumber` set to `null` indicates that the release is stable. To represent this property in C#, I would use `Nullable<int>`, so both states can be captured.

pitermarx:
I agree. Pls reject this PR. I just wanted to raise attention to this

--
Pedro Marques

No dia 15/09/2017, às 12:59, Asbjørn Ulsberg <notifications@github.com> escreveu:

PR #1297, opened 2017-09-25
honour increment config value for Mainline

### LINKED ISSUE #1295: [Question] Is this Version flow possible?
I've been playing around with GitVersion trying to get the following version calculation flow to work

```
master
  * (HEAD, calulated version: 1.1.0)
  |
  * (tag: release/v1.0.0, calculated version: 1.0.0)
  |
  * (calulated version: 0.4.1)
  |\
  | \
  |  * (branch: feature/bar, calculated version: 0.4.1-bar)
  |  |
  |  * (branch: feature/bar, +semver: patch, calculated version: 0.4.1-bar)
  | /
  |/
  * (calulated version: 0.4.0)
  |\
  | \
  |  * (branch: feature/foo, calculated version: 0.4.0-foo)
  |  |
  |  * (branch: feature/foo, calculated version: 0.4.0-foo)
  | /
  |/
  * (calculated version: 0.3.0)
  |
  * (calculated version: 0.2.0)
  |
  * (initial commit, calculated version: 0.1.0)
```

but haven't had much luck. Is it possible with v3 or v4?

It seems like I want to use `Mainline` but that seems to only increment `Patch` and I want it to increment `Minor`

JonCubed:
Adds support to Mainline mode to honour `Increment` value in config. Previously if not a tag or commit message bump would default to `Patch` increment.

Includes:
* Added new method to `IncrementStrategyFinder` to work out default Increment Strategy for a branch
* Updated `MainlineDevelopmentMode` to use new method instead of defaulting to `Patch`
* Added unit test to cover setting increment in config

Fixes #1295

### DIFF
--- a/src/GitVersionCore.Tests/IntegrationTests/MainlineDevelopmentMode.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/MainlineDevelopmentMode.cs
@@ -339,6 +339,104 @@ public class MainlineDevelopmentMode
             fixture.AssertFullSemver(currentConfig, "1.1.0");
         }
     }
+
+    [Test]
+    public void VerifyIncrementConfigIsHonoured()
+    {
+        var minorIncrementConfig = new Config
+        {
+            VersioningMode = VersioningMode.Mainline,
+            Increment = IncrementStrategy.Minor,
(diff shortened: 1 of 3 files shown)


asbjornu:
LGTM! Thanks!

PR #1324, opened 2017-11-16
Add VS extension updater

twsouthwick:
This adds a task to update VSIX manifest files if they are present in the project. This is done by hooking itself such that it runs right before CreateVsixPackage if it exists. The update is done on the intermediate manifest, so the original copy is not changed and thus the version update is transparent to the user.

### DIFF
--- a/src/GitVersionTask/GitVersionTask.csproj
+++ b/src/GitVersionTask/GitVersionTask.csproj
@@ -66,6 +66,7 @@
   <ItemGroup>
     <Compile Include="BuildLogger.cs" />
     <Compile Include="GenerateGitVersionInformation.cs" />
+    <Compile Include="UpdateVsixManifest.cs" />
     <Compile Include="GitVersionTaskBase.cs" />
     <Compile Include="WriteVersionInfoToBuildLog.cs" />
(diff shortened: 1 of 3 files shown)


JakeGinnivan:
I am in two minds about this, @gep13 @asbjornu feel free to disagree, but I think we should have less of these specific updates.

GitVersion should focus on generating the version number, in the Task it should be a msbuild variable which can be used by tasks like this, but they probably shouldn't live inside this repo.

This is an extension to the discussion if should we update .net core versions etc. Thoughts? Should be maybe create another repo with a heap of tasks and helpers to update various types of projects?

asbjornu:
I agree that the different outputs should be separated from the core and from each other.

gep13:
@JakeGinnivan yes, I totally agree with you.  GitVersion should be the tool to assert the version number.  There are plenty of other tools, like MagicChunks, that can then be used to take that version number and update the required files.

twsouthwick:
That sounds fine. I wanted to look at contributing this here first for easier discoverability if there was interest, but will package it elsewhere. Thanks for the work with GitVersion!

PR #1340, opened 2017-12-06
Fix addin/tool versions

gep13:
- This will make the builds more reliable and repeatable

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -1,5 +1,5 @@
-#tool "nuget:?package=NUnit.ConsoleRunner"
-#tool "nuget:?package=GitReleaseNotes"
+#tool "nuget:https://www.nuget.org/api/v2?package=NUnit.ConsoleRunner&version=3.7.0"
+#tool "nuget:https://www.nuget.org/api/v2?package=GitReleaseNotes&version=0.7.0"
 
 var target = Argument("target", "Default");
 var configuration = Argument("configuration", "Release");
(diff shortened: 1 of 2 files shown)


gep13:
@JakeGinnivan can you take a look at this?

gep13:
@pascalberger yeah, that sounds like another bug that needs to be resolved.  I can try to take a look at that later, but can't guarantee it :-(

JakeGinnivan:
Looks good to me, but we can't release beta13/14 yet. I have been testing and had scenarios where it isn't picking up merge messages, I think we have a decent regression somewhere

gep13:
@JakeGinnivan ouch! Is this showing up in the tests, or just from your own testing?

JakeGinnivan:
Just from my own testing. 

Also, there is a branch config inheritance issue.. See https://travis-ci.org/GitTools/GitVersion/jobs/312274320 - got stuck in a loop..

May take me a day or two, but I will try to add the tests and ensure we don't have those regressions.

It would be handy if everyone could build and do some basic testing of master on their own code bases to see if there are issues

JakeGinnivan:
Hopefully this should go green now, then can merge and we should be good to release beta.14

gep13:
@JakeGinnivan looks like there is still an issue with the uploading of artifacts.

JakeGinnivan:
Odd that it's the branch build failing but all others succeed. I recon we just merge. It's probably a gem precondition not creating it for that build

PR #1342, opened 2017-12-07
Fix/release issues

JakeGinnivan:
Some of these fixes may not be needed because of other fixes, but trying to stamp out any chance of infinite config lookups..

### DIFF
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -15,7 +15,7 @@ namespace GitVersion
         /// </summary>
         public static BranchConfig GetBranchConfiguration(GitVersionContext context, Branch targetBranch, IList<Branch> excludedInheritBranches = null)
         {
-            var matchingBranches = context.FullConfiguration.GetConfigForBranch(targetBranch.FriendlyName);
+            var matchingBranches = context.FullConfiguration.GetConfigForBranch(targetBranch.NameWithoutRemote());
             
             if (matchingBranches == null)
             {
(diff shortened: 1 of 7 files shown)


asbjornu:
I'd rather have too many than too few checks in place to avoid infinite loops, so this looks good to me.

JakeGinnivan:
Ok, looked into this a bit harder. Am now actually excluding branches which have no config or their branch config is inherit. This should mean that we don't resolve a branch with increment of inherit.

I think this is actually the real fix for now.

src/GitVersionCore/BranchConfigurationCalculator.cs
```diff
@@ -27,11 +27,20 @@ public static BranchConfig GetBranchConfiguration(GitVersionContext context, Bra
                 ConfigurationProvider.ApplyBranchDefaults(context.FullConfiguration, matchingBranches, "", new List<string>());
             }
 
-            return matchingBranches.Increment == IncrementStrategy.Inherit ?
-                InheritBranchConfiguration(context, targetBranch, matchingBranches, excludedInheritBranches) :
-                matchingBranches;
+            if (matchingBranches.Increment == IncrementStrategy.Inherit)
```
JakeGinnivan: This shouldn't happen now because we filter the conditions which would cause this code to be executed. But still, I think it's worth keeping in just in case

src/GitVersionCore/BranchConfigurationCalculator.cs
```diff
@@ -50,17 +59,17 @@ static BranchConfig InheritBranchConfiguration(GitVersionContext context, Branch
                 {
                     excludedInheritBranches = repository.Branches.Where(b =>
                     {
-                        var branchConfig = config.GetConfigForBranch(b.FriendlyName);
+                        var branchConfig = config.GetConfigForBranch(b.NameWithoutRemote());
```
JakeGinnivan: This I think is the real fix. This will now exclude local and remote branches which do not have config, or the config is Inherit.

Previously we only excluded branches which had config and it was Inherit, but our infinite loops were caused by branches without config..

PR #1344, opened 2017-12-08
Docker improvements

neilinger:
This will be another addition to #1218.

As i entered via https://opensourcefriday.com/ so please be gentle if I missed sth.
I don't have experience with the Software on this project itself, but know a bit about Docker.
Hope this will help/improve sth.


JakeGinnivan:
Awesome, thanks heaps!

Will get this released soon.

PR #1351, opened 2018-01-05
Make sure GetVersion is called for project references

bording:
Adding this property ensures that the GetVersion target will be called in the right place when NuGet's Pack target is calculating the version for project references.

This is using the new feature in https://github.com/NuGet/NuGet.Client/pull/1915 to do this, so you won't see it actually doing anything until an updated version of NuGet is released.


bording:
Can we get this merged? We need to this be able to use the new NuGet feature coming soon in the VS 2017 15.6 update.

bording:
FYI Visual Studio 2017 15.6 is out now, so we really need to get this merged in. Is anyone around anymore?

erik-am:
Thank you @bording! I tried it out in my project and it works great. 😄 
I hope this can be merged soon.

PR #1365, opened 2018-01-24
switch GetMainlineTip from using branch config key to using the regex

### LINKED ISSUE #1364: Finding mainline tip uses key instead of regex from branch config
When finding possible tips for the mainline branch the key is used instead of the regex defined in the branch configuration.

eg. with the default regex for master branch configuration of `master$` it will still match using the key (`master`) as shown here:

`INFO [01/25/18 8:36:43:86] Found possible mainline branches: feature/masterdetail, master`
`INFO [01/25/18 8:36:43:86] Mainline for current branch is feature/masterdetail`

tvoyle:
When finding possible tips for the mainline branch the key is used instead of the regex defined in the branch configuration. This PR corrects that problem. Fixes #1364.


asbjornu:
Thanks for this, but can you please explain the rationale behind the change?

asbjornu:
Thanks!

Issue #1355, opened 2018-01-16
When using mode: Mainline, the develop branch is set to mode: ContinuousDeployment


FamularoA:
Hello, 

I have to get an atomic version in each branch in Major.Minor.Patch without the patch tail.

System:
TFS 2017 Build, with Gitversion 4, Beta 12

With the current configuration I get for the develop branch following versions.
2.4.1-alpha0004
2.4.1-alpha0005
2.4.1-alpha0006

What I need is this, increment the Patch with each Pull-Request:
2.4.1-alpha0004
2.4.2-alpha0005
2.4.3-alpha0006

Pumping with +semver :patch do not the job. 

This is my configuration:

assembly-versioning-scheme: MajorMinorPatch
assembly-file-versioning-scheme: MajorMinorPatch
mode: Mainline
tag-prefix: '[vV]'
continuous-delivery-fallback-tag: ''
legacy-semver-padding: 4
build-metadata-padding: 4
commits-since-version-source-padding: 4
commit-message-incrementing: MergeMessageOnly
increment: Patch
branches:
  master:
    tag: ''
    regex: master
    increment: Patch
  develop:
    tag: alpha
    regex: develop 
    increment: Patch
  release:
    tag: beta
    regex: release?[/-]
    increment: Patch
  feature:
    tag: feature
    tag-number-pattern: '[//](?<number>\d{5})'    
    regex: feature?[/-]        
    increment: Patch
  hotfix:
    tag: bug
    regex: hotfix?[/-]  
    increment: Patch   
  bug:
    tag: bug
    tag-number-pattern: '[//](?<number>\d{5})'     
    regex: bug?[/-]    
    increment: Patch   
  support:
    tag: ''
    regex: support[/-]    
    increment: Patch
ignore:
  sha: []

The configuration shown with gitversion /showconfig is:

assembly-versioning-scheme: MajorMinorPatch
assembly-file-versioning-scheme: MajorMinorPatch
mode: Mainline
tag-prefix: '[vV]'
continuous-delivery-fallback-tag: ''
major-version-bump-message: '\+semver:\s?(breaking|major)'
minor-version-bump-message: '\+semver:\s?(feature|minor)'
patch-version-bump-message: '\+semver:\s?(fix|patch)'
no-bump-message: '\+semver:\s?(none|skip)'
legacy-semver-padding: 4
build-metadata-padding: 4
commits-since-version-source-padding: 4
commit-message-incrementing: Enabled
branches:
  master:
[...]

PR #1371, opened 2018-02-15
#1355 Fix develop branch config in mainline mode

Dobriy33:
Issue #1355 .
Develop branch is always set in the ContinuousDeployment mode in config and even when we use the Mainline mode. Consequently, we can not get the expected version in the develop branch.
If we try to override branch mode in config, we get an error: "Mainline mode only works at the repository level, a single branch cannot be put into mainline mode..."
By default, all other branches inherit the global mode.

### DIFF
--- a/src/GitVersionCore/Configuration/ConfigurationProvider.cs
+++ b/src/GitVersionCore/Configuration/ConfigurationProvider.cs
@@ -109,7 +109,7 @@ If the docs do not help you decide on the mode open an issue to discuss what you
                 new List<string>(),
                 defaultTag: "alpha",
                 defaultIncrementStrategy: IncrementStrategy.Minor,
-                defaultVersioningMode: VersioningMode.ContinuousDeployment,
+                defaultVersioningMode: config.VersioningMode == VersioningMode.Mainline? VersioningMode.Mainline : VersioningMode.ContinuousDeployment,
                 defaultTrackMergeTarget: true,
                 tracksReleaseBranches: true);
             ApplyBranchDefaults(config,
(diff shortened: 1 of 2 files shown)


Dobriy33:
@asbjornu can you review this?

src/GitVersionCore/Configuration/ConfigurationProvider.cs
```diff
@@ -109,7 +109,7 @@ public static void ApplyDefaultsTo(Config config)
                 new List<string>(),
                 defaultTag: "alpha",
                 defaultIncrementStrategy: IncrementStrategy.Minor,
-                defaultVersioningMode: VersioningMode.ContinuousDeployment,
+                defaultVersioningMode: config.VersioningMode == VersioningMode.Mainline? VersioningMode.Mainline : VersioningMode.ContinuousDeployment,
```
asbjornu: Why can't `defaultVersioningMode` always be set to `config.VersioningMode`?

Dobriy33: Because documentation says that the develop branch is set to ContinuousDeployment mode by default. And all tests require develop branch in ContiniousDeployment mode. But in mainline mode we get wrong behaviour.

asbjornu: Aha, that makes sense.

asbjornu:
Can you please do a rebase of this branch on top of the latest `HEAD` of `master`?

Dobriy33:
Done.

asbjornu:
Excellent, thanks!

PR #1375, opened 2018-02-19
Support for bitbucket pull request merge message

Dobriy33:
Hi, I detected that merge message like "Merge pull request #xxx in BIA/project_Name from feature/branch_Name to develop" not supported. This message generates by bitbucket.
I added new regex for support this.


Dobriy33:
@asbjornu can you review this?

asbjornu:
This is a nice fix that we can have in right away, but it would be nice if this regex was configurable, yes. Thanks!

PR #1384, opened 2018-03-12
Parallelize Unit Tests

jkingry:
Seems like at least most NUnit fixtures can be run in parallel

Added assembly level attributes so that all fixtures run in parallel by default, can be disabled on a per-fixture basis. My tun all tests time went from 133 seconds to 67 seconds, 50% faster


src/GitVersionCore.Tests/AssemblyAttributes.cs
```diff
@@ -0,0 +1,3 @@
+﻿using NUnit.Framework;
+
+//[assembly: Parallelizable(ParallelScope.Fixtures)]
```
asbjornu: Did you mean to comment this?

src/GitVersionTask.Tests/AssemblyAttributes.cs
```diff
@@ -0,0 +1,3 @@
+﻿using NUnit.Framework;
+
+//[assembly: Parallelizable(ParallelScope.Fixtures)]
```
asbjornu: And this?

src/GitVersionExe.Tests/AssemblyAttributes.cs
```diff
@@ -0,0 +1,3 @@
+﻿using NUnit.Framework;
+
+//[assembly: Parallelizable(ParallelScope.Fixtures)]
```
asbjornu: And this?

jkingry:
@asbjornu I messed up and committed the code I was running to double check the timing without the attributes.
Fixed that, and used `AssemblyInfo.cs` as the file for assembly attributes as I see that is what's being done in other projects in the solution.

asbjornu:
@jkingry: I see. Unfortunately, bringing those back broke the build. Could you take a look at it, please?

jkingry:
Yeah. I was afraid of that. Will look into it.

Issue #1366, opened 2018-01-26
Montonically increasing AssemblyFileVersion to satisfy the MSI 


ruhullahshah:
We use GitFlow as our branching model and GitVersion 3 for versioning our Assemblies. 

The versioning scheme in use is:
assembly-file-versioning-scheme: MajorMinorPatchTag.

GitVersion resets the Tag in beta releases at times and MSI requirement of strictly increasing FileVersion is violated. 

Is there a solution existing for this issue? In case it is not, how about reading the Tag from environment if  presented with a special syntax such as {$TAG}? We can then use the build number which is strictly increasing as the Tag

Issue https://github.com/GitTools/GitVersion/issues/1157 has a similar requirement and so does Issue https://github.com/GitTools/GitVersion/issues/648

asbjornu:
Yes, I think this is a duplicate of #1157. We would love a PR solving this.

ruhullahshah:
Fixed with the https://github.com/GitTools/GitVersion/pull/1385

Issue #1157, opened 2017-02-20
Identify multiple builds against same commit/tag/branch on Team Foundation Build (vNext)


ralphhendriks:
Dear all,

I am trying to figure out the best practice to distinguish between multiple builds against the same commit/tag/branch.  My current project is on TFS (on premise). TFS (as well VSTS) stores the build configuration separately from the source repository.  Hence, multiple builds against the same commit/tag/branch could result in different artifacts.  I would like the build metadata to reflect this.

Before using GitVersion we have solved this by using the revision counter from the TFS Build.  The default mechanism of TFS determines the build number before any build task is invoked.  There is a build number replacement token `$(Rev:.r)` that can be used to make the build number unique when nothing else changed ([docs](https://www.visualstudio.com/en-us/docs/build/define/general#tokens)).  We used to rely on the build number, and parse the SemVer and build metadata from it.

As explained in the [docs](http://gitversion.readthedocs.io/en/latest/build-server-support/build-server/tfs-build-vnext/#using-gitversion-variables-in-build-name), GitVersion doesn't play nice with the revision number replacement token on the build number because of delayed expansion in the GitVersion TFS task.  That leaves the problem that multiple builds against the same commit/tag/branch will result in the same output of GitVersion.

I am wondering how others cope with this issue.  I am willing to contribute to the docs if there is a clear consensus on how to handle this.  If there isn't, then adding something to the GitVersion output that is sufficiently unique and sortable (i.e. a timestamp) might be a way forward.

I am interested in hearing you opinions.

JakeGinnivan:
I think the ideal solution is to allow GitVersion to take some sort of build_number or revision parameter which can be used, surfaced in variables and in the build metadata. That way you can pass in an external variable into GitVersion and it will affect the build number.

Would love a PR for this :)

JakeGinnivan:
The discussion at https://github.com/GitTools/GitVersion/issues/648 raised a similar issue. Going to close it in favor of this. I am now more keen on the idea than I was then

asbjornu:
Where would the build number be used? If it's in any of the `Assembly*Version` attributes, I think the route of adding support for environment variables and adding a `-format` version for the ones that currently only have `-scheme` variants in the configuration might be a good solution (as mentioned in https://github.com/GitTools/GitVersion/issues/1145#issuecomment-372080318 and https://github.com/GitTools/GitVersion/issues/1367#issuecomment-372081440).

I'm envisioning something like this:

```yaml
[...]

PR #1385, opened 2018-03-13
Fix issue Montonically increasing AssemblyFileVersion to satisfy the …

ruhullahshah:
Fix for issue https://github.com/GitTools/GitVersion/issues/1366

Implementation methodology same as suggested by @jbaehr in https://github.com/GitTools/GitVersion/issues/1157

### DIFF
--- a/src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
+++ b/src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
@@ -1,0 +1,75 @@
+﻿using System;
+
+using GitVersion;
+using NUnit.Framework;
+
+namespace GitVersionCore.Tests
+{
+    [TestFixture]
+
+    public class StringFormatWithExtensionTests
+    {
+        [Test]
+        public void FormatWith_NoTokens()
+        {
+            var propertyObject = new { };
+            var target = "Some String without tokens";
+            var expected = target;
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_SingleSimpleToken()
+        {
+            var propertyObject = new { SomeProperty = "SomeValue" };
+            var target = "{SomeProperty}";
+            var expected = "SomeValue";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_MultipleTokensAndVerbatimText()
+        {
+            var propertyObject = new { SomeProperty = "SomeValue", AnotherProperty = "Other Value" };
+            var target = "{SomeProperty} some text {AnotherProperty}";
+            var expected = "SomeValue some text Other Value";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_EnvVarToken()
+        {
+            Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR", "Env Var Value");
+            var propertyObject = new { };
+            var target = "{$GIT_VERSION_TEST_VAR}";
+            var expected = "Env Var Value";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_EnvVarTokenWithFallback()
+        {
+            Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR", "Env Var Value");
+            var propertyObject = new { };
+            var target = "{$GIT_VERSION_TEST_VAR??fallback}";
+            var expected = "Env Var Value";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_UnsetEnvVarTokenWithFallback()
+        {
+            Environment.SetEnvironmentVariable("GIT_VERSION_UNSET_TEST_VAR", null);
+            var propertyObject = new { };
+            var target = "{$GIT_VERSION_UNSET_TEST_VAR??fallback}";
+            var expected = "fallback";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+    }
+}
--- a/src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
+++ b/src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
@@ -55,7 +55,7 @@ namespace GitVersionCore.Tests
         {
             Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR", "Env Var Value");
             var propertyObject = new { };
-            var target = "{env:GIT_VERSION_TEST_VAR??fallback}";
+            var target = "{env:GIT_VERSION_TEST_VAR ?? fallback}";
             var expected = "Env Var Value";
             var actual = target.FormatWith(propertyObject);
             Assert.AreEqual(expected, actual);
@@ -66,7 +66,7 @@ namespace GitVersionCore.Tests
         {
             Environment.SetEnvironmentVariable("GIT_VERSION_UNSET_TEST_VAR", null);
             var propertyObject = new { };
-            var target = "{env:GIT_VERSION_UNSET_TEST_VAR??fallback}";
+            var target = "{env:GIT_VERSION_UNSET_TEST_VAR ?? fallback}";
             var expected = "fallback";
             var actual = target.FormatWith(propertyObject);
             Assert.AreEqual(expected, actual);
@@ -89,7 +89,7 @@ namespace GitVersionCore.Tests
         {
             var propertyObject = new { };
             //Test the greediness of the regex in matching env: char
-            var target = "{env:env:GIT_VERSION_TEST_VAR_1} and {env:DUMMY_VAR??fallback}";
+            var target = "{env:env:GIT_VERSION_TEST_VAR_1} and {env:DUMMY_VAR ?? fallback}";
             var expected = "{env:env:GIT_VERSION_TEST_VAR_1} and fallback";
             var actual = target.FormatWith(propertyObject);
             Assert.AreEqual(expected, actual);
@@ -100,7 +100,7 @@ namespace GitVersionCore.Tests
         {
             var propertyObject = new { };
             //Test the greediness of the regex in matching env: and ?? chars
-            var target = "{env:env:GIT_VERSION_TEST_VAR_1} and {env:DUMMY_VAR???fallback}";
+            var target = "{env:env:GIT_VERSION_TEST_VAR_1} and {env:DUMMY_VAR ??? fallback}";
             var actual = target.FormatWith(propertyObject);
             Assert.AreEqual(target, actual);
         }
@@ -111,9 +111,19 @@ namespace GitVersionCore.Tests
             Environment.SetEnvironmentVariable("DUMMY_ENV_VAR", "Dummy-Val");
             var propertyObject = new { };
             //Test the sanity of the regex when there is a grammar mismatch
-            var target = "{en:DUMMY_ENV_VAR} and {env:DUMMY_ENV_VAR?fallback}";
+            var target = "{en:DUMMY_ENV_VAR} and {env:DUMMY_ENV_VAR??fallback}";
             var actual = target.FormatWith(propertyObject);
             Assert.AreEqual(target, actual);
+        }
+
+        [Test]
+        public void FormatWIth_NullPropagationWithMultipleSpaces()
+        {
+            var propertyObject = new { SomeProperty = "Some Value" };
+            var target = "{SomeProperty} and {env:DUMMY_ENV_VAR  ??  fallback}";
+            var expected = "Some Value and fallback";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
         }
     }
 }
--- a/src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
+++ b/src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
@@ -71,5 +71,48 @@ namespace GitVersionCore.Tests
             var actual = target.FormatWith(propertyObject);
             Assert.AreEqual(expected, actual);
         }
+
+        [Test]
+        public void FormatWith_MultipleEnvVars()
+        {
+            Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR_1", "Val-1");
+            Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR_2", "Val-2");
+            var propertyObject = new { };
+            var target = "{$GIT_VERSION_TEST_VAR_1} and {$GIT_VERSION_TEST_VAR_2}";
+            var expected = "Val-1 and Val-2";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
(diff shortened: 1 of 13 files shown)


jbaehr:
Thanks a lot for your work. I gave it a try and it works like a charm (just remember to call `GitVersion.exe` with `-nocache` when playing around ;-) ).

I would love a comment from the maintainers whether this approach has chances to get merged. @asbjornu, you already gave thumbs up on my comment in #1157... Please tell us if you request further changes to the implementation, format string syntax, or the like.

ruhullahshah:
@jbaehr You are welcome and thanks for testing out the changes :)

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_EnvVarToken()
+        {
+            Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR", "Env Var Value");
+            var propertyObject = new { };
+            var target = "{$GIT_VERSION_TEST_VAR}";
```
asbjornu: To avoid collisions, I would prefer if we could use a more explicit prefix here, like `env:`.

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            var expected = "SomeValue";
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_MultipleTokensAndVerbatimText()
+        {
+            var propertyObject = new { SomeProperty = "SomeValue", AnotherProperty = "Other Value" };
+            var target = "{SomeProperty} some text {AnotherProperty}";
```
asbjornu: Will these "bare" format strings work for the variables provided by GitVersion? Perhaps we could make that explicit with a `gv:` prefix?

src/GitVersionCore/Configuration/Config.cs
```diff
@@ -26,6 +26,9 @@ public Config()
         [YamlMember(Alias = "assembly-informational-format")]
         public string AssemblyInformationalFormat { get; set; }
 
+        [YamlMember(Alias = "assembly-file-versioning-format")]
+        public string AssemblyFileVersioningFormat { get; set; }
```
asbjornu: Shouldn't we provide `AssemblyVersioningFormat` as well?

src/GitVersionCore/Helpers/EnvironmentHelper.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace GitVersion.Helpers
+{
+    public class EnvironmentHelper
+    {
+        public static string GetEnvironmentVariableForProcess(string envVar)
+        {
+            return Environment.GetEnvironmentVariable(envVar, EnvironmentVariableTarget.Process);
```
asbjornu: Could we make an abstraction on top of `Environment` here so it's translated to something less specific like `IDictionary<string, string>`? That would make testing easier and the code more portable.

src/GitVersionCore/StringFormatWith.cs
```diff
@@ -8,7 +8,7 @@ namespace GitVersion
 
     static class StringFormatWithExtension
     {
-        private static readonly Regex TokensRegex = new Regex(@"{\w+}", RegexOptions.Compiled);
+        private static readonly Regex TokensRegex = new Regex(@"{\$??\w+(\?\?\w+)??}", RegexOptions.Compiled);
```
asbjornu: Since this is based on a regex, I would love to see some negative tests as well that challenges the regex engine a bit, just to make sure we don't have a too greedy expression or something that causes performance problems.

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
ruhullahshah: Sure we could but I think from the reaction on approach suggested by @jbaehr on https://github.com/GitTools/GitVersion/issues/1157 we agreed to use $ as the prefix for enviroment variables and I am not sure as to what the collisions would be here since the usage of $ occurs in the format string and we parse them internally.

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
ruhullahshah: This testcase just tests the existing functionality and specifically tests the `FormatWith` method. I guess this should not be a concern

src/GitVersionCore/Configuration/Config.cs
ruhullahshah: Addressed in the latest patch with some code cleanup

src/GitVersionCore/StringFormatWith.cs
ruhullahshah: Addressed with some additional testcases

src/GitVersionCore/Helpers/EnvironmentHelper.cs
ruhullahshah: I am not sure exactly as to how could I improve here and provide a useful abstraction.  Seems like Mono https://github.com/mono/mono/blob/master/mcs/class/corlib/System/Environment.cs uses the same approach as we have. In case there are multiple env vars, a dictionary would make sense but in this case we seek the value of a single env var and this should be fine

ruhullahshah:
@asbjornu You are welcome. I have addressed some of your review comments and provided my opinion on others. In my opinion, we have the functionality in place that addresses the theme of this issue.

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
jbaehr: Note that introducing a "gv:"-prefix for git version properties would break existing config files. It may be acceptable to introduce breaking changes while still in beta phase, but that is something I'd rather do in a dedicated PR.

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
asbjornu: The collision I'm thinking of is between non-environment variables and GitVersion's variables. We might introduce system variables and whatnot in the future, so having a prefix that clearly identifies where the variable is coming from is more future proof and intuitive than just `$` imho.

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
asbjornu: True, scrap that. No prefix means GitVersion's own variables.

src/GitVersionCore/Helpers/EnvironmentHelper.cs
asbjornu: I just want a translation so our "core" isn't dependent on the `System.Environment` class itself. Perhaps `Func<string, string>` is a better abstraction?

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
jbaehr: Using `$` for dereferencing env vars was my initial proposal as for me it is very intuitive --
 it's was bash is using.
Anyway, I have no strong opinion on that; I'm also fine with `env:`.

src/GitVersionCore/Helpers/EnvironmentHelper.cs
jbaehr: Using `Func<string, string>` has the advantage that one can use `System.Environment.GetEnvironmentVariable` (or even the currently used `Helpers.EnvironmentHelper.GetEnvironmentVariableForProcess`) directly at the call site.

What is your preference here? Turning this formatter into a class and use ctor injection or rather give the currently used string extension method an additional parameter?
(e.g. `public static string FormatWith<T>(this string template, T source, Func<string, string> getEnvironmentVariable)`)

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
ruhullahshah: Addressed in the latest commit

src/GitVersionCore/Helpers/EnvironmentHelper.cs
ruhullahshah: `GitVersionCore\Authentication.cs`,  `GitVersionCore\BuildServers\*.cs` etc all use a naked call to `System.Environment.GetEnvironmentVariable`. How exactly does my patch create a problem when there are already existing reference to System.Environment in the GitVersionCore? Maybe abstracting all the System.Environment references in the GitVersionCore can be a part of a cleanup but I am not sure how drastically does it change things here.

asbjornu: True, I did not think of the existing calls to `System.Environment`. I agree; let's address that abstraction in a future PR.

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_EnvVarTokenWithFallback()
+        {
+            Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR", "Env Var Value");
+            var propertyObject = new { };
+            var target = "{env:GIT_VERSION_TEST_VAR??fallback}";
```
asbjornu: Are spaces allowed around the null propagation operator? That would be a readability improvement, imho. Is that hard to add?

ruhullahshah: Addressed in the latest patch.

ruhullahshah:
You are welcome. I have included the changes for the null propagation operator readability and updated some test cases.

src/GitVersionCore/OutputVariables/VariableProvider.cs
```diff
@@ -100,5 +94,24 @@ static void PromoteNumberOfCommitsToTagNumber(SemanticVersion semanticVersion)
             semanticVersion.BuildMetaData.CommitsSinceVersionSource = semanticVersion.BuildMetaData.CommitsSinceTag ?? 0;
             semanticVersion.BuildMetaData.CommitsSinceTag = null;
         }
+
+        static void CheckAndFormatString<T>(string formatString, T source,  string defaultValue, string formatVarName, out string formattedString)
```
asbjornu: Neat, I like this. But is there a reason why the `formattedString` is an `out` parameter instead of a return value?

src/GitVersionCore.Tests/StringFormatWithExtensionTests.cs
```diff
@@ -0,0 +1,129 @@
...
+            var actual = target.FormatWith(propertyObject);
+            Assert.AreEqual(expected, actual);
+        }
+
+        [Test]
+        public void FormatWith_EnvVarTokenWithFallback()
+        {
+            Environment.SetEnvironmentVariable("GIT_VERSION_TEST_VAR", "Env Var Value");
+            var propertyObject = new { };
+            var target = "{env:GIT_VERSION_TEST_VAR ?? fallback}";
```
asbjornu: 👍

src/GitVersionCore/StringFormatWith.cs
```diff
@@ -27,16 +27,44 @@ public static string FormatWith<T>(this string template, T source)
 
             // {MajorMinorPatch}+{Branch}
             var objType = source.GetType();
+            const string ENV_VAR_INDICATOR = "env:";
```
asbjornu: Perhaps this `const` can be moved up alongside the `TokensRegex`?

src/GitVersionCore/StringFormatWith.cs
```diff
@@ -27,16 +27,44 @@ public static string FormatWith<T>(this string template, T source)
...
             {
                 var memberAccessExpression = TrimBraces(match.Value);
-                Func<object, string> expression = CompileDataBinder(objType, memberAccessExpression);
-                string propertyValue = expression(source);
+                string propertyValue = null;
+
+                // Support evaluation of environment variables in the format string
+                // For example: {env:JENKINS_BUILD_NUMBER??fall-back-string}
+
+                if (memberAccessExpression.StartsWith(ENV_VAR_INDICATOR))
```
asbjornu: Can't you just check for the match of `env:` in the regex instead of using this constant? With [Named Matched Subexpressions](https://docs.microsoft.com/en-us/dotnet/standard/base-types/grouping-constructs-in-regular-expressions#named-matched-subexpressions) in the regex you can reference the [match group](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.group?view=netframework-4.7.1) by name and check its [`.Success` property](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.group.success).

src/GitVersionCore/OutputVariables/VariableProvider.cs
ruhullahshah: No preferences. Matter of taste

ruhullahshah:
@asbjornu I would like to address all the changes you request in one go. This is getting repetitive. The review before this should have had these comments because these changes have been there before the last commit. I will give you some time to make up your mind about the PR as a whole and once you have reviewed all the code let me know, i will address your changes.

asbjornu:
@ruhullahshah: Sorry about that. I was sure the latest review was for new code. I should of course have reviewed that in the first go, but if I remember correctly, I had limited time when I did the first review so it wasn't as thorough as I would have liked.

src/GitVersionCore/StringFormatWith.cs
ruhullahshah: Addressed in the latest commit.

src/GitVersionCore/StringFormatWith.cs
ruhullahshah: Addressed in the latest commit

src/GitVersionCore/OutputVariables/VariableProvider.cs
ruhullahshah: Addressed in the latest commit

asbjornu:
Excellent. Thanks for this!

ruhullahshah:
You are welcome :)

vishalvatsal:
Hi Team,

I have a need to use build number in the assembly version for my TeamCity build. 

Following is my gitversion config

`next-version: 1.0.0
assembly-versioning-format: '{Major}.{Minor}.{Patch}.{env:BUILD_COUNTER??0}'
assembly-file-versioning-scheme: MajorMinorPatch
continuous-delivery-fallback-tag: ''
mode: ContinuousDeployment
branches:
    release:
        tag: ''        
    develop:
        tag: 'dev'
    master:
        tag: 'master'
ignore:
  sha: []
merge-message-formats: {}`


In Team City project I have an environment variable

`Name: env.BUILD_COUNTER
 Value: %build.counter%'

After the build this is what I get
**GitVersion.AssemblySemVer**  =  **1.0.0.{env:BUILD_COUNTER??0}**


Can you please let me know what I am doing wrong here? Or what do I need to do to make sure env:BUILD_COUNTER gets substituted with the correct value?

Thank you

jbaehr:
@vishalvatsal The implementation in this PR explicitly requires spaces around the `??` in order to treat it as a fallback-separator. Your config sample seem to omit those spaces.

Note however, that the currently latest GitVersion release features an enhanced implementation, that further allows fallback strings for regular GitVersion variables, too, and also allows for quoted fallback strings containing special characters. cf. #2179

vishalvatsal:
Thanks a ton @jbaehr.

It works now.

Issue #1132, opened 2017-01-10
Feature request: GitVersionTask with C++, WixToolset projects support


sergey-s-betke:
For WixToolset projects - just environment variables is enough. Or `version.wxi` with 

    <?xml version="1.0" encoding="utf-8"?>
    <Include xmlns="http://schemas.microsoft.com/wix/2006/wi">
        <?define ProductVersion="02.04.3369.0007"?>
        ...
    </Include>

Or `version.wxs` with

    <?xml version="1.0" encoding="utf-8"?>
    <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
        <?include version.wxi ?>
        <WixVariable Id="ProductVersion" Value="$(var.ProductVersion)"/>
        ...
    </Wix>

For .vcxproj - `version.h` for с++ compiler and **resource compiler** is enough. Minimum - project variables for using in `/D` directive for c++ compiler and resource compiler...

sergey-s-betke:
For C++ custom build step:

          <ItemDefinitionGroup>
            <CustomBuildStep>
              <Command>$(SolutionDir)packages\GitVersion.CommandLine.3.6.5\tools\GitVersion.exe /exec update_version.h.cmd /execargs version.h</Command>
              <Message>GitVersion: retrieving version information from git repository</Message>
              <Outputs>version.h</Outputs>
              <Inputs>$(SolutionDir).git/logs/HEAD</Inputs>
              <TreatOutputAsContent>false</TreatOutputAsContent>
            </CustomBuildStep>
          </ItemDefinitionGroup>

and `update_version.h.cmd`:

				SETLOCAL EnableDelayedExpansion

				SET VERSION_H_FILE=%~f1 

				echo // version data file > %VERSION_H_FILE%

				FOR %%V IN (
					Major,Minor,Patch,PreReleaseTag,PreReleaseTagWithDash,PreReleaseLabel,PreReleaseNumber,
					BuildMetaData,BuildMetaDataPadded,FullBuildMetaData,MajorMinorPatch,SemVer,LegacySemVer,LegacySemVerPadded,
					AssemblySemVer,FullSemVer,InformationalVersion,BranchName,Sha,
					NuGetVersionV2,NuGetVersion,
					CommitsSinceVersionSource,CommitsSinceVersionSourcePadded,CommitDate
				) DO (
					echo #define GITVERSION_%%V !GitVersion_%%V! >> %VERSION_H_FILE%
				)

				ENDLOCAL

[...]

PR #1386, opened 2018-03-15
Adding support to generate WiX version files

ruhullahshah:
- Create WiX version files from GitVersion variables using the CLI which can allow smoother integration with WiX project workflows.
- Partially addresses https://github.com/GitTools/GitVersion/issues/1132
- Usage: `GitVersion.exe /updatewixversionfile`
- Sample output attached
[GitVersion_WixVersion.zip](https://github.com/GitTools/GitVersion/files/1816254/GitVersion_WixVersion.zip)


ruhullahshah:
You are welcome. I have added some testcases that cover the use cases scenarios for this patch

src/GitVersionExe.Tests/UpdateWixVersionFileTests.cs
```diff
@@ -0,0 +1,102 @@
+﻿namespace GitVersionExe.Tests
```
asbjornu: Why only `GitVersionExe` tests and not for `GitVersionCore`? Also, don't you agree approval tests such as this is better suited for file manipulation testing such as this?

https://github.com/GitTools/GitVersion/blob/ee5b77a3439b7e905ca8401c7d8ed15b2495b0d6/src/GitVersionCore.Tests/AssemblyInfoFileUpdaterTests.cs#L26-L40

src/GitVersionExe/HelpWriter.cs
```diff
@@ -47,6 +47,12 @@ public static void WriteTo(Action<string> writeAction)
                     it be created with these attributes: AssemblyFileVersion, AssemblyVersion and AssemblyInformationalVersion
                     ---        
                     Supports writing version info for: C#, F#, VB    
+
+    # Create or update Wix version file
+    /updatewixversionfile
```
asbjornu: Does it make sense updating WiX files from GitVersionTask as well? If so, perhaps we should add a way to do that as well?

ruhullahshah: For our use cases, changes to GitVersion exe suffice and while it is a valid use case but I don't aim to address GitVersionTask here.

src/GitVersionExe.Tests/UpdateWixVersionFileTests.cs
ruhullahshah: Could you please explain this in detail?

asbjornu: Yes, sure. Approval tests allows you to create a file on disk that is what the test is compared against. If the content matches, the test is successful, otherwise it fails and a diff is presented.

The above test will verify the manipulated file content against the following approved text file:

https://github.com/GitTools/GitVersion/blob/ee5b77a3439b7e905ca8401c7d8ed15b2495b0d6/src/GitVersionCore.Tests/Approved/cs/AssemblyInfoFileUpdaterTests.ShouldCreateAssemblyInfoFileWhenNotExistsAndEnsureAssemblyInfo.approved.txt#L1-L13

src/GitVersionExe/HelpWriter.cs
asbjornu: Ok.

asbjornu:
@ruhullahshah, can you please find time to address the code review comment regarding approval testing, rebase and resolve the conflicts of this PR?

ruhullahshah:
I will resolve the pending work in https://github.com/GitTools/GitVersion/pull/1599,

Issue #1401, opened 2018-04-20
TfsTask - Add 'path' parameter


Rora:
## Context
We're using the gitversion tfs task, gitversion with multiple solutions in 1 repository. Our repository tree is the following:

Root
   - Solution A
      - gitversion.yml
      - Project A1
         - Program.cs
   - Solution B
      - gitversion.yml

We have a separate build for each solution. Since the build uses the root of our repo as working directory we have to setup most build steps to look in the solution dir. For example, when building we provide 'Solution A/**/*.csproj'.

Since this isn't supported at the moment by the gitversion task, gitversion isn't able to find our gitversion file.

## Desired behavior
At the moment the GitVersion task calls GitVersion.exe with the path to the working directory of the build. It would be great if we could add a parameter 'Target path' that allows users to change the path to a deeper level. The gitversion would use this path and combine it with the working directory path changing:

``C:\Agent-1\_work\_tasks\GitVersion_e5983830-3f75-11e5-82ed-81492570a08e\3.6.5\GitVersion.exe C:\Agent-1\_work\16\s /output buildserver /nofetch``

to

``C:\Agent-1\_work\_tasks\GitVersion_e5983830-3f75-11e5-82ed-81492570a08e\3.6.5\GitVersion.exe C:\Agent-1\_work\16\s\Solution A /output buildserver /nofetch`` 

**I'm willing to write a PR if you agree with this enhancement**

asbjornu:
This sounds related to #1170. Is it?

Roraa:
Correct, I however try to fix it by running gitversion in a specific directory. If we'd allow for gitversion.yml to exist in subdirectories, then how would we handle multiple gitversion files? For example:

/SolutionA/Gitversion.yml
/SolutionB/Gitversion.yml


________________________________
Van: Asbjørn Ulsberg <notifications@github.com>
Verzonden: maandag 28 mei 2018 11:54
Aan: GitTools/GitVersion
CC: Rob; Author
Onderwerp: Re: [GitTools/GitVersion] TfsTask - Add 'path' parameter (#1401)


This sounds related to #1170<https://github.com/GitTools/GitVersion/issues/1170>. Is it?

asbjornu:
The closest one to the project/solution being built, wins?

Roraa:
That could result in a tie, also when running in a tfs build, the gitversion task is ran without context of any project or solution. We use it as follows:

1. Start build
[...]

PR #1402, opened 2018-04-20
TfsTask - Add 'path' parameter

Rora:
Resolves issue #1401


asbjornu:
@Rora, thanks for this. Is it possible to test this in any way? If so, could you please add a test, rebase and resolve conflicts? 🙏

arturcic:
As far as I remember, this is a nodejs app so the only way to test it is by running as nodejs app. Not sure we can include a unit test

asbjornu:
I don't mind having a test suite executed with `npm test`. If that's the only way to test this, then I would appreciate if we got something like that up and running.

Rora:
I agree that it would be best to get a test suite up and running to be able to test this code. We should IMO log this as a separate issue as I don't have any experience in setting a nodejs test suite up and it would unnecessary prolong this PR.

arturcic:
+1 for a separate PR

asbjornu:
While I agree on doing a test suite as a separate PR, I would have preferred it the other way around; test suite first, features second. But seeing how long this PR has waited, I think it's fine merging this first.

PR #1410, opened 2018-05-17
 [Bugfix] GitVersion crashes if there are multiple branches containing the keyword develop and the current branch has Increment policy set to Inherit with current commit being a merge commit

ruhullahshah:
- Problem
In case there are multiple branches containing the keyword `develop` and the current branch has an increment policy set to inherit e.g. `feature/x`, if the current commit is a merge commit, GitVersion crashes.

- Root Cause
In case of merge commits, GitVersion derives the branch configuration from `master` or `develop`. When multiple develops exist, any one of them can be selected, in case of the attached test repo it is `bob_develop`. While trying to derive the configuration of `bob_develop`, the regex is modified in `Config.cs` as shown in the changelist. This causes a regex mismatch and hence the invalid configuration attached to `bob_develop`

- Fix
Remove the `^` added to the branch regex. Either the `^` should be present in the branch regex itself or not at all.

Attached is a repo where this problem can be reproduced.
[test-repo.zip](https://github.com/GitTools/GitVersion/files/2013139/test-repo.zip)


tvoyle:
@asbjornu  @arturcic This change to the default branch regexes breaks determination of master or develop if there is no local tracking branch as it will not match origin/(master|develop)

The BranchConfigurationCalculator checks against Branch.FriendlyName, which includes the remote, when trying to find master or develop.

Should the checks be switched to use Branch.NameWithoutRemote instead?

asbjornu:
@tvoyle: Thanks for both testing and alerting us! Are you up for submitting a PR changing from `Branch.FriendlyName` to `Branch.NameWithoutRemote`?

PR #1248, opened 2017-06-30
Upgrade LibGit2Sharp from v0.23 to v0.24

asbjornu:
This PR upgrades `LibGit2Sharp` from version `0.23` to version `0.24` and `LibGit2Sharp.NativeBinaries` from version `1.0.160` to version `1.0.185`.

PR #1416, opened 2018-05-30
Update LibGit2Sharp to 0.25.0-preview-0033

### LINKED ISSUE #1415: Update libgit2sharp to 0.24+ to fix path issues with escape sequences
`libgit2sharp` can't load the native DLL if the path contains `%2F`, which happens frequently on Jenkins when building pipelines in nested folders.

libgit2/libgit2sharp#1389 was fixed in libgit2/libgit2sharp#1409, which appears to be in 0.24+. It looks like GitVersion `master` is still targeting a prerelease of `0.23`. Any chance we can upgrade to 0.24+ and cut a release? I'd be happy to send a PR if that is acceptable.

We can work around this by setting a `customWorkspace` on the Jenkins node definition, but I'd really like to not have to update all my builds with custom workspace naming logic just to get nice versioning.

nlowe:
Based on #1248, this Upgrades `LibGit2Sharp` to `0.25.0-preview-0033`. Getting tests to pass required upgrading `GitTools.Testing` and `GitTools.Core` to the latest version. `GitTools.Testing`'s minimum target is `4.5.2`, so the testing projects had to be upgraded.

Starting with `0.25.0-preview-0073`, `LibGit2Sharp` is `netstandard2.0` only. I ran into target file issues with the targets that add references to the native binaries. I have to look into this more but I believe the simplest solution is to go straight to `netstandard2.0` here (#1175 ?). If you're uncomfortable running on a pre-release, I can re-do this to target 0.24.1 as it still fixes #1415 for me. Let me know what you think.

Fixes #1415

### DIFF
--- a/src/GitVersionCore.Tests/GitVersionCore.Tests.csproj
+++ b/src/GitVersionCore.Tests/GitVersionCore.Tests.csproj
@@ -70,5 +70,4 @@
   <ItemGroup>
     <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
   </ItemGroup>
-
 </Project>
--- a/src/GitVersionCore.Tests/Mocks/MockQueryableCommitLog.cs
+++ b/src/GitVersionCore.Tests/Mocks/MockQueryableCommitLog.cs
@@ -36,16 +36,6 @@ public class MockQueryableCommitLog : IQueryableCommitLog
     {
         throw new NotImplementedException();
     }
-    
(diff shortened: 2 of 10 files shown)


nlowe:
Rebased on master, there were a lot of conflicts with the recent dotnet core merge.

dazinator:
Sorry guys, it looks like multiple people have tried to do the same thing and there are overlapping pr's. Im closing this one in favour of #1422

PR #1438, opened 2018-07-03
Update GitReleaseNotes

arturcic:
- Updated to latest version of GitReleaseNotes to fix the ssl issue when generating release notes
- fixed the version back to 4.0.0

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -3,4 +3,6 @@ branches:
   master:
     tag: beta
 ignore:
-  sha: []
+  sha: 
+    - 0e77e2e16a136a1cd0f0474fe5710f617d508d81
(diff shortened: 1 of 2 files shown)


gep13:
@arturcic thank you for taking the time to update this.  Let's get this merged, and see if it helps with the error that was occurring.

gep13:
@arturcic so, I don't "think" the failed build is unexpected, as I think it still expecting that the path to the releasenotes is present, which it isn't when the releasenotes are being created.

I will try to take the current build for a spin later to see if I can get that part to work.  Unless someone gets to it before me.

gep13:
@arturcic looks like the build number is correctly working now, so thanks for making that change!

arturcic:
I guess we just need to check if **build** folder exists and create it if not

PR #1446, opened 2018-07-24
[Bugfix] Wrong version number observed in a feature branch after a tagged develop is merged into it

ruhullahshah:
- The testcase (failing as of now) is pretty self-explanatory and outlines the problem.


ruhullahshah:
@JakeGinnivan 
- While analysing the code, I saw that you guys to set the `Increment` to `Patch` [here](https://github.com/GitTools/GitVersion/blob/c7c03329ef0ae21496552219a38caa6d16dfb73f/src/GitVersionCore/BranchConfigurationCalculator.cs#L148), is there a specific reason for that? 
- This causes the the tagged commit strategy to yield `16.23.1-feat-featX.4` instead of `16.24.0-feat-featX.4`

jbaehr:
@JakeGinnivan ping! would you please have a look at Ruh's questions above? unfortunately this issue still pops up from time to time...

asbjornu:
It seems like we have quite a few loosely (or closely) related bugs here in this, #1600, #1598 and #1595 (which are rebased versions of #1261, #1533 and #1394). It would of course be great if all birds could be killed with one stone, but I'm not sure how cloesly related these bugs are. I would love to dig into what the source of the problems are and figure out one or more fixes, but I just don't have the time. Perhaps adding all of the failing tests into one PR that I can maintain might be a good idea to make it easier to pick up for someone to fix?

ruhullahshah:
Added a patch that fixes the issue.
### Analysis:  
While trying to inherit the branch configuration for a merged branch, `develop` or `master` are the chosen candidates, however, "for some reason", their defaults are not used and the `IncrementStrategy` was set to `Patch`, which is incorrect. 

### Fix: 
In case the chosen branch for resolution is `master` or `develop`, use the respective `IncrementStrategy`, if it is not set to `Inherit`

asbjornu:
Fantastic, @ruhullahshah. Thanks!

ruhullahshah:
@asbjornu  You are welcome :D

asbjornu:
I had hoped this would fix #1595, #1598 and #1600, but unfortunately not. 😕

ruhullahshah:
@asbjornu  The only issue, in my opinion,  that is related to this merge request is https://github.com/GitTools/GitVersion/issues/1525 and after taking a brief look at the testcases 
for the issues you mentioned above, I do not see the fix related to them,

ruhullahshah:
https://github.com/GitTools/GitVersion/pull/1598 is composed of two testcases:
- One is being addressed here
-  and the second one is related to `CommitsSinceVersionSource`, which is a problem for us as well. 

I can fix try to fix the `CommitsSinceVersionSource` , if I get some feedback on the [comment](https://github.com/GitTools/GitVersion/issues/1526#issuecomment-466460003)

PR #1451, opened 2018-08-07
Support for generic build servers, improved scripting support

ravenpride:
Dear development team,

my team is just setting up a [GoCD](https://www.gocd.org/) build server and we found GitVersion really useful to keep versioning as easy and consistent as possible. Although many build servers are already supported by GitVersion, GoCD is not in the list. The main reason seems to be that GoCD does not support injecting retrieved version information into the build process.

To circumvent this missing feature in GoCD and probably other builder servers we've developed a small .NET tool called 'EnvRun'. It simply wraps processes and scans their output (stdout/stderr) for key expressions that control setting and resetting environment variables for following runs:
```
@@envrun[set name='MyVar' value='MyValue']
@@envrun[reset name='MyVar']
```

This pull request contains a new build server implementation for GitVersion that emits the messages EnvRun needs to operate. A short example on how to use EnvRun:

```
EnvRun.exe gitversion /l console /output buildserver
EnvRun.exe cmd.exe /C "echo %%GitVersion_MajorMinorPatch%%"
```

Please see the [EnvRun project site](https://github.com/GriffinPlus/EnvRun) for details.


ravenpride:
Could someone from the GitVersion team please give a short statement on this pull request? Any idea/improvement that is needed to convince you to merge it into the GitVersion codebase? We actively use *EnvRun* and it works quite well when it comes to supporting building using simple scripts or build servers that do not support injecting environment variables (like GoCD).

Thank you very much.

asbjornu:
@ravenpride, thanks for this. I see no problem in merging this besides the lack of tests. Can you please add a few tests for this new addition alongside the existing [build server tests](https://github.com/GitTools/GitVersion/tree/master/src/GitVersionCore.Tests/BuildServers)?

ravenpride:
@asbjornu Thank you for your clearance. I've added the unit tests, so everything should be ready to merge.

asbjornu:
Thanks for your contributions! 🙏

PR #1405, opened 2018-04-27
Add "ShortSha" to the data provided by GitVersion

pmiossec:
from the `ShortenObjectId()` provided by LibGit2Sharp

(because always handling long sha1 could be a little painfull)

Should fix  #887

PR #1452, opened 2018-08-07
add ShaShort as new Variable

macphil:
At some purposes, the full SHA is to long.
That's why i'm added a new variable with name `ShaShort`.

The default are the first six characters of the SHA.

The length is configurable with `commit-sha-shortlength`


pmiossec:
Same effort than mine ;) #1405

>The default are the first six characters of the SHA.

Is it not better to keep the default behavior of git (and I imagine that libgit2sharp is doing the same).
That is that git is automatically determining the right length...

santiagovm:
code looks good, it would be very useful to have this pull request merged and published to avoid problems with long versions and NuGet v2 feeds (e.g. TeamCity, ProGet)

santiagovm:
@pmiossec,

I understand your argument about avoiding sha collisions, however, I think both options should be allowed: (1) GitVersion calculates a safe short sha and (2) user is willing to take the risk of collisions and specify the short sha length.

I'm looking to have a short sha of a specific length to work around the limitation in NuGet v2 feeds (OData) like TeamCity and ProGet. I understand that collisions would be possible, but one could further resolve the collision using the NuGet Package's publish date to determine what commit is the correct one.

When using NuGet v3 feeds I would use the full sha, no need for a safe short sha.

asbjornu:
I agree that a configurable length of the SHA is a bad idea, so I consider #1405 to fix this. Since that's merged, I'll close this.

PR #1459, opened 2018-08-24
Fix detection of multi targeting

ermshiperete:
The previous code always set GitVersionTaskBuildTools_IsMultiTargeting
to `false`, and the property we set didn't get used anywhere.

### DIFF
--- a/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
+++ b/src/GitVersionTask/NugetAssets/build/GitVersionTask.targets
@@ -1,13 +1,8 @@
 ﻿<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
 
-  <ItemGroup>
-    <PathSeparators Include="/"/>
-    <PathSeparators Include="\"/>
-  </ItemGroup>
-
   <PropertyGroup>
     <GitVersionTaskBuildTools_IsMultiTargeting>false</GitVersionTaskBuildTools_IsMultiTargeting>
-    <GitVersionTaskBuildTools_IsMultiTargeting Condition="$(MSBuildThisFileDirectory.Substring($(MSBuildThisFileDirectory.LastIndexOfAny(@(PathSeparators)))).Substring(1).Contains('buildMultiTargeting'))">true</GitVersionTaskBuildTools_IsMultiTargeting>
+    <GitVersionTaskBuildTools_IsMultiTargeting Condition="$(MSBuildThisFileDirectory.TrimEnd('\').TrimEnd('/').EndsWith('buildMultiTargeting'))">true</GitVersionTaskBuildTools_IsMultiTargeting>
   </PropertyGroup>
 
   <!-- Import infrastructure props for util pack. -->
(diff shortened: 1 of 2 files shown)


dazinator:
Thank you. Not sure how this slipped through. I imagine this should resolve some issues.

dazinator:
I have tried out this PR package. `dotnet build` was fine.
Building from VS resulted in the same error as below (logging level set to diagnostic):

> Using "WriteVersionInfoToBuildLog" task from the task factory "NuGetTaskRunnerFactory".
1>  Task "WriteVersionInfoToBuildLog"
1>    Task Parameter:SolutionDirectory=C:\Users\darre\Documents\Visual Studio 2017\Projects\ClassLibrary7\ClassLibrary7
1>    Task Parameter:NoFetch=false
1>    Resolved GitVersionTask, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null located in C:\Users\darre\.nuget\packages\gitversiontask\4.0.0-pullrequest1459-1614\build\net461\GitVersionTask.dll and loaded from same path.
1>    Resolved Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a located in C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Build.dll and loaded from same path.
1>    Resolved Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a located in C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\MSBuild\15.0\Bin\Microsoft.Build.dll and loaded from same path.
1>    MSBUILD : error MSB4166: Child node "2" exited prematurely. Shutting down. Diagnostic information may be found in files in "C:\Users\darre\AppData\Local\Temp\" and will be named MSBuild_*.failure.txt. This location can be changed by setting the MSBUILDDEBUGPATH environment variable to a different

Inspection of that log file was the same error about binary writer and null.

Something stands out though:

>  Task Parameter:SolutionDirectory=C:\Users\darre\Documents\Visual Studio 2017\Projects\ClassLibrary7\ClassLibrary7

That path is the project directory not the solution directory. I'll look into this - it could be that the wrong solution dir is being computed and passed to the task.

dazinator:
Even when the correct SolutionDir is passed, it doesn't solve this issue. I think it might be a utilpack issue. Going to raise something over there.

dazinator:
I'm going to accept this PR as it fixes what it set out too, even if GitVersionTask is still problematic, I have a feeling the issue was introduced by utilpack.nuget.msbuild version 2.6.0 but I will test that seperately.

dazinator:
@ermshiperete - Unfortunately this PR didn't solve the issue. 

The problem is, $(MSBuildThisFileDirectory) is evaluated as: `C:\Users\Daz\.nuget\packages\gitversiontask\4.0.0-beta0014\build\functionality`

i.e this is the path to the targets file inside the nuget package - and it doesn't appear to be in a `buildMultiTargeting` in the nuget package.

Issue #1466, opened 2018-09-03
GitVersionTask 4.0.0-beta0014 does not work with MSBuild-14.0


jbaehr:
When I reference the GitVersionTask in version 4.0.0-beta0014 from a csproj I can't build it any more using MSBuild v14. It used to work fine with beta13, but now I get the following error
```
C:\...\packages\GitVersionTask\build\UtilPack.Version.props(1,1): error MSB4041: The default XML namespace of the project must be the MSBuild XML namespace. If the project is authored in the MSBuild 2003 format, please add xmlns="http://schemas.microsoft.com/developer/msbuild/2003" to the <Project> element. If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format. [C:\...\Foo.csproj]
```

The issue is quite easy to fix: just add the namespace declaration to `src/GitVersionTask/UtilPack.Version.props`:
```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
....
```

dazinator:
Any chance of a PR?

jbaehr:
@dazinator Any chance of taking care of the PR? (and #1465, too)

dazinator:
@jbaehr  PR merged, sorry for the delay, and thank you for the contribution!

dazinator:
Will have a look next at #1465

PR #1471, opened 2018-09-12
Add the xml namespace to work with MSBuild v14

jbaehr:
This fixes GitTools/GitVersion#1466


dazinator:
Thanks for this, sorry for the delay. Can't see any harm in merging this as its just a namespace addition, shouldn't break anything, builds and tests all look good, so merging.

PR #1487, opened 2018-10-02
Refactoring of merge message class

Kieranties:
Refactoring of the `MergeMessage` class to enable simpler (and consistent) approach to adding new regular expressions.  This is a prelude to an additional PR where I hope to enable custom patterns through config.

- All patterns are now stored in a collection.  The collection is enumerated to apply the order in which they are checked (currently the same order previously used)
- Each pattern has a required _SourceBranch_ capture group
- Pull request patterns have a required _PullRequestNumber_ capture group
- All patterns (except Bit Bucket pull request) have an optional _TargetBranch_ capture group.  _TargetBranch_ is required for Bit Bucket pull requests

This introduces the following **breaking changes**:

| Merge message | Previous result | Breaking Change |
| --- | --- | --- | 
| Merge pull request # from feature/one | Pull Request Number = 0 <br /> Merged Branch = feature/one |  **No match** |
| Merge pull request # in feature/one from feature/two to master | Pull Request Number = 0 <br /> Merged Branch = feature/two <br /> Target Branch = master | **No match** |
| Merge pull request #1234 from feature/one into dev | Pull Request Number = 1234 <br /> Merged Branch = **feature/one into dev** <br /> Target Branch = dev| Pull Request Number = 1234 <br /> Merged Branch = **feature/one** <br /> Target Branch = dev|
| Merge pull request #1234 from feature/one from feature/two to master | Pull Request Number = 1234 <br /> Merged Branch = feature/two <br /> Target Branch = **null** | Pull Request Number = 1234 <br /> Merged Branch = feature/two <br /> Target Branch = **master** |
| Finish V4.0.0 into master | Merged Branch = **V4.0.0 into master** <br /> Target Branch = master | Merged Branch = **V4.0.0** <br /> Target Branch = master |

PR #1488, opened 2018-10-02
Enable custom merge message formats via configuration

Kieranties:
**NOTE**: This PR has is dependent on the breaking changes introduced by #1487 and should not be approved unless the related breaking changes are approved.

This PR enables custom merge message formats to be added via configuration.

I've now used GitVersion on a few build systems but only ran into issues with merge messages when migrating to TFS.  TFS merge messages vary depending on if the merge is a PR, has one or more commits, or is customised by the user completing the PR.  In many cases, the merge message will be _Merged PR 1234: <original commit message>_.  this sadly does not fit the baked-in messages that come with GitVersion.

Instead of simply adding another entry into the defaults, this PR enables user to add custom messages formats into their configuration.


Kieranties:
@asbjornu All synched up and green 🎉

docs/configuration.md
```diff
@@ -48,6 +48,7 @@ commit-date-format: 'yyyy-MM-dd'
 ignore:
   sha: []
   commits-before: yyyy-MM-ddTHH:mm:ss
+merge-messages: {}
```
asbjornu: Wouldn't it be more consistent with existing config to name this `merge-message-formats`?

docs/configuration.md
```diff
@@ -179,6 +180,24 @@ Date and time in the format `yyyy-MM-ddTHH:mm:ss` (eg `commits-before:
...
 
+### merge-messages
+Custom merge message formats to enable identification of merge messages that do not
+follow the built-in conventions.  Entries should be added as key-value pairs where
+the value is a regular expression.
+e.g.
+
+```
+merge-messages:
+    tfs: ^Merged (?:PR (?<PullRequestNumber>\d+)): Merge (?<SourceBranch>.+) to (?<TargetBranch>.+)
```
asbjornu: How is the `tfs` key defined and used and which other keys do you envision that we add?

src/GitVersionCore/MergeMessage.cs
```diff
@@ -4,56 +4,74 @@
...
+                {
+                    return;
                 }
             }
         }
 
-        public string MatchDefinition { get; }
-        public string TargetBranch { get; }
-        public string MergedBranch { get; } = "";
+        public string MatchDefinition { get; private set; }
```
asbjornu: Is it possible to leave these properties immutable and instead return a new instance of the `MergeMessage` class when mutation is required?

docs/configuration.md
Kieranties: Yes it would!  I'll get that fixed.

src/GitVersionCore/MergeMessage.cs
Kieranties: These properties are now set in `ApplyPattern` outside of the constructor so required private set access.  I could refactor things so that they are set in the constructor again if you wish (e.g. making `ApplyPattern` a local function and/or merging the pattern collections together).

They are properties of the instance being actively constructed. An internal context class could be used for immutability purposes, but any previous mutations would be redundant as they caller of  would never receive them.

docs/configuration.md
Kieranties: I don't envisage the project adds _any_ more at all.  The intent was that the user would add entries as needed.  This means new or modified external services don't require a new version of Gitversion to support them.

The key is intended to be an identity so that the match used for the commit can be logged or debugged more easily.  I've not implemented this for logging, but can pick that up.

Kieranties:
@asbjornu I've some time tomorrow to tidy things up as required, then will be AFK for a few days.  Let me know what you want to do on the last two points above

docs/configuration.md
Kieranties: Changed `merge-message` -> `merge-message-formats`

Kieranties:
@asbjornu Did you have any further response on the code review?  I'll be able to address things this evening.  It seems the appveyor build failed for an unrelated reason.

docs/configuration.md
Kieranties: @asbjornu I've added a line to log the matching merge message definition in the `MergeMessageBaseVersionStrategy`

src/GitVersionCore/MergeMessage.cs
Kieranties: @asbjornu This is the only thing left on this review.  Let me know what you want to do here.

asbjornu: It's not very important, but if it's not much work to retain immutability, I'd prefer it.

docs/configuration.md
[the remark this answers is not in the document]
Kieranties: It should, will fix

src/GitVersionCore.Tests/MergeMessageTests.cs
```diff
@@ -228,15 +229,15 @@ public void EmptyTagPrefix(string prefix)
...
+              sut.MatchDefinition.ShouldBe("TfsMergeMessageEnglishUS");
             sut.TargetBranch.ShouldBe(expectedTargetBranch);
             sut.MergedBranch.ShouldBe(expectedMergedBranch);
             sut.IsMergedPullRequest.ShouldBeFalse();
             sut.PullRequestNumber.ShouldBeNull();
             sut.Version.ShouldBe(expectedVersion);
         }
 
-        private static readonly object[] ParsesTfsGermanDEMergeMessages =
+private static readonly object[] ParsesTfsGermanDEMergeMessages =
```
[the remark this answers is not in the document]
Kieranties: Merge artefact.  I'll get that fixed.

src/GitVersionCore.Tests/MergeMessageTests.cs
```diff
@@ -228,15 +229,15 @@ public void EmptyTagPrefix(string prefix)
             var sut = new MergeMessage(message, _config);
 
             // Assert
-            sut.MatchDefinition.ShouldBe("TfsMergeMessageEnglishUS");
+              sut.MatchDefinition.ShouldBe("TfsMergeMessageEnglishUS");
```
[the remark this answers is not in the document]
Kieranties: Note to self, don't use the github merge tool in the future...

Kieranties:
> I think this mostly looks good. But I'm thinking, could perhaps `merge-message-formats` replace the entire TFS merge message support that was implemented in #1591?

Absolutely, but you'll have a regression for consumers of that change.  They will need to update their configuration to cater for the TFS messages when they update to 5.0

asbjornu:
> Absolutely, but you'll have a regression for consumers of that change. They will need to update their configuration to cater for the TFS messages when they update to 5.0

Couldn't the TFS merge messages be added as defaults in `merge-message-formats`?

Kieranties:
> Couldn't the TFS merge messages be added as defaults in `merge-message-formats`?

I might be confused... Do you mean adding to the default configuration class itself?  I'm not sure I see the benefit.

asbjornu:
> I might be confused... Do you mean adding to the default configuration class itself? I'm not sure I see the benefit.

Don't worry. If anything, I'm the confused one. 😄 I'm thinking that the `merge-message-formats` implementation can add the TFS merge messages as defaults so the TFS merge message support isn't lost, it's just migrated over to the `merge-message-formats` property.

Kieranties:
@asbjornu Anything else to do here?

asbjornu:
@Kieranties, I don't feel we've got the same understanding of the relationship between `merge-message-formats` and the TFS merge messages. I still have a feeling we could implement the support for TFS merge messages with the `merge-message-formats` feature, I'm just unsure how.

Think of it this way: If this PR was merged before #1591, could #1591 have looked differently? If so, how? If we're able to describe "how", I would love to have that implemented in this PR.

Kieranties:
@asbjornu if this had been completed before #1591 then the user would be able to add the following to their configuration.
```yaml
merge-message-formats:
  TfsMergeMessageEnglishUS: ^Merge (?<SourceBranch>[^\s]*) to (?<TargetBranch>[^\s]*)
  TfsMergeMessageGermanDE: ^Zusammengeführter PR ""(?<PullRequestNumber>\d+)""\: (?<SourceBranch>.*) mit (?<TargetBranch>.*) mergen")
```
However, as #1591 has already been implemented, the above already exist as defaults in the `MergeMessage` class.  So now there is no need for any user to ever specify them, unless they are removed from the list of defaults.

asbjornu:
Ok, thanks for the explanation, @Kieranties. It all makes sense now. Then we only have to decide whether we want to undo the changes introduced in #1591 just to keep the core of GitVersion as simple as possible. How do you feel about that, @joergmetzler?

joergmetzler:
I'm fine with that.

asbjornu:
Awesome. Can you please effectively revert #1591 then, @Kieranties? I'll be quick to approve and merge after.

Kieranties:
I'll pick this up this evening.  You'll need to highlight the regression
though right?

Kieranties:
Ah!  Just spotted it's a new thing in 5.0 as well.  No regression then 👍

Kieranties:
@asbjornu Good to go!

src/GitVersionCore/MergeMessage.cs
```diff
@@ -72,17 +76,7 @@ private SemanticVersion ParseVersion(string branchName, string tagPrefix)
             return null;
         }
 
-        private class MergeMessagePattern
```
asbjornu: I'm sorry for not noticing this before, but why have you removed the `MergeMessagePattern` class?

Kieranties: I'd introduced the private class in the refactoring and unit testing completed in #1487. The class was unnecessary as it was simple a keyvalue pair so I switched out for that instead.

asbjornu: I quite like having utility classes like that. Makes the code much more readable. Can you please bring it back? :pray: :smiley:

Kieranties: Done.  This highlighted that there were a few different words being used inconsistently (Definition, Format, Pattern).  I've tidied that up too:
Each entry from the config and defaults creates a `MergeMessageFormat` instance.  A `MergeMessageFormat` consists of a `Name` and a `Pattern`. Where a format matches, the `MergeMessage.FormatName` property is populated from the `MergeMessageFormat.Name` property.

I also spotted that the `ParseVersion` function was receiving an irrelevant parameter, so fixed that.

asbjornu:
Awesome, thank you so much for the great work and patience, @Kieranties!

Issue #1464, opened 2018-08-30
Publish as .NET Core Global Tool


kll:
I think it would be a good idea to publish GitVersion as a [.NET Core Global Tool](https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools).

I'm willing to do the work if it is agreed upon that it is a good idea. I haven't looked into it in detail yet but I assume it would probably need some outstanding work like #1445 and #1422 to get wrapped up first. It would obviously require getting actual stable releases out to nuget.org again. :smile: 

First order of business would be picking the name of the package.  I'm personally pretty confused on what all the various GitVersion related packages already on nuget.org are for so I think some care to pick a good meaningful is important. I like what Cake did in [v0.30.0](https://cakebuild.net/blog/2018/08/cake-v0.30.0-released) and suggest `GitVersion.Tool` for the package name. It's not taken but there is an existing [GitVersion.Tools](https://www.nuget.org/packages/GitVersion.Tools/) which might be confusing.

The end result would be easily installing GitVersion on any platform that supports .NET core with the following command:
```
dotnet tool install -g GitVersion.Tool
```

Then you can run it from anywhere with this command:
```
dotnet gitversion
```

dazinator:
@arodus 

> The command would be just gitversion without the dotnet.

According to https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools it could be run either way depending on how it's written! 

@arturcic  - this would definitely be a welcomed PR but you are right those issues you linked are high on the agenda right now for obvious reasons

kll:
This works for me in Windows but not in Linux (Debian 9). It's the same LibGit2Sharp error I've seen before. I'll try to debug it but I'm a bit out of my depth.

```
System.TypeInitializationException: The type initializer for 'LibGit2Sharp.Core.NativeMethods' threw an exception. ---> System.DllNotFoundException: Unable to load shared library 'git2-b0d9952' or one of its dependencies. In order to help diagnose loading problems, consider setting the LD_DEBUG environment variable: libgit2-b0d9952: cannot open shared object file: No such file or directory
   at LibGit2Sharp.Core.NativeMethods.git_libgit2_init()
   at LibGit2Sharp.Core.NativeMethods.InitializeNativeLibrary()
   at LibGit2Sharp.Core.NativeMethods..cctor()
   --- End of inner exception stack trace ---
   at LibGit2Sharp.Core.NativeMethods.git_buf_free(GitBuf buf)
[...]

PR #1422, opened 2018-06-14
Feature/netstandard2 (NETCore 2.1 support)

kll:
Changing the target to netstandard2.0 allows upgrading to the latested versions of LibGit2Sharp and its associated LibGit2Sharp.NativeBinaries packages.

This allows the dotnetcore version to run on all the [.NET Core 2.1 supported OS](https://github.com/dotnet/core/blob/master/release-notes/2.1/2.1-supported-os.md) versions. 

I left the unit tests targeting the full desktop framework but bumped them to net461 so they could consume the netstandard2.0 GitVersionCore project. The reason I didn't convert everything over is that there are several nuget packages used in the test projects that haven't been updated in a long time and do not have any netstandard support at all.

PR #1528, opened 2018-11-09
Publish as .NET Core Global Tool

arturcic:
Implements #1464 

I decided to implement it as a separate PR not as part of #1422 as that PR is blocked, and we can have a global tool without upgrading to .net core 2.1.

I have used the `GitVersion.Tool` as the package ID, even thought there is a GitVersion.Tools - this one has one version and a few downloads and I don't think it will be too confusing.

@gep13 , @dazinator  what do you think?


src/GitVersionExe/GitVersion.Tool.csproj
```diff
@@ -0,0 +1,15 @@
+<Project>
+  <PropertyGroup>
+    <PackAsTool>true</PackAsTool>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+    <ToolCommandName>dotnet-gitversion</ToolCommandName>
+    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
```
dazinator: nitpick: Guessing this `GenerateAssemblyInfo` prop can be deleted here, as it's already imported from GitVersionExe.csproj ?

src/GitVersionExe/GitVersionExe.csproj
```diff
@@ -6,7 +6,7 @@
     <OutputType>Exe</OutputType>
     <RootNamespace>GitVersion</RootNamespace>
     <AssemblyName>GitVersion</AssemblyName>
-    <TargetFrameworks>net40;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks Condition=" '$(PackAsTool)' != 'true'">net40;netcoreapp2.0</TargetFrameworks>
```
dazinator: Very minor thing, I understand why you have added this condition, shame it is necessary though, from a purity perspective I would find it nicer if this project file was oblivious to global tool builds, and all global tool build stuff was handled in `GitVersion.Tool.csproj`. However one condition like this is no biggie really. It makes me wonder though if it would work if you removed this condition , and then in GitVersion.Tool.csproj, after the `GitVersion.Tool.proj` file is imported, redeclare the TargetFrameworks property there - thinking it would override the imported value from here..
However this is not a blocker for merging this.

src/GitVersionExe/GitVersion.Tool.csproj
arturcic: You're right, I'll change that

src/GitVersionExe/GitVersionExe.csproj
arturcic: I have changed this by importing the project first, and then overriding with the netcoreapp2.1 for the global tool

arturcic:
@dazinator  thanks for the feedback, I made the changes, if you consider them ok I'll squash the commits and push them as one commit before merging

dazinator:
LGTM!

arturcic:
ok then, squashing the changes

arturcic:
It's ready for the merge. Looking forward to see the global tool published.

dazinator:
Merging, thanks @arturcic

dazinator:
@arturcic might also be worth adding this here: https://github.com/natemcmaster/dotnet-tools/blob/master/README.md

arturcic:
Yeah, that would be a great addition. Thanks

arturcic:
Here it is https://github.com/natemcmaster/dotnet-tools/pull/84

arturcic:
@dazinator It was merged

Crono1981:
I can't seem to be able to run it on my end.

```
dotnet tool install GitVersion.Tool -g --version 5.0.0-beta3-4
dotnet gitversion
```

I get the following error:

```
No executable found matching command "dotnet-gitversion"
```

I should mention that this is the first time I ever used `dotnet tool` so I might be missing something obvious. Any clue?

gep13:
@Crono1981 once installed globally, you should be able to simply run:

`dotnet-gitversion`

there is no need to run `dotnet` first.

Run the command `dotnet tool list -g` to see all the tools that you have installed, and the commands that can be ran to execute them.

Hope that helps!

Crono1981:
Thanks @gep13, I tried `dotnet-gitversion` too but then I get a message saying that it doesn't exist. Even though I do see it when I run `dotnet tool list -g`.

Maybe something is wrong with my PATH environment variable? I'll look it up later when I'm back to my office.

Anyway, now at least I'm sure that the correct way to run the tool is `dotnet-gitversion`. Thanks!

Crono1981:
Yep, bad PATH value. Sorry I even bothered you with this. 😞🤦‍♂️

gep13:
@Crono1981 no need to apologise. Glad you got it working!

PR #1534, opened 2018-11-16
nuget update

arturcic:
Updated the nuget packages as part of csproj cleanup. Have updated to the latest possible, except some that have latest targeting netstandard 2.0 (those probably will be updated as part of the PR for netstandard2)


arturcic:
@dazinator could you please have a look?

arturcic:
@dazinator, the required changes included

arturcic:
I'll squash as one commit and then you can merge it

dazinator:
My only concern is the update to msbuild dependencies. I have found them unusually fragile in the past. However if all tests pass then thats the only solid metric we have.I still think its brave to update them ;-)

arturcic:
Thanks, just trying to move the project forward and preparing the ground for the PR #1422

PR #1544, opened 2018-12-03
Updating libgit2sharp binaries (removes curl/zlib dependency)

pauldotknopf:
See [this](https://github.com/libgit2/libgit2sharp.nativebinaries/pull/84) and [this](https://github.com/libgit2/libgit2sharp.nativebinaries/pull/82).

Using ```gitversion``` on Ubuntu 18.10 currently is broken (because of libcurl4). Updating to this version of the libgit2sharp native binaries should resolve my issue.

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -2,6 +2,6 @@
  <PropertyGroup>
    <PackageVersion_GitToolsCore>1.3.1</PackageVersion_GitToolsCore>
    <PackageVersion_YamlDotNet>5.2.1</PackageVersion_YamlDotNet>
-   <PackageVersion_LibGit2SharpNativeBinaries>[1.0.185]</PackageVersion_LibGit2SharpNativeBinaries>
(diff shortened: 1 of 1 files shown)


pauldotknopf:
Ah, it looks like https://github.com/GitTools/GitTools.Core needs to be updated. I'll look into that.

pauldotknopf:
It looks like the latest libgit2sharp uses netstandard2.0 and net461.

This would affect both GitTools.Core and this repo.

Is this a no go? Would you guys want to drop netstandard1.3 and net40?

kll:
See #1422

asbjornu:
@pauldotknopf, thanks fort this! I assume #1422 will change the prerequisites for your work here so perhaps wait until that's merged, rebase and move from there?

pauldotknopf:
Yup, I'll wait.

pauldotknopf:
In order for this to work, we'd have to drop net461 and target only netstandard.

pauldotknopf:
Actually, it looks like it was already done. Sweet!

PR #1571, opened 2019-01-19
CentOS 7 docker based image

mr-exz:
added Dockerfile with: 
- clear cenots/7
- dotnet core 2.2
- gitversion 4.0.0


asbjornu:
Thanks for this, @mr-exz. Is it possible to test the image somehow?

mr-exz:
> Thanks for this, @mr-exz. Is it possible to add an automatic test that is run by Travis or similar, somehow?

i will check how you did it, and will try fit my Dockerfile into build process

mr-exz:
@asbjornu @arturcic pls check cahnges

src/Docker/linux/centos/Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM centos:7
+LABEL maintainers="GitTools Maintainers"
+
+ENV DOTNET_VERSION='2.2'
```
arturcic: Could you set the version to 2.1 to have the same version as for other docker images? I'll make some changes later to add support for building both 2.1 and 2.2 versions

build.cake
```diff
@@ -622,6 +623,7 @@ Task("Publish-DockerHub")
     {
         DockerPush("linux", "dotnetcore", parameters);
         DockerPush("linux", "fullfx", parameters);
+        DockerPush("linux", "centos", parameters);
```
asbjornu: Please excuse my ignorance, but wow is this going to affect the naming on [DockerHub](https://hub.docker.com/u/gittools)? I don't quite understand where the existing names repositories [`gitversion-fullfx`](https://hub.docker.com/r/gittools/gitversion-fullfx), [`gitversion-dotnetcore`](https://hub.docker.com/r/gittools/gitversion-dotnetcore), [`gitversion`](https://hub.docker.com/r/gittools/gitversion) and [`libgit2sharp-mono`](https://hub.docker.com/r/gittools/libgit2sharp-mono) come from and how they relate to these `DockerPush()` statements. Will `centos` become a new `tag` in the `gitversion-dotnetcore` repository?

arturcic: > @asbjornu  [`gitversion`](https://hub.docker.com/r/gittools/gitversion) and [`libgit2sharp-mono`](https://hub.docker.com/r/gittools/libgit2sharp-mono) are already obsolete, only fullfx and dotnetcore are maintained. 

As for centos, that is right, the current changes will not create a tag under dotnetcore - good catch, we'll need to adjust the scripts to allow several distros of linux under dotnetcore and fullfx

arturcic: this is the method where the tags are created https://github.com/GitTools/GitVersion/blob/master/build/utils.cake#L174

asbjornu: I see. I find it a bit unfortunate that we have 4 different repositories for variations of the same executable. Can we get rid of every repository that isn't called simply `gitversion` and follow [NGINX](https://hub.docker.com/_/nginx?tab=tags)' naming scheme for labels to distinguish different versions of the image? That would give us one repository with tags following the scheme `{version}-{platform}-{framework}`, for example:

- `4.0.0-centos`
- `4.0.1-beta2-debian-fx`
- `4.0.0-beta15-windows-dotnetcore`

I know doing anything about this scheme is outside the scope of this PR, I just want to have a discussion on this so I know which options we have.

arturcic: I suggest creating a separate issue, I can then implement it, and after that we can adjust this PR to the new scheme.

arturcic: @asbjornu please create an issue with the proposed scheme and we can discuss it there

asbjornu: Sure, I've created #1572 for further discussion.

arturcic:
@asbjornu can we postpone merging this till I get #1572 implemented ?

asbjornu:
@arturcic, yep, that's the plan. 😃

arturcic:
@mr-exz the #1572  was implemented and merged, can you rebase on top of master and add the CentOS support? Do you think we can also add Fedora support?

mr-exz:
> @mr-exz the #1572 was implemented and merged, can you rebase on top of master and add the CentOS support? Do you think we can also add Fedora support?

i think we can, i'll take a look

arturcic:
@mr-exz , I think you should move the Dockerfile to the src/Docker/linux/centos/netcoreapp2.1 folder similar to debian version. This way we'll be able to add net core 2.2 version as well

arturcic:
Looks great! Thank you for your contribution!

mr-exz:
Not at all

mr-exz:
@arturcic can you provide link on new centos7/fedora27 docker images link ?

arturcic:
Yesterday when the PR was merged the Linux build has failed, some unit tests for some reason failed, now they work again. The tags can be found at https://hub.docker.com/r/gittools/gitversion/tags

arturcic:
@mr-exz can you confirm that those docker images are valid and work as expected as I don't have time to test them these days?

mr-exz:
i will, in process

arturcic:
Thanks

mr-exz:
For me working fine
`
# docker run --rm -v "$(pwd):/repo" gittools/gitversion:5.0.0-linux-centos7-netcoreapp2.1 /repo
# docker run --rm -v "$(pwd):/repo" gittools/gitversion:5.0.0-linux-fedora27-netcoreapp2.1 /repo
`

arturcic:
great, thanks for the fedora image as well.

mr-exz:
Not at all :)

arturcic:
Btw, are you in a position to add a new PR for netcoreapp2.2? for linux images? That would be a great addition

mr-exz:
ok will do, it shouldn't be so difficult

asbjornu:
Hm, I just discovered that the tag `gittools/gitversion:5.0.0-linux-centos7-netcoreapp2.1` has the version number `5.0.0` which indicates it’s a stable version. Since it isn’t, that’s a problem, no? Can we get the pre-release label into there so we only tag stable releases as such?

arturcic:
There is also a tag `gittools/gitversion:5.0.0-beta2-18-linux-centos7-netcoreapp2.1` that is the same as the one you mentioned. Basically on every new build we re-tag the 5.0.0-linux-centos7-netcoreapp2.1 to the newest version, as well as we tag with a pre-release tag. That means the pre-release tag will not change, but the the other tag gets updated with every new commit. (it's is similar to how latest tag is used in other projects, meaning it's always pointing to the latest, stable or not stable)

arturcic:
@asbjornu, you can review this https://github.com/GitTools/GitVersion/blob/master/build/utils.cake#L163-L196

arturcic:
And the versioning here https://github.com/GitTools/GitVersion/blob/master/build/version.cake#L23-L24

mr-exz:
> Btw, are you in a position to add a new PR for netcoreapp2.2? for linux images? That would be a great addition

https://github.com/GitTools/GitVersion/pull/1633

asbjornu:
@arturcic, thanks for the explanation. That sounds sensible to me. 👍

PR #1584, opened 2019-02-15
Parallelize unit tests

asbjornu:
This is a rebase of #1384 to see if the failing tests can be fixed somehow. Resolved #1384.

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -66,40 +66,47 @@ Setup<BuildParameters>(context =>
 
 Teardown<BuildParameters>((context, parameters) =>
 {
-    Information("Starting Teardown...");
-
-    Information("Repository info : IsMainRepo {0}, IsMainBranch {1}, IsTagged: {2}, IsPullRequest: {3}",
-        parameters.IsMainRepo,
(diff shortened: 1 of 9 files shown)


arturcic:
@asbjornu can you squash the commits as one commit?

asbjornu:
@arturcic: I'm personally not a fan of squashing commits, since we loose the thought process and meaning behind many of the code changes. I find the commit messages in 9d001910d39b9a25f6905bb7b19e425053ea5bb1, 4c4c08d13024f67d3a6eb43dacc868044ff53cbf and 63cb9f132e05402ca5d3ac96c7a8a5eaef9eb2f4 valuable in the case similar problems surface in the future. If we squash, those messages go away or get attached to a single large commit that also deals with other stuff, so the intent becomes unclear.

I'm all for rebasing a PR to get rid of merge-commits (updates) from `master`, though.

arturcic:
Agree with the point, but in this particular case it's about testing the CI if the build fails, that's why I thought better to have one commit that represents the actual change to the code base that improves it

Issue #1054, opened 2016-10-02
Rename tag to label


asbjornu:
In GitVersion, a [`tag`](http://gitversion.readthedocs.io/en/latest/configuration/#tag) is not the same thing as a Git tag, which I find a bit confusing. Wanting to reference a **Git tag** and use it as the value of the **pre-release tag** makes it even more confusing. It is currently not possible to do this, but [discussions on Gitter](https://gitter.im/GitTools/GitVersion?at=57f0fb76b0ff456d3ada81a4) might indicate that something like this would be very handy:

``` yaml
branches:
  release:
    tag: useLatestTagLabel
```

However, it would be easier to understand if we renamed the pre-release `tag` to `label`:

``` yaml
branches:
  release:
    label: useFromLatestTag
```

This would allow for alternating pre-release labels on the same branch, such that a `beta` could be promoted to `rc` and stay at `rc` without having to tag every commit on that branch.

Thoughts?

JakeGinnivan:
Maybe just `pre-release`, that is what semver.org calls it?

We can probably include this change in v4 if we move on it soon. Thoughts @asbjornu

asbjornu:
@JakeGinnivan: What SemVer says about this that I can find, is:

> Additional **labels** for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

And:

> A pre-release version MAY be denoted by appending a hyphen and a series of dot separated **identifiers** immediately following the patch version.

So it seems to call these pre-release strings for **labels** or **identifiers**. Perhaps `pre-release-identifier` or `pre-release-label` could work? I think just `label` works too, but `identifier` is too generic.

asbjornu:
If no one objects, I'd like to conclude this discussion with renaming `tag` to `label`. PRs welcome. 😃 🙏

HHobeck:
> If no one objects, I'd like to conclude this discussion with renaming tag to label. PRs welcome.

I like your idea to rename it to label. :)

PR #1600, opened 2019-02-20
Pre-release label in git tag

### LINKED ISSUE #1255: The same version generated twice
I have stumbled on smth that looks like a bug to me .

Given this configuration (abbreviated from my unit tests for gitversion.yml):

```
next-version: 5.0
branches:
  master:
    regex: master
    tag: beta
    increment: Patch
    mode: ContinuousDeployment
```
I am getting this in my unit test log:

```
0001 Initial commit - expect beta.0
     git init .
     commit = True
          SemVer:               5.0.0-beta.0
          FullSemVer:           5.0.0-beta.0
          AssemblySemVer:       5.0.0.0
          AssemblySemFileVer:   5.0.0.0
          InformationalVersion: 5.0.0-beta.0+Branch.master.Sha.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

0002 No committed changes - expect beta.0
     commit = False
          SemVer:               5.0.0-beta.0
          FullSemVer:           5.0.0-beta.0
          AssemblySemVer:       5.0.0.0
          AssemblySemFileVer:   5.0.0.0
          InformationalVersion: 5.0.0-beta.0+Branch.master.Sha.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

0003 Commit changes - expect beta.1
     commit = True
          SemVer:               5.0.0-beta.1
          FullSemVer:           5.0.0-beta.1
          AssemblySemVer:       5.0.0.0
          AssemblySemFileVer:   5.0.0.0
          InformationalVersion: 5.0.0-beta.1+Branch.master.Sha.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

0004 Another commit - expect beta.2
     commit = True
          SemVer:               5.0.0-beta.2
          FullSemVer:           5.0.0-beta.2
          AssemblySemVer:       5.0.0.0
          AssemblySemFileVer:   5.0.0.0
          InformationalVersion: 5.0.0-beta.2+Branch.master.Sha.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

0005 Create tag for the release candidate 1, no changes committed after the tag
     git tag v5.0.0-rc.1
     commit = False
          SemVer:               5.0.0-rc.1
          FullSemVer:           5.0.0-rc.1
          AssemblySemVer:       5.0.0.0
          AssemblySemFileVer:   5.0.0.0
          InformationalVersion: 5.0.0-rc.1+Branch.master.Sha.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx

0006 Simple commit - expect 5.0.1-beta.1
     commit = True
          SemVer:               5.0.0-beta.1
          FullSemVer:           5.0.0-beta.1
          AssemblySemVer:       5.0.0.0
          AssemblySemFileVer:   5.0.0.0
          InformationalVersion: 5.0.0-beta.1+Branch.master.Sha.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

Note that in step 0003 and 0006 gitversion has returned the same semantic version:  *5.0.0-beta.1*
And then it continues the count, duplicating beta.2 and so on.

I see two problems with the observed behaviour:

- Gitversion generated non-unique version numbers which violates a pronciple of unique identification
- Gitversion generated a lower version after the higher one, which breaks correct sorting order (*5.0.0-rc.1* is higher than *5.0.0-beta.1*)

There is also an inconsistency: in step 0001 it started to count beta from 0 (which surprised me), but after the RC in step 0006 it started to count beta from 1.

In addition (I do not show this in the log to keep things simpler) after +semver:fix even though the patch number was increased, the beta suffix did not restart the count, which IMHO is not the right behaviour.

I am testing against version: 
`4.0.0-beta.12+1315.Branch.master.Sha.9c70946f68973b3b31b2861913c247fbe6c3f0ec`

I would expect GitVersion to figure out that it needs to increase the version as every 5.0.0-beta would be lower than 5.0.0-rc.1, so given my configuration it should bump the patch number and re-start the beta count. Using this logic, the expected version would be: *5.0.1-beta.1*

Any tips on how to get the configuration right will be appreciated, as I need to come up with working configuration for a team adopting this tool. I have only started to learn about gitversion; maybe I am not using it correctly?

asbjornu:
Rebase and conflict resolvement of #1261. Resolves #1261, fixes #1255.

### DIFF
--- a/src/GitVersion.Core.Tests/IntegrationTests/DevelopScenarios.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/DevelopScenarios.cs
@@ -556,4 +556,17 @@ public class DevelopScenarios : TestBase
         // ✅ succeeds as expected
         fixture.AssertFullSemver("1.0.0", configurationBuilder.Build());
     }
+
+    [Test]
+    public void ShouldProvideTheCorrectVersionEvenIfPreReleaseLabelExistsInTheGitTag()
+    {
+        using (var fixture = new EmptyRepositoryFixture())
+        {
+            fixture.Repository.MakeACommit();
+            fixture.ApplyTag("1.0.0-oreo.1");
+            fixture.BranchTo("develop");
+            fixture.Repository.MakeACommit();
+            fixture.AssertFullSemver("1.1.0-alpha.1");
+        }
+    }
 }
--- a/src/GitVersion.Core.Tests/IntegrationTests/DevelopScenarios.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/DevelopScenarios.cs
@@ -560,13 +560,11 @@ public class DevelopScenarios : TestBase
     [Test]
     public void ShouldProvideTheCorrectVersionEvenIfPreReleaseLabelExistsInTheGitTag()
     {
-        using (var fixture = new EmptyRepositoryFixture())
-        {
-            fixture.Repository.MakeACommit();
-            fixture.ApplyTag("1.0.0-oreo.1");
-            fixture.BranchTo("develop");
-            fixture.Repository.MakeACommit();
-            fixture.AssertFullSemver("1.1.0-alpha.1");
-        }
+        using var fixture = new EmptyRepositoryFixture();
+        fixture.Repository.MakeACommit();
+        fixture.ApplyTag("1.0.0-oreo.1");
+        fixture.BranchTo("develop");
+        fixture.Repository.MakeACommit();
+        fixture.AssertFullSemver("1.1.0-alpha.1");
     }
 }
--- a/src/GitVersion.Core.Tests/IntegrationTests/MainScenarios.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/MainScenarios.cs
@@ -406,4 +406,36 @@ public class MainScenarios : TestBase
         // ✅ succeeds as expected
         fixture.AssertFullSemver(expectedSemanticVersion, configuration);
     }
+
+    [Test]
+    public void Bug1255()
+    {
+        var config = new Config
+        {
+            NextVersion = "5.0",
+            Branches =
+            {
+                {
+                    "master", new BranchConfig
+                    {
+                        Regex = "master",
+                        Tag = "beta",
+                        Increment = IncrementStrategy.Patch,
+                        VersioningMode = VersioningMode.ContinuousDeployment
+                    }
+                }
+            }
+        };
+        using (var fixture = new EmptyRepositoryFixture())
+        {
+            fixture.Repository.MakeACommit();
+            fixture.AssertFullSemver(config, "5.0.0-beta.0"); // why not "5.0.0-beta.1"?
+            fixture.Repository.MakeACommit();
+            fixture.AssertFullSemver(config, "5.0.0-beta.1");
+            fixture.Repository.MakeATaggedCommit("v5.0.0-rc.1");
+            fixture.AssertFullSemver(config, "5.0.0-rc.1");
+            fixture.Repository.MakeACommit();
+            fixture.AssertFullSemver(config, "5.0.1-beta.1"); // test fails here, it generates "5.0.0-beta.1" which is not unique and lower than "5.0.0-rc.1"
+        }
+    }
 }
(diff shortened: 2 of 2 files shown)


ruhullahshah:
The reason for the observed incorrect version number is that GitVersion is somehow sensitive to tags are of the form `v5.0.0-rc.1` or `v5.0.0-beta.1`. In case the `PreReleaseTag` is dropped from the Git tag, for instance naming the tag simply as `v5.0.0` leads to a successful execution of the testcase reported here.

I see that the chosen versioning strategy is correct. however, this part of the [code](https://github.com/GitTools/GitVersion/blob/14b736584840bae5878b565e7d92caea6afc6c7d/src/GitVersionCore/SemanticVersion.cs#L293) causes an incorrect version increment, even though it is set to `Patch`. I think that this part of the code would generally hold true, for instance, bumping the version number on `develop` from `5.0.0-alpha.1` to `5.0.0-alpha.2`, but in this case it is doing the wrong thing.

Any thoughts?

asbjornu:
Thanks for digging into this, @ruhullahshah. I'm a bit confused, since the tag should only take presedence if it is on the current commit. For the test case in this PR, there is a commit after the tag `v5.0.0-rc.1` is added, so according to my understanding, we should end up on [this `case VersionField.Patch`](https://github.com/GitTools/GitVersion/blob/14b736584840bae5878b565e7d92caea6afc6c7d/src/GitVersionCore/SemanticVersion.cs#L308). However, that does not seem to be the case.

ruhullahshah:
@asbjornu  You are welcome.  Your argument is correct, if the tag name did not have `PreReleaseTag` in its name. However, (this)[https://github.com/GitTools/GitVersion/blob/14b736584840bae5878b565e7d92caea6afc6c7d/src/GitVersionCore/SemanticVersion.cs#L293] line of code will not allow it to pass through. Try removing the `-rc*` from the test and everything works as expected.

asbjornu:
Aha, I was caught yet again by the naming confusion discussed in #1054. I instinctively think of `git tag` every time I read `PreReleaseTag`, which is why I argue in #1054 that it should be called `PreReleaseLabel` instead. Okay, so in the presence of a pre-release label, the label's `Number` will be incremented instead of applying the chosen increment strategy. I'm not sure why that should be correct for any other commits than the one where the `git tag` is applied, though. So my thought is still the same: Why should the pre-release label affect the increment strategy at all?

asbjornu:
Also, when thinking of how the test is formulated and https://github.com/GitTools/GitVersion/issues/1255#issuecomment-465548324, isn't `mode: Mainline` more appropriate for the given use case? I'll test how it affects the version number.

asbjornu:
Nope, [`mode: Mainline` does not support pre-release labels](https://ci.appveyor.com/project/GitTools/gitversion/builds/22682682#L253): `System.NotSupportedException : Mainline development mode doesn't yet support pre-release tags on master`.

I still think `mode: Mainline` is correct for the use case, but we don't support the given use case with `Mainline`. 🤷‍♂️

ruhullahshah:
Your questions are valid and relate more to the specifications and usage of GitVersion. It "might" be true that in this particular case, GitVersion is abused. However,  I would like to demonstrate the general problem with a simpler test case:
```csharp
[Test]
public void ShouldProvideTheCorrectVersionEvenIfPreReleaseLabelExistsInTheGitTag()
{
    using (var fixture = new EmptyRepositoryFixture())
    {
        fixture.Repository.MakeACommit();
        fixture.ApplyTag("1.0.0-oreo.1");
        fixture.BranchTo("develop");
        fixture.Repository.MakeACommit();
        fixture.AssertFullSemver("1.1.0-alpha.1");
    }
}
```
This simple test fails due to the presence of a `PreReleaseLabel` (I agree that this is a better naming convention)  in the `git tag`

asbjornu:
Great point, @ruhullahshah. I've added your test to this PR as further proof that something needs to be done about this. I'm just not sure what.

ermshiperete:
#1844/ #1845 seem to be related

arturcic:
Closed in favor of #3445

Issue #1572, opened 2019-01-22
Docker naming scheme


asbjornu:
As discussed in https://github.com/GitTools/GitVersion/pull/1571#discussion_r249751741, it would be beneficial for us if [GitTools](https://hub.docker.com/u/gittools) didn't have 4 different repositories for GitVersion:

- [`gitversion-fullfx`](https://hub.docker.com/r/gittools/gitversion-fullfx)
- [`gitversion-dotnetcore`](https://hub.docker.com/r/gittools/gitversion-dotnetcore)
- [`gitversion`](https://hub.docker.com/r/gittools/gitversion)
- [`libgit2sharp-mono`](https://hub.docker.com/r/gittools/libgit2sharp-mono)

I believe having these four options publicly available makes it unnecessary hard to choose which Docker image to use and the naming standard of the tags in each repository is also a bit hard to understand.

I think it would be beneficial to us and our users if we ended up with just one Docker image and associated repository, simply called [`gitversion`](https://hub.docker.com/r/gittools/gitversion) and that the tag naming scheme of this repository followed that of [NGINX](https://hub.docker.com/_/nginx?tab=tags), which would give us the tag format `{version}-{platform}-{framework}`. For example:

- `4.0.0-centos`
- `4.0.1-beta2-debian-fx`
- `4.0.0-beta15-windows-dotnetcore`
- `latest-ubuntu`
- `latest-windows`
- `stable-ubuntu`
- `stable-windows`

I have a whole heap of unanswered questions related to this:

- Is this at all possible?
- Can we obsolete repositories so they aren't listed, but still have their contained images be usable for those who are using them?
- Can we clean up the **visible** tags of a repository without making the images for those tags unavailable?
- Will the proposed naming scheme support all the different variations we want and need?
- Are there anything I haven't thought of?

arturcic:
> * Is this at all possible?
I think we can.

> * Can we obsolete repositories so they aren't listed, but still have their contained images be usable for those who are using them?
As far as I know, on Dockerhub there is no option to unlist a repo - it is either delete the repo or make it private, in any case the images will not be available for consumption externally

> * Can we clean up the **visible** tags of a repository without making the images for those tags unavailable?
There is a different option, we can keep the images from `gitversion-dotnetcore` and `gitversion-fullfx` by re-publishing to `gitversion` repository all the tags, using the new scheme - for that we'll need a script to pull them all and re-tag with new tags and then push them to new repo.

[...]

PR #1603, opened 2019-02-25
Docker naming scheme

arturcic:
Implements #1572 

I have already migrated the old images from the gittools/gitversion-dotnetcore and gittools/gitversion-fullfx to gittools/gitversion and adapted the tag names to use the new scheme


arturcic:
In the `README.md` I have formatted the table of the artifacts, so now there is only one link to dockerhub, previously there were 2.

You should `Hide whitespace changes` in the Diff Settings when code reviewing

asbjornu:
I see. I also did not understand the removal of the Travis badge, but I notice it has stopped reporting builds. Any idea why?

arturcic:
There is some issue building with Travis for the last month, for Linux image most often, and because on Azure Pipelines we also use the linux and macos VMs I decided to disable building on Travis. The second reason for disabling Travis was it takes much longer to build with it as it first needs to install mono, and it usually takes ~60% of the build time. And from the point of view of Azure pipeline I have one configuration for all 3 OSes (and later planning to add the validation Docker jobs as well).

arturcic:
Still I kept the travis configuration in code, in case we need to go back with Travis

asbjornu:
Well, with [this](https://twitter.com/asbjornu/status/1098873841273708544) in mind, I think it's wise to move away from Travis altogether, I was just wondering. Azure Pipelines sure seems like a great solution. 👍

arturcic:
We're using Azure Pipelines here for Gitversion for half a year already I guess, and it seems to work properly, especially in the context of Docker building

arturcic:
Hey @asbjornu , do you think we can already merge this PR?

asbjornu:
405c7008ad45472097ec80b4a6812ef4776a7c9e caused the [Windows build on Azure DevOps to ](https://dev.azure.com/GitTools/GitVersion/_build/results?buildId=240) to fail:

```
Executing: "C:/Program Files/Docker/docker.exe" build --build-arg "contentFolder=/content" --file "src/Docker/windows/windowservercore/net472/Dockerfile" --rm --tag "gittools/gitversion:5.0.0-windows-windowservercore-net472" --tag "gittools/gitversion:5.0.0-beta1-79-windows-windowservercore-net472" src/Docker/windows/windowservercore/net472
unable to prepare context: unable to evaluate symlinks in Dockerfile path: GetFileAttributesEx D:\a\1\s\src\Docker\windows\windowservercore\net472\Dockerfile: The system cannot find the file specified.
An error occurred when executing task 'Docker-Build'.
```

A bit strange that the build was successful in the PR, but not on `master`. Ideas, @arturcic?

arturcic:
When there is a PR we don't run publish Docker, that calls Docker Build, that's why it did not fail. The issue was the misspelling of the distro, fixed in another commit

asbjornu:
Aha, excellent.

pauldotknopf:
Did the you guys delete docker images?

```manifest for gittools/gitversion:v4.0.0-beta.12 not found```

arturcic:
I guess we did when migrating from other repositories. Can you use the 4.0.0 tag instead?

pauldotknopf:
Is 4.0.0 still mono?

arturcic:
You can use 4.0.0-linux-debian-net472, as this one is on top of mono image, the other netcoreapp is for dotnetcore

PR #1605, opened 2019-02-25
Update GitVersionBuild.targets

### LINKED ISSUE #1604: NuGetPackOutput has incorrect version when project is being build using Rebuild with GitVersionTask 4.0
GetVersion task runs some time later after _GetOutputItemsFromPack (because it is run by _CleanPackageFiles after Clean target), and it causes NuGetPackOutput to have incorrect file paths  (e.g.  something.1.0.0.nupkg instead of .0.1.0.nupkg). However because the actual output path is calculated again in the PackTask after build, the real package is named with correct version.
Because of that the package is not being properly deleted during Clean, and I can't use the output package path in other targets.
Here are screenshots from MSBuild Structured Log Viewer:
image

image

Mephistofeles:
Added _GetOutputItemsFromPack to GetVersion target's BeforeTargets items. Fixes #1604.


arturcic:
@Mephistofeles, I merged this change as it seems to be a safe change, and will fix the issue you created. Thank you for the contribution, sorry for late review.

PR #1613, opened 2019-03-01
Use NameWithoutRemote() instead of FriendlyName

### LINKED ISSUE #1606: GitVersion cannot find remote master or develop from inherited branch config
The most recent beta release changes the default branch regexes in a way that prevents them from matching remote branches if GitVersion can't find the parent.  

https://github.com/GitTools/GitVersion/blob/14b736584840bae5878b565e7d92caea6afc6c7d/src/GitVersionCore/BranchConfigurationCalculator.cs#L130

The use of Branch.FriendlyName with the default master regex of ^master$ means that it will never match a remote branch.

see also
https://github.com/GitTools/GitVersion/pull/1410

asbjornu:
Fixes #1606.

### DIFF
--- a/src/GitVersionCore/BranchConfigurationCalculator.cs
+++ b/src/GitVersionCore/BranchConfigurationCalculator.cs
@@ -192,14 +192,14 @@ namespace GitVersion
             }
             else if (branches.Count > 1)
             {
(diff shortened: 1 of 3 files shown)


arturcic:
LGTM

Issue #1554, opened 2018-12-17
MSBUILD : Task factory warning NMSBT010: There is a mismatch between SDK NuGet version (4.9.2) and the NuGet version the task factory was compiled against (4.8.0). There might occur some exotic errors.


czeslav87:
Hello,

Issue strikes back with new dotnet core sdk (i'm using 2.2.101). Warning is followed by an error:

```
MSBUILD : Task factory error NMSBT001: Exception in initialization: System.MissingMethodException: Method not found: 'Void NuGet.Commands.RestoreRequest..ctor(NuGet.ProjectModel.PackageSpec, NuGet.Commands.RestoreCommandProviders, NuGet.Protocol.Core.Types.SourceCacheContext, NuGet.Common.ILogger)'. [C:\Users\mjasni\source\repos\PokInvoices\InvoicesApp\InvoicesApp.csproj]
MSBUILD : Task factory error NMSBT001:    at UtilPack.NuGet.BoundRestoreCommandUser.PerformRestore(ValueTuple`2[] targets, CancellationToken token) [C:\Users\mjasni\source\repos\PokInvoices\InvoicesApp\InvoicesApp.csproj]
MSBUILD : Task factory error NMSBT001:    at System.Runtime.CompilerServices.AsyncMethodBuilderCore.Start[TStateMachine](TStateMachine& stateMachine) [C:\Users\mjasni\source\repos\PokInvoices\InvoicesApp\InvoicesApp.csproj]
MSBUILD : Task factory error NMSBT001:    at UtilPack.NuGet.BoundRestoreCommandUser.PerformRestore(ValueTuple`2[] targets, CancellationToken token) [C:\Users\mjasni\source\repos\PokInvoices\InvoicesApp\InvoicesApp.csproj]
MSBUILD : Task factory error NMSBT001:    at UtilPack.NuGet.BoundRestoreCommandUser.RestoreIfNeeded(CancellationToken token, ValueTuple`2[] packageInfo) [C:\Users\mjasni\source\repos\PokInvoices\InvoicesApp\InvoicesApp.csproj]
MSBUILD : Task factory error NMSBT001:    at UtilPack.NuGet.MSBuild.NuGetTaskRunnerFactory.Initialize(String taskName, IDictionary`2 parameterGroup, String taskBody, IBuildEngine taskFactoryLoggingHost) [C:\Users\mjasni\source\repos\PokInvoices\InvoicesApp\InvoicesApp.csproj]
[...]

Issue #1537, opened 2018-11-20
Stability of MSBuild integration for future .NET Core releases


stazz:
Hi,

When helping with issue #1503 the concern was arised that any new .NET Core releases have potential to break the UtilPack task factory used by GitVersionTask. This stems from the fact that while it is possible to lock down the NuGet assemblies version on .NET Desktop (and Mono), it is not possible to do so on .NET Core. This because NuGet assemblies are part of Trusted Platform Assembly list when `dotnet build` command is invoked, and it is not possible to load another version of them by the task factory. This will cause crashes as new versions of .NET Core introduce potentially binary-incompatible changes to the used APIs of NuGet assemblies.

So the root problem can be solved if there is a way to lock down NuGet assembly version on .NET Core `build` environment. One such way is to use .NET Core Global Tools, since they run using plain `dotnet` instead of `dotnet build`.

It seems that you are publishing GitVersionTask already as .NET Core Global Tool, according to #1463 . Maybe the most straightforward way is to just invoke the GitVersionTask as global tool from .targets file, installing it if it is missing from local system? Would it produce all information that is required by the .targets file?

This would essentially split the behaviour of GitVersionTask in MSBuild usage into two parts:
- On .NET Desktop (and Mono) environment, the UtilPack task factory would be continued to be used
- On .NET Core environment, the global tool would be installed if needed, and invoked, parsing the potential output parameters from the output or file. The UtilPack task factory would not be used at all.

arturcic:
Thanks @stazz for detailed explanation, we might consider the approach you suggested

dazinator:
Hi @stazz it's definately something to consider.

Couple of points thinking out loud. A csproj file built with dot net would need the global tool installed, but built with msbuild it wouldnt. Today we can build the same csproj file with either, there are no platform specific instructions which is nicer imho but probs not a deal breaker.

If we invoked the global tool as part of the build presumably we would have to pass any msbuild properties needed to gitversion and then get its STDOUT and then set the same msbuild properties during the build that the task sets today. This means the task could be dumbed down, and dependencies removed (presumably we'd need to remove all dependencies because we will no longer have utilpack to restore them for us). 

We'd have to think whether we need to extend gitversion command line for this mode of operation (i.e to support gitversiontask in its new endeavours)

[...]

PR #1634, opened 2019-03-10
Updating UtilPack.NuGet.MSBuild to NuGetUtils.MSBuild.Exec

stazz:
Hi,

As discussed in #1554 and #1537 , there is a need to change UtilPack's way of executing everything in MSBuild process, as that causes big problems with NuGet versioning (binary-incompatible changes even on minor version change, and the `dotnet build` command lock NuGet version to specific one).

This change contains various modifications in GitVersionTask project, which makes it:
- Compatible with NuGetUtils.MSBuild.Exec, and
- Independant of Microsoft.Build.Utilities.Core package.

The GitVersionTask is now invoked via separate process by task factory of NuGetUtils.MSBuild.Exec package.

I have run only very preliminary tests so far, but starting PR process already to see how automated tests fare as well.


stazz:
FYI, I've found out that the CI build process generated faulty `.(dll|exe).config` files, and `NuGetUtils.MSBuild.Exec` currently doesn't run on .NET Desktop. I'm in a process of fixing it, I found out what I was missing and I just need to write some code to patch things up. I'll keep you posted.

stazz:
For my part, the pull request is ready for review. I've tested with small test .csproj with both .NET Desktop and .NET Core, and I think Mono should be fine too, since AppDomains are no longer used anyway.

I'm not sure why the checks fail - based on messages, they seem quite random (SSL errors or npm failing).

stazz:
Also, one thing I should mention: the `NuGetUtils.MSBuild.Exec` package has now the following minimum supported frameworks:
- `.NET 4.6` for .NET Desktop, and
- `.NET Core 2.1` for .NET Core.

These can be worked on, though, if the minimum requirements are gonna cause problems.

asbjornu:
Regarding the upgrade of requirements to .NET 4.6 and .NET Core 2.1, I think it's fine. We already have breaking changes planned for 5.0.0, so I think it's a good fit.

stazz:
Glad to hear the requirements upgrade is ok! Related to #1644 , yeah, the GitVersionTask no longer uses MSBuild logging at all. In some way, this is a bit of regression, but I'm sure it can be fixed by introducing concept of verbosity to input parameters of tasks, which can they propagate it to loggers.

Right now, I can't think of any quick fixes which could bring any major improvement to NuGetUtils or to this PR. @dazinator , just ping me if you need any help with the review, I think the changes are quite straightforward.

asbjornu:
I see. Removing the integration with MSBuild might be a bummer for some users. What are your thoughts on this, @dazinator and @arturcic?

stazz:
@asbjornu The detailed discussion about this is in issues #1554 and #1537 , but to summarize:
- When `dotnet build` command is invoked, it locks down the NuGet assembly versions (marking them Trusted Platform Assemblies) so that the process can not load any NuGet assemblies with different version than the ones shipped with current version of `dotnet`.
- The ex-UtilPack, now NuGetUtils, task factory which loads and executes `GitVersionTask` uses NuGet assemblies to deduce dependencies and load them on-the-fly from their package directories within NuGet repository. This way, GitVersionTask does not need to ship all dependencies within its package.
- NuGet assemblies often get binary-incompatible changes even in minor releases, causing the task factory to fail uglily when users run `GitVersionTask` using `dotnet` with binary-incompatible NuGet assemblies than the ones which task factory was compiled against
- One solution is to always re-release the task factory whenever NuGet gets new versions, but that is not scalable solution. So, to fix this properly, the task factory was refactored to handle all NuGet-related stuff (restoring, dependency deduction, etc) in separate processes which it invokes.
- As a direct consequence of this, the `GitVersionTask` no longer has access to `IBuildEngine` instance.

When I talked with @dazinator some months ago about this, he said that losing access to MSBuild stuff (`ITask`, `IBuildEngine`) was not a big issue. When I did code changes, I didn't notice anything MSBuild-dependant, other than logging.

If access to MSBuild-specific stuff is critical, I guess you would need to give up on using NuGetUtils task factory, and ship dependencies within the NuGet package, and handle native assembly loading logic. I guess the @dazinator and @arturcic know the best route to proceed here.

asbjornu:
Thanks for the thorough explanation, @stazz. Personally, I have no problems with dropping the MSBuild integration, but I'm not sure I'm representative of GitVersion's full audience. Let's wait for @dazinator and @arturcic's input before we move ahead.

dazinator:
Sorry I haven't had a chance to look at this yet, will try and pick up this weekend!

asbjornu:
Now that #1644 is merged, I'm afraid we need you to rebase and resolve the conflicts of this PR, @stazz. To avoid the merge commits, could I ask you to please do the following?

```bash
git remote add upstream https://github.com/GitTools/GitVersion.git
git fetch upstream
git checkout stazz/NuGetUtilsUpdate
git rebase upstream/master
# Resolve conflicts, add and commit
git push --force
```

asbjornu:
Seems like that rebase didn't quite work out. No worries, I'll fix it in #1645, which will resolve this when merged.

dazinator:
@arturcic I'm holding off the review for this as it looks like there are conflicts to resolve on the proj files still. Once this PR is in good shape again i'll happily review.

asbjornu:
@dazinator, I've resolved the conflicts and performed a fresh rebase in #1645. Can you please review that?

Issue #1483, opened 2018-09-27
GitVersion crashes on Ubuntu 18.04 with System.BadImageFormatException: Method has no body


alexeyzimarev:
My build script runs fine in Docker mono image, but when I try to build on my machine, I get this error:

```
 ✘  ~/dev/infra/email   master ●  mono packages/tools/GitVersion.CommandLine/tools/GitVersion.exe 
INFO [09/27/18 11:39:09:75] Working directory: /home/alexey/dev/infra/email
INFO [09/27/18 11:39:09:76] IsDynamicGitRepository: False
ERROR [09/27/18 11:39:09:77] An unexpected error occurred:
System.BadImageFormatException: Method has no body
File name: 'LibGit2Sharp'
  at LibGit2Sharp.Core.Proxy.git_buf_free (LibGit2Sharp.Core.Handles.GitBuf buf) [0x00000] in <8cb98e2882834fdfae1774f5509e9705>:0 
  at LibGit2Sharp.Core.Handles.GitBuf.Dispose () [0x00000] in <8cb98e2882834fdfae1774f5509e9705>:0 
  at LibGit2Sharp.Core.Proxy.ConvertPath (System.Func`2[T,TResult] pathRetriever) [0x0002e] in <8cb98e2882834fdfae1774f5509e9705>:0 
  at LibGit2Sharp.Core.Proxy.git_repository_discover (LibGit2Sharp.Core.FilePath start_path) [0x00005] in <8cb98e2882834fdfae1774f5509e9705>:0 
  at LibGit2Sharp.Repository.Discover (System.String startingPath) [0x00006] in <8cb98e2882834fdfae1774f5509e9705>:0 
  at GitVersion.GitPreparer.GetDotGitDirectory () [0x0000f] in <d3d987301fcb4b25b3d107f5a758a97d>:0 
  at GitVersion.GitPreparer.GetProjectRootDirectory () [0x00048] in <d3d987301fcb4b25b3d107f5a758a97d>:0 
  at GitVersion.ConfigurationProvider.Verify (GitVersion.GitPreparer gitPreparer, GitVersion.Helpers.IFileSystem fileSystem) [0x00014] in <d3d987301fcb4b25b3d107f5a758a97d>:0 
  at GitVersion.Program.VerifyConfiguration (GitVersion.Arguments arguments, GitVersion.Helpers.IFileSystem fileSystem) [0x00023] in <d3d987301fcb4b25b3d107f5a758a97d>:0 
[...]

PR #1642, opened 2019-03-19
Upgrade LibGit2Sharp to 0.26.0

asbjornu:
Upgrade LibGit2Sharp from version `0.25.0-preview-0033` to version `0.26.0`.


arturcic:
Hey @asbjornu what about the nuspec? there are several nuspec that we use for generating the nuget packages. We need to update that as well. Still the newer version of Libgit2sharp has different layout of the dependencies that we need to check (I mean the way it resolves the RID for different distros). That's why I postponed updating Libgit2sharp for now.

asbjornu:
Good point, @arturcic. And as expected, the build failed. I just wanted to try it out in order to address the comments in #1483.

arturcic:
It's on my todo list tp update to latest libgit2sharp and fix the packaging. I hope to find some time for that

trondhindenes:
it would be awesome to have this fixed. We have to keep old build agents and build on old container images (based on Ubuntu 16) in order for gitversion to work.

bmargula:
Yes, GitVersion Task V4 on Azure DevOps always fails when using the Microsoft hosted Ubuntu 1604 based agent. Please update this as soon as possible

PR #1645, opened 2019-03-21
NuGetUtils update

asbjornu:
Rebase of #1634. Resolves #1634.


asbjornu:
This is puzzling. This PR is just a rebase of #1634 (so all the merge commits are removed), but the build fails on this while it's successful in #1634. Any ideas why, @arturcic?

arturcic:
Looks like everywhere the same test is failing OutputsShouldMatchVariableProvider

asbjornu:
Any ideas why that fails here and not in #1634, @arturcic? I don't get it.

arturcic:
Not really, we need to check if there is any difference between these 3 branches

asbjornu:
@arturcic, I did a fresh `git reset` and `git rebase` now and that seems to have fixed the problem. Things are at least green. 🤷‍♂️

arturcic:
Great! My question is why do we have 2 PRs?

asbjornu:
Just to get rid of these: c2f36105a1ccf7f4ec13f8e97ec3743bfae85bab, 9d8eb3806d868300a92382bf2ea82ea4eb1b2641, b608f96175e90c5dbc8cddaa371c6439ad2ff3f7, 854f16b6753d84214c78af52224f90500da7ac11.

arturcic:
So I guess this PR will get merged instead of the other one

asbjornu:
Yes, this resolves #1634 so as soon this is merged, it will be closed automatically. It would be neat if GitHub supported `rebase` instead of `merge` for the "Update branch" button, but alas it doesn't. I've just installed [Autorebase](https://github.com/tibdex/autorebase) which might help us with that. I'll try that on #1634 shortly.

asbjornu:
Sigh. Seems like my rebase failed for some reason.

src/GitVersionTask/GitVersionTask.csproj
```diff
@@ -37,9 +38,7 @@
     <PackageReference Include="YamlDotNet" Version="$(PackageVersion_YamlDotNet)">
         <PrivateAssets>All</PrivateAssets>
     </PackageReference>
-    <!--See https://github.com/Microsoft/msbuild/issues/3671 for reason for the ExcludeAssets against msbuild packagerefernces -->
-    <PackageReference Include="Microsoft.Build.Utilities.Core" Version="15.9.20" ExcludeAssets="runtime" />
-    <PackageReference Include="UtilPack.NuGet.MSBuild" Version="$(PackageVersion_UtilPackNuGetMSBuild)">
+    <PackageReference Include="NuGetUtils.MSBuild.Exec" Version="$(PackageVersion_UtilPackNuGetMSBuild)">
```
dazinator: @asbjornu @arturcic  This looks like a problem with the rebase. If you compare this csproj file to @stazz's PR you can see his is different: https://github.com/GitTools/GitVersion/pull/1634/files#diff-bca0d1ca1ea8fbf39936384490e12d86R41

asbjornu:
Thanks for figuring that out, @dazinator. I think I got the rebase right now, but notice that 3b28212a117c8c51cd4f24b28f354afccc0ed2fc (and my rebased de3112488b7ea63ad4f34b13e3d0da38e92cb8c1) reintroduces the `Logger.Reset()` calls that @tpaxatb removed in #1644). I suppose that's a mistake, @stazz?

tpaxatb:
> Thanks for figuring that out, @dazinator. I think I got the rebase right now, but notice that [3b28212](https://github.com/GitTools/GitVersion/commit/3b28212a117c8c51cd4f24b28f354afccc0ed2fc) (and my rebased [de31124](https://github.com/GitTools/GitVersion/commit/de3112488b7ea63ad4f34b13e3d0da38e92cb8c1)) reintroduces the `Logger.Reset()` calls that @tpaxatb removed in #1644. I suppose that's a mistake, @stazz?

I'm going to assume that it's actually ok as is, since the logger.reset is actually paired with a setloggers call at the beginning of the function (rather than in a constructor as it was previously...)

asbjornu:
Thanks for the insight, @tpaxatb. Then I guess this just needs review and approval from @dazinator and we should be good to go. Do you agree, @arturcic?

arturcic:
Agree

stazz:
@asbjornu Yes, that is indeed a mistake. Rebase ended up being a bit messy even though I followed instructions, sorry about that. I have close to zero experience with rebasing stuff, as I've never worked with GIT repo which follows SVN workflow.

But yeah if @tpaxatb was ok with that, then it's fine. At least nothing should really interact directly with the MSBuild things anymore. The question is, does the NuGetUtils.MSBuild.Exec task factory handle logging using `IBuildEngine` in some special way? I think no, but maybe @tpaxatb can shed some insight on this? Currently the task factory is using the `IBuildEngine` provided in `Initialize` method when it performs pre-checks, and the task proxy will just redirect all output of invoked task to the `IBuildEngine` that it receives.

@dazinator Do I need to do any changes anywhere? The change request about the .csproj file seems to be not-relevant anymore, at least when looking at current state of https://github.com/asbjornu/GitVersion/blob/feature/NuGetUtilsUpdate/src/GitVersionTask/GitVersionTask.csproj .

stazz:
Hmm, actually, it seems I have some bug introduced in recent commits, despite the fact that I've added tests for the NuGetUtils.MSBuild.Exec. I'll investigate and back to you.

stazz:
> 
> 
> Hmm, actually, it seems I have some bug introduced in recent commits, despite the fact that I've added tests for the NuGetUtils.MSBuild.Exec. I'll investigate and back to you.

Found the bug, it was only appearing in some quite special circumstances (e.g. having multiple assembly files in same package or having certain dependencies). I've fixed it now for `2.0.5`, and updated [my fork](https://github.com/stazz/GitVersion/tree/stazz/NuGetUtilsUpdate).

feitzi:
any updates?

stazz:
Hey @feitzi , for my part, everything should be ok, I think. The NuGetUtils task factory has been tested and put into use by GitVersionTask. I am not sure what the conflicts mentioned here are exactly about - it seems that GitVersionTask is using workflow from SVN originally (based on some message I saw in this or another issue).
@dazinator @asbjornu if I need to still do something, feel free to instruct what needs to be done. :)

src/GitVersionTask/GenerateGitVersionInformation.cs
```diff
@@ -1,42 +1,99 @@
 namespace GitVersionTask
 {
+    using System;
     using System.IO;
     using GitVersion;
     using GitVersion.Helpers;
-    using Microsoft.Build.Framework;
 
-    public class GenerateGitVersionInformation : GitVersionTaskBase
+    public static class GenerateGitVersionInformation
```
asbjornu: @stazz, why the change to `static class` here?

src/GitVersionTask/GenerateGitVersionInformation.cs
```diff
@@ -1,42 +1,99 @@
...
+            };
+            var generator = new GitVersionInformationGenerator( fileWriteInfo.FileName, fileWriteInfo.WorkingDirectory, versionVariables, new FileSystem());
+            generator.Generate();
 
-            var workingDirectory = IntermediateOutputPath ?? TempFileTracker.TempPath;
+            return output;
+        }
 
-            GitVersionInformationFilePath = Path.Combine(workingDirectory, fileName);
+        public sealed class Input
```
asbjornu: @stazz, why isn't any form of inheritance used across the different `Input` classes (and `Output` classes, for that matter)?

src/GitVersionTask/GenerateGitVersionInformation.cs
```diff
@@ -1,42 +1,99 @@
...
+            public string ProjectFile { get; set; }
+
+            public string IntermediateOutputPath { get; set; }
+
+            public string Language { get; set; }
+
+            public bool NoFetch { get; set; }
+        }
+
+        private static Boolean ValidateInput( this Input input )
```
asbjornu: @stazz, why isn't this method placed inside the `Input` class itself?

src/GitVersionTask/GetVersion.cs
```diff
@@ -1,108 +1,129 @@
 namespace GitVersionTask
 {
-    using Microsoft.Build.Framework;
+    using System;
+    using GitVersion;
 
-    public class GetVersion : GitVersionTaskBase
+    public static class GetVersion
```
asbjornu: @stazz, why the change to `static` here?

stazz:
The `static` is because there is no longer need to extend `ITask`. Instead, the entrypoint acts just like a normal C# entrypoint, with an addition of having complex typed parameters (and not just `String[]`). That pretty much eliminates the need to use class hierarchy for the tasks contained in the assembly as well.

So did I undestand you correctly @asbjornu that are you suggesting that I make a fresh fork out of the latest tip of GitVersion repository, and re-apply changes for NuGetUtils? That makes sense for me, as I have a feeling that the change log is now way too convoluted. The work load of doing so is not too big, IMO, as the changes were relatively straightforward, but I think I'll have time for that only on this Sunday.

With the full re-apply of changes, I also hope it will be easier to see the reasoning behind the resulting `static` entrypoints and no class hierarchy for tasks.

src/GitVersionTask/GenerateGitVersionInformation.cs
stazz: The reason is that the class has no state on its own anymore (instead, the `Input` and `Output` inner classes are used), so no instances of it will be generated either.

src/GitVersionTask/GenerateGitVersionInformation.cs
stazz: I simply forgot to do that. I'll fix this on Sunday.

src/GitVersionTask/GetVersion.cs
stazz: Just like `GenerateGitVersionInformation` (and one other task too), this class has no state on its own anymore, so no instances of it will be created either.

src/GitVersionTask/GenerateGitVersionInformation.cs
stazz: No particular reason. I guess the validity input depends a bit on a context (which I admit in this case is exactly one). Can move it as instance method of `Input` class on Sunday.

src/GitVersionTask/GenerateGitVersionInformation.cs
asbjornu: I see. Thanks for the explanation.

src/GitVersionTask/GenerateGitVersionInformation.cs
asbjornu: Great!

asbjornu:
@stazz:

>So did I undestand you correctly @asbjornu that are you suggesting that I make a fresh fork out of the latest tip of GitVersion repository, and re-apply changes for NuGetUtils?

Yes, exactly.

>That makes sense for me, as I have a feeling that the change log is now way too convoluted. The work load of doing so is not too big, IMO, as the changes were relatively straightforward, but I think I'll have time for that only on this Sunday.

I'm very happy to hear that! Awesome!

>With the full re-apply of changes, I also hope it will be easier to see the reasoning behind the resulting `static` entrypoints and no class hierarchy for tasks.

👍

stazz:
@asbjornu This is now opened as #1677 . :)

asbjornu:
Resolved by #1677.

PR #1599, opened 2019-02-20
Adding support to generate WiX version files

ruhullahshah:
Will take up the pending work of PR (https://github.com/GitTools/GitVersion/pull/1386) here

PR #1647, opened 2019-03-22
Add WiX define constants

tpaxatb:
When a WiX project has been detected via the property WixTargetsImported
and GenerateGitVersionWixDefines is true, the GitVersion_* properties
are passed to the WiX compiler so that they me be used as
$(var.GitVersion_*) in .wxs files.  This avoids the need to run the
executable version to generate a .wxi file.

-----

Now that I can build within Visual Studio...
This adds the GitVersion_* values as defines to WiX projects when using the build task (via the targets).  This seems a lot cleaner and "out of the box" than instructing people to do a manual step (either running the executable to generate a .wxi file, or to manually edit the wixproj file).

My question is: the executable, when generating the wxi, was introduced in #1599.  However, there is a discrepancy between the defines produced and the defines the Task produces.  The task creates the defines as matching the actual MSBuild property (e.g. -dGitVersion_Major=$(GitVersion_Major).  The defines generated by the executable, however, do not have these prefixed.  I believe they need to be consistent so that people running the executable with the include file can also utilize the same .wxs in a msbuild environment.  Personally, I believe that the defines created by the executable via the GitVersionCore/WixVersionFileUpdater.cs should be prefixed with the GitVersion_ string to avoid potential collisions (especially with Major, Minor, and Patch), however, since that has been already merged, I can easily remove the prefix from the targets.  Or if someone has a different preferred methodology, I can use that.  Regardless, before this PR is merged, the discrepancy should be fixed.

This way the "Using" section of the documentation just needs to state they need to reference $(var.GitVersion_*) whether they use the GitVersionExe generate include .wxi method OR the GitVersionTask method.

Neither PR #1645 nor PR #1634 are affected by this change.

### DIFF
--- a/src/GitVersionTask/NugetAssets/build/functionality/GitVersionBuild.targets
+++ b/src/GitVersionTask/NugetAssets/build/functionality/GitVersionBuild.targets
@@ -86,6 +86,36 @@
       <FileVersion Condition=" '$(FileVersion)' == '' ">$(GitVersion_AssemblySemFileVer)</FileVersion>
     </PropertyGroup>
 
+    <PropertyGroup Condition=" '$(WixTargetsImported)' == 'true' And '$(GenerateGitVersionWixDefines)' == 'true' ">
+      <DefineConstants Condition=" '$(GitVersion_Major)' != '' ">GitVersion_Major=$(GitVersion_Major);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_Minor)' != '' ">GitVersion_Minor=$(GitVersion_Minor);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_Patch)' != '' ">GitVersion_Patch=$(GitVersion_Patch);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_PreReleaseTag)' != '' ">GitVersion_PreReleaseTag=$(GitVersion_PreReleaseTag);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_PreReleaseTagWithDash)' != '' ">GitVersion_PreReleaseTagWithDash=$(GitVersion_PreReleaseTagWithDash);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_PreReleaseLabel)' != '' ">GitVersion_PreReleaseLabel=$(GitVersion_PreReleaseLabel);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_PreReleaseNumber)' != '' ">GitVersion_PreReleaseNumber=$(GitVersion_PreReleaseNumber);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_BuildMetaData)' != '' ">GitVersion_BuildMetaData=$(GitVersion_BuildMetaData);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_BuildMetaDataPadded)' != '' ">GitVersion_BuildMetaDataPadded=$(GitVersion_BuildMetaDataPadded);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_FullBuildMetaData)' != '' ">GitVersion_FullBuildMetaData=$(GitVersion_FullBuildMetaData);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_MajorMinorPatch)' != '' ">GitVersion_MajorMinorPatch=$(GitVersion_MajorMinorPatch);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_SemVer)' != '' ">GitVersion_SemVer=$(GitVersion_SemVer);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_LegacySemVer)' != '' ">GitVersion_LegacySemVer=$(GitVersion_LegacySemVer);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_LegacySemVerPadded)' != '' ">GitVersion_LegacySemVerPadded=$(GitVersion_LegacySemVerPadded);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_AssemblySemVer)' != '' ">GitVersion_AssemblySemVer=$(GitVersion_AssemblySemVer);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_AssemblySemFileVer)' != '' ">GitVersion_AssemblySemFileVer=$(GitVersion_AssemblySemFileVer);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_FullSemVer)' != '' ">GitVersion_FullSemVer=$(GitVersion_FullSemVer);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_InformationalVersion)' != '' ">GitVersion_InformationalVersion=$(GitVersion_InformationalVersion);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_BranchName)' != '' ">GitVersion_BranchName=$(GitVersion_BranchName);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_Sha)' != '' ">GitVersion_Sha=$(GitVersion_Sha);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_NuGetVersionV2)' != '' ">GitVersion_NuGetVersionV2=$(GitVersion_NuGetVersionV2);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_NuGetVersion)' != '' ">GitVersion_NuGetVersion=$(GitVersion_NuGetVersion);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_NuGetPreReleaseTagV2)' != '' ">GitVersion_NuGetPreReleaseTagV2=$(GitVersion_NuGetPreReleaseTagV2);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_NuGetPreReleaseTag)' != '' ">GitVersion_NuGetPreReleaseTag=$(GitVersion_NuGetPreReleaseTag);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_CommitDate)' != '' ">GitVersion_CommitDate=$(GitVersion_CommitDate);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_CommitsSinceVersionSource)' != '' ">GitVersion_CommitsSinceVersionSource=$(GitVersion_CommitsSinceVersionSource);$(DefineConstants)</DefineConstants>
+      <DefineConstants Condition=" '$(GitVersion_CommitsSinceVersionSourcePadded)' != '' ">GitVersion_CommitsSinceVersionSourcePadded=$(GitVersion_CommitsSinceVersionSourcePadded);$(DefineConstants)</DefineConstants>
+    </PropertyGroup>
+
   </Target>
 
   <!--Support for ncrunch-->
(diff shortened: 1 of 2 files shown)


jbaehr:
I think this is a great idea to round up the WiX support.

@asbjornu what do you thing about adding this to the 5.0.0 milestone?

asbjornu:
@jbaehr, I have no way to validate that this does anything useful to anyone, so if you vouch for the usefulness of this PR, I suppose we can merge it. :)

arturcic:
@asbjornu I don't mind adding these changes. If we merge them I'll update my PR #1659 to include these changes as well

asbjornu:
Good to know, @arturcic. I suppose it would be best if we merge #1677 before merging this.

arturcic:
Agree, but we can do the other way around, this one merged is not affecting the other one. I'd prefer this one to merge first till we do the review of the other PR, and do a rebase of the #1677.

asbjornu:
Ok, I just thought #1677 "deserved" to be merged first since it's been lingering for a while, but since there won't be any conflicts, I suppose it's the same.

arturcic:
Yep. Just let merge this one and we can continue on the other one.

asbjornu:
Thank you for your contributions, @tpaxatb! 🙏

tpaxatb:
Sweet, thanks for taking this mainline so we don't have to remember to add this manually here.  This actually has been very useful for us because you are able to do something like:

` <Product Id="*" Name="MyProduct $(var.GitVersion_FullSemVer)" Language="1033" Version="$(var.GitVersion_MajorMinorPatch)" Manufacturer="MyCompany" UpgradeCode="[GUID]">
    <Package InstallerVersion="500" Compressed="yes" InstallScope="perMachine" Description="$(var.GitVersion_InformationalVersion)" />
...
</Product>`

And useful if you want to version specific unversioned files to the current version for patching rather than using the hash/date method.  Manually updating the .wxs files was a royal pain for sure.

The only thing you have to do is either a post build step to rename the file (since the file name property is set before the git version task is run) OR (like us) have the CI system generate the same git version and do the rename on deployment.

I will submit changes in the documentation (I was waiting to do so until this was merged) after I get back I will be in China on business for the next week then I go on vacation for another week...

arturcic:
Please create an issue and assign it on you regarding documentation. Thanks for your time!

tpaxatb:
#1681 Opened (Apparently I can't assign to myself though haha)

NameOfTheDragon:
This is (would be) an awesome feature but I can't figure out how to make it work. If I do this:
```lang-xml
  <?define ProductVersion="$(var.GitVersion_MajorMinorPatch)" ?>
```
Then I get this:

>     Error CNDL0150: Undefined preprocessor variable '$(var.GitVersion_MajorMinorPatch)'. (33)

So it isn't "Plug-n-Play" and I can't find WiX referenced in the docs... what am I missing?

asbjornu:
@NameOfTheDragon, #1681 is still unresolved, so that's why we have no documentation for this. Why it doesn't work, though, @tpaxatb would have to help out with as I don't use Wix myself.

tpaxatb:
I'm sorry I haven't had a chance to do anything because of my "real" job haha.

@NameOfTheDragon I will need to know your exact version of GitVersionTask, as support for the $(var.GitVersion*) things didn't exist until a later beta of the 5.0 cycle.

In addition, there is an issue in Visual Studio 2019 that I am aware of in that if you download a nuget package and attach it to a project, the project doesn't get reloaded and the internal targets for the project aren't updated.  The only solution I have found for that is to shut down and restart Visual Studio.

I also know that up until fairly recently, the nuget version numbers on the packages would be not sorted correctly, so if you had beta.127 and did a package upgrade, it would think beta.4 would be the update and actually downgrade your package.

Ensure you are at 5.0.0 release (I honestly don't recall what label was the first appearance of the parameters but it was anything after mid-may i blieve) and look at the output msbuild command line. If you do not see the command line parameters "-dGitVersion_MajorMinorPatch=x.y.z on the candle command line, you are using a version of GitVersionTask that does not have this feature.

This is a "Works For Me" thing until I know more information, as my company utilizes this feature on a daily basis for our product versioning and I know it works with stable 5.0.0 as I literally just tested it to make sure that there wasn't a typo somewhere...

tpaxatb:
And this may be a silly question, @NameOfTheDragon but are you sure that you have pulled the GitVersionTask package and it is attached to the WiX project itself? i.e. there's a packages.config in the project folder?

Be aware that if you have divided your WiX projects into several sub projects, you will need to attach the GitVersionTask package to any project that is referencing the $(var.GitVersion_*) defines, as the stage that uses these references is the candle (compile) stage.

NameOfTheDragon:
Yes, I have pulled 5.0.1 and have done so in the WiX project. Here's the packages.config from that project:
```lang=xml
<packages>
  <package id="GitVersionTask" version="5.0.1" developmentDependency="true" />
  <package id="TA.Ascom.WixCustomActions" version="0.1.0-Beta15" />
  <package id="UtilPack.NuGet.MSBuild" version="2.9.1" developmentDependency="true" />
  <package id="WiX" version="3.11.1" />
</packages>
```

Is it necessary to also update the command line version of GitVersion, or should the NuGet be completely self-contained? Because my command line version is reporting

> 4.0.0+Branch.master.Sha.e715eb5dd515037d90a257f527e62442cae8c1c5

NameOfTheDragon:
OK, I'm definitely using package version 5.0.1 (and command line version 5.0.1).

I created a new project, and _that works_. I can see all the -dGitVersion_XXX properties being defined.

However in my "real" solution, _those properties don't appear_. I'm using a few project references and I do get defines for all of those, but nothing for GitVersion.

I've compared the two .wixProj files and although there are obviously some differences, I really can't see why one works and the other doesn't. I'm attaching my full `.wixProj` file here in case you can see anything that I've missed.

[TA.NexDome.Setup.zip](https://github.com/GitTools/GitVersion/files/3518219/TA.NexDome.Setup.zip)

jbaehr:
@NameOfTheDragon: Maybe [MSBuild Binary and Structured Log Viewer](http://msbuildlog.com) can help you to track the issue down.

SeanWilliamson2004:
This worked for me after a bit of effort.
I did a basic project first, but I couldn't get the GitVersionTask to apply to the basic wix project even using old style packages.config so I applied old style packages.config to a vb.net project, then copied the vbproj updates over to the wix project.  It worked once I sorted out the git repo for the test project.  I did set ToolsVersion to 15.0 but I don't know if this helped.

I then applied this to my actual project.  I made a few mistakes not correcting paths to the package folder which will pass without warning if incorrect.  But once they were sorted it worked and it's great.
Binary logging wouldn't work for me with WIX projects so I had to use the detailed log.  But it was pretty obvious in this log that the gitversion props and targets were not loading when the paths were incorrect.
Note that for both projects I don't have the WIX toolset installed but use the WIX nuget packages.

tpaxatb:
Ahhhh. this could very well be the issue.  I am using the toolset, not the nuget packages.

In other news the define constants section that is in the build subdirectory does NOT exist in the buildMultiTargeting folder.

It would help to get your exact solution setup and installation and steps to replicate, because it is not repeatable even when I grab the nuget package and apply it to my wix project file.

SeanWilliamson2004:
We use WIX on a CI Server, so I keep the toolset off my PC to avoid inadvertently relying on it.  I don't know if using the nuget package made any difference. 
Here's my test project wixproj file...

`<?xml version="1.0" encoding="utf-8"?>
<Project ToolsVersion="15.0" DefaultTargets="Build" InitialTargets="EnsureWixToolsetInstalled" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <Import Project="packages\GitVersionTask.5.0.1\build\GitVersionTask.props" Condition="Exists('packages\GitVersionTask.5.0.1\build\GitVersionTask.props')" />

  <Import Project="packages\WiX.3.11.1\build\wix.props" Condition="Exists('packages\WiX.3.11.1\build\wix.props')" />
  <PropertyGroup>
    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
    <Platform Condition=" '$(Platform)' == '' ">x86</Platform>
    <ProductVersion>3.10</ProductVersion>
    <ProjectGuid>5523a0df-f585-4a6a-9c2e-8e61d5221d95</ProjectGuid>
    <SchemaVersion>2.0</SchemaVersion>
    <OutputName>WIXGitVersionTest</OutputName>
    <OutputType>Package</OutputType>
    <Name>WIXGitVersionTest</Name>
    <NuGetPackageImportStamp>
    </NuGetPackageImportStamp>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x86' ">
    <OutputPath>bin\$(Configuration)\</OutputPath>
    <IntermediateOutputPath>obj\$(Configuration)\</IntermediateOutputPath>
    <DefineConstants>Debug</DefineConstants>
  </PropertyGroup>
  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|x86' ">
    <OutputPath>bin\$(Configuration)\</OutputPath>
    <IntermediateOutputPath>obj\$(Configuration)\</IntermediateOutputPath>
  </PropertyGroup>
  <ItemGroup>
    <Compile Include="Product.wxs" />
  </ItemGroup>
  <ItemGroup>
    <Content Include="packages.config" />
  </ItemGroup>
  <PropertyGroup>
    <UpdateVersionProperties>false</UpdateVersionProperties>
  </PropertyGroup>
  <Import Project="$(WixTargetsPath)" Condition=" '$(WixTargetsPath)' != '' " />
  <Import Project="$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\Wix.targets" Condition=" '$(WixTargetsPath)' == '' AND Exists('$(MSBuildExtensionsPath32)\Microsoft\WiX\v3.x\Wix.targets') " />
  <Target Name="EnsureWixToolsetInstalled" Condition=" '$(WixTargetsImported)' != 'true' ">
    <Error Text="The WiX Toolset v3.11 (or newer) build tools must be installed to build this project. To download the WiX Toolset, see http://wixtoolset.org/releases/" />
  </Target>
  <Target Name="EnsureNuGetPackageBuildImports" BeforeTargets="PrepareForBuild">
    <PropertyGroup>
      <ErrorText>This project references NuGet package(s) that are missing on this computer. Use NuGet Package Restore to download them.  For more information, see http://go.microsoft.com/fwlink/?LinkID=322105. The missing file is {0}.</ErrorText>
    </PropertyGroup>
    <Error Condition="!Exists('packages\WiX.3.11.1\build\wix.props')" Text="$([System.String]::Format('$(ErrorText)', 'packages\WiX.3.11.1\build\wix.props'))" />
  </Target>
  
  <Import Project="packages\GitVersionTask.5.0.1\build\GitVersionTask.targets" Condition="Exists('packages\GitVersionTask.5.0.1\build\GitVersionTask.targets')" />

</Project>`

and my packages.config...

`<?xml version="1.0" encoding="utf-8"?>
<packages>
  <package id="GitVersionTask" version="5.0.1" developmentDependency="true" />
  <package id="WiX" version="3.11.1" />
</packages>`

> 

My product.wxs is an out-of-the-box generated V3 one except with the changes at the top...

`<?xml version="1.0" encoding="UTF-8"?>
<Wix xmlns="http://schemas.microsoft.com/wix/2006/wi">
	<Product Id="*" Name="WIXGitVersionTest $(var.GitVersion_FullSemVer)" Language="1033" Version="$(var.GitVersion_MajorMinorPatch)" Manufacturer="Codis Limited" UpgradeCode="592a1e65-56bb-4358-aaf1-234d79f46553">
		<Package InstallerVersion="200" Compressed="yes" InstallScope="perMachine" />

		<MajorUpgrade DowngradeErrorMessage="A newer version of [ProductName] is already installed." />
		<MediaTemplate />

		<Feature Id="ProductFeature" Title="WIXGitVersionTest" Level="1">
			<ComponentGroupRef Id="ProductComponents" />
		</Feature>
	</Product>

	<Fragment>
		<Directory Id="TARGETDIR" Name="SourceDir">
			<Directory Id="ProgramFilesFolder">
				<Directory Id="INSTALLFOLDER" Name="WIXGitVersionTest" />
			</Directory>
		</Directory>
	</Fragment>

	<Fragment>
		<ComponentGroup Id="ProductComponents" Directory="INSTALLFOLDER">
			
			
				
			
		</ComponentGroup>
	</Fragment>
</Wix>`

I did a git init and committed my changes to master, but I got pretty clear errors that I needed to do that.

NameOfTheDragon:
Sorry it's taken me a while to get back. When I create a brand new WiX project, it seems to work, but it seems to be pre-existing projects that are having issues, and I can't see why. I have just found a second project that doesn't work. It;s one I haven't worked on for a couple of years but I've updated all my NuGet packages and installed GitVersionTask for the first time in this project. No versioning. This one is open source so I guess you could check it out and try to build it if that would help? It's at https://github.com/Tigra-Astronomy/TA.ArduinoPowerController.AscomServer/tree/feature/embedded-driver (feature/embedded-driver branch)

SeanWilliamson2004:
I've just converted another project and I noticed that I got the

jbaehr:
@NameOfTheDragon 
Currently the GitVersion target [GetVersion](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionTask/build/GitVersionTask.targets#L63) states that it should run before the "EnsureWixToolsetInstalled" target. The problem is, that this target is not provided by the WiX-toolchain itself, but only via WiX-Project-Templates. And even there, its sole purpose is to error when the toolchain is not available; pretty useless if consumed via nuget. We should really find a WiX-Target better suited for hooking the GetVersion-target in.

So for now you can either update your WiX-Projects to the newer templates (basically what @SeanWilliamson2004 did), or just set GitVersion's "GetVersion" as [InitialTagets](https://docs.microsoft.com/en-us/visualstudio/msbuild/target-build-order?view=vs-2019) of your project.

PR #1659, opened 2019-04-18
LibGit2Sharp update to 0.26.0

arturcic:
This PR is meant to update `GitVersion` to the latest `LibGit2Sharp` version `0.26.0`.

This means the command line tools will not require anymore to have the workaround (you can see the Dockerfiles, the `libgit2-devel.x86_64` dependency is not installed).

As for the `GitVersionTask`, I had to change the way we package the nuget artifact, and inspired from https://github.com/dotnet/sourcelink I have created a separate package called `GitVersion.MsBuild`. 

`GitVersion.MsBuild` is meant to replace `GitVersionTask` as it is using the latest `LibGit2Sharp` and has the mechanism of loading the native libraries in place. In order to do that I had to create a custom `AssemblyLoadContext` that for .net core knows how to load the native dependencies.

For now `GitVersion.MsBuild` package was tested on .net full framework on windows and .net core on window and linux (ubuntu 16.04) with single TargetFramework.

I'm planning to add multi TargetFrameworks support and also some docker files for all supported linux OSes `Libgit2sharp` supports to validate the `GitVersion.MsBuild` package.


src/GitVersion.MsBuild.Task.Tests/GitVersionTaskDirectoryTests.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        try
+        {
+            executeCore.ExecuteGitVersion(null, null, null, null, true, workDirectory, null);
+        }
+        catch (Exception ex)
+        {
+            // `RepositoryNotFoundException` means that it couldn't find the .git directory,
+            // any other exception means that the .git was found but there was some other issue that this test doesn't care about.
+            Assert.IsNotAssignableFrom<RepositoryNotFoundException>(ex);
```
asbjornu: Can’t this, including the try/catch, be written as `Assert.That(() => …, Does.Not.Throw<RepositoryNotFoundException>());`?

arturcic: I took the GitVersionTask as a starting point for building GitVersion.MsBuild. The goal was to make the new task compatible with the old one. As for the Unit tests, they are the same as for GitVersionTask except they use the new Tasks.

asbjornu:
This is fantastic, @arturcic! Thank you so much for the work you’ve put into figuring this out. :pray: :tada: :love:

src/GitVersion.MsBuild.Task.Tests/GitVersionTaskDirectoryTests.cs
asbjornu: If we just took your `GitVersion.Msbuild` and just called it `GitVersionTask`, what would break?

arturcic:
> This is fantastic, @arturcic! Thank you so much for the work you’ve put into figuring this out. 🙏 🎉 :love:

I took me a while to understand how all of these works, had to understand how the dependencies are loaded, and I had good inspiration. 

If everything goes well I'll move the Libgit2Sharp specific code that loads native dependencies into a separate nuget package as suggested [here ](https://github.com/libgit2/libgit2sharp/issues/1649)

src/GitVersion.MsBuild.Task.Tests/GitVersionTaskDirectoryTests.cs
arturcic: well eventually we can rename it back to `GitVersionTask` when it gets to a stable and tested level. As I said in the description I want to add some dockerfiles to validate it on all linux OSes Libgit2sharp is able to run, including .net core and mono, docs. I decided to create the PR first, give it a try and meanwhile work on the remaining stuff. As for the naming, to be honest I would keep `GitVersion.MsBuild` as it is clear it's about msbuild, `GitVersionTask` is not that suggestive, but I don't mind changing the name.

asbjornu: I don’t mind the name change, I just think it’s enough with one solution. I completely agree with your approach, but at one point I would like to see one of `GitVersionTask` or `GitVersion.Msbuild` to go away so we only have one or the other.

Could we get this to preserve the Git history from `GitVersionTask` somehow? If we continue with this as is and remove everything related to the old `GitVersionTask`, we will lose all of its history. Ideas?

arturcic: Btw, GitVersionTask is still not able to resolve dependencies, so for now it will be broken. So either we go with GitVersion.MsBuild and remove GitVersionTask, or we just rename GitVersion.MsBuild to GitVersionTask

arturcic: I would think of a way of transforming GitversionTask into GitVersion.MsBuild keeping the history on the way.

arturcic: I'll try to do that and push again the changes.

asbjornu: Perfect!

arturcic:
@asbjornu I have reworked the PR, the GitVersionTask transforms into GitVersion.MsBuild keeping the history on the ways as discussed. Can you have a new review?

src/GitVersion.MsBuild.Task/GitVersionTaskBase.cs
```diff
@@ -0,0 +1,13 @@
+namespace GitVersion.MsBuild.Task
+{
+    using Microsoft.Build.Framework;
+    using Microsoft.Build.Utilities;
+
+    public abstract class GitVersionTaskBase : Task
```
asbjornu: Perhaps you can add this file with the same contents as the old `GitVersionTaskBase` and then move in the new content in two separate commits so Git detects it as a move?

src/GitVersionTask/GenerateGitVersionInformation.cs
```diff
@@ -1,42 +0,0 @@
-namespace GitVersionTask
-{
-    using System.IO;
-    using GitVersion;
-    using GitVersion.Helpers;
-    using Microsoft.Build.Framework;
-
-    public class GenerateGitVersionInformation : GitVersionTaskBase
```
asbjornu: It would be nice if Git detected this as a move and not delete/add.

arturcic: Actually that's what you see in GitHub the changes made in the PR. If you look closer in each of the commits there are separate commits, and from one commit to another you can see in the history the rename

arturcic: you can see the commit 046621700e94f6fb901c9fdd866995558c28fa32 that does the rename

asbjornu: Ah, ok. Not the easiest thing in the world to do code reviews on a mobile phone. :smile:

arturcic: Agree

arturcic:
Looks like the 0.26.0 version of mono is not properly working with mono https://github.com/dotnet/sourcelink/issues/155. 

I guess for now till we have a version that does work, we can move forward with .net core (all supported oses) + windows full framework. I'll follow the progress on that thread and update with mono support.

This PR makes sense to be merged as it adds support for a lot of OSes, and it does not have dependency on libcurl

@asbjornu ideas?

asbjornu:
This PR is the only sensible step forward, imho. :shipit:

dazinator:
Is the intention to drop UtilPack and handle the dependency resolution ourselves in GitVersionTask? If so, better let @stazz know so he doesn't waste any further time on PR's related to this area!

dazinator:
One other scenario we should try and test is with Omnisharp (i.e build a csproj that has the msbuild task installed from vscode)

arturcic:
> One other scenario we should try and test is with Omnisharp (i.e build a csproj that has the msbuild task installed from vscode)

I will give it a try

arturcic:
> Is the intention to drop UtilPack and handle the dependency resolution ourselves in GitVersionTask? If so, better let @stazz know so he doesn't waste any further time on PR's related to this area!

To be honest I know about that PR and I also was playing with the artifacts from that PR and still for .net core on linux it was not working even with the latest Libgit2Sharp version. I started the work as a proof of concept. That's why initially I have created a separate nuget package GitVersion.MsBuild. I ended up with the current code that does work for .net core windows and linux, but it's not working on mono as it seems like the latest Libgit2Sharp has some issues with it.

twsI:
Is anything currently blocking this from getting merged?

seanamosw:
Want to confirm that using the `GitVersion.CommandLine.DotNetCore.5.0.0-PullRequest1659-79.nupkg` artifact from AppVeyor hosted on our internal nuget resolves the issue on our Ubuntu 18 dev machines.

Would be great if we could get this in preview/stable on nuget.org.

arturcic:
@seanamosw thanks for the feedback. Unfortunately this PR is not compatible with the recent changes on master. So I will need to re-implement the PR, but that will be challenging.

arturcic:
Closing in favor of #1713

PR #1683, opened 2019-05-17
Bug fixes and small cleanup

Genbox:
Each commit describes each change. They are small general fixes, spelling and cleanup of redundant code.


src/GitVersionCore.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
```diff
@@ -40,7 +40,6 @@ public void ShouldNotAllowIncrementOfVersion()
         [TestCase("Merge branch 'Releases-0.2.0'", true, "0.2.0")]
         [TestCase("Merge branch 'Releases/0.2.0'", true, "0.2.0")]
         [TestCase("Merge branch 'release-4.6.6' into support-4.6", true, "4.6.6")]
-        [TestCase("Merge branch 'release-10.10.50'", true, "10.10.50")]
```
asbjornu: Why is this test case deleted?

src/GitVersionCore/BuildServers/AppVeyor.cs
```diff
@@ -23,11 +23,7 @@ public override string GenerateSetVersionMessage(VersionVariables variables)
 
             var data = string.Format("{{ \"version\": \"{0}.build.{1}\" }}", variables.FullSemVer, buildNumber);
             var bytes = Encoding.UTF8.GetBytes(data);
-            if (request.Headers == null)
```
asbjornu: Is `request.Headers` never `null` since you deleted this?

src/GitVersionCore.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
Genbox: I deleted the wrong one I think. Line 37 and line 40 are the same test case, it was supposed to be one of those! Thanks for spotting that one.

src/GitVersionCore/BuildServers/AppVeyor.cs
Genbox: That is correct. It returns a pre-initialized collection, and even if you set it to null, it will fail when it tries to iterate the null value. See https://source.dot.net/#System.Net.Requests/System/Net/HttpWebRequest.cs,835

src/GitVersionCore.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
Genbox: You know what, I just checked it again.

```
line35: [TestCase("Merge branch 'release-10.10.50'", true, "10.10.50")]
line43: [TestCase("Merge branch 'release-10.10.50'", true, "10.10.50")]
```
I think I hit the correct one the first time. line 37 and 40 have an "s" different, but 35 and 43 are identical. I removed line 43 in the commit there. Sorry about the confusion.

asbjornu: Indeed. Good catch.

PR #1686, opened 2019-05-21
Enhance logic of getting current branch in case of drone

vhatsura:
Hey, @arturcic!

It's the 2nd PR related to Drone.

Little introduction:
`DRONE_BRANCH` environment variable is equal to destination branch in case of pull request [details](https://discourse.drone.io/t/getting-the-branch-a-pull-request-is-created-from/670). By this reason, in case of pull request, we need to fetch another environment variable to know current branch. Unfortunately, this is also not simple. In the `1.x.x` Drone version it can be fetched from `DRONE_SOURCE_BRANCH`, however in older version it can be done by parsing `CI_COMMIT_REFSPEC` environment variable, which has the following format: `{sourceBranch}:{destinationBranch}`.

The PR implement such logic of fetching current branch name.

### DIFF
--- a/src/GitVersionCore.Tests/BuildServers/DroneTests.cs
+++ b/src/GitVersionCore.Tests/BuildServers/DroneTests.cs
@@ -46,5 +46,106 @@ namespace GitVersionCore.Tests.BuildServers
             // Assert
             result.ShouldBeFalse();
         }
+
+        [Test]
+        public void GetCurrentBranch_ShouldDroneBranch_InCaseOfPush()
+        {
+            // Arrange
+            const string droneBranch = "droneBranch";
+
+            Environment.SetEnvironmentVariable("DRONE_PULL_REQUEST", "");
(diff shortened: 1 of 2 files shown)


arturcic:
LGTM! Would be great if we can check the merge messages that Drone creates when PR is merged, and adjust the unit tests in case it is needed. For now I'll merge these changes. Thank you for your time!

vhatsura:
@arturcic, could point. I will try to check it, however, I don't understand how it connected with GitVersion?

arturcic:
GitVersion can use the merge messages when computing the next version. And each of the supported CI have different message format for the merge commit. Please have a look [here](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs)

vhatsura:
Got it, thanks.

PR #1691, opened 2019-05-28
Allow overriding of the config file path

### LINKED ISSUE #894: Add support for GitVersion.yml outside of git repo root
The ability to either use a global gitversion.yml, or the ability to point gitversion at a config file outside of the source tree, eg:

```
gitversion -config $env:HOME/GitVersion.yml
```

Would enable us to avoid duplicating a bunch of files in many repos. For companies with many git repos that use the same git patterns, requiring the config files to be in the root of each repo is a repo setup and maintenance burden.

If you're interested in accepting a PR for this I can provide it.

chuseman:
This adds a new abstract `ConfigFileLocator` class, which has two implementations:
- `DefaultConfigFileLocator`, which is the current logic (`GitVersion.yml` and `GitVersionConfig.yaml` in working and project root directories).
- `NamedConfigFileLocator` that will only look for a specific file path (it can be a relative or absolute path).

`GitVersionExe` has a new optional `/config` argument which takes a path.

`GitVersionTask` has a new `ConfigFilePath` property on all inputs and is exposed via the MSBuild `GitVersionConfig` property.

`GitVersionTfsTask` has a new `configFilePath` input.

This resolves #894.

### DIFF
--- a/src/GitVersionCore.Tests/ConfigProviderTests.cs
+++ b/src/GitVersionCore.Tests/ConfigProviderTests.cs
@@ -14,19 +14,18 @@ using YamlDotNet.Serialization;
 [TestFixture]
 public class ConfigProviderTests : TestBase
 {
-    private const string DefaultRepoPath = "c:\\MyGitRepo";
-    private const string DefaultWorkingPath = "c:\\MyGitRepo\\Working";
+    private const string DefaultRepoPath = @"c:\MyGitRepo";
 
     string repoPath;
-    string workingPath;
     IFileSystem fileSystem;
+    ConfigFileLocator configFileLocator;
 
     [SetUp]
     public void Setup()
     {
         fileSystem = new TestFileSystem();
+        configFileLocator = new DefaultConfigFileLocator();
         repoPath = DefaultRepoPath;
-        workingPath = DefaultWorkingPath;
 
         ShouldlyConfiguration.ShouldMatchApprovedDefaults.LocateTestMethodUsingAttribute<TestAttribute>();
     }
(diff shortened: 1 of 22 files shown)


chuseman:
Looks like I'm running into some flaky tests here?

Should I drop my last commit? I think Shouldly will give you better error messages for failures if you use it this way, but it isn't related to my PR.

arturcic:
@chuseman could you please rebase on top of master as there are some changes to the vsix version.

asbjornu:
Thanks for your contributions, @chuseman!

Skoucail:
I believe this feature introduced a bug.
If i leave the Config file field empty i believe it should look for GitVersion.yml files in the working directory and the project directory.
But as the task always calls gitversion with /config <workingdirectory> it doesn't look for any GitVersion.yml files. And doesn't even use the GitVersion.yml in the working directory.

I believe the problem is that configFilePath is of type filePath. So if you enter something relative (or empty) Azure DevOps will always pretent the working directory. So in the GitVersion.ts the property configFilePath will always contain a value (working directory) and will result in always adding the /config param.

asbjornu:
@Skoucail, this should be fixed in #1777. Can you please verify?

Skoucail:
@asbjornu I can test it on Monday. But i’m unsure how i can test an unreleased version of an azure devops task. As the tasks are downloaded from the marketplace.

arturcic:
@Skoucail you can use the preview version of the task - https://marketplace.visualstudio.com/items?itemName=gittools.gitversion-preview

arturcic:
If the code is merged it should be available in the preview version

chuseman:
FWIW, I've verified the fix on our Azure DevOps Server 2019 instance with the preview version. The /config parameter is no longer passed unless the task specifies it.

Skoucail:
> FWIW, I've verified the fix on our Azure DevOps Server 2019 instance with the preview version. The /config parameter is no longer passed unless the task specifies it.

Same here, looks fine now. Thx for fixing @chuseman

PR #1702, opened 2019-06-05
Bugfix/fix only tracked branches parameter implementation

crazycrank:
Hey there

I'm currently playing around with GitVersion because we have some weird behaviors with our current branching strategy and what better way to learn something about git, branching and versioning then getting muddy in the code.

Anyway, it's late, my jokes get terrible and I found small bug. 
In the class GitRepoMetadataProvider, the method GetBranchesContainingCommit has a parameter onlyTrackedBranches that parameter is actually implemented exactly the other way arround, so when set to true, all branches get scanned, with false only the tracked ones. I'll add a comment into the PR at the point.

I thought about renaming the parameter to allBranches or something like that, but that felt wrong naming wise, as it doesn't refer to tracking branches which this parameter is about. So I fixed up the implementation, and changed existing functionality to adapt the original behavior again. Three tests seem to be dependent on non-tracked branches which have been fixed up too. 

If you have a better alternative name and prefer to just rename the parameter, just discard this PR. but it's not actually less work as the naming spreads through the hole call chain too :)

Have a good night.

### DIFF
--- a/src/GitVersionCore/GitRepoMetadataProvider.cs
+++ b/src/GitVersionCore/GitRepoMetadataProvider.cs
@@ -81,7 +81,7 @@ namespace GitVersion
                 // TODO: It looks wasteful looping through the branches twice. Can't these loops be merged somehow? @asbjornu
                 foreach (var branch in branches)
                 {
-                    if (branch.Tip != null && branch.Tip.Sha != commit.Sha || (onlyTrackedBranches && !branch.IsTracking))
+                    if (branch.Tip != null && branch.Tip.Sha != commit.Sha || ((onlyTrackedBranches && branch.IsTracking) || !onlyTrackedBranches))
                     {
                         continue;
                     }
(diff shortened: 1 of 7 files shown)


src/GitVersionCore/GitRepoMetadataProvider.cs
```diff
@@ -77,7 +77,7 @@ public IEnumerable<Branch> GetBranchesContainingCommit(Commit commit, IList<Bran
                 // TODO: It looks wasteful looping through the branches twice. Can't these loops be merged somehow? @asbjornu
                 foreach (var branch in branches)
                 {
```
crazycrank: This is the change I talked about in the PR. What the original dev probably meant was to use `onlyTrackedBranches || !ibranch.IsTracking`. I used the longer version `(onlyTrackedBranches && branch.IsTracking) || !onlyTrackedBranches` as its more verbose and makes sure this bug doesn't happen :)

arturcic:
Looks good to me

asbjornu:
@crazycrank, can you please look at my code review and rebase this PR to the `HEAD` of `master`?

asbjornu:
@crazycrank: Poke. 👉 😃

crazycrank:
Hey @asbjornu 

I'm terribly sorry, this got burried in my inbox and I'm only rarely doing stuff on GitHub, so I have missed all your request. I'm trying to take a look at the PR on the weekends!

asbjornu:
No problem, @crazycrank! Thanks for picking it up. 😃

crazycrank:
Hey @asbjornu

I rebased my branch and aligned the naming throughout the solution (at least everything I could find). Take a look and write if you need any more changes!

asbjornu:
Awesome, thanks for your contributions, @crazycrank! 🙏

Issue #1520, opened 2018-10-31
Bug: Branch contains / is not replaced by - in build metadata


melcloud:
GitVersion version: 4.0, 3.6.4
Problem:
Given a branch name contains `/`, e.g. `hotfixes/build-script`, the build metadata does not convert it to hyphen. This is not a valid semver 2.

```json
{
  "Major":0,
  "Minor":2,
  "Patch":1,
  "PreReleaseTag":"beta.1",
  "PreReleaseTagWithDash":"-beta.1",
  "PreReleaseLabel":"beta",
  "PreReleaseNumber":1,
  "BuildMetaData":11,
  "BuildMetaDataPadded":"0011",
  "FullBuildMetaData":"11.Branch.hotfixes/build-script.Sha.301ff2de375da97133ab544c9561cbd909039877",
  "MajorMinorPatch":"0.2.1",
  "SemVer":"0.2.1-beta.1",
  "LegacySemVer":"0.2.1-beta1",
  "LegacySemVerPadded":"0.2.1-beta0001",
  "AssemblySemVer":"0.2.1.1",
  "FullSemVer":"0.2.1-beta.1+11",
  "InformationalVersion":"0.2.1-beta.1+11.Branch.hotfixes/build-script.Sha.301ff2de375da97133ab544c9561cbd909039877",
  "BranchName":"hotfixes/build-script",
  "Sha":"301ff2de375da97133ab544c9561cbd909039877",
  "NuGetVersionV2":"0.2.1-beta0001",
  "NuGetVersion":"0.2.1-beta0001",
  "CommitsSinceVersionSource":11,
  "CommitsSinceVersionSourcePadded":"0011",
  "CommitDate":"2018-10-31"
}
```

Skoucail:
Any update on this?

NuGet 4.3.0+ (with support for SemVer 2.0) actually has issues with this.
[Nuget pack defaults to version 1.0.0 when AssemblyInformationalVersion metadata contains forward slash](https://github.com/NuGet/Home/issues/4082)

And according to the [Semantic Versioning 2.0.0](https://semver.org/) spec '/' isn't an allowed character.

[...]

PR #1704, opened 2019-06-07
Comply with the SemVer2.0 format in Metadata

Skoucail:
Hey,

I found some time to do the fix for #1520.

And according to the Semantic Versioning 2.0.0 spec in the metadata they only allow a-z A-Z 0-9 '.' and '-' All other chars should be replaced by '-' as '.' is to split identifiers.

`Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphen [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85.`

(Its my first ever pull request, so hope i did everything the correct way)


asbjornu:
Thanks for your contribution, @Skoucail!

PR #1708, opened 2019-06-18
Try updating PackageVersion_LibGit2Sharp to see if it will work now

derwasp:


arturcic:
There is already a work in progress PR for #1659, it's not only about nuget package update, it's also about the way it's packaged the msbuild Task and the way the native binaries are loaded

derwasp:
> There is already a work in progress PR for #1659, it's not only about nuget package update, it's also about the way it's packaged the msbuild Task and the way the native binaries are loaded

Yeah, I saw that. That's why it's a draft :) It's just you already have the CI set up so I can get feedback on where what's failing while I poke around it.

derwasp:
@arturcic I wanted to ask you and @asbjornu (as I saw his messages regarding this update issue) if you guys considered releasing the updated version w/o msbuild support for the time being?
I saw that @asbjornu made an attempt of doing it in #1642, but something didn't work out for him and the build didn't pass.
I see that currently there are a few issues (if I understood everything correctly of course):
 - GitVersion doesn't work on e.g. Ubuntu 18.04 (my case).
 - Updating LibGit2Sharp breaks the msbuild task.
 - Fixing the msbuild task requires fiddling, because of the way new native dependencies are distributed.

In any case, it's a bit broken at the moment. If you can release the version w/o msbuild support as even alpha and not beta it could fix the issue I have at the moment. And in the mean time the work on creating a proper msbuild task can continue.
As in #1483 people mentioned that the new LibGit2Sharp fixes the Ubuntu problem. So I am not the only one who would benefit.

Is there a chance that you can consider it?

Issue #1710, opened 2019-06-19
"GitVersion for Bamboo" is no longer supported.


GrayPockets:
The "GitVersion for Bamboo" app is no longer supported, and has been officially archived.

/docs/build-server-support/build-server/bamboo.md

Clicking the download link leads you to the archive: https://marketplace.atlassian.com/plugins/com.carolynvs.gitversion

Links to this app should be removed from the documentation.

arturcic:
Closed by #1711

PR #1711, opened 2019-06-19
Remove reference to the "GitVersion for Bamboo" app

GrayPockets:
This is a direct fix to #1710 with command line instructions for using Linux and the .NET Core GitVersion global tool.


arturcic:
@GrayPockets Thank you for your contribution!! 👍

PR #1722, opened 2019-06-27
(doc) add a hint on using - vs. / for cli switches

pniederlag:
improve user experience. thx for @asbjornu for guidance


arturcic:
Thank you for your contribution. Documentation always remains on later stage. 👍

PR #1724, opened 2019-06-27
added option to disable gitversiontask on build

### LINKED ISSUE #1723: Add option to disable the complete GitVersion.Task for building with no existing repositories
Hi,

since v5.0.0-beta-3-27 it is not possible to use a GitVersion with tools like NCrunch, which copies the source files for building and don't have a copy of the git repository.

@onovotny fixed this in #1193 very NCrunch Specific, these changes are not in the current version anymore. Maybe because of NCrunch is too specific.

Now i must disable the GitVersion.Task with something like

```xml
  <PropertyGroup Condition="'$(NCrunch)' == '1'">
    <WriteVersionInfoToBuildLog>false</WriteVersionInfoToBuildLog>
    <UpdateAssemblyInfo>false</UpdateAssemblyInfo>
    <GenerateGitVersionInformation>false</GenerateGitVersionInformation>
    <GetVersion>false</GetVersion>
    <GenerateGitVersionWixDefines>false</GenerateGitVersionWixDefines>
    <UpdateVersionProperties>false</UpdateVersionProperties>
    <UseFullSemVerForNuGet>false</UseFullSemVerForNuGet>
  </PropertyGroup>
```
which is plenty of stuff, and may change over time.

So I propose a general off switch. So i can do something like.
```xml
  <PropertyGroup Condition="'$(NCrunch)' == '1'">
    <DisableGitVersionTask>true</DisableGitVersionTask>
  </PropertyGroup>
```

I willing to provide a PR, but would like to check the naming, or if there is a strong "we will never do that" within the team.

DerAlbertCom:
- this resolves #1723
- setting the DisableGitVersionTask property to true disables all activities if needed (e.g. NCrunch)
- there is no conditional disabling in the .props file because it must have priority and simple disable it

### DIFF
--- a/src/GitVersionTask.Tests/GitVersionTaskPropertiesTests.cs
+++ b/src/GitVersionTask.Tests/GitVersionTaskPropertiesTests.cs
@@ -1,0 +1,154 @@
+namespace GitVersionTask.Tests
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Text;
+    using System.Xml;
+    using Microsoft.Build.Evaluation;
+    using Microsoft.Build.Framework;
+    using Microsoft.Build.Logging;
+    using NUnit.Framework;
+    using Shouldly;
+
+    [TestFixture]
+    public class GitVersionTaskPropertiesTests
+    {
+        const string projectXml =
+            @"<Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003'>
+<PropertyGroup>
+            <TargetFramework>net461</TargetFramework>
+            <Configuration>Debug</Configuration>
+            <GitVersionTaskRoot>..\GitVersionTask\build\</GitVersionTaskRoot>
+            <GitVersionAssemblyFile>..\..\GitVersionTask.MsBuild\bin\$(Configuration)\$(TargetFramework)\GitVersionTask.MsBuild.dll</GitVersionAssemblyFile>
+</PropertyGroup>
+            <Import Project='$(GitVersionTaskRoot)GitVersionTask.props'/>
+            <Import Project='$(GitVersionTaskRoot)GitVersionTask.targets'/>
+            </Project>";
+
+        (bool failed, Project project) CallMsBuild(IDictionary<string, string> globalProperties)
+        {
+            using (var stringReader = new StringReader(projectXml))
+            {
+                StringBuilder builder;
+                Project project;
+                bool result;
+                using (var collection = new ProjectCollection(globalProperties))
+                {
+                    builder = new StringBuilder();
+                    var writer = new StringWriter(builder);
+                    WriteHandler handler = x => writer.WriteLine(x);
+                    var logger = new ConsoleLogger(LoggerVerbosity.Quiet, handler, null, null);
+                    collection.RegisterLogger(logger);
+                    XmlReader reader = new XmlTextReader(stringReader);
+                    project = collection.LoadProject(reader);
+                    result = project.Build();
+                    collection.UnregisterAllLoggers();
+                }
+
+                var consoleOutput = builder.ToString();
+                Console.WriteLine(consoleOutput);
+
+                return (!result, project);
+            }
+        }
+
+        [Test]
+        public void Zero_properties_should_fail_the_build_because()
+        {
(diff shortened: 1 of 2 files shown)


DerAlbertCom:
i have no idea how to test this in the test project. I found no clue about testing setting build properties. i tried even to break the tests, with setting some properties in the .props file to random values, seems not be used at all 😄

arturcic:
You can run the project using http://msbuildlog.com/ and see the props and the targets that are used

DerAlbertCom:
Didn't know that tool, but seems I'm not worthy ;) Also, i'm not familiar with creating Build Targets.

I've run `msbuild /bl` on the GitVersionTask, GitVersionTask.Test and the complete solution, viewed the logfiles, and found zero evidence that any properties of the .props file is set.

the build.ps1 is stopping on Pack-Gem because i don't have ruby, this can i comment out, but it is failing on other stuff. But this all only for one manual test. This is not a good idea.

Currently i'm trying running msbuild in memory in the unit, that works, but have problems with adding the build targets. will check this out a little bit further. 

Or any other Ideas?

arturcic:
I used that tool when I tested the changes I did when updated to Libgi2sharp 0.26.0. It's really easy to use it.

DerAlbertCom:
I'm not against that tool, i search for the properties in the log, nothing found. i have to create a test project which includes the nuget package. and it is still a manual test.

So i decided to give a unit test a try, so this is currently building. if the build now succeeds (worked with dotnet test, but not with nunit in the build script) I rebase that stuff.

arturcic:
@DerAlbertCom I like the idea of how you tested the properties, I guess we might be using this approach for testing the tasks by running the msbuild the same way. These way we can cover the msbuild tasks as well. I gave a try the test locally with Visual studio test adapter as well as Resharper and the Unit test were failing on my machine. I can see they are also failing on the appveyor.

I would suggest to make some small changes to the code so that the tests can run everywhere (VS + resharper)

``` c#
(bool failed, Project project) CallMsBuild(IDictionary<string, string> globalProperties)
        {
            var currentDirectory = AppDomain.CurrentDomain.BaseDirectory;
            var projectXml =
                $@"<Project xmlns='http://schemas.microsoft.com/developer/msbuild/2003'>
<PropertyGroup>
            <TargetFramework>net461</TargetFramework>
            <Configuration>Debug</Configuration>
            <GitVersionTaskRoot>{currentDirectory}\..\..\..\..\GitVersionTask\build\</GitVersionTaskRoot>
            <GitVersionTaskRoot Condition=""!Exists('$(GitVersionTaskRoot)GitVersionTask.props')"">.\src\GitVersionTask\build\</GitVersionTaskRoot>
            <GitVersionAssemblyFile>{currentDirectory}\..\..\GitVersionTask.MsBuild\bin\$(Configuration)\$(TargetFramework)\GitVersionTask.MsBuild.dll</GitVersionAssemblyFile>
</PropertyGroup>
            <Import Project='$(GitVersionTaskRoot)GitVersionTask.props'/>
            <Import Project='$(GitVersionTaskRoot)GitVersionTask.targets'/>
            </Project>";

            using (var stringReader = new StringReader(projectXml))
            {
                StringBuilder builder;
                Project project;
                bool result;
                using (var collection = new ProjectCollection(globalProperties))
                {
                    builder = new StringBuilder();
                    var writer = new StringWriter(builder);
                    void Handler(string x) => writer.WriteLine(x);
                    var logger = new ConsoleLogger(LoggerVerbosity.Quiet, Handler, null, null);
                    collection.RegisterLogger(logger);
                    XmlReader reader = new XmlTextReader(stringReader);
                    project = collection.LoadProject(reader);
                    result = project.Build();
                    collection.UnregisterAllLoggers();
                }

                var consoleOutput = builder.ToString();
                Console.WriteLine(consoleOutput);

                return (!result, project);
            }
        }
```

DerAlbertCom:
the change with the current directory at runtime made also the NUnit Test Run Workaround obsolete.

If the tests failed on AppVeyour then the PR check did not really worked.

arturcic:
@DerAlbertCom, I have integrated the changes in the props file, but left the tests out, looks like there are some issues with the tasks running in a unit test. I'll investigate that later. Thanks for your contribution. Please check this [version](https://www.nuget.org/packages/GitVersionTask/5.0.0-beta5.1)

PR #1730, opened 2019-07-01
Use dot instead of dash to separate build metadata

### LINKED ISSUE #1717: Fix the version numbers for beta versions of GitVersion
Because the beta releases are not padded, they return the wrong version.
e.g. If I want latest, I get Beta3+4 instead of Beta3+35 because it's sorted A-Z!

.\nuget install GitVersion.CommandLine -ExcludeVersion -PreRelease -OutputDirectory /tools

Should it follow the GitVersion schema instead of 0035 ?

image

asbjornu:
Use dot instead of dash to separate build metadata from the version number. Should resolve #1717.

### DIFF
--- a/build/version.cake
+++ b/build/version.cake
@@ -14,7 +14,7 @@ public class BuildVersion
         var vsixVersion = gitVersion.MajorMinorPatch;
 
         if (!string.IsNullOrWhiteSpace(gitVersion.BuildMetaData)) {
-            semVersion += "-" + gitVersion.BuildMetaData;
+            semVersion += "." + gitVersion.BuildMetaData;
             vsixVersion += "." + DateTime.UtcNow.ToString("yyMMddHH");
         }
 


build/version.cake
```diff
@@ -14,7 +14,7 @@ public class BuildVersion
         var vsixVersion = gitVersion.MajorMinorPatch;
```
asbjornu: Why are we using `LegacySemVer` here, @arturcic? Shouldn't we be able to just use `FullSemVer` for the `semVersion` variable and rely on GitVersion's built-in logic of adding metadata, etc.? As far as I can tell, had we done that, #1717 shouldn't have happened in the first place.

asbjornu:
I would like to not build a version number inside `version.cake` and instead just rely on `gitVersion.FullSemVer`. If we can do that, I would rather change this PR and not have it merged in its current state.

build/version.cake
arturcic: I think the reason I used the `LegacySemVer` instead of `FullSemVer` is because the same variable is used for the `gem` file, and if I'm not wrong there were some issues when a gem with that version was pushed to the respective repository. So for now I would prefer to use the `.` and hopefully the build does not break. I'll merge it as is now.

arturcic:
Looks like chocolatey is not using semver2. Build has failed https://dev.azure.com/GitTools/GitVersion/_build/results?buildId=706&view=logs&j=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&t=c37637b8-ed57-5a02-c556-da2855a7e2d3&l=12930

arturcic:
probably we need to make the change for nuget only, and keep chocolatey the old way

asbjornu:
Yes, that sounds like a plan.

arturcic:
Do you mind if we postpone this after the release 5.0.0?

arturcic:
I want to focus on the remaining issue for release 5.0.0 and prepare the new release.

asbjornu:
Yep, you can just revert this so we get back to a working build. At least now we know it's not a quick-fix. 😄

PR #1767, opened 2019-08-07
Allow head to move for dynamic repositories

GeertvanHorrik:
I've fixed the tests for the dynamic repositories, they should succeed now. They are "slow" (max 30 seconds on my machine), but I think they are required in order to make sure we don't break this functionality in the future.


GeertvanHorrik:
It finally succeeds on the build agents as well... Ready for review!

GeertvanHorrik:
Going to merge this so we can use this in beta, all unit tests succeed.

Issue #1457, opened 2018-08-21
Why is the Build number overwritten in Azure Dev Ops


kirkone:
Hi,

I am using the VSTS task to get my nuget packages versioned on build in VSTS ans I was wondering why the "BUILD_BUILDNUMBER" environment variable is being overwritten completely when there is no GitVersion variable in it. I can not see the benefit of this behavior. Did I miss something?

It would be nice to have a name for a build while using GitVersion.

I think here should be a slightly different logic:
https://github.com/GitTools/GitVersion/blob/8b302e164aefc09d50c3465f7cc8eb48b9f0d136/src/GitVersionCore/BuildServers/VsoAgent.cs#L42

Thank you for your help.
KirK

kirkone:
@JakeGinnivan I know it is a long time ago, but what was the intention behind this change?
I found this commit:
https://github.com/GitTools/GitVersion/commit/e98f68b89148e3e0353a5e80d0c3b75ca293c616
but there I can not see why this change had be made.

If someone want his build number to be replaced I think the better solution would be to set his build number to something like `$(GITVERSION_NuGetVersionV2)`

Did I miss something here?

Best regards
KirK

paule96:
any updates @JakeGinnivan?

JakeGinnivan:
Don't remember really, but a wild guess would be that we couldn't detect we were in VSTS, or there were restrictions on updates. I see no reason to change

kirkone:
hi,
there are cases where you don not need the GitVersion string in the name of the build number. I would even claim it's more the exception to have that. The problem is, that changing the `BUILD_BUILDNUMBER` variable also overrides the name of the build itself. This is not a good behavior that a random task renames the build even it has nothing to do with it.

Maybe for better compatibility it is an option to have an parameter like `--skipbuildnumber` or something equal.
I looked into it if I can provide an PR but I did not find the point where to add such a thing.

I hope this explains a bit more why I opened this issue.

rodchenkov:
I have the same problem, any update on this?

kirkone:
@rodchenkov I opend a PR #1770  for this. We will see how this will come out.

roryprimrose:
This probably should be an option to set on the task in the pipeline. Some people won't want the build number to be updated, others will, others will set the build number to a GitVersion variable for which a substitution could happen.

[...]

PR #1770, opened 2019-08-09
[Azure DevOps] Leave the original Build Number when no GitVersion variable in Build Number

kirkone:
The Build number should not replaced when there is no GitVersion variable in it.

This is discussed in #1457

### DIFF
--- a/docs/input/docs/configuration.md
+++ b/docs/input/docs/configuration.md
@@ -213,6 +213,10 @@ Sets the format which will be used to format the `CommitDate` output variable.
 
 The header for ignore configuration.
 
+### update-build-number
+
+Configures GitVersion to update the build number or not when running on a build server.
+
 #### sha
 
 A sequence of SHAs to be excluded from the version calculations. Useful when
--- a/src/GitVersionCore.Tests/BuildAgents/BuildServerBaseTests.cs
+++ b/src/GitVersionCore.Tests/BuildAgents/BuildServerBaseTests.cs
@@ -50,6 +50,10 @@ namespace GitVersionCore.Tests.BuildAgents
             buildServer.WriteIntegration(writes.Add, variables);
 
             writes[1].ShouldBe("1.2.3-beta.1+5");
+
+            writes = new List<string>();
+            buildServer.WriteIntegration(writes.Add, variables, false);
+            writes.ShouldBeEmpty();
         }
 
         private class BuildAgent : BuildAgentBase
--- a/src/GitVersionCore.Tests/BuildAgents/GitHubActionsTests.cs
+++ b/src/GitVersionCore.Tests/BuildAgents/GitHubActionsTests.cs
@@ -149,6 +149,23 @@ namespace GitVersionCore.Tests.BuildAgents
         }
 
         [Test]
+        public void ShouldNotWriteIntegration()
+        {
+            // Arrange
+            var vars = new TestableVersionVariables("1.0.0");
+
+            var list = new List<string>();
+
+            // Assert
+            environment.GetEnvironmentVariable("GitVersion_Major").ShouldBeNullOrWhiteSpace();
+
+            // Act
+            buildServer.WriteIntegration(s => { list.Add(s); }, vars, false);
+
+            list.ShouldBeEmpty();
+        }
+
+        [Test]
         public void GetEmptyGenerateSetVersionMessage()
         {
             // Arrange
--- a/src/GitVersionCore.Tests/Configuration/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
+++ b/src/GitVersionCore.Tests/Configuration/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
@@ -123,3 +123,4 @@ ignore:
   sha: []
 commit-date-format: yyyy-MM-dd
 merge-message-formats: {}
+update-build-number: false
--- a/src/GitVersionCore.Tests/Configuration/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
+++ b/src/GitVersionCore.Tests/Configuration/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
@@ -123,4 +123,4 @@ ignore:
   sha: []
 commit-date-format: yyyy-MM-dd
 merge-message-formats: {}
-update-build-number: false
+update-build-number: true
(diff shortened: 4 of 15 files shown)


kirkone:
> Could you please add tests to make sure this change works as expected?

I can not find a place where to place these tests. Any help?

kirkone:
I it possible to get a review on this PR?

arturcic:
Sorry for not coming back. To be honest the idea to disable BuildNumber override looks good to know, but I'd prefer to have a more consistent approach, we need to add this ability to all supported CICD not only Azure pipeline, what do you think?

kirkone:
As far as I can see the other pipelines do not behave like this already.
This seems to be only in the code for Azure DevOps.

I can have a look into the other pipelines again and make this sure.

kirkone:
I do not know the other pipelines enough to understand what setting the Build number will have for konequences.
Maybe the Solution mentioned here is more consistent:
https://github.com/GitTools/GitVersion/issues/1457#issuecomment-477467822

I do not know how to implement such a switch but this would allow to disable the the setting of the BuildNumber completly.

The change I made here is only the leave the BuildNumber untouched when there is no GitVersion variable in it. In my opinion this has no side effects to the other pipelines.

kirkone:
@arturcic I checked again and I think the behavior of replacing the whole string when no variable is in it is a thing only in the implementation of the Azure DevOps part.

If you still want the current behavior you can use something like that in your build name: "$(GitVersion.SemVer)".

What do you think?

arturcic:
As far as I can see, these are the CI providers that also override the build number:

AppVeyor - runs a PUT http request to override the build number
ContinuaCi - sets the `setBuildVersion` using their mechanism of messaging
Teamcity - sets the `buildNumber` using their mechanism of messaging
VsoAgent (aka AzurePipelines) - set the `build.updatebuildnumber` using their mechanism of messaging

So as you can see there are at least 4 CIs that by default overrides the build number. I do agree that VsoAgent does that in it's own way, it's trying to resolve GITVERSION variables in the string, and we should fix that to be consistent with the others.

And I do agree that we need to be able to disable the build number overriding in CIs but that should be done on the BuildServerBase level

kirkone:
OK, you are right, but can you give me a hint how to implement such a switch?

I also think that the difference ist how the build environment will handle the change of the buildnumber value. What I want to say is that only Azure Pipelines will use the build number also as the name for the build. I asked for splitting this up on their side but this is not possible as far as I know.
The other environments will not behave like this as far as I know.

asbjornu:
@kirkone, can you please submit a proposal for how you would like this implemented in a least common multiple of what all build servers support? Just a sketch so we have something concrete to discuss, because now it's all a bit too abstract for me to even comment on.

arturcic:
Bump

kirkone:
Hi @asbjornu ,

sorry for the late response.

I think the solution with the least impact on other Pipelines would be to add a new switch for the tool.
Something like `--skipbuildnumber`. If this is set the code block with the replacement will be skipped.

Aside from that I am very sure that replacing the whole string when there is no GitVersion reference in it is not the way most people would expect. If you want to use the replacement put a variable in the build name.

arturcic:
@kirkone seems to be a good idea. I would go with the --updateBuildNumber instead, meaning better to explicitly specify you need to update the build number. Or we can better have a configuration in the config file

`updateBuildNumber` - true|false
`updateBuildNumberFormat` - the build string format

kirkone:
@arturcic but this would be an breaking change then? So the tool would behave in a different way as before when it is called without the parameter?
Or do you mean the default value when not set is true?

arturcic:
Yes it looks like a breaking change. What I mean is if you want to enable a feature, better to specify you need to enable it, instead of not specifying.  We can include in the release notes it's breaking change

arturcic:
@asbjornu Thoughts?

kirkone:
Is there also a way to include this setting in the .yaml file?

arturcic:
> Is there also a way to include this setting in the .yaml file?

That will be the second option. Use config instead of cmdline switch. And I would prefer to have it as 2 configs, one to enable updating, and the other the format

arturcic:
Still this will be a breaking change

kirkone:
The format might not be necessary because if you want the replacement to happen then you easily can add variables to your Build name. That is the same way it works now.
I only want to change what happens if there is no variable set in the BuildName.

arturcic:
Ok, was thinking for all the Build Server we support. Are you in a position to work on a PR in that direction?

kirkone:
besides the changes I made so far I do not have any idea where to add new config values.
I will have a look into it again and update this PR.

arturcic:
You can start looking here https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/Configuration/Config.cs

kirkone:
Startet a new approach ...

kirkone:
I added the configuration value but can not find an example how to access the configuration.

Should I get it from the DI?

arturcic:
> I added the configuration value but can not find an example how to access the configuration.
> 
> Should I get it from the DI?
Check this https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore.Tests/ConfigProviderTests.cs

arturcic:
> I added the configuration value but can not find an example how to access the configuration.
> 
> Should I get it from the DI?
Check this https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore.Tests/ConfigProviderTests.cs

arturcic:
@kirkone I have rebased on top of master and fixed/ added some tests. Thank you for the initial work and the thought process

kirkone:
Thank you @arturcic for finishing this!

I am very happy to see this. Sorry for not getting this ready, but I was not able to get those last changes running on my system.

Again, thanks! I will now prepare all my Pipelines to use this new feature.

kirkone:
hello @arturcic 

sorry for coming back to this but I missed a little detail in your changes.
setting `update-build-number: false` will disable all the output to the build Server. So there are all variables will not be set.

I do not think this is working like it should.

This here:
https://github.com/GitTools/GitVersion/blob/24eba665fa7c3f0b0fd5d718feb34c1e1212ca7d/src/GitVersionCore/Core/BuildAgentBase.cs#L33
should look like:

```csharp
if (writer == null)
{
    return;
}

if (updateBuildNumber)
{
    writer($"Executing GenerateSetVersionMessage for '{GetType().Name}'.");
    writer(GenerateSetVersionMessage(variables));
}
writer($"Executing GenerateBuildLogOutput for '{GetType().Name}'.");
foreach (var buildParameter in GenerateBuildLogOutput(variables))
```

Should I provide a new PR?

arturcic:
Yes please

kirkone:
Sorry for beeing so late here. I had some trouble in real live.

I created a new PR: #2457 

@arturcic can you please have a look?

arturcic:
Glad you came back, hope you're better now

Issue #1765, opened 2019-08-06
GitVersion for PR branch on Azure DevOps not working as expected


onlyutkarsh:
Not entirely sure that it is a gitversion issue but it used to work for us before. Our `gitversion.ym`l is as below

```yaml
next-version: 1.0
assembly-informational-format: '{MajorMinorPatch}'
mode: ContinuousDeployment
branches:
  master:
    tag: ''
  release:
    tag: rc
  feature:
    tag: alpha
    increment: Patch
  pull-request:
    tag: pr
  hotfix:
    tag: rc
  support:
    tag: s
  develop:
    tag: beta
```

We now use GitVersion task in Azure DevOps pipelines and I see that pull requests are being tagged as `PullRequest` instead of what is defined in `gitversion.yml` file `pr`

I see following in the logs, any idea what is causing this?

```
##[section]Starting: GitVersion
==============================================================================
Task         : GitVersion Task
Description  : Easy Semantic Versioning (http://semver.org) for projects using Git
Version      : 5.0.0
Author       : GitVersion Contributors
Help         : See the [documentation](http://gitversion.readthedocs.org/en/latest/) for help
==============================================================================
[command]"C:\Program Files\dotnet\dotnet.exe" C:\a\vm1-1\_work\_tasks\GitVersion_e5983830-3f75-11e5-82ed-81492570a08e\5.0.0\core\GitVersion.dll C:\a\vm1-1\_work\40\s /output buildserver /nofetch /config C:\a\vm1-1\_work\40\s
INFO [08/06/19 11:09:10:85] Working directory: C:\a\vm1-1\_work\40\s
INFO [08/06/19 11:09:10:88] IsDynamicGitRepository: False
INFO [08/06/19 11:09:10:90] Returning Project Root from DotGitDirectory: C:\a\vm1-1\_work\40\s\.git - C:\a\vm1-1\_work\40\s
INFO [08/06/19 11:09:10:90] Running on Windows.
INFO [08/06/19 11:09:10:90] Applicable build agent found: 'VsoAgent'.
INFO [08/06/19 11:09:10:91] Branch from build environment: refs/pull/3378/merge
INFO [08/06/19 11:09:10:91] Begin: Normalizing git directory for branch 'refs/pull/3378/merge'
  INFO [08/06/19 11:09:10:96] One remote found (origin -> 'https://xxxxxxx.visualstudio.com/PDTO/_git/xxxxxxxx').
[...]

Issue #1774, opened 2019-08-13
AzureDevOps: GitVersion Configuration File is Ignored


brunomartinspro:
After upgrading to GitVersion 5 in AzureDevOps, the GitVersion.yml is ignored.
Develop branch always gets the alpha tag when it was specified that the tag would be "unstable".

Is the documentation outdated? https://gitversion.readthedocs.io/en/latest/configuration/#branch-configuration

```
branches:  
  master:
    regex: master
    mode: ContinuousDelivery
    tag: ''
    increment: Patch
    prevent-increment-of-merged-branch-version: true
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
  release:
    regex: releases?[/-]
    mode: ContinuousDelivery
    tag: beta
    increment: Patch
    prevent-increment-of-merged-branch-version: true
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: true
  feature:
    regex: features?[/-]
    mode: ContinuousDelivery
    tag: unstable
    increment: Inherit
    prevent-increment-of-merged-branch-version: false
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
  task:
    regex: task?[/-]
    mode: ContinuousDelivery
    tag: unstable
    increment: Inherit
    prevent-increment-of-merged-branch-version: false
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
    source-branches: ['feature']
    is-source-branch-for: ['feature']
  pull-request:
    regex: (pull|pull\-requests|pr)[/-]
    mode: ContinuousDelivery
    tag: PullRequest
    increment: Inherit
    prevent-increment-of-merged-branch-version: false
    tag-number-pattern: '[/-](?<number>\d+)[-/]'
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
  hotfix:
    regex: hotfix(es)?[/-]
    mode: ContinuousDelivery
    tag: beta
    increment: Patch
    prevent-increment-of-merged-branch-version: false
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
  support:
    regex: support[/-]
    mode: ContinuousDelivery
    tag: ''
    increment: Patch
    prevent-increment-of-merged-branch-version: true
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
  develop:
    regex: dev(elop)?(ment)?$
    mode: ContinuousDeployment
    tag: unstable
[...]

PR #1777, opened 2019-08-13
Only append /config if the user specified a configFilePath

chuseman:
When reading the new configFilePath parameter, there wasn't a check to see if the file path was actually supplied by the user.

This should fix the bug reported by @Skoucail at https://github.com/GitTools/GitVersion/pull/1691#issuecomment-520844538


chuseman:
Looks like this will also fix #1765 and #1774

arturcic:
@chuseman looks good, thanks for fixing it 👍

PR #1785, opened 2019-08-20
Add NoUnix test category

chuseman:
The logic for excluding `NoMono` categorized tests really only checks if running on Unix. Instead, add a new `NoUnix` category for that and make `NoMono` actually mean skip Mono (by detecting a full framework test on Unix).

See https://github.com/GitTools/GitVersion/pull/1759#discussion_r315531632 for discussion.

### DIFF
--- a/build/pack.cake
+++ b/build/pack.cake
@@ -63,10 +63,17 @@ Task("Test")
                     CoverletOutputName = $"{projectName}.coverage.xml"
                 };
 
+                var filters = new List<string>();
+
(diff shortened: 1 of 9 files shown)


arturcic:
I fixed the unix tests for .net core in af4e947b20e85c5cc0ca03e4a5ce7e904695e330 and cleaned up the NoMono category attribute, now only the tests that really needs to be skipped on mono are marked. Thanks anyway for pointing to the right direction. 👍

PR #1787, opened 2019-08-21
Fixed entrypoint for github actions

mcasperson:
Github actions can take advantage of the GitVersion docker image, but they ignore the working directory, so the entrypoint `["dotnet", "GitVersion.dll"]` fails to load. Pointing to the absolute path for the DLL solves the issue.

### DIFF
--- a/src/Docker/Dockerfile
+++ b/src/Docker/Dockerfile
@@ -8,4 +8,4 @@ ARG contentFolder
 WORKDIR /app
 COPY $contentFolder/ ./
 
-ENTRYPOINT ["dotnet", "GitVersion.dll"]
+ENTRYPOINT ["dotnet", "/app/GitVersion.dll"]


arturcic:
Looks good to me! Thanks for the contribution 👍. Could you please provide details on how use use the Github actions with GitVersion docker image? Just curious.

gep13:
@mcasperson @arturcic I would be curious about how you are doing this as well.  Would a dedicated GitHub Action for GitVersion make sense?

mcasperson:
Getting values out of a Github action seems a little clunky at the moment, but you can save the version to a file and consume that in another step like this:

```
name: Python package

on: [push]

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
    - uses: actions/checkout@v1
    - name: Get Git Version
      uses: docker://gittools/gitversion:5.0.2-linux-centos-7-netcoreapp2.2
      with:
        args: /github/workspace /nofetch /exec /bin/sh /execargs "-c \"echo $GitVersion_MajorMinorPatch > /github/workspace/version.txt\""
```

mcasperson:
https://octopus.com/blog/versioning-with-github-actions describes how we have taken advantage of this feature.

arturcic:
Great article @mcasperson! A quick question, have you considered using 
`/showvariable FullSemVer` as argument instead of calling bash to output the env variable?

mcasperson:
@arturcic The challenge with GitHub actions is sharing the result with subsequent jobs. GA doesn't capture the output of a command, so for that variable to be made available it has to be saved in a file because the filesystem is shared between GA jobs. Calling bash via `/exec` looked like the easiest way to use an unmodified docker image and generate a file with the version contents.

Issue #1063, opened 2016-10-06
Doesn't work with git worktrees


AuthorProxy:
On build there is an error while working in git worktree:

> Failed to prepare or find the .git directory in path

```
MyProject/
-- .git/
-- folder_with_project/
---- MyProj.proj
-- worktree/
-- MySln.sln
```

Inside worktree the same structure as in `MyProject/` directory

ghuntley:
Confirmed.

![image](https://user-images.githubusercontent.com/127353/31337174-cfbc3c24-ad45-11e7-8a83-998eb8fcde25.png)

ermshiperete:
Would be nice if GitVersion would work with worktrees.

ermshiperete:
> Inside worktree the same structure as in `MyProject/` directory

The difference is that instead of a .git subdirectory, in a worktree there is a file `.git` that contains the path to the main working tree:

```
gitdir: /path//to/MyProject/.git/worktrees/worktree
```

MarkBinghamS:
This is a lot closer to working now.

This is the output I get:-
INFO [08/19/19 14:55:20:09] Working directory: C:\worktreedir
INFO [08/19/19 14:55:20:10] IsDynamicGitRepository: False
INFO [08/19/19 14:55:20:11] Returning Project Root from DotGitDirectory: C:\repodir\.git\worktrees\worktreedir - C:\repodir\.git\worktrees
INFO [08/19/19 14:55:20:11] Running on Windows.
INFO [08/19/19 14:55:20:13] IsDynamicGitRepository: False
INFO [08/19/19 14:55:20:13] Returning Project Root from DotGitDirectory: C:\repodir\.git\worktrees\worktreedir - C:\repodir\.git\worktrees
INFO [08/19/19 14:55:20:13] Project root is: C:\repodir\.git\worktrees
INFO [08/19/19 14:55:20:13] DotGit directory is: C:\repodir\.git\worktrees\worktreedir
ERROR [08/19/19 14:55:20:13] An unexpected error occurred:
System.ArgumentException: Value does not fall within the expected range.
   at GitVersion.GitVersionCacheKeyFactory.CalculateDirectoryContents(String root)
   at GitVersion.GitVersionCacheKeyFactory.GetGitSystemHash(GitPreparer gitPreparer)
   at GitVersion.GitVersionCacheKeyFactory.Create(IFileSystem fileSystem, GitPreparer gitPreparer, Config overrideConfig
[...]

PR #1793, opened 2019-08-27
Fix several issues related to worktree

ermshiperete:
This PR includes my previous PR (#1792) and generalizes the fix from chuseman (#1759).

Besides fixing the cache key creation when using worktrees it will also fix the location of the GitVersion cache so that the cache directory gets shared between all the worktrees.

### DIFF
--- a/src/GitVersionCore/GitPreparer.cs
+++ b/src/GitVersionCore/GitPreparer.cs
@@ -135,17 +135,14 @@ namespace GitVersion
 
         public string GetDotGitDirectory()
         {
-            if (IsDynamicGitRepository)
-                return DynamicGitRepositoryPath;
+            var dotGitDirectory = IsDynamicGitRepository ? DynamicGitRepositoryPath : Repository.Discover(targetPath);
 
-            var dotGitDirectory = Repository.Discover(targetPath);
-
+            dotGitDirectory = dotGitDirectory?.TrimEnd('/', '\\');
             if (string.IsNullOrEmpty(dotGitDirectory))
                 throw new DirectoryNotFoundException("Can't find the .git directory in " + targetPath);
 
-            dotGitDirectory = dotGitDirectory.TrimEnd('/', '\\');
-            if (string.IsNullOrEmpty(dotGitDirectory))
-                throw new DirectoryNotFoundException("Can't find the .git directory in " + targetPath);
+            if (dotGitDirectory.Contains(Path.Combine(".git", "worktrees")))
+                return Directory.GetParent(Directory.GetParent(dotGitDirectory).FullName).FullName;
 
             return dotGitDirectory;
         }
@@ -159,11 +156,15 @@ namespace GitVersion
                 return targetPath;
             }
 
-            var dotGetGitDirectory = GetDotGitDirectory();
-            using (var repo = new Repository(dotGetGitDirectory))
+            var dotGitDirectory = Repository.Discover(targetPath);
+
+            if (string.IsNullOrEmpty(dotGitDirectory))
+                throw new DirectoryNotFoundException($"Can't find the .git directory in {targetPath}");
+
+            using (var repo = new Repository(dotGitDirectory))
             {
                 var result = repo.Info.WorkingDirectory;
-                Logger.WriteInfo($"Returning Project Root from DotGitDirectory: {dotGetGitDirectory} - {result}");
+                Logger.WriteInfo($"Returning Project Root from DotGitDirectory: {dotGitDirectory} - {result}");
                 return result;
             }
         }
--- a/src/GitVersionCore/GitPreparer.cs
+++ b/src/GitVersionCore/GitPreparer.cs
@@ -160,9 +160,12 @@ namespace GitVersion
             }
 
             var dotGetGitDirectory = GetDotGitDirectory();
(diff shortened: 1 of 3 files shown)


src/GitVersionCore/GitPreparer.cs
```diff
@@ -150,6 +150,21 @@ public string GetDotGitDirectory()
             return dotGitDirectory;
         }
 
+        public string GetDotGitDirectoryOfMain()
+        {
+            var dotGitDir = GetDotGitDirectoryOfWorktreeOrMain();
```
asbjornu: It feels a bit weird to me that something called "get dot git directory **of main**" invokes a method called "get dot git directory **of worktree or main**". It seems like it should be the other way around?

src/GitVersionCore/GitPreparer.cs
```diff
@@ -150,6 +150,21 @@ public string GetDotGitDirectory()
...
+
+            if (!dotGitDir.Contains(Path.Combine(".git", "worktrees")))
+                return dotGitDir;
+
+            return Directory.GetParent(Directory.GetParent(dotGitDir).FullName).FullName;
+        }
+
+        public string GetDotGitDirectory()
+        {
+            return GetDotGitDirectoryOfMain();
```
asbjornu: A method just invoking another method feels a bit unnecessary?

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -42,7 +42,7 @@ static List<string> CalculateDirectoryContents(string root)
 
             if (!Directory.Exists(root))
             {
-                throw new ArgumentException();
+                throw new ArgumentException($"Root directory does not exist: {root}");
```
asbjornu: Why not `DirectoryNotFoundException`?

asbjornu:
With this merged, does it mean we can close #1792, #1759 and #1063? Please rebase this on `HEAD` of `master` and force-push.

src/GitVersionCore/GitVersionCacheKeyFactory.cs
ermshiperete: Done

src/GitVersionCore/GitPreparer.cs
ermshiperete: I agree. I tried to minimize the changes to existing code, but it would be clearer if we'd replace `GetDotGitDirectory` with `GetDotGitDirectoryOfMain` throughout the code base.

src/GitVersionCore/GitPreparer.cs
ermshiperete: We have two types of directories that can contain `.git`: 

a) a regular git repo with a `.git` subdirectory. This might be a normal repo, or the main working tree if worktrees are used.
b) a linked working tree repo with a `.git` file. This file contains the path to the real `.git` directory which resides under `.git/worktrees/<worktreename>` of the main working tree.

In some cases it doesn't matter if we're dealing with case A or B, but in other cases we need to get the `.git` directory of the main working tree.

I tried to name the methods so that this distinction is clear, but maybe you can think of better names?

src/GitVersionCore/GitVersionCacheKeyFactory.cs
```diff
@@ -42,7 +42,7 @@ static List<string> CalculateDirectoryContents(string root)
 
             if (!Directory.Exists(root))
             {
-                throw new ArgumentException();
+                throw new DirectoryNotFoundException($"Root directory does not exist: {root}");
```
asbjornu: 👍

src/GitVersionCore/GitPreparer.cs
ermshiperete: I'm not to fond of these names either. Maybe we can come up with something better.

_Main working tree_ and _linked working tree_ is what is used in the git documentation: "A repository has one main working tree (if it’s not a bare repository) and zero or more linked working trees." (https://git-scm.com/docs/git-worktree).

We could use `GetDotGitDirectoryOfRepo` (instead of `GetDotGitDirectoryOfMain`) and `GetDotGitDirectoryOfWorktree` (instead of `GetDotGitDirectoryOfWorktreeOrMain`), although the latter method name doesn't make it obvious that it returns `.git` directory of the main or linked working tree.

asbjornu: That would be an improvement, yes. 👍

ermshiperete: Done.

arturcic:
Looks like there was an issue when running the windows azure pipeline build, but I ran it manually again and it succeeds.

ermshiperete:
How about now?

arturcic:
@ermshiperete there is a failing test on linux and macos, related to a "/" GetProjectRootDirectory_NoWorktree

ermshiperete:
@arturcic Can you trigger the build again please? Doesn't look like the test failures now are related after I fixed the real failure on linux and macos.

arturcic:
It seems like the pipeline has some intermittent issues. I already triggered one and it succeeded

arturcic:
I think we can merge it. There is a unit test that is randomly failing, but that is not an issue

asbjornu:
Thank you so much @ermshiperete!

PR #1794, opened 2019-08-28
Feature/unit tests vsix

arturcic:
Added initial unit testing for vsix, written in nodejs with mocha.

### DIFF
--- a/build/pack.cake
+++ b/build/pack.cake
@@ -39,7 +39,7 @@ Task("Test")
     .Does<BuildParameters>((parameters) =>
 {
     var frameworks = new[] { parameters.CoreFxVersion, parameters.FullFxVersion };
-    var testResultsPath = parameters.Paths.Directories.TestResultsOutput + "/";
+    var testResultsPath = parameters.Paths.Directories.TestResultsOutput;
 
     foreach(var framework in frameworks)
     {
@@ -60,7 +60,8 @@ Task("Test")
 
                 if (!parameters.IsRunningOnMacOS) {
                     settings.TestAdapterPath = new DirectoryPath(".");
-                    settings.Logger = $"nunit;LogFilePath={MakeAbsolute(new FilePath($"{testResultsPath}{projectName}.results.xml"))}";
+                    var resultsPath = MakeAbsolute(new FilePath($"{testResultsPath}/{projectName}.results.xml"));
+                    settings.Logger = $"nunit;LogFilePath={resultsPath}";
                 }
 
                 var coverletSettings = new CoverletSettings {
@@ -90,7 +91,8 @@ Task("Test")
 
     var workDir = "./src/GitVersionVsixTask";
     var npmSettings = new NpmRunScriptSettings { WorkingDirectory = workDir, LogLevel = NpmLogLevel.Silent, ScriptName = "test" };
-    npmSettings.Arguments.Add($"--reporter-options mochaFile={MakeAbsolute(new FilePath($"{testResultsPath}vsix.results.xml"))}");
+    var vsixResultsPath = MakeAbsolute(new FilePath($"{testResultsPath}/vsix.results.xml"));
+    npmSettings.Arguments.Add($"--reporter-options mochaFile={vsixResultsPath}");
     NpmRunScript(npmSettings);
 })
 .ReportError(exception =>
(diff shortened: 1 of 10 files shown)


build/pack.cake
```diff
@@ -32,6 +39,7 @@ Task("Test")
     .Does<BuildParameters>((parameters) =>
 {
     var frameworks = new[] { parameters.CoreFxVersion, parameters.FullFxVersion };
+    var testResultsPath = parameters.Paths.Directories.TestResultsOutput + "/";
```
asbjornu: I prefer using `Path.Combine` to build paths and also think the path can be built entirely up here instead of when it's used?

```suggestion
    var testResultsPath = Path.Combine(parameters.Paths.Directories.TestResultsOutput, "vsix.results.xml");
    var absoluteTestResultsPath = MakeAbsolute(new FilePath(testResultsPath));
```

build/pack.cake
```diff
@@ -80,6 +87,11 @@ Task("Test")
 
         Parallel.Invoke(options, actions.ToArray());
     }
+
+    var workDir = "./src/GitVersionVsixTask";
+    var npmSettings = new NpmRunScriptSettings { WorkingDirectory = workDir, LogLevel = NpmLogLevel.Silent, ScriptName = "test" };
+    npmSettings.Arguments.Add($"--reporter-options mochaFile={MakeAbsolute(new FilePath($"{testResultsPath}vsix.results.xml"))}");
```
asbjornu: Having a lot of statements inside a string interpolation makes it hard to read. What do you think about this solution?

```suggestion
    npmSettings.Arguments.Add($"--reporter-options mochaFile={absoluteTestResultsPath}");
```

build/pack.cake
arturcic: Good suggestion, I'll review the code and update where possible to Path.Combine or use the Cake extensions for that.

build/pack.cake
arturcic: Looks good to me, I'll adjust, thanks

arturcic:
The tests are actually executed, they do execute the GitVersion.js wrapper, but for now checking all the inputs are set correct and the generated command is the expected. Not testing GitVersion.dll/exe as we do it in the .net unit tests

build/pack.cake
arturcic: Actually `testResultsPath` is used below when computing the path for nunit test results

arturcic:
@asbjornu changes applied

build/pack.cake
```diff
@@ -53,7 +60,8 @@ Task("Test")
 
                 if (!parameters.IsRunningOnMacOS) {
                     settings.TestAdapterPath = new DirectoryPath(".");
-                    settings.Logger = $"nunit;LogFilePath={MakeAbsolute(new FilePath($"{testResultsPath}{projectName}.results.xml"))}";
+                    var resultsPath = MakeAbsolute(new FilePath($"{testResultsPath}/{projectName}.results.xml"));
```
asbjornu: `Path.Combine()` won't be better here?

build/pack.cake
```diff
@@ -80,6 +88,12 @@ Task("Test")
 
         Parallel.Invoke(options, actions.ToArray());
     }
+
+    var workDir = "./src/GitVersionVsixTask";
+    var npmSettings = new NpmRunScriptSettings { WorkingDirectory = workDir, LogLevel = NpmLogLevel.Silent, ScriptName = "test" };
+    var vsixResultsPath = MakeAbsolute(new FilePath($"{testResultsPath}/vsix.results.xml"));
```
asbjornu: `Path.Combine()` won't be better here?

build/pack.cake
arturcic: usually in cake they use FilePath instead of Path, as it's much flexible. All the extensions like MakeAbsolute use FilePath or DirectoryPath

asbjornu: Yes, but I'm thinking about using `Path.Combine()` (or similar variants) so we don't hard code `/` and other OS-dependent path building code, but instead leave that up to the framework and OS.

arturcic: If I change

arturcic: Inspired from the cake team's script `MakeAbsolute(testResultsPath.CombineWithFilePath("vsix.results.xml"));`

PR #1805, opened 2019-09-08
Add PrivateAssets="All" on JetBrains.Annotations for GitVersionCore

derwasp:
The change is pretty self-explanatory.

I am using `GitVersionCore` in my project and saw that it has a reference on the annotations package, which probably is not required during runtime.


arturcic:
Thank you so much for your contributions, @derwasp 👍

PR #1810, opened 2019-09-12
Working/repo directories with named configuration throw warning

### LINKED ISSUE #1808: dotnet Tool: Named Config file throws warning if Working and Repo root directories are the same
GitVersion.Tool 5.0.1

If I have a named config file in the root of my repository, and attempt to specify it, an exception is thrown.  For example:
`dotnet gitversion /config ./myGitVersion.yml`
throws a warning:
`Ambiguous config file selection from 'C:\Projects\MyRepo\GitVersion-foobar.yml' and 'C:\Projects\MyRepo\GitVersion-foobar.yml'`

pianomanjh:
When specifying a named configuration, and the working directory and repository root are the same, do not throw a warning.

For example, my repo root is
`c:\projects\repo`
My named config is at `c:\projects\repo\myConfig.yml`
When running this command, from the repo root:
`dotnet gitversion /config myConfig.yml`
A warning is thrown, and versioning fails.

Additionally, the `TestFileSystem` abstraction was broken, causing one of my tests to erroneously pass.  This is due to:
a) it isn't case insensitive
b) doesn't resolve/nomalize pathing constructs

I've only updated the `GitVersionCore.Tests` copy.  Let me know if you'd like me to fix the other copies of this.  Ultimately, it would be better to go with a well-supported library for this, like `System.IO.Abstractions`

Fixes #1808.


src/GitVersionCore/Configuration/NamedConfigFileLocator.cs
```diff
@@ -35,8 +35,12 @@ private void WarnAboutAmbiguousConfigFileSelection(string workingDirectory, stri
             var workingConfigFile = GetConfigFilePath(workingDirectory, fileSystem);
             var projectRootConfigFile = GetConfigFilePath(projectRootDirectory, fileSystem);
 
+            if (Path.GetFullPath(workingConfigFile).Equals(Path.GetFullPath(projectRootConfigFile)))
```
asbjornu: Will `GetFullPath` normalize the path so a `String.Equals` is sufficient here? I.e. are trailing slashes removed (or added) consistently, folder names adjusted to the physical casing (on case-insensitive file systems), etc.?

pianomanjh: it will normalize the paths (i.e. slashes and paths get normalized) however, casing does not.  
Trailing slashes aren't normalized, but these are filenames, so trailing slashes are not supported.  It would be worth making this case insensitive, however, so I'll post a commit for that.

pianomanjh: I've updated this to behave correctly for case-insensitive platforms.

asbjornu:
Sorry for not merging this earlier, @pianomanjh. Can you please take a look at why the build is failing and perhaps performing a rebase to get rid of my merge commits 855ec8c21cb009f53068d90b1a9abb23cb1d294d and 0702712183b48c30e0da43b7881933eae14c6d0c?

asbjornu:
Thank you for your contributions, @pianomanjh! 🙏

Issue #1208, opened 2017-04-27
Gitversion performance for big repositories


tihobrazov:
For our repo running gitversion takes about 2 mins. Recently we imported another repo with history to our main repo and now it takes up to 15 min to run gitversion. Is it possible to add lightweight mode for version generation? 
Possible options:
- add mode for skipping commits count since last version tag (so, 1.0.0-beta.1 tag will force next commit to be 1.0.0-beta.2+1 - now it counts since the beginning of branch. What is more important here is for gitversion to stop calculating any other details)
- optimize commits-before configuration option, now it continues walking over commits, but just skips them (according to log)

hazzik:
Have you tried `/nofetch`?

JakeGinnivan:
I'd love to optimise this stuff, are you able to post the full log in a gist so we can see what is slow?

Any pull requests in this area welcome.

zpavic:
We have a similar situation with a large team and repo and GitVersion takes minutes to run and sometimes crashes. It may be exacerbated with our team workflow of personal forks, PRs, and merge commits but it still makes using GitVersion basically unworkable.

JakeGinnivan:
Can you post your config and log?

Issue #1066, opened 2016-10-12
GitVersion newer than v3.4.1 has severe performance problems


roryprimrose:
We use GitVersionTask (3.6.3) on all the projects in a solution to ensure that all deployed assemblies reflect the correct version from a build (local or build agent). Something has happened in our git branching whereby GitVersionTask for a project has now gone from ~10 seconds to run WriteVersionInfoToBuildLog, UpdateAssemblyInfo and GetVersion to ~60 seconds. 

This is a big issue for us because there a 16 projects in the solution that use GitVersionTask. Our overall build time has just jumped from 8 minutes to 18-24 minutes.

Is it possible for GitVersionTask to cache the calculation of version metadata and share this between the projects being built? Re-calculating the same information on each project build is very expensive.

discomurray:
If you roll back to GitVersion 3.4.1 does the performance improve?

roryprimrose:
3.6.3 ran for 24.2 minutes
3.4.1 ran for 8.6 minutes

roryprimrose:
Should we be rolling back to 3.4.1?

asbjornu:
@roryprimrose: If you don't need any of the features or bug fixes implemented by 3.4.1+, then by all means downgrade until we figure this one out. 😃 

Caching has been a repeated issue, but usually the other way around since we implemented it in #737; people have complained about the cache being too eager and not being able to invalidate it. So a lot of stuff has happened around cache invalidation lately that I guess affects you in unexpected ways.

If you have a look at [this search for "cache"](https://github.com/GitTools/GitVersion/pulls?utf8=%E2%9C%93&q=is%3Apr%20cache), you can see that there has been done a lot of stuff related to caching, so any one of those pull requests can be the source of your troubles. 

The [diff between 3.4.1 and 3.6.3](https://github.com/GitTools/GitVersion/compare/v3.4.1...v3.6.3) is too big to list on GitHub, so I guess wading through the swathes of commits is going to be too much work, but the source of your problems is hidden in there somewhere. 😄

roryprimrose:
I've been able to narrow it down a little. I've incrementally updated each version and pushed builds. The build against 3.5.4 took 10 minutes. Updating to 3.6.0 took 24 minutes.

The following is a view of the GitVersionTask executing for just one of the projects in the solution. The execution time goes from 14 seconds to 57 seconds.

3.5.4
```
2016-10-13T23:44:49.5597356Z WriteVersionInfoToBuildLog:
[...]

Issue #360, opened 2015-01-31
Wrong version is applied on a branch from a hotfix branch


SeanFeldman:
When branching from `support-x.y` to create a `hotfix-x.y.z` and branching again from `hotfix-x.y.z` to `issue-abc` to have a WIP PR for discussions / code review, GitVersion is applying wrong version.

JakeGinnivan:
Will be fixed in 3.0

PR #1838, opened 2019-10-02
Performance improvements - profiling and simple caching

erikbra:
Used the profiler to find hotspots, most were in Regex matching of
commit messages. Added simple static cache to `IncrementStrategyFinder`.

Runtime halved on medium-sized git codebase (~3 years of history). Same on GitVersion's own codebase.

There are several issues raised concerning performance issues, e.g. #1208, #1066, and more. I am also experiencing issues with a large-ish codebase, so I started looking into it. This one is a low-hanging fruit. Hopefully more to come (e.g. I also suspect that #360 has introduced rather heavy performance issues, but I'll look further into this in a separate PR) 

Medium-sized codebase:

before:
```
       61.21 real        58.71 user         2.26 sys
```

after: 
```
       29.75 real        29.63 user         0.18 sys
```

GitVersion codebase:

before:
```
real	0m2.364s
user	0m2.113s
sys	0m0.262s
```

after: 
```
real	0m0.968s
user	0m0.869s
sys	0m0.097s
```


asbjornu:
@erikbra, thank you so much for your time and effort looking into this! Highly appreciated! The AZDO build is failing, though – would you care to look into why?

erikbra:
Really strange. The build fails only on Linux, and the error is not anywhere near where I've touched: 
```
Mocks/MockRepository.cs(214,16): error CS0104: 'Index' is an ambiguous reference between 'LibGit2Sharp.Index' and 'System.Index' [/home/vsts/work/1/s/src/GitVersionCore.Tests/GitVersionCore.Tests.csproj]
Mocks/MockRepository.cs(7,35): error CS0738: 'MockRepository' does not implement interface member 'IRepository.Index'. 'MockRepository.Index' cannot implement 'IRepository.Index' because it does not have the matching return type of 'Index'. [/home/vsts/work/1/s/src/GitVersionCore.Tests/GitVersionCore.Tests.csproj]
```

I did try to fix it, even though it's unrelated. Don't have a linux box available right here, and it built on my Mac even before the change. Buut let's see if I nailed it anyway ;)

erikbra:
Looks like the build error was introduced in master a couple of commits before my PR. But, nevermind, I fixed it in the PR  :)

erikbra:
Nice build pipeline in Azure DevOps btw... where does the "Release" task put stuff? Is it possible for  me to use a pre-release of the GitVersion task in our own DevOps builds? Or is that a manual process? I'm thinking of using something like this in my .yml?

```
    - task: GitVersion@5.0.2-beta.1+77
      displayName: Get nuget version
      inputs:
        runtime: 'core'
```

arturcic:
You have to install https://marketplace.visualstudio.com/items?itemName=gittools.gitversion-preview into your organization in order to use it. This is the preview version we publish on non release version, basically on every not tagged commit

erikbra:
Thanks, @arturcic . Worked like a charm. And run time on all our builds' GitVersion step is now down from 60-70s to ~20s. Happy, happy!!

PR #1842, opened 2019-10-07
A failing test for incorrect is-mainline setting in mainline development.

da9l:
PR with failing test for issue #1671 
There are two new tests. 

* A non-failing test that set a correct configuration for mainline development (master is-mainline=true)
* A failing test that set an incorrect configuration for mainline development (master is-mainline=false& release is-mainline=true)

Expected behavior: If not supported, this should give a relevant error message instead of an exception.

Disclaimer: This failing test might not cover all cases when this might happen.


src/GitVersionCore.Tests/IntegrationTests/MainlineDevelopmentMode.cs
```diff
@@ -530,6 +530,69 @@ public void VerifyIncrementConfigIsHonoured()
         }
     }
 
+    public class MainlineDevelopmentModeWithBadConfig : TestBase
```
asbjornu: Could you please place this class in its own file?

da9l: Sure. Let me know if there is anything else.

PR #1843, opened 2019-10-07
Change the implementation of AzureDevops task.

arturcic:
Change the implementation of AzureDevops task. Instead of embedding the executables, installing GitVersion.Tool with a versionSpec

``` yaml
  - task: UseGitVersion@5
    inputs:
      versionSpec: '>5.0.2-beta1'
      includePrerelease: true
```

This way we can install different versions of the Gitversion by either specifying the exact version, or a semver range. You can have a look here on the versionSpec

https://semver.npmjs.com/


arturcic:
I have merged @asbjornu , need to check the build works as expected. Thanks for reviewing

PR #1847, opened 2019-10-09
Add the xml namespace to work with MSBuild v14

### LINKED ISSUE #1823: GitVersionTask 5.0.1 does not work with MSBuild-14.0
When I reference the GitVersionTask in version 5.0.1 from a csproj I can't build it any more using MSBuild v14. It is the same error as in #1466, just another file: `GitVersionTask/build/GitVersionTask.props`. It appears that my fix for 4.0 (#1471) did not survive the commit 152de0d

Again, the issue is quite easy to fix: just add the namespace declaration to `src/GitVersionTask/build/GitVersionTask.props`:
```xml
<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
...
```
(the .target file misses the namespace declaration, too, but apparently msbuild does not care in that case)

jbaehr:
This fixes #1823.

Tests have been done with with MSBuild-14, MSBuild-15.8 and NetSdk-2.2 (aka MSBuild-15.9).
MSBuild-15 works without this patch, but MSBuild-14 seems to require the namespace, otherwise you'll encounter the following error:
```
C:\...\GitVersionTask.props(2,1): error MSB4041: The default XML namespace of the project must be the MSBuild XML namespace. If the project is authored in the MSBuild 2003 format, please add xmlns="http://schemas.microsoft.com/developer/msbuild/2003" to the <Project> element. If the project has been authored in the old 1.0 or 1.2 format, please convert it to MSBuild 2003 format. [C:\...\MyProject.csproj]
```

### DIFF
--- a/src/GitVersionTask/build/GitVersionTask.props
+++ b/src/GitVersionTask/build/GitVersionTask.props
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
-<Project>
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
     <PropertyGroup>
         <GitVersionAssemblyFile Condition="$(GitVersionAssemblyFile) == '' And '$(MSBuildRuntimeType)' != 'Core'">$(MSBuildThisFileDirectory)..\tools\net472\GitVersionTask.MsBuild.dll</GitVersionAssemblyFile>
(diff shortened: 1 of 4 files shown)


arturcic:
@jbaehr, thanks for taking the time to implement it again. I'll merge it when the build passes

PR #1849, opened 2019-10-09
Feature/netcore3.0

arturcic:
This PR adds the following:
- [x] Added .net core 3.0 support
- [x] Drop .net 2.2 support, we'll support current version (3.0) and lts version (2.1)
- [x] Made GitVersionCore target netstandard2.0
- [x] Generate SingleExecutable File for each os platform, for now the windows executable is used in the portable and cmdline package. Later we could add homebrew for macos as well as snaps https://snapcraft.io. This will allow to run the GitVersion on macos and linux distros without mono or docker installed as it's self contained.
- [x] Publish tar.gz files, later on we'll add npm support using these tar.gz files


arturcic:
@asbjornu if you don't mind I'll merge the PR

PR #1853, opened 2019-10-11
Fix typo

omerfarukdogan:


asbjornu:
@arturcic, seems like this is failing due to the following:

```
##[error]Artifact storage already exists for build 1092.
```

The source seems to be the following:

https://github.com/GitTools/GitVersion/blob/bc04fad56c7ba913d313af404999b68cb584de60/build/stages/build.yml#L13-L16

The best would probably be to fail that task silently, or to not execute it at all on other branches than `master`. Ideas?

arturcic:
The storage is needed, it is part of the build flow, it is used to store the artifacts between stages, We publish the artifacts on `build` stage and then later on consume them on the next stages. Usually that error happen when you retry a failed `build windows` job, as the windows job is publishing the artifacts, not the linux or macos. In that case it's better to trigger a new build for the same commit

arturcic:
Thank you so much for your contributions, @omerfarukdogan 👍

Issue #1539, opened 2018-11-28
+semver: major bumps minor, should bump major change


MiroKov:
> Adding +semver: breaking or +semver: major will cause the major version to be increased, +semver: feature or +semver: minor will bump minor and +semver: patch or +semver: fix will bump the patch.

We use GitVersion 4.0 nuget package.
When we add message +semver: major, it causes a minor change. Should cause a major change.

waaromikniet:
We are having the same issue

pietergheysens:
Same issue here on my side while building in Azure DevOps using GitVersion 4.0.3. Adding +semver:major in commit message only increments the minor version.

npiskunov:
Also facing the same problem

FixRM:
Same problem, any fixes? I use yaml configuration to override semver regex:
``` YAML
mode: ContinuousDelivery
major-version-bump-message: '(breaking|major):'
minor-version-bump-message: '(feature|minor):'
patch-version-bump-message: '(fix|patch):'
no-bump-message: '(none|skip):'
ignore:
  sha: []
```
'breaking' updates minor version

martinhoza:
Same here. This way we have to start every new project with the version 1.0.0, which can be confusing. Or set 1.0.0 for the first time manually.

jim-paton:
I've just lost a couple of hours because of this issue. The problem appears to be here: https://github.com/GitTools/GitVersion/blob/9e5d4827d57fb60e93f34f8a122c9ef69662b9b1/src/GitVersionCore/IncrementStrategyFinder.cs#L37-L41

I have no problem with this restriction apart from the fact that I can't find any reference to it in the documentation. If it's intentional, then the fact tat the commit message can't be used to increment from `0.x.y` to `1.0.0` needs to be clearly stated in the documentation.

jim-paton:
Actually, I think this restriction should be removed. I tried to work around it by tagging my branch with `v1.0.0` before it got merged - however that then resulted in a version of `1.0.1` when picked up by our build system, so I've actually ended up with a skipped version.

martinhoza:
[...]

PR #1858, opened 2019-10-16
Clarify version bumping for alpha version

Thaoden:
As laid out in this issue #1539, bumping the major version via commit messages only works once the current version hits 1.0.0. This behaviour is not yet documented.


asbjornu:
Thank you so much for your contribution, @Thaoden! 🙏

Fresa:
This should really be added to stdout, in fact, there should be more details on the version selection process in general to clarify why a certain version was chosen, it's really confusing sometimes, specifically this case.

PR #1861, opened 2019-10-16
Feature/di

arturcic:
Refactored code to prepare the addition of DI to improve the modularity/testability.

Some of the changes introduced:

- [x] Introduce GitVersionApplication
- [x] Introduce several interfaces instead of creating class instances directly (IArgumentParser, IConfigFileLocator)
- [x] Replaced the old Logger static class with ILog interface with 2 appenders - ConsoleAppender and FileAppender. Updated the code to use the new ILog

This changes will allow gradually to introduce a DI (Autofac) so that we can later on split in modules some of the functionalities we currently support as addins/plugins call them whatever you want.

The GitVersion itself will just take as input the arguments and return as output the GitVersion information that later can be consumed by these plugins.

Example of plugins could be: 
- AssemblyInfo updater
- BuildServer integration
- In the future we'll be able to decouple the Git metadata reading from the version computation in separate module and so on


src/GitVersionCore/Log/NullLog.cs
```diff
@@ -0,0 +1,24 @@
+using System;
+
+namespace GitVersion.Log
```
asbjornu: Using [gerund](https://en.wikipedia.org/wiki/Gerund) form on words in namespaces avoid collision with class names. Since there could very well exist a class called `Log`, I therefore think `Logging` instead of `Log` is preferable as the namespace.

src/GitVersionCore/SemanticVersionPreReleaseTag.cs
```diff
@@ -95,7 +95,8 @@ public static SemanticVersionPreReleaseTag Parse(string preReleaseTag)
             var match = Regex.Match(preReleaseTag, @"(?<name>.*?)\.?(?<number>\d+)?$");
             if (!match.Success)
             {
-                Logger.WriteWarning($"Unable to successfully parse semver tag {preReleaseTag}");
+                // TODO check how to log this
+                Console.WriteLine($"Unable to successfully parse semver tag {preReleaseTag}");
```
asbjornu: Why the change from `Logger` to `Console`?

src/GitVersionExe/GitVersionOptions.cs
```diff
@@ -0,0 +1,7 @@
+namespace GitVersion
+{
+    public class GitVersionOptions
```
asbjornu: What's the plan with `GitVersionOptions`?

arturcic: I forgot to delete it. Removed

src/GitVersionCore/Log/NullLog.cs
arturcic: I changed the namespace to Logging

src/GitVersionCore/SemanticVersionPreReleaseTag.cs
arturcic: I left it for now like this as it was not easy to inject ILog into the static method. I will consider changing that later. There is a second TODO in Config.cs class

arturcic:
> I whole-heartedly condone and applaud the changes you're doing here, @arturcic! 🎉 👏 ❤️
> 
> I just have a few questions and comments made inline. I also wonder whether with all the refactoring you're doing here, there might be an opportunity to move code from `GitVersionExe` and `GitVersionCore` into `GitVersionCore` in order to tackle #883?

I think you meant `GitVersionExe` and `GitVersionTask` into `GitVersionCore`. That makes sense, but as part of a different refactoring.

asbjornu:
Are these changes so massive that they warrant a new major bump?

arturcic:
It's more a cleanup/refactor than big changes. There are no major breaking changes. There is only one important change, but backward compatible for now. The `verbosity` flag is using both LogLevel and Verbosity enums, but at some point we'll remove the loglevel enum and keep verbosity. The old Verbosity became LogLevel

arturcic:
This change was made to avoid the confusion between verbosity and loglevel. Now the LogLevel is used for the type of the log - Warn, Info and so on, and Verbosity for how verbose the log output is.

arturcic:
and I think we can already merge this, I have a pending draft PR based on this PR that introduce the DI container that you suggested, and I will need some feedback on that one

asbjornu:
Excellent. Great work!

Issue #1875, opened 2019-10-30
GitVersion 5.1.1 AppVeyor: Input string was not in a correct format.


casz:
Fails on this https://ci.appveyor.com/project/nmklotas/gitlabapiclient/builds/28479124#L77

https://github.com/nmklotas/GitLabApiClient/blob/082e81db571dd1ce894e79cacb2ff2734f6ef2c9/appveyor.yml

```
dotnet :   ERROR [10/30/19 6:38:07:03] An unexpected error occurred:
At line:1 char:1
+ dotnet gitversion /l console /output buildserver
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (  ERROR [10/30/...error occurred::String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
 
System.FormatException: Input string was not in a correct format.
   at System.Text.StringBuilder.FormatError()
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, Stri
ng format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(String format, Object[] args)
   at GitVersion.Logging.Log.Write(Verbosity verbosity, LogLevel level, String format, Object[] args) in D:\a\1\s\src\GitVersionCore\Logging\Log.cs:line 45
   at GitVersion.Logging.LogExtensions.Write(ILog log, LogLevel level, String format, Object[] args) in D:\a\1\s\src\GitVersionCore\Logging\LogExtensions.cs:line 100
   at GitVersion.Logging.LogExtensions.Info(ILog log, String format, Object[] args) in D:\a\1\s\src\GitVersionCore\Logging\LogExtensions.cs:line 51
   at GitVersion.Helpers.GitRepositoryHelper.CreateFakeBranchPointingAtThePullRequestTip(ILog log, Repository repo, AuthenticationInfo authentication) in D:\a\1\s\src\GitVersionCore\Helpers\GitRepositoryHelper.cs:line 208
   at GitVersion.Helpers.GitRepositoryHelper.NormalizeGitDirectory(ILog log, IEnvironment environment, String gitDirectory, AuthenticationInfo authentication, Boolean noFetch, String currentBranch, Boolean isDynamicRepository) 
[...]

PR #1878, opened 2019-11-01
Don't try to format non-format strings.

eric-winkler:
Many consumers are doing a `log.Info("somestring")` and have no desire to trigger formatstring behavior.  This change allows them to use `{` or `}` without running into formatting exceptions.

Fixes https://github.com/GitTools/GitVersion/issues/1875 - This was triggered by attempts to log remote tips using the `<rev>^{}` revision syntax.


arturcic:
@eric-winkler thank you very much for having a lóok. LGTM. Merging it.

Issue #1876, opened 2019-10-30
GitVersionTask 5.1.1: object reference not set to an instance of object.


casz:
Fails with 5.1.1, downgrading to 5.0.1 works.

`GitVersionTask.targets(56,9): error : Object reference not set to an instance of an object.`

When building this csproj

```xml
<Project Sdk="Microsoft.Build.NoTargets/1.0.85">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <IncludeBuildOutput>false</IncludeBuildOutput>
    <UpdateAssemblyInfo>false</UpdateAssemblyInfo>
    <SuppressDependenciesWhenPacking>true</SuppressDependenciesWhenPacking>
  </PropertyGroup>

  <PropertyGroup>
    <developmentDependency>true</developmentDependency>
    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
  </PropertyGroup>

  <ItemGroup>
    <Content Include="build\*">
      <PackagePath>build\</PackagePath>
    </Content>
    <Content Include="buildMultiTargeting\*">
      <PackagePath>buildMultiTargeting\</PackagePath>
    </Content>
    <Content Include="tools\**">
      <PackagePath>tools\</PackagePath>
    </Content>
  </ItemGroup>

  <ItemGroup>
    <PackageReference Include="GitVersionTask" Version="5.1.1">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
    </PackageReference>
  </ItemGroup>

</Project>
```

Bit-Chaos:
I do get the same when using GitVersionTask from within a .NET core 3 container. The log shows:
/home/vscode/.nuget/packages/gitversiontask/5.1.1/build/GitVersionTask.targets(16,9): Error: Object reference not set to an instance of an object
Project gets loaded from VSCode and the Omnisharp-plugin fails to load the project because of GitVersionTask.
Using version 5.0.1 I do get the following stack-trace:
[warn]: OmniSharp.MSBuild.ProjectLoader
        WARN [10/31/19 7:21:51:99] Could not determine assembly version: System.BadImageFormatException: Method has no body
File name: 'LibGit2Sharp'
[...]

PR #1887, opened 2019-11-05
GH-1876, object reference not set to an instance of object

arturcic:
Fixes GH-1876


arturcic:
I'll add one test at least.

arturcic:
I added the unit test @asbjornu. I'll merge the fix.

PR #1892, opened 2019-11-06
on exception show stack trace

casz:
@arturcic @asbjornu I would prefer this was the default.

Makes debugging and reporting easier.


arturcic:
If You could fix the build I would merge the change, it makes perfect sense to me

casz:
@arturcic There needs to be some sort of retry: https://dev.azure.com/GitTools/GitVersion/_build/results?buildId=1227&view=results Would be great if we could retrigger it but it seems to require authentication.

arturcic:
@casz, re-trriggered the failed job and it was published

PR #1905, opened 2019-11-13
Improved warning if multiple configurations match

ivan-danilov:
`BranchConfig` does not have `ToString()` overridden, causing `GitVersion.BranchConfig` in the message; the list was ill-formed as well.


asbjornu:
Thank you for your contributions, @ivan-danilov! 🙏

PR #1924, opened 2019-11-20
Added Azure DevOps to the list of build platforms 

ShvetsovZE:


docs/build-server-support/build-server-support.md
```diff
@@ -1,6 +1,7 @@
 # Build Server Support
 GitVersion has support for quite a few build servers out of the box. Currently we support:
 
+ - [Azure DevOps](build-server/azure-devops.md)
```
gep13: To keep the list in alphabetical order, I would suggest that this needs to go after AppVeyor.

arturcic: Agree

arturcic:
Thank you so much for your contributions, @ShvetsovZE 👍

docs/build-server-support/build-server-support.md
ShvetsovZE: Will moved in the next PR, once this one already closed. How soon the changes are going to available on the web site? @arturcic

arturcic: We need to ask @asbjornu as I don't know how the publishing of docs work, or maybe @gep13  knows that

gep13: @arturcic if I recall correctly, there is a webhook going over to readthedocs, so once merged, the docs build runs, and assuming it completes successfully, it happening very quickly.  However, it doesn't look like this has happened yet.  I don't have access to the readthedocs project, so can't see if the docs build has happened or not.  I did notice this:

image

When I recently switched away from readthedocs for GitReleaseManager.  I honestly don't know if that has been done for GitVersion or not.

arturcic: @gep13 do you think we should also switch from readthedocs as you did for GRM?

gep13: @arturcic honestly not sure.  I made the switch because I pretty much use Wyam for everything, so for me, it made sense to change over.  It would be good to figure out "what" is going on with readthedocs, and to get access to administer it, so that we "can" make the switch.  I checked with @GeertvanHorrik and he doesn't have access.  I have just reached out to @JakeGinnivan in the hope that he has access, and can add the team.

PR #1933, opened 2019-11-23
(build) Add posix bootstrapper

gep13:
To allow exectuion of build on Mac.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+
+##########################################################################
+# This is the Cake bootstrapper script for Linux and OS X.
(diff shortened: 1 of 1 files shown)


gep13:
Hmm, ok, looks like there is more work to be done here...

I will need to think about this one some more 😄

arturcic:
Hey @gep13 I wonder why you need the posix script? you can use the powershell version as it runs with powershell core both on windows and posix

gep13:
@arturcic I guess it is just habit.  I tend to have iTerm open, rather than a PowerShell Core session.  I will switch to using that.

PR #1945, opened 2019-12-03
silently skip non-commit tags

cyanite:
Instead of crashing.


cyanite:
To test this I need to be able to make non-commit tags. The testing infrastructure used, however, doesn't seem to allow that, so I didn't write a test. Any testing suggestions are welcome.

arturcic:
Thank you so much for your contributions, @cyanite

PR #1966, opened 2019-12-04
Support for GitHub Actions

### LINKED ISSUE #1931: Add Github Actions BuildAgent

### LINKED ISSUE #1993: Failure after shallow clone via GitHub actions/checkout@v2
See https://github.com/actions/checkout/issues/113. I presume GitVersion can't do anything about this, except maybe improve the error message at non-diagnostic verbosity, although that may be more of a Cake build issue?

StanleyGoldman:
Fixes #1931 
Fixes #1993

Functionality:
- Detects GitHub Actions as a build server
- Outputs environment variables to future GitHub Actions build steps

Notes:
- Unable to set environment variables for current build step.
- GitHub Actions doesn't have the concept of a build version as I am familiar with it in something like AppVeyor
- Changed a resharper setting which locates imports in a strange position

Using Documentation from:

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#default-environment-variables

https://help.github.com/en/actions/automating-your-workflow-with-github-actions/development-tools-for-github-actions#set-an-environment-variable-set-env


arturcic:
@StanleyGoldman thanks for taking the time to implement, I'll have a look later. One think I did not see is the registration in the DI system. Please add the registration here https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/GitVersionCoreModule.cs#L59

StanleyGoldman:
No problem.

StanleyGoldman:
I'm still testing this myself, I'll let you know when I feel confident in it.

StanleyGoldman:
@arturcic I'm ready for your thoughts

src/GitVersionCore/BuildServers/GitHubActions.cs
```diff
@@ -0,0 +1,74 @@
...
+
+                console.WriteLine($"::set-env name={key}::{value}");
+
+                return new[]
+                {
+                    $"Adding Environment Variable to future steps. name='{key}' value='{value}'"
+                };
+            }
+
+            return new string[0];
```
arturcic: I think we need to use the same approach as for the [AzurePipilines ](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/BuildServers/AzurePipelines.cs#L22-L25)
```suggestion
return new[]
{
    $"::set-env name=GitVersion_{name}::{value}"
};
```

src/GitVersionCore/Common/Environment.cs
```diff
@@ -11,5 +13,10 @@ public void SetEnvironmentVariable(string variableName, string value)
         {
             System.Environment.SetEnvironmentVariable(variableName, value);
         }
+
```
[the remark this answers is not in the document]
StanleyGoldman: Yep, you are right, I was trying to set environment variables in the current step and realized that it wouldn't be possible.

src/GitVersionCore/BuildServers/GitHubActions.cs
StanleyGoldman: No problem, I had it this way first, just wasn't sure what would be right.

StanleyGoldman:
@arturcic Fixed up things after your review

src/GitVersionCore/BuildServers/GitHubActions.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            this.console = console;
+        }
+
+        public const string EnvironmentVariableName = "GITHUB_ACTION";
+        protected override string EnvironmentVariable { get; } = EnvironmentVariableName;
+        public override bool CanApplyToCurrentContext()
+        {
+            return !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(EnvironmentVariable));
+        }
```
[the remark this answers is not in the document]
arturcic: Could you remove this? as it was not removed.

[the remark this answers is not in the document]
StanleyGoldman: Eek..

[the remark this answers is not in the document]
StanleyGoldman: Fixed.

StanleyGoldman:
@arturcic fixed up again 😅

arturcic:
Thank you so much for your contributions, @StanleyGoldman. Thanks for the time you spent 👍

src/GitVersionCore/BuildServers/GitHubActions.cs
```diff
@@ -0,0 +1,65 @@
...
+{
+    public class GitHubActions: BuildServerBase
+    {
+        // https://help.github.com/en/actions/automating-your-workflow-with-github-actions/using-environment-variables#default-environment-variables
+
+        public GitHubActions(IEnvironment environment, ILog log) : base(environment, log)
+        {
+        }
+
+        public const string EnvironmentVariableName = "GITHUB_ACTION";
```
TingluoHuang: Should this be `GITHUB_ACTIONS`? 


GITHUB_ACTIONS | Always set to true when GitHub Actions is running the workflow. You can use this variable to differentiate when tests are being run locally or by GitHub Actions.
-- | --

arturcic: Makes sense, are you in a position to send a PR with the fix?

TingluoHuang: https://github.com/GitTools/GitVersion/pull/2006

PR #2008, opened 2019-12-24
Replace [^a-zA-Z0-9-] with - in BranchName variable

KiLLeRRaT:
Otherwise the branch name variable can never be used in any SemVer spec since it can contain a / when using nested branches.

My reasoning to just always replace it is that people will only use the branch name variable when trying to put together a GitVersion tag, which should be SemVer compliant.  So no need to make this setting driven or introduce a new variable just for this.

Cheers,

### DIFF
--- a/src/GitVersionCore/OutputVariables/VariableProvider.cs
+++ b/src/GitVersionCore/OutputVariables/VariableProvider.cs
@@ -5,6 +5,7 @@ using GitVersion.VersionCalculation;
 using GitVersion.VersioningModes;
 using GitVersion.Configuration;
 using GitVersion.Helpers;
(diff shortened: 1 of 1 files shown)


Skoucail:
The BranchName variable has not direct connection with the SemVer.
If people are using the BranchName variable to do checks or reporting this change will break those setups.
Its only the SemVer spec that doesn't allow most special chars.
Variables like BranchName, InformationalVersion, etc all have no problem with those chars.

Skoucail:
The issue will be resolved in #2014

KiLLeRRaT:
@Skoucail Excellent, thank you for that! You can then close this PR.

Cheers,

PR #2013, opened 2020-01-02
Change default for UseFullSemVerForNuGet to true

### LINKED ISSUE #1181: When is NuGetVersionV3 coming?
Browsing through the discussions and the code, I have noticed that the `NuGetVersion` property is currently pointing at the `NuGetVersionV2`, returning the same thing. Also there is this comment `// TODO When NuGet 3 is released: public string NuGetVersionV3 { get { return ??; } }`
https://github.com/GitTools/GitVersion/blob/416ed5ad41a2699bc1dc9fd0731aae1909e1cfdd/src/GitVersionCore/SemanticVersionFormatValues.cs

_NuGet 3.5_ does support _SemVer2_ (well, most of it), and has been out for a while. So I'm wondering when this feature `NuGetVersionV3` might be coming to GitVersion?

For the moment I've resorted to using the `SemVer` property, BUT it can generate version strings longer than 20 characters, which are not supported even by _NuGet 3.5_.

OTOH _NuGet 4.0.0-rc4_ does accept really long strings, solving the problem for now. But that should be the `NuGetVersionV4`, right? :)

casz:
fixes #1181


asbjornu:
@arturcic, don't you agree this should be safe to set `UseFullSemVerForNuGet` to `true` by now?

arturcic:
@asbjornu looks good, sorry on vacation till next week, have no time to have a look

PR #2014, opened 2020-01-03
Replace special chars in InformationalVersion with dashes

Skoucail:
Replacing all characters not allowed in SemVer 2.0 by dashes in the InformationVersion field.
This was already the case for the default format but not if a custom format was used.

2 tests were added to check the fix: one with and one without a format.

Discussion in #1520


src/GitVersionCore.Tests/VariableProviderTests.cs
```diff
@@ -1,15 +1,14 @@
...
-using GitVersion;
 using GitVersion.VersionCalculation;
 using GitVersion.VersioningModes;
+using NUnit.Framework;
+using Shouldly;
+using System;
 
-namespace GitVersionCore.Tests
-{
+namespace GitVersionCore.Tests {
```
casz: Please don't deviate from code formatting

src/GitVersionCore.Tests/VariableProviderTests.cs
```diff
@@ -1,15 +1,14 @@
-using System;
+using GitVersion;
```
casz: Please revert the sort.

src/GitVersionCore/OutputVariables/VariableProvider.cs
```diff
@@ -1,13 +1,13 @@
...
 using GitVersion.VersionCalculation;
 using GitVersion.VersioningModes;
-using GitVersion.Configuration;
-using GitVersion.Helpers;
+using System;
+using System.Text.RegularExpressions;
 
-namespace GitVersion.OutputVariables
-{
+namespace GitVersion.OutputVariables {
```
casz: Ditto

src/GitVersionCore.Tests/VariableProviderTests.cs
Skoucail: Sorry its how my visual studio has been configured. Does it auto on save.
I fixed it with the github editor.

src/GitVersionCore.Tests/VariableProviderTests.cs
Skoucail: Ditto

src/GitVersionCore/OutputVariables/VariableProvider.cs
Skoucail: Ditto

src/GitVersionCore/OutputVariables/VariableProvider.cs
```diff
@@ -56,7 +57,7 @@ public VersionVariables GetVariablesFor(SemanticVersion semanticVersion, Effecti
             var semverFormatValues = new SemanticVersionFormatValues(semanticVersion, config);
 
             var informationalVersion = CheckAndFormatString(config.AssemblyInformationalFormat, semverFormatValues,
-                environment, semverFormatValues.DefaultInformationalVersion, "AssemblyInformationalVersion");
+                environment, semverFormatValues.DefaultInformationalVersion, "AssemblyInformationalVersion").RegexReplace("[^0-9A-Za-z-.+]", "-");
```
asbjornu: Is it not possible for `CheckAndFormatString` to perform the `RegexReplace`? The problem can in theory occur for `assemblyFileSemVer` and `assemblySemVer` as well if their formats or schemes include the prerelease label (tag)?

Skoucail: Hmm.
I see someone removed the fields assembly-versioning-format and assembly-file-versioning-format from the documentation but not from the code.

I would say it not needed to move the regex because for assembly version and assembly file version the rules are way more restricting.
Basicly only the schemes described in https://github.com/GitTools/GitVersion/blob/0a64063fd57099e6ffb96b8bbc5fc731a530aae5/src/GitVersionCore/Extensions/AssemblyVersioningScheme.cs#L5-L9
and 
https://github.com/GitTools/GitVersion/blob/0a64063fd57099e6ffb96b8bbc5fc731a530aae5/src/GitVersionCore/Extensions/AssemblyFileVersioningScheme.cs#L5-L9
are allowed by Microsoft specs.

I guess that is also the reason why the fields are no longer in the documentation for GitVersion 4.x and 5.x.

Also see [Microsoft docs](https://docs.microsoft.com/en-us/dotnet/standard/assembly/versioning)

asbjornu: Good point. Would moving the `RegexReplace` hurt, though? I would prefer it inside the method if it doesn't actually destroy anything.

Skoucail: Nop would not hurt i guess.
Just a theoretical performance hit. But i don't think that will be noticeable.

asbjornu:
Thank you so much for your contributions, @Skoucail! 🙏

PR #2016, opened 2020-01-03
fix empty xml body

casz:


casz:
could properly apply [skip ci]

PR #2021, opened 2020-01-05
Turn Microsoft.Extensions.DependencyInjection into a private dependency

### LINKED ISSUE #2020: Dependency Microsoft.Extensions.DependencyInjection copied to project output folder
GitVersionTask v5.1.3 introduces a dependency on `Microsoft.Extensions.DependencyInjection` for .NETFramework 4.7.2 as well as .NETCoreapp 2.1 and 3.1. 

After retargeting my solution to net472, the above dependency (and indirectly `Microsoft.Extensions.DependencyInjection.Abstraction`) is added to my solution as well. Further, both DLLs are copied to the project output folders, where they will be harvested by my WiX installer and, thus, included in the redistributable I am building.

I would expect `GitVersionTask` and all of its dependencies to be build-time dependencies. Like `GitVersionTask` itself, those DLLs should not be copied to the output folders.

ThomasBarnekow:
This commit adds `PrivateAssets="All"` to the `PackageReference` for the
`Microsoft.Extensions.DependencyInjection` NuGet package.

Fixes #2020.

### DIFF
--- a/src/GitVersionTask/GitVersionTask.csproj
+++ b/src/GitVersionTask/GitVersionTask.csproj
@@ -11,10 +11,21 @@
         <DevelopmentDependency>true</DevelopmentDependency>
 
         <LangVersion>8.0</LangVersion>
-
     </PropertyGroup>
-    <ItemGroup>
+
+    <ItemGroup Condition=" '$(TargetFramework)' == 'net472' ">
         <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.0.1" PrivateAssets="All" />
+    </ItemGroup>
+
+    <ItemGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.1' ">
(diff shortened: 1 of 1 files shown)


arturcic:
Is it possible to have a targetframework == net472 and one !=472? Meaning to have 2 options only

ThomasBarnekow:
Sure, I thought about that as well but then decided to be explicit for each target framework.

Will make the change.

arturcic:
LGTM!

arturcic:
Thank you so much for your contributions, @ThomasBarnekow :+1:

ThomasBarnekow:
You are welcome, @arturcic. Thanks for merging so quickly!

arturcic:
Let me know if the published version when available is working for you

ThomasBarnekow:
@arturcic, I just checked the published version and it works for me. There is still the known issue of updating the NuGet package for projects using `packages.config`, but outside of that, the dependency on `Microsoft.Extensions.DependencyInjection` is removed.

PR #2054, opened 2020-01-19
Fixes #1796 by using a constant file name

### LINKED ISSUE #1796: GitVersionTask breaks incremental builds
Everytime the UpdateAssemblyInfo or GenerateGitVersionInformation target is executed, a randomized file is generated in the intermediate output directory (as implemented in the GetFileWriteInfo extension).
However, this interferes with the update logic at the end of AssemblyInfoFileUpdater.Update() and GitVersionInformationGenerator.Generate(), where the respective file is only updated if its content has changed.
The effect is that even if there were no changes to the project between two builds, the assembly will always be recompiled.

batzen:
When creating a file inside `intermediateOutputPath` the file name must be constant.
When creating a file inside `TempPath` the file name must or should be random.

I am not even sure if using a random file name every time when using `TempPath` is a good idea as we would be polluting the temp path that way.

Fixes #1796.

### DIFF
--- a/src/GitVersionTask.Tests/FileHelperTests.cs
+++ b/src/GitVersionTask.Tests/FileHelperTests.cs
@@ -1,0 +1,46 @@
+using GitVersion.MSBuildTask;
+using NUnit.Framework;
+
+namespace GitVersionTask.Tests
+{
+    [TestFixture]
+    public class FileHelperTests
+    {
+        [Test]
+        [TestCase("C#", "cs")]
+        [TestCase("F#", "fs")]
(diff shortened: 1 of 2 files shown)


asbjornu:
This looks good, but since it was undetected by tests previously, would it be possible to add a test that provokes the old and bad behaviour and thus ensures we won't have any regressions in the future?

batzen:
Added tests for my fixes and also for `GetFileExtension`.

asbjornu:
Superb! Thanks so much for your contributions, @batzen!

PR #2058, opened 2020-01-21
Add json serializer

casz:


casz:
@asbjornu @arturcic it seems to fail on https://github.com/GitTools/GitVersion/blob/master/build.cake#L62

casz:
I suspect it fails on commit message with `:` 🤔

casz:
Reopening old PR

PR #2062, opened 2020-01-22
removing newtonsoft json from tests

casz:


casz:
Newtonsoft Json is doing too much magic 🤔

PR #2067, opened 2020-01-30
fix formatting

casz:
in my few PRs I have done, my editor and I have been fighting over formatting and gratuitous changes!

So here I am fixing them once and for all! 🤪

PR #2069, opened 2020-01-30
change prerelease tag to local.timestamp when building locally

### LINKED ISSUE #1570: Increment version number when local changes are detected
Checking out a project that produces nuget packages on build with GitVersionTask where master branch's latest commit is tagged with 2.0.0. It would be nice to increment the patch number when local changes are detected.

Or when you create a feature branch from master that is still tagged with 2.0.0 it should bump the minor version when local changes are detected.

casz:
fixes #1570


src/GitVersionCore/GitVersionFinder.cs
```diff
@@ -30,7 +30,7 @@ public SemanticVersion FindVersion(GitVersionContext context)
 
             var filePath = Path.Combine(context.Repository.GetRepositoryDirectory(), "NextVersion.txt");
             if (File.Exists(filePath))
-            {
```
casz: one of the gratuitous changes i mentioned in #2067 😓

src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs
```diff
@@ -77,6 +78,12 @@ public SemanticVersion FindVersion(GitVersionContext context)
                 }
             }
 
+            if (context.IsLocalBuild && taggedSemanticVersion == null && context.Repository.RetrieveStatus().IsDirty)
```
casz: This is why this is superior to our current solution we have with our own msbuild target.
We can check if the repo is dirty in GitVersion 😍 

cc @zionyx

casz:
Again seems azure pipelines is sick 😓 

However I think this is ready for review!

src/GitVersionExe.Tests/ExecCmdLineArgumentTest.cs
```diff
@@ -33,7 +33,8 @@ public void RunExecViaCommandLine()
             var result = GitVersionHelper.ExecuteIn(fixture.RepositoryPath, ExecCommand.BuildTool, "RunExecViaCommandLine.csproj /target:OutputResults");
 
             result.ExitCode.ShouldBe(0, result.Log);
-            result.Log.ShouldContain("GitVersion_FullSemVer: 1.2.4+1");
+            var now = DateTime.Now.ToString("yyyyMMddHH");
+            result.Log.ShouldMatch($".+GitVersion_FullSemVer: 1.2.4-local.{now}\\d+\\+1.+");
```
casz: great we already had a test that constituted as a dirty commit 😍

casz: Should properly add a dirty one and none dirty? to see when build is local 👍

src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs
casz: Though I wonder if that's what we want.
We properly want to have all local build marked with `local.timestamp` when I think about our use case for local nuget repository source.

src/GitVersionCore/GitVersionCalculator.cs
```diff
@@ -75,13 +76,13 @@ private string ResolveCurrentBranch(IBuildServer buildServer, string targetBranc
             return currentBranch;
         }
 
-        private VersionVariables GetCachedGitVersionInfo(string targetBranch, string commitId, Config overrideConfig, bool noCache)
+        private VersionVariables GetCachedGitVersionInfo(string targetBranch, string commitId, Config overrideConfig, bool noCache, bool isCiBuild)
```
asbjornu: Shouldn't we stick to either `isLocalBuild` or `isCiBuild` to avoid confusion?

```suggestion
        private VersionVariables GetCachedGitVersionInfo(string targetBranch, string commitId, Config overrideConfig, bool noCache, bool isLocalBuild)
```

casz: Was indecisive on wording I think `isLocalBuild` is best. I must have missed this one.

casz:
@asbjornu the hour thing is only for tests that ignores minutes and seconds cause it would be cause test failures.

The local version number is timestamped down to seconds so everytime you build you will get a new package version with the new version.

Together with custom `PackageOutputPath` and `GeneratePackageOnBuild` this will generate a new package with a new timestamp that you can use in your local package source.

The desire for this is ease local development between libraries.

casz:
I will agree that this may not be something for GitVersion to accept as it is very different premise.

Perhaps it could be hidden behind a GitVersion configuration otherwise we will continue to use our current solution we have in our specialized msbuild targets.

arturcic:
> @asbjornu the hour thing is only for tests that ignores minutes and seconds cause it would be cause test failures.
> 
> The local version number is timestamped down to seconds so everytime you build you will get a new package version with the new version.
> 
> Together with custom `PackageOutputPath` and `GeneratePackageOnBuild` this will generate a new package with a new timestamp that you can use in your local package source.
> 
> The desire for this is ease local development between libraries.

To be honest I don't think we need to cover this specific scenario

arturcic:
> > @asbjornu the hour thing is only for tests that ignores minutes and seconds cause it would be cause test failures.
> > The local version number is timestamped down to seconds so everytime you build you will get a new package version with the new version.
> > Together with custom `PackageOutputPath` and `GeneratePackageOnBuild` this will generate a new package with a new timestamp that you can use in your local package source.
> > The desire for this is ease local development between libraries.
> 
> To be honest I don't think we need to cover this specific scenario

This specific scenario should be handled outside if the GitVersion tool

casz:
No worries, thanks for taking the time to review the PR ❤️

asbjornu:
I think #1385 points to a direction you might take to solve this, @casz. See [`assembly-file-versioning-format`](https://gitversion.net/docs/configuration#assembly-file-versioning-format) for details. It may not support your exact use-case now, but a continuation of it in the right direction might and will be more likely to be something we'll accept and merge.

casz:
Thanks may look into `assembly-file-versioning-format` and whether it is helpful 😅

PR #2070, opened 2020-01-31
set eol to lf to avoid conflicting with editorconfig

casz:
editorconfig says `end_of_line = lf`
https://github.com/GitTools/GitVersion/blob/f0e5275d974d7bd694dad8dce0ac85e57e33b19f/.editorconfig#L4-L10


casz:
not sure why this even fails on linux ONLY.

PR #2071, opened 2020-01-31
fix editorconfig

casz:
fix the syntax to be `*.yml` and no need to override existing values from `*`


casz:
I think azure pipelines are sick 🤔

PR #2072, opened 2020-02-01
use dotnet format

casz:
using dotnet format we can avoid gratuitous changes.

Using GitHub action to do the dotnet format check on pull request to give developer faster feedback rather than adding it to existing ci such as appveyor or azure pipeline.


asbjornu:
Sorry, but I don't understand this PR. Is it perhaps that your local Git config is out of whack?

casz:
The configuration inside editorconfig will try and convert crlf to lf and git on checkout will convert all files from lf to crlf.
Creating an initial conversion in each file when you save them and on every commit git converts them back to crlf before commiting them as lf into the git repo 😓 

by having both gitattributes and editorconfig align on lf the conversion will be avoided.
Also Microsoft recommends lf and UTF-8 without bom for all due to dotnet core.

Also using dotnet format forces clean code no gratulations whitespace permitted into the repo.

src/GitVersionCore.Tests/BuildServers/TeamCityTests.cs
```diff
@@ -19,7 +19,7 @@ public void SetUp()
             });
             buildServer = sp.GetService<TeamCity>();
         }
-
```
casz: These are the gratitious whitespaces that are lying around in the repo waiting for an IDE/Editor to come and cleanup whitespace that needs trimming.

src/GitVersionCore.Tests/Helpers/PathHelper.cs
```diff
@@ -16,15 +16,15 @@ public static string GetExecutable()
 #if NETFRAMEWORK
             var executable = Path.Combine(GetExeDirectory(), "GitVersion.exe");
 #else
-        var executable = "dotnet";
+            var executable = "dotnet";
```
casz: here someone forgot to indent properly and dotnet format takes care for that

src/GitVersionExe/ExecCommand.cs
```diff
@@ -44,12 +44,10 @@ public void Execute()
             switch (arguments.Output)
             {
                 case OutputType.BuildServer:
-                {
```
casz: unnecessary brackets that wasn't even indented correctly

.github/workflows/format.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      # workaround for dotnet tools not being on path https://github.com/actions/virtual-environments/issues/213
+      - name: add .dotnet/tools to path
+        run: echo "::set-env name=PATH::${PATH}:${HOME}/.dotnet/tools"
+
+      - run: dotnet tool install -g dotnet-format
+
+      - run: dotnet format --folder ./ --dry-run --check
```
casz: This PR will fail if formatting is not done correctly.
See: https://github.com/nmklotas/GitLabApiClient/runs/420703815#step:5:7 for example

asbjornu:
I see. If we are able to get this green, I suppose this is a nice addition. What do you think @arturcic?

arturcic:
Adding automated formatting sounds good to me. But I would prefer to have several PRs instead of one huge PR. 

- One PR that will just format the existing code
- One PR that will add workflow for the Github
- One PR that will add the formatting configs + adjust the build script

They could be like part 1-3

@casz  can you think you can split it like that?

casz:
No problem I can split it.

casz:
I think I have sufficiently split the PR 😅

arturcic:
@casz I think this one can be closed

PR #2092, opened 2020-02-11
Fix parallel tests execution, fix potential collisions, cache only regexes

Inok:
Let me explain all these changes:

1. I replaced the static Dictionary with ConcurrentDictionary since its content could be updated by multiple threads. Looks like it is not happening in real scenarios, but the Dictionary breaks almost every time I run unit tests because of `[assembly: Parallelizable(ParallelScope.Fixtures)]`. The more CPU cores you have, the more chances to break the Dictionary.

2. Previous approach to cache something by the GetHashCode() value is wrong because of possible collisions. The more commits you have - the more chances to face a collision (for example, "fix foo +semver:major" and "my commit +semver:patch" messages potentially could have the same hash code, and if it so, the result will depend on the order of this commits in history).

3. Looks like there is no reason to cache increments by commit message text. I believe that it's unusual to have a lot of commits with the same name in real world scenarios. Besides, when the repo has a lot of commits, there will be a huge memory footprint.

4. Bump messages are likely to stay default, so it's reasonable to turn default patterns to static compiled regexes, and also compile and cache custom regexes.

5. I also copy reference to intermediateCommitCache to the local variable before use it because of possible issues with multithreading.

### DIFF
--- a/src/GitVersionCore/VersionCalculation/IncrementStrategyFinder.cs
+++ b/src/GitVersionCore/VersionCalculation/IncrementStrategyFinder.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text.RegularExpressions;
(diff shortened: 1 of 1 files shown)


asbjornu:
The changes look good, but there's a few tests that fail for some reason. I don't quite understand why, but can you please look into it?

arturcic:
> The changes look good, but there's a few tests that fail for some reason. I don't quite understand why, but can you please look into it?

The tests were failing starting with last week, I have enabled back the testing, and it start failing on linux again.

We can disable for now till I figure out the reason and continue merging the changes. What do you think @asbjornu ?

asbjornu:
Sounds good to me, @arturcic.

arturcic:
@asbjornu the build succeeds now, you can continue with the review/merge

asbjornu:
Thanks for your contributions, @Inok! 🙏

Issue #1031, opened 2016-09-02
LibGit2Sharp.LockedFileException: The index is locked. This might be due to a concurrent or crashed process


michaelnoonan:
Wasn't sure if this was an issue or whether we shouldn't be using `msbuild /m` to enable concurrent builds.

```
[11:45:48]  [BuildSolution]     17>MSBUILD : warning : WARN [09/02/16 1:45:48:65] Could not determine assembly version: LibGit2Sharp.LockedFileException: The index is locked. This might be due to a concurrent or crashed process [Y:\Work\refs\heads\master\source\Octopus.E2ETests\Octopus.E2ETests.csproj]
[11:45:48]  [BuildSolution] MSBUILD : warning :    at LibGit2Sharp.Core.Ensure.HandleError(Int32 result) [Y:\Work\refs\heads\master\source\Octopus.E2ETests\Octopus.E2ETests.csproj]
[11:45:48]  [BuildSolution] MSBUILD : warning :    at LibGit2Sharp.Core.Proxy.git_checkout_tree(RepositoryHandle repo, ObjectId treeId, GitCheckoutOpts& opts) [Y:\Work\refs\heads\master\source\Octopus.E2ETests\Octopus.E2ETests.csproj]
[11:45:48]  [BuildSolution] MSBUILD : warning :    at LibGit2Sharp.Repository.CheckoutTree(Tree tree, IList`1 paths, IConvertableToGitCheckoutOpts opts) [Y:\Work\refs\heads\master\source\Octopus.E2ETests\Octopus.E2ETests.csproj]
[11:45:48]  [BuildSolution] MSBUILD : warning :    at LibGit2Sharp.Repository.Checkout(Tree tree, CheckoutOptions checkoutOptions, String refLogHeadSpec) [Y:\Work\refs\heads\master\source\Octopus.E2ETests\Octopus.E2ETests.csproj]
[11:45:48]  [BuildSolution] MSBUILD : warning :    at LibGit2Sharp.Repository.Checkout(Branch branch, CheckoutOptions options) [Y:\Work\refs\heads\master\source\Octopus.E2ETests\Octopus.E2ETests.csproj]
[11:45:48]  [BuildSolution] MSBUILD : warning :    at LibGit2Sharp.Repository.Checkout(String committishOrBranchSpec, CheckoutOptions options) [Y:\Work\refs\heads\master\source\Octopus.E2ETests\Octopus.E2ETests.csproj]
[...]

Issue #983, opened 2016-07-28
Add additional output type for saving asserted version numbers to a file


gep13:
It would allow some consistency when calling GitVersion with `/output buildserver` where the variables are not output to the stdout, since this is not possible.

We are already creating the cache file on each run, so it should be simple enough to take a file path from the user, and save the file there also.

As discussed on Gitter here:

https://gitter.im/GitTools/GitVersion?at=5799fe2c93148c6b21e580a5

kritisingh1:
@gep13 Hi. I would like to work on this issue. Please let me know if I may. I am a beginner and this would be my first contribution to open source.

asbjornu:
@kritisingh1: Awesome! Please go ahead and:

1. [Fork the repository](https://help.github.com/articles/fork-a-repo/)
2. [Create a new branch](https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging)
3. Commit the fixes to the branch
4. Push the branch to your remote fork
5. [Create a pull request](https://help.github.com/articles/creating-a-pull-request/)

If you have any questions, don't hesitate to ask them here or on [Gitter](https://gitter.im/GitTools/GitVersion).

WhiteWaterCoder:
Is this issue still up for grabs? I might try and have a stab at it in my minimal spare time.

asbjornu:
@WhiteWaterCoder: Yes, absolutely! That would be awesome!

PR #2128, opened 2020-02-22
GH-2127 - allow json and buildserver output at the same time

### LINKED ISSUE #1236: "/output json" does not work in build server environment
Hello

We would like to execute GitVersion from withing a build script, but would like to get the json output. If the build environment is detected, the option json is ignored.

This was already issued once: https://github.com/GitTools/GitVersion/issues/675 

Apparentely, this bug seems to be still open.

You can repro it by yourself by executing the following inside a PowerShell session:
```
$env:TF_BUILD = "True"
GitVersion.exe /output json /nofetch
```

The output is still the "buildserver" one.

### LINKED ISSUE #2127: Allow specifying -output json and -output buildserver at the same time
Might be a continuation of #1236

arturcic:
Closes #2127 Closes #1236

### DIFF
--- a/src/GitVersionCore/Arguments.cs
+++ b/src/GitVersionCore/Arguments.cs
@@ -8,6 +8,10 @@ namespace GitVersion
 {
     public class Arguments
     {
+        public Arguments()
+        {
+            Output.Add(OutputType.Json);
+        }
         public AuthenticationInfo Authentication;
 
         public Config OverrideConfig;
@@ -44,7 +48,7 @@ namespace GitVersion
         public bool NoCache;
         public bool NoNormalize;
 
-        public OutputType Output = OutputType.Json;
+        public ISet<OutputType> Output = new HashSet<OutputType>();
         public Verbosity Verbosity = Verbosity.Normal;
 
         public bool UpdateAssemblyInfo;
--- a/src/GitVersionCore/Arguments.cs
+++ b/src/GitVersionCore/Arguments.cs
@@ -8,10 +8,6 @@ namespace GitVersion
 {
     public class Arguments
     {
-        public Arguments()
-        {
-            Output.Add(OutputType.Json);
-        }
         public AuthenticationInfo Authentication;
 
         public Config OverrideConfig;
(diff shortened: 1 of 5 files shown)


arturcic:
> What happens if you duplicate `-output json`, for instance?

We use ISet, that means only one instance will be saved.


> Also, it's now impossible to **not** output JSON, true?

We add by default in the constructor of `Arguments` the json because we need to preserve the way it worked before

arturcic:
But **not** using json at all, I think the old behavior allowed that. Let me check

arturcic:
@asbjornu now it's possible to add only `-output buildserver` without `output json` as well as multiple times `-output json`  or `-output buildserver` or a combination of those

gep13:
Interesting! This will help with one of the long time issues with using GitVersion and Cake on CI system. Namely that you would have to run GitVersion twice, once to set the BuildServer version and then again to get the asserted variables to use in rest of script.

arturcic:
> Interesting! This will help with one of the long time issues with using GitVersion and Cake on CI system. Namely that you would have to run GitVersion twice, once to set the BuildServer version and then again to get the asserted variables to use in rest of script.

Right, that was one of the reasons I implemented this, to avoid running twice.

arturcic:
@gep13 I think this change will break cake alias for GitVersion, am I right?

gep13:
Off the top of my head, I am not sure. I would need to check. Travelling at the minute, so not in a position to look at this just now.

arturcic:
@gep13 I think this will still work with the cake aliases, just whoever uses the cake aliases will not be able to specify both the `json` and `buildserver` output. So I think we can merge this. @asbjornu ?

asbjornu:
Sure. But just to be clear: This closes #1236 and probably makes #983 obsolete as well, no? Writing the console output to a file is just `> GitVersion.json` in most console environments.

asbjornu:
Unless we use the implementation of #983 to fix #1031. 🤔

arturcic:
I think it will close #1236, 
As for #983  just > GitVersion.json might work except when the tool fails it will write to error to the GitVersion.json file so I'd prefer a new output instead.
As for #1031 it might partially fix the issue as we will only call it once instead of twice, making it less probable, but in the case of GitVersionTask it might occur as for each task it calls the GitVersion (something to refactor I think)

asbjornu:
Good analysis. Please add `Closes #1236` in your PR description so it gets closed automatically on merge and then just merge ahead. :shipit

PR #2141, opened 2020-02-26
Remove obsolete config file

arturcic:
As the legacy configuration file was deprecated long time ago, I think in version 4.x we're removing it in this version


asbjornu:
I think this change warrants a new major version. What do you think?

arturcic:
Yeah, or it might be 5.2.0 as it's not breaking, it just removes the obsolete

arturcic:
I was considering to change the milestone from 5.1.4 to 5.2.0 as there are a lot of issues/changes

asbjornu:
Yeah, agreed. As you write, the deprecation warning has been there since v4, so let's bump to 5.2 and get rid of it.

PR #2154, opened 2020-03-04
Fix for GH-2153

### LINKED ISSUE #2153: The "WriteVersionInfoToBuildLog" task could not be loaded from the assembly
Build Issue after installing `5.1.4-beta1.216`

When I install a fresh copy of GitVersionTask on a .NET Framework 4.7.2 project, I get the following error when trying to build

```
Severity	Code	Description	Project	File	Line	Suppression State
Error		The "WriteVersionInfoToBuildLog" task could not be loaded from the assembly C:\ProjectsG.Git\Gouws.WakeOnLan\packages\GitVersionTask.5.1.4-beta1.216\build\..\tools\net472\GitVersionTask.MsBuild.dll. Could not load file or assembly 'System.Runtime.Loader, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.	Gouws.WakeOnLan	C:\ProjectsG.Git\Gouws.WakeOnLan\packages\GitVersionTask.5.1.4-beta1.216\build\GitVersionTask.targets	10	
```

Any ideas? I have been regularly trying the latest pre-release versions because I'm after some functionality in there, but each time I'm hit with a build error.

Cheers,

arturcic:
Closes #2153

### DIFF
--- a/src/GitVersionTask.MsBuild/GitVersionTask.MsBuild.csproj
+++ b/src/GitVersionTask.MsBuild/GitVersionTask.MsBuild.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0</TargetFrameworks>
+        <TargetFrameworks>net472;netstandard2.0</TargetFrameworks>
         <LangVersion>8.0</LangVersion>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     </PropertyGroup>
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -22,7 +22,7 @@
         <PackageVersion_JetBrainsAnnotations>2019.1.3</PackageVersion_JetBrainsAnnotations>
         <PackageVersion_YamlDotNet>8.1.0</PackageVersion_YamlDotNet>
         <PackageVersion_NewtonsoftJson>12.0.3</PackageVersion_NewtonsoftJson>
-        <PackageVersion_MicrosoftExtensions>3.1.2</PackageVersion_MicrosoftExtensions>
+        <PackageVersion_MicrosoftExtensions>3.0.0</PackageVersion_MicrosoftExtensions>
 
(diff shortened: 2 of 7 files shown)


src/GitVersionTask.MsBuild/GitVersionTask.MsBuild.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>netstandard2.0</TargetFrameworks>
+        <TargetFrameworks>net472;netstandard2.0</TargetFrameworks>
```
dazinator: I think it was this that was the main part of the problem. This assembly needs to be cross compiled for full net framework and .net core, .net core msbuild ends up using the .net standard compiled one, and .net framework msbuild needs to use the full .net (net472) compiled version of the assembly - that's becuase in this assembly there is stuff like this:
```
  static TaskProxy()
        {
#if !NETFRAMEWORK
            GitLoaderContext.Init("GitVersionCore", "LibGit2Sharp");
#endif
            LibGit2SharpLoader.LoadAssembly("GitVersionTask");
```
If we only build the .net standard version of the binary to be used by both .net core and net framework msbuild variants, then it will now not include the full .NET dependency resolution logic. 

I had fixed this in my PR, but I broke other things when trying to refactor task proxy. I have decided to start my PR again from a clean branch off latest master, and I'll just do some more targetted fixes.

arturcic: I'm working now on adding some integration tests using msbuild and dotnet build to validate the GitVersionTask,  when that is in place I think we can release, and you can use that for validating the refactoring you're doing. As for the support for preview builds of msbuild I think we need to handle that after this release,

PR #2155, opened 2020-03-05
Some light enhancements for GitVersionTask

dazinator:
- Added a README to GitVersionTask with hopefully some useful tidbits of info that may help devs.
- Build Tweaks:
    - PackageReference for `System.Runtime.Loader` now conditional to .NET Core only.
    - Removed unnecessary target framework form GitVersionTask's csproj - didn't seem to be used anywhere.

- Removed dependency on `Microsoft.Build.Tasks.Core` which we used for its base `Task` implementation. Not sure how much this will help with keeping compatibility accross different msbuild versions, but it can't hurt the cause.

- Debugging tweak: was having trouble debugging the static initialisers as I couldn't get a breakpoint to be hit within them - made a slight tweak to capture the exception that occurs in the static type initilaiser, and then re-throw it on Task execution in an effort to aid debug-ability.

I tested this version of GitVersionTask locally, and it seems to work the same as current master: 5.1.4-beta1.220 package. Both versions seem to work fine for me, except under `dotnet sdk 3.1.200-preview-014977` - for which both versions seem to exhibit the same bug - which isn't addressed by this PR.

### DIFF
--- a/src/GitVersionTask.MsBuild/GitVersionTask.MsBuild.csproj
+++ b/src/GitVersionTask.MsBuild/GitVersionTask.MsBuild.csproj
@@ -5,11 +5,11 @@
         <LangVersion>8.0</LangVersion>
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     </PropertyGroup>
-
+   
     <ItemGroup>
-        <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+        <PackageReference Condition="'$(TargetFramework)' == 'netstandard2.0'" Include="System.Runtime.Loader" Version="4.3.0" PrivateAssets="All" />
         <PackageReference Include="Microsoft.Build" Version="16.4.0" NoWarn="NU1701" />
-        <PackageReference Include="Microsoft.Build.Tasks.Core" Version="16.4.0" />
+        <PackageReference Include="Microsoft.Build.Framework" Version="16.4.0" />
     </ItemGroup>
 
     <ItemGroup>
--- a/src/GitVersionTask.MsBuild/Tasks/GenerateGitVersionInformation.cs
+++ b/src/GitVersionTask.MsBuild/Tasks/GenerateGitVersionInformation.cs
@@ -16,6 +16,7 @@ namespace GitVersion.MSBuildTask.Tasks
         [Output]
         public string GitVersionInformationFilePath { get; set; }
 
-        public override bool Execute() => TaskProxy.GenerateGitVersionInformation(this);
+        protected override bool OnExecute() => TaskProxy.GenerateGitVersionInformation(this);
+
     }
 }
--- a/src/GitVersionTask.MsBuild/Tasks/GetVersion.cs
+++ b/src/GitVersionTask.MsBuild/Tasks/GetVersion.cs
@@ -94,6 +94,7 @@ namespace GitVersion.MSBuildTask.Tasks
         [Output]
         public string CommitsSinceVersionSourcePadded { get; set; }
 
-        public override bool Execute() => TaskProxy.GetVersion(this);
+        protected override bool OnExecute() => TaskProxy.GetVersion(this);
+
     }
 }
--- a/src/GitVersionTask.MsBuild/Tasks/UpdateAssemblyInfo.cs
+++ b/src/GitVersionTask.MsBuild/Tasks/UpdateAssemblyInfo.cs
@@ -19,6 +19,6 @@ namespace GitVersion.MSBuildTask.Tasks
         [Output]
         public string AssemblyInfoTempFilePath { get; set; }
 
-        public override bool Execute() => TaskProxy.UpdateAssemblyInfo(this);
+        protected override bool OnExecute() => TaskProxy.UpdateAssemblyInfo(this);
     }
 }
--- a/src/GitVersionTask.MsBuild/Tasks/WriteVersionInfoToBuildLog.cs
+++ b/src/GitVersionTask.MsBuild/Tasks/WriteVersionInfoToBuildLog.cs
@@ -2,6 +2,6 @@ namespace GitVersion.MSBuildTask.Tasks
 {
     public class WriteVersionInfoToBuildLog : GitVersionTaskBase
     {
-        public override bool Execute() => TaskProxy.WriteVersionInfoToBuildLog(this);
+        protected override bool OnExecute() => TaskProxy.WriteVersionInfoToBuildLog(this);
     }
 }
--- a/src/GitVersionTask/GitVersionTask.csproj
+++ b/src/GitVersionTask/GitVersionTask.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
     <PropertyGroup>
-        <TargetFrameworks>net472;netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+        <TargetFrameworks>net472;netcoreapp2.1</TargetFrameworks>
 
         <PackageId>GitVersionTask</PackageId>
         <Title>GitVersionTask</Title>
@@ -11,10 +11,12 @@
         <DevelopmentDependency>true</DevelopmentDependency>
 
         <LangVersion>8.0</LangVersion>
-    </PropertyGroup>
+    </PropertyGroup>   
 
     <ItemGroup>
-        <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(PackageVersion_MicrosoftExtensions)" PrivateAssets="All" />
+        <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(PackageVersion_MicrosoftExtensions)" PrivateAssets="All">
+            <Publish>true</Publish>
+        </PackageReference>
         <PackageReference Include="LibGit2Sharp" Version="$(PackageVersion_LibGit2Sharp)" PrivateAssets="All" />
     </ItemGroup>
 
--- a/src/GitVersionTask/README.md
+++ b/src/GitVersionTask/README.md
@@ -1,0 +1,17 @@
+## UNDERSTAND THIS
+
+1. On .NET Core, our MSBuild tasks cannot be executed within the AssemblyLoadContext that they are loaded into my MsBuild.
+This is because our MSBuild tasks have dependencies that must be loaded (like libgit2sharp etc) and we have no way to extend
+the MsBuild AssemblyLoadContext that MsBuild has loaded them into, to help satisfy loading these dependencies from the nuget package folder.
+
+2. When running on the full .NET framework this isn't a problem because in that case we can access the current AppDomain and add AssemblyResolve handlers to load these dependencies as needed.
+
+3. On .NET Core then, we create a new AssemblyLoadContext, and load try to execute our task within it but this is tricky because:
+   - If you try to load the same task Type and use it in another AssemblyloadContext (for example by creating an instance of it and calling Execute) you will
+   get weird errors along the lines of "cannot cast type A to type B"even though the two types are the same.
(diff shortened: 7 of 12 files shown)


global.json
```diff
@@ -0,0 +1,5 @@
+{
+  "sdk": {
+    "version": "3.1.101"
```
arturcic: Can you put the same version as in the build.config?

src/GitVersionTask.MsBuild/GitVersionTask.MsBuild.csproj
```diff
@@ -6,8 +6,12 @@
         <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     </PropertyGroup>
 
+    <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0' ">
+        <DefineConstants>UseSystemRuntimeLoader</DefineConstants>
+    </PropertyGroup>
+
     <ItemGroup>
-        <PackageReference Include="System.Runtime.Loader" Version="4.3.0" />
+        <PackageReference Condition="$(DefineConstants.Contains(UseSystemRuntimeLoader))" Include="System.Runtime.Loader" Version="4.3.0" PrivateAssets="All" />
```
arturcic: I wonder if you can use directly the Targetframework property instead of creating a new one

src/GitVersionTask.MsBuild/Tasks/GenerateGitVersionInformation.cs
```diff
@@ -16,6 +16,14 @@ public class GenerateGitVersionInformation : GitVersionTaskBase
         [Output]
         public string GitVersionInformationFilePath { get; set; }
 
-        public override bool Execute() => TaskProxy.GenerateGitVersionInformation(this);
+        public override bool Execute()
```
arturcic: Can we move this check into the base class?

src/GitVersionTask.MsBuild/Tasks/GenerateGitVersionInformation.cs
```diff
@@ -16,6 +16,14 @@ public class GenerateGitVersionInformation : GitVersionTaskBase
...
+        {
+            if (TaskProxy.InitialiseException != null)
```
arturcic: Can we move this check into the base class?

src/GitVersionTask.MsBuild/Tasks/GetVersion.cs
```diff
@@ -94,6 +94,14 @@ public class GetVersion : GitVersionTaskBase
         [Output]
         public string CommitsSinceVersionSourcePadded { get; set; }
 
-        public override bool Execute() => TaskProxy.GetVersion(this);
+        public override bool Execute()
+        {
+            if (TaskProxy.InitialiseException != null)
+            {
```
arturcic: Can we move this check into the base class?

src/GitVersionTask.MsBuild/Tasks/UpdateAssemblyInfo.cs
```diff
@@ -19,6 +19,14 @@ public class UpdateAssemblyInfo : GitVersionTaskBase
         [Output]
         public string AssemblyInfoTempFilePath { get; set; }
 
-        public override bool Execute() => TaskProxy.UpdateAssemblyInfo(this);
+        public override bool Execute()
+        {
+            if (TaskProxy.InitialiseException != null)
```
arturcic: Same here

src/GitVersionTask.MsBuild/Tasks/WriteVersionInfoToBuildLog.cs
```diff
@@ -2,6 +2,14 @@ namespace GitVersion.MSBuildTask.Tasks
 {
     public class WriteVersionInfoToBuildLog : GitVersionTaskBase
     {
-        public override bool Execute() => TaskProxy.WriteVersionInfoToBuildLog(this);
+        public override bool Execute()
+        {
+            if (TaskProxy.InitialiseException != null)
```
arturcic: Same here

src/GitVersionTask.MsBuild/GitVersionTask.MsBuild.csproj
dazinator: I can do, this was personal preference as I prefer to keep condition expressions simple. In this case there is only one target framework involved so my approach isn't really paying much dividends. I'll change it.

global.json
arturcic: I think the artifacts test that test against 2.1 will fail because of this file, I remember I removed it because of that. I added a global json in the parent folder that contains my local copy of the repo

dazinator:
@arturcic looks like dotnet core sdk `3.1.102` now needs to be installed here: https://github.com/GitTools/GitVersion/pull/2155/checks?check_run_id=488081585 - is this something you can do?

arturcic:
> @arturcic looks like dotnet core sdk `3.1.102` now needs to be installed here: https://github.com/GitTools/GitVersion/pull/2155/checks?check_run_id=488081585 - is this something you can do?

One more reason to avoid global.json in the repo

dazinator:
> > @arturcic looks like dotnet core sdk `3.1.102` now needs to be installed here: https://github.com/GitTools/GitVersion/pull/2155/checks?check_run_id=488081585 - is this something you can do?
> 
> One more reason to avoid global.json in the repo

I think a requirement on a specific patch level version of the SDK is probably too specific.
However a global.json with say `2.1` or `3.1` in, stops developers potentially hitting breaking changes when they checkout and build the repo and they have a newer (or older) version of the sdk installed on their environment locally - and so it ends up building with their version. This may not be a problem for this repository - so I can remove it. 

In terms of the version of the SDK used by the github workflows - how is this controlled? And shouldn't it be the same version used by build.config anyway?

arturcic:
> > > @arturcic looks like dotnet core sdk `3.1.102` now needs to be installed here: https://github.com/GitTools/GitVersion/pull/2155/checks?check_run_id=488081585 - is this something you can do?
> > 
> > 
> > One more reason to avoid global.json in the repo
> 
> I think a requirement on a specific patch level version of the SDK is probably too specific.
> However a global.json with say `2.1` or `3.1` in, stops developers potentially hitting breaking changes when they checkout and build the repo and they have a newer (or older) version of the sdk installed on their environment locally - and so it ends up building with their version. This may not be a problem for this repository - so I can remove it.
> 
> In terms of the version of the SDK used by the github workflows - how is this controlled? And shouldn't it be the same version used by build.config anyway?

It's using the agent's dotnet installation, same for azdo. They periodically update.

dazinator:
> It's using the agent's dotnet installation, same for azdo. They periodically update.

It seems to be happy with "3.1" in the global.json.
However if you want to allow a bit more flexibility, other than removing it, I could set it to:

```
  "sdk": {
        "allowPrerelease": false
    }
```

That would eliminate a trust issue where we currently trust that the agents won't install a preview version that ends up breaking the build in perhaps exotic ways.. ?

arturcic:
> > It's using the agent's dotnet installation, same for azdo. They periodically update.
> 
> It seems to be happy with "3.1" in the global.json.
> However if you want to allow a bit more flexibility, other than removing it, I could set it to:
> 
> ```
>   "sdk": {
>         "allowPrerelease": false
>     }
> ```
> 
> That would eliminate a trust issue where we currently trust that the agents won't install a preview version that ends up breaking the build in perhaps exotic ways.. ?

Fine for me, they usually install only stable releases.

dazinator:
... Changing it from 3.1 to "allowPreRelease: false" has broken: https://github.com/GitTools/GitVersion/pull/2155/checks?check_run_id=488177987

Not sure why. I'm going to remove global.json.

arturcic:
@dazinator can you rebase on top of master?

dazinator:
> @dazinator can you rebase on top of master?

done

src/GitVersionTask/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+2. When running on the full .NET framework this isn't a problem because in that case we can access the current AppDomain and add AssemblyResolve handlers to load these dependencies as needed.
+
+3. On .NET Core then, we create a new AssemblyLoadContext, and load try to execute our task within it but this is tricky because:
+   - If you try to load the same task Type and use it in another AssemblyloadContext (for example by creating an instance of it and calling Execute) you will
+   get weird errors along the lines of "cannot cast type A to type B"even though the two types are the same.
+
+   To workaround this issue then, the Task running within MsBuild's AssemblyLoadContext must load a Type that is not being used in it's AssemblyLoadContext. It can then interact with that type by invoking methods via reflection.
+   It's the responsibility of this "reverse proxy" to execute the required task logic within our custom AssemblyLoadContext, where any dependencies that are needed will be loaded on demand from the Nuget package folder location.
```
arturcic: I like the way you explained here!

arturcic:
@dazinator could you please cleanup a bit the commit history by squashing some of the commits and leave only the important steps?

dazinator:
> @dazinator could you please cleanup a bit the commit history by squashing some of the commits and leave only the important steps?

If I reduce the number of commits, will this result in a build producing the same SemVer version (as commit count would have deprecated) - and a subsequent failure of a build check that tries to publish a nuget package who's version was already published?

arturcic:
> > @dazinator could you please cleanup a bit the commit history by squashing some of the commits and leave only the important steps?
> 
> If I reduce the number of commits, will this result in a build producing the same SemVer version (as commit count would have deprecated) - and a subsequent failure of a build check that tries to publish a nuget package who's version was already published?

nope we publish from master only

dazinator:
> @dazinator could you please cleanup a bit the commit history by squashing some of the commits and leave only the important steps?

Done

arturcic:
@dazinator is this PR fixing the issue you had when using a preview sdk?

dazinator:
> @dazinator is this PR fixing the issue you had when using a preview sdk?

No. I had hoped that removing the additional msbuild dependency may sort it after some quick googling. It seems it didn't though - but I still think it's probably a worthwhile change.

arturcic:
> > @dazinator is this PR fixing the issue you had when using a preview sdk?
> 
> No. I had hoped that removing the additional msbuild dependency may sort it after some quick googling. It seems it didn't though - but I still think it's probably a worthwhile change.

Agree. Leaving the previews support for another release

arturcic:
thank you so much for your contributions 👍 @dazinator

src/GitVersionTask/GitVersionTask.csproj
```diff
@@ -11,10 +11,12 @@
         <DevelopmentDependency>true</DevelopmentDependency>
 
         <LangVersion>8.0</LangVersion>
-    </PropertyGroup>
+    </PropertyGroup>
```
Jetersen: I do not like these random whitespace inconsistency left around in the project. 😭

dazinator: I wonder if this is something dotnet-format and editorconfig rules could catch in future.

Jetersen: editorconfig should already trim whitespace: https://github.com/GitTools/GitVersion/blob/master/.editorconfig#L9

dotnet format possible could.

dazinator: dotnet-format docs mentions trailing whitespace isnt supported but states the tool removes trailing whitespace by default: https://github.com/dotnet/format/blob/master/docs/Supported-.editorconfig-options.md

dazinator: I'm not sure how the whitespace came about but I do have a habbit of ctrl + k + d to auto format documents in VS, but I would hope that should also honour editorconfig rules, so perhaps this was just a flourish of the fingers.

PR #2156, opened 2020-03-05
Integration tests for msbuild task artifact

arturcic:
Added integration test for msbuild task


dazinator:
I like your inventiveness here - building a console app with gitversiontask, which reports on it's version when executed - seems like a nice way to do this check.

This LGTM.

arturcic:
> I like your inventiveness here - building a console app with gitversiontask, which reports on it's version when executed - seems like a nice way to do this check.
> 
> This LGTM.

I'll merge this one so that we can then check your PR if still works

PR #2159, opened 2020-03-06
Use GitVersion.Commandline package as source for the chocolatey package

arturcic:


gep13:
@arturcic is there a reason that we are no longer embedding the exe into the Chocolatey Package?

arturcic:
> @arturcic is there a reason that we are no longer embedding the exe into the Chocolatey Package?

The reason is that I want to reduce the variants of packages we produce to global tool and the single file self content. All the others will just use one of these 2. Another reason is that I want to be able to specify the version I want to install, basically the chocolatey package will be a simple download and install tool.

gep13:
I see where you are coming from, but you are going against the principles of Chocolatey.  A package is meant to be a single immutable thing, it is not intended that the script within a Chocolatey package will take  additional actions in terms of deciding what version should be installed.  For example, what would this mean:

```
choco install gitversion.portable --version 1.1.1 --params="/version=1.1.2"
```

This is a little bit confusing, and will likely result in lots of questions about what "should" be installed.  It also potentially will cause problems during the upgrade of the GitVersion package.

arturcic:
> I see where you are coming from, but you are going against the principles of Chocolatey. A package is meant to be a single immutable thing, it is not intended that the script within a Chocolatey package will take additional actions in terms of deciding what version should be installed. For example, what would this mean:
> 
> ```
> choco install gitversion.portable --version 1.1.1 --params="/version=1.1.2"
> ```
> 
> This is a little bit confusing, and will likely result in lots of questions about what "should" be installed. It also potentially will cause problems during the upgrade of the GitVersion package.

Good point, I knew you're the right person for the review, what if we remove the param and just use the package version as the version to install?

arturcic:
Originally the idea was to just use the tar.gz from the GitHub release as source, but that means only stable releases can be referenced, and probably that is a good thing, but I wanted to also be able to install a preview version, but where the tar.gz to be stored? That's why the idea evolved into this PR

arturcic:
Maybe we should publish only stable versions..

gep13:
>@arturcic said...
what if we remove the param and just use the package version as the version to install?

The `--version` is available within the chocolateyInstall.ps1 file.  It is what is populated into this environment variable:

$env:ChocolateyPackageVersion

However, this approach would only work in new packages going forward, that had the modification to the chocolateyInstall.ps1 file.  You wouldn't be able to install any older versions using this approach.

In terms of what Chocolatey, specifically, would prefer, it would be to continue to embed the exe into the package, although I understand the reluctance to do this.  Embedding the assembly(s) into the Chocolatey package means that it is a completely self contained package, that doesn't need to reach out to anywhere else in order to get the assemblies.

Doing what you are doing in the package currently could be achieved by doing:

```
choco install GitVersion.CommandLine --source https://nuget.org/api/v2
```

However, that isn't as "nice" as it means configuring an extra command, or source, in your Chocolatey configuration.

arturcic:
> > @arturcic said...
> > what if we remove the param and just use the package version as the version to install?
> 
> The `--version` is available within the chocolateyInstall.ps1 file. It is what is populated into this environment variable:
> 
> $env:ChocolateyPackageVersion
> 
> However, this approach would only work in new packages going forward, that had the modification to the chocolateyInstall.ps1 file. You wouldn't be able to install any older versions using this approach.
> 
> In terms of what Chocolatey, specifically, would prefer, it would be to continue to embed the exe into the package, although I understand the reluctance to do this. Embedding the assembly(s) into the Chocolatey package means that it is a completely self contained package, that doesn't need to reach out to anywhere else in order to get the assemblies.
> 
> Doing what you are doing in the package currently could be achieved by doing:
> 
> ```
> choco install GitVersion.CommandLine --source https://nuget.org/api/v2
> ```
> 
> However, that isn't as "nice" as it means configuring an extra command, or source, in your Chocolatey configuration.

Got it

gep13:
>@arturcic said...
Maybe we should publish only stable versions..

Personally, I think this is a good idea.  I see NuGet.org and Chocolatey.org as the production locations.  I would much prefer to see usage of something like MyGet, Feedz, Azure Artifacts, as the destination for the preview packages.

If we did want to push a beta version to NuGet/Chocolatey to widen the audience as we are getting closer to a release, that can be done as well.

arturcic:
> > @arturcic said...
> > Maybe we should publish only stable versions..
> 
> Personally, I think this is a good idea. I see NuGet.org and Chocolatey.org as the production locations. I would much prefer to see usage of something like MyGet, Feedz, Azure Artifacts, as the destination for the preview packages.
> 
> If we did want to push a beta version to NuGet/Chocolatey to widen the audience as we are getting closer to a release, that can be done as well.

As we're already publishing to GitHub packages already, I guess we can make the GitHub as the previews channel and keep chocolatey/nuget as the stable channel

arturcic:
> In terms of what Chocolatey, specifically, would prefer, it would be to continue to embed the exe

Ok agree, then closing this PR and keep embedding, thanks for the feedback, there is one more change I want to address and we can release 5.2.0

asbjornu:
Using GitHub for prerelease packages sounds like a good strategy to me. The reason we’ve published prereleases to NuGet so far is because we haven’t had any good alternatives. With GitHub Packages, now we do.

PR #1848, opened 2019-10-09
Fix HEAD for worktrees

chuseman:
When running inside a worktree, the HEAD considered is from the default worktree's branch/commit instead of the worktree that GitVersion is executed for. This behavior was introduced in #1793 when `GitPreparer.GetDotGitDirectory()` was changed to be more worktree friendly.

This commit switches to use `GetProjectRootDirectory()` instead of `GetDotGitDirectory()` so that we can get the correct branch. I can't think of a scenario where this isn't what you'd expect.

Also, when troubleshooting I found a problem when specifying `/b` for a branch. If you have a local branch and a remote branch of the same name, you'll get an exception because `SingleOrDefault()` is used to find the branch you meant. I changed it over to use `FirstOrDefault()` while preferring local branches first.

PR #2166, opened 2020-03-10
Issue #2165 WorkingTree uses incorrect commit.

### LINKED ISSUE #2165: Worktrees incorrectly calculated
When running gitversion within a worktree, the version that is calculated is against the HEAD of the main repository rather than the HEAD of the worktree itself.  This is caused by the fact that the Repository object that is created is using the return from the main git directory rather than the worktree path.

I don't know how to fix this, because if I change it to either GetWorkingDirectory or GetProjectDirectory (either will work), the tests:
GitVersionCore.Tests.DynamicRepositoryTests.FindsVersionInDynamicRepo
DynamicRepositoriesShouldNotErrorWithFailedToFindGitDirectory
fail.

Obviously, the GetDotGitDirectory needs to return the actual directory so that caching works, however when in a worktree, the repository object that is created needs to be based on the worktree.

See also #1063

tpaxatb:
The repository opened for use to do versioning should be the Project Directory, not the DotGit directory.
Also fixes it so the project directory used is based on the DynamicGitDirectory if it is a dynamic repo.

Fixes #2165. Closes #1848.

### DIFF
--- a/src/GitVersionCore/Extensions/ArgumentExtensions.cs
+++ b/src/GitVersionCore/Extensions/ArgumentExtensions.cs
@@ -26,12 +26,7 @@ namespace GitVersion.Extensions
 
         public static string GetProjectRootDirectory(this Arguments arguments)
         {
-            if (arguments.IsDynamicGitRepository())
-            {
-                return arguments.GetWorkingDirectory();
-            }
-
-            var dotGitDirectory = Repository.Discover(arguments.GetWorkingDirectory());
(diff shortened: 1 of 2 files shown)


asbjornu:
Since this seems to tackle the same problems as #1848, would it be possible for you to incorporate the tests added there in this PR? I'm not confident that we have sufficient coverage for this, for the reasons outlined in https://github.com/GitTools/GitVersion/issues/2165#issuecomment-597548377.

arturcic:
Closing this one in favor of #2281

PR #2168, opened 2020-03-10
Issue #2167 Generated files not soon enough

### LINKED ISSUE #2167: UpdateAssemblyInfo and GenerateGitVersionInformation need to be done earlier
Similar to #1465, the UpdateAssemblyInfo and GenerateGitVersion targets, which write out the appropriate code classes and attributes, need to be done prior to MarkupCompilePass1.  In addition, in order for them to be compiled into the pass, the files need to be added to the _GeneratedCodeFiles item list so they are added to the compile line.

tpaxatb:
To allow for the generated files to be available during the first pass,
they must be generated before the pass1.  In addition, the files that are
generated must be added to the `_GeneratedCodeFiles` items.

Fixes #2167.


arturcic:
Currently the tests we have don't cover the order of the tasks, and to be honest I don't know if we can automate that.

asbjornu:
I'm more concerned with the task still working at all than the order being as expected. As long as we have tests that dogfood GitVersionTask and check the generated version variables, I think we're good.

arturcic:
We do have tests that run the msbuild artifact and checks the generated version

asbjornu:
@arturcic, great! Thanks for your contributions, @tpaxatb! 🙏

PR #2178, opened 2020-03-13
Add PrivateAssets="All" to package reference

### LINKED ISSUE #2177: GitVersionTask should not have a package dependency on Microsoft.NETFramework.ReferenceAssemblies
As of the 5.2.0 release, the GitVersionTask package has a dependency on Microsoft.NETFramework.ReferenceAssemblies. This is incorrect because this package is for enabling  compiling .NET Framework assemblies on Linux. It has no runtime impact.

The reason for the unneeded dependency is because the package reference is not marked with `PrivateAssets="All"` in your [Directory.Build.props](https://github.com/GitTools/GitVersion/blob/master/src/Directory.Build.props#L42) file.

## Expected Behavior
The GitVersionTask package should have no dependencies.

## Actual Behavior
The GitVersionTask package does have a dependency:
image

## Possible Fix
Add `PrivateAssets="All"` to package reference in [Directory.Build.props](https://github.com/GitTools/GitVersion/blob/master/src/Directory.Build.props#L42) file.

bording:
This adds `PrivateAssets="All"` to the Microsoft.NETFramework.ReferenceAssemblies package reference. This removes this package dependency from the GitVersionTask package.

Fixes #2177

### DIFF
--- a/src/Directory.Build.props
+++ b/src/Directory.Build.props
@@ -39,6 +39,6 @@
     </ItemGroup>
 
     <ItemGroup Condition=" '$(TargetFramework)' == 'net472' ">
-        <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" />
+        <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.0" PrivateAssets="All" />
     </ItemGroup>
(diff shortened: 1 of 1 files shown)


arturcic:
thank you so much for your contributions 👍 @bording

PR #2179, opened 2020-03-13
feat: accept a fallback value in all formatted expressions

piedar:
## Description

This PR extends the custom formatting syntax to support a fallback value in a variable expression, to be consistent with the fallback syntax for environment variables.  It also adds the ability to specify a more complex fallback value in a quoted string.

For example

```yaml
assembly-file-versioning-format: '{Major}.{Minor}.{Patch}.{WeightedPreReleaseNumber ?? 0}'
assembly-informational-format: '{AssemblySemFileVer}+{env:BUILD_TAG ?? "fancy-string*with%special#characters"}'
```

## Motivation and Context

I tried to use

```yaml
assembly-file-versioning-format: '{Major}.{Minor}.{Patch}.{WeightedPreReleaseNumber}'
```

but that fails for versions where there is no pre-release label because `WeightedPreReleaseNumber` is null.  So then I tried

```yaml
assembly-file-versioning-format: '{Major}.{Minor}.{Patch}.{WeightedPreReleaseNumber ?? 0}'
```
but discovered that this syntax was only supported for environment variables.

## How Has This Been Tested?

- added unit tests to better cover existing cases
- added unit tests to cover new cases
- examined references to discover that the changed code is used only for custom version formatting

## Checklist:


asbjornu:
Thank you so much for improving GitVersion, @piedar! 🙏

Issue #2181, opened 2020-03-15
[Bug] #2167 breaks building WPF applications


batzen:
**Describe the bug**
The fixes made for #2167 seem to break building WPF applications because version attributes are added as duplicates.

## Expected Behavior
Build should work and assembly info should not have duplicate attributes.

## Actual Behavior
Assemblyinfo has duplicate version attributes.

## Possible Fix
`MarkupCompilePass1DependsOn` should not be modified to contain `UpdateAssemblyInfo`.

## Steps to Reproduce
- Clone https://github.com/snoopwpf/snoopwpf
- Compile

## Context
Tried to compile with the latest gitversiontask version.
Version 5.2.2 works fine. Only 5.2.3 shows this issue.

## Your Environment
.NET Core SDK (reflecting any global.json):
 Version:   3.1.101
 Commit:    b377529961

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.18363
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\3.1.101\

Host (useful for support):
  Version: 3.1.2
  Commit:  916b5cba26

- Version Used: 5.2.3
- Operating System and version (Windows 10): 1909
- Link to your project: https://github.com/snoopwpf/snoopwpf
- Link to your CI build (if appropriate):

arturcic:
@tpaxatb can you check this bug?

tpaxatb:
Yeah I realized after the fact that netcore apps seem to build slightly different (i didn't have netcore 3 sdk until friday). I think maybe on the .net core that the UpdateAssemblyInfo should be turned off in the props file or something.  I'll see about taking a look later if/when I get a chance, as our office is in the process of transforming to mandatory work at home due to government directives yesterday and today.

arturcic:
Sure, take your time, any help is welcomed

bording:
Looks like the PR I just opened should fix this: #2183

arturcic:
@bording, ok then, after we merge your PR we will validate it here

arturcic:
closing, please reopen if still reproduces

PR #2183, opened 2020-03-16
Don't enable UpdateAssemblyInfo for SDK-style projects

### LINKED ISSUE #2182: UpdateAssemblyInfo target should not be run for SDK-style projects
The `UpdateAssemblyInfo` target is currently always being run, even for SDK-style projects that don't need it. The SDK-style project system handles that automatically as long as the appropriate properties have been set. This is already happening in the [`GetVersion` target](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionTask/build/GitVersionTask.targets#L104). 

This behavior used to work correctly, and you can even see a comment in the [GitVersionTask.props](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionTask/build/GitVersionTask.props#L20) file indicating that this was the behavior.

This appears to have been broken in the 5.0.0 release, but fixing #2167 in 5.2.3 brought it to my attention because that change broke one of my projects.

## Expected Behavior
The `UpdateAssemblyInfo` target should not be run when using an SDK-style project.

## Actual Behavior
The `UpdateAssemblyInfo` target should not be run when using an SDK-style project.

## Possible Fix
Add a condition to the `UpdateAssemblyInfo` property in GitVersionTask.props that disables the `UpdateAssemblyInfo` when the `UsingMicrosoftNETSdk` property is `true`.

## Context
I have an SDK-style project that is using the MSBuild.Sdk.Extras SDK to build a WPF app, and using 5.2.3 is causing a build error because of duplicate version attributes from GitVersion trying to create them instead of letting the project system do it.

bording:
Fixes #2182 by adding a condition disable the target when `UsingMicrosoftNETSdk` is `true`.

### DIFF
--- a/src/GitVersionTask/build/GitVersionTask.props
+++ b/src/GitVersionTask/build/GitVersionTask.props
@@ -19,12 +19,8 @@
 
         <!-- Property that enables UpdateAssemblyInfo. Default to off for SDK builds -->
         <UpdateAssemblyInfo Condition=" '$(DisableGitVersionTask)' == 'true' ">false</UpdateAssemblyInfo>
+        <UpdateAssemblyInfo Condition=" '$(UsingMicrosoftNETSdk)' == 'true' ">false</UpdateAssemblyInfo>
         <UpdateAssemblyInfo Condition=" '$(UpdateAssemblyInfo)' == '' ">true</UpdateAssemblyInfo>
-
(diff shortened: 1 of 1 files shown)


bording:
I believe this also fixes #2181.

batzen:
I haven't tested this but won't this disable the updating completely when using the sdk and having GenerateAssemblyInfo=false? I guess people might want to have the update from gitversiontask. How about some unittests that verify what happens?

bording:
@batzen I would say that if you are using the SDK project system, you should be relying on the project system to provide those values. This is how GitVersionTask worked back in 4.x, and it's a bug that it stopped working that way.

PR #2195, opened 2020-03-19
Generate assembly info for System.Web precompiled projects

### LINKED ISSUE #2192: [Feature] Support injecting assembly information in precompiled website binary
This is for `GitVersion.Task`.

**Is your feature request related to a problem? Please describe.**
For System.Web projects, during publish, it is possible to precompile the files. Usually the build process generates a temporary assembly info file, however, it does not when GitVersion is used.

## Detailed Description
Generate the proper temporary assembly info file so it will be incorporated in the generated precompiled DLL.

## Context
Currently *no* version information is stored at all in the precompiled DLL, so you can't tell from which commit, or codebase it has been compiled.

## Possible Implementation
Hook onto the correct MSBuild target, but I will need to investigate.

---

I'm happy to provide a pull request, if you give me a few pointers how you would like to integrate it in your codebase.

Sebazzz:
This only works for precompiled and merged publications. Unmerged published DLLs don't allow any custom handling of assembly information.

Fixes #2192.

## Description
Added targets that hook into the correct stage of the System.Web precompilation publishing pipeline. The first target is needed because a `XxxDependsOn` property is used and we must inject our task execution at the right stage. Then, the other target injects the attributes which are used in the merge assembly compilation.

## Related Issue
[#2192 - Support injecting assembly information in precompiled website binary](https://github.com/GitTools/GitVersion/issues/2192)

## Motivation and Context
Currently no version information is stored at all in the precompiled DLL, so you can't tell from which commit, or codebase it has been compiled.

## How Has This Been Tested?

1. Create a ASP.NET Web Forms or MVC project with content.
2. Create a publish profile and enable "precompilation", disable "updateable" and enable merging. Eventually you end up with a publish profile that looks like this:
```
<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
  <PropertyGroup>
    <WebPublishMethod>FileSystem</WebPublishMethod>
    <PublishProvider>FileSystem</PublishProvider>
    <LastUsedBuildConfiguration>Release</LastUsedBuildConfiguration>
    <LastUsedPlatform>Any CPU</LastUsedPlatform>
    <SiteUrlToLaunchAfterPublish />
    <LaunchSiteAfterPublish>True</LaunchSiteAfterPublish>
    <PrecompileBeforePublish>True</PrecompileBeforePublish>
    <EnableUpdateable>False</EnableUpdateable>
    <DebugSymbols>True</DebugSymbols>
    <WDPMergeOption>MergeAllOutputsToASingleAssembly</WDPMergeOption>
    <UseMerge>True</UseMerge>
    <SingleAssemblyName>_MergedWebAssembly</SingleAssemblyName>
    <publishUrl>bin\Release\Publish</publishUrl>
  </PropertyGroup>
</Project>
```
3. Publish the project, either through Visual Studio, or by invoking MSBuild with these parameters:

    /t:Rebuild /p:DeployOnBuild=true /p:PublishProfile=[filename of publish profile without extension]

4. Find the `_MergedWebAssembly.dll` file in the folder of the `publishUrl` parameter in the publish profile. Note it has no assembly information (version is set to 0.0.0.0).

5. Include the updated GitVersion targets in the project or install the compiled NuGet package. 

6. Re-execute step 3. Like in step 4, find the file again and notice the assembly information has now been injected.

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation. --> I added a flag to allow disabling the code generation, but I don't see many of the current flags documented either. Do we want to document this flag?
- [ ] I have added tests to cover my changes. --> I'm not sure if your existing tests allow full integration testing of the MSBuild publish pipeline. 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,4 @@ config.wyam.hash
 config.wyam.packages.xml
 /test/core/build
 /test/full/build
+/test/systemweb/build
--- a/build/artifacts-test.cake
+++ b/build/artifacts-test.cake
@@ -48,9 +48,9 @@ Task("Artifacts-MsBuildCore-Test")
     }
 });
 
-Task("Artifacts-MsBuildFull-Test")
-    .WithCriteria<BuildParameters>((context, parameters) => parameters.IsRunningOnWindows, "Artifacts-MsBuildFull-Test can be tested only on Windows agents.")
-    .WithCriteria<BuildParameters>((context, parameters) => parameters.IsReleasingCI,      "Artifacts-MsBuildFull-Test works only on Releasing CI.")
+Task("Artifacts-MsBuildFull-Library-Test")
+    .WithCriteria<BuildParameters>((context, parameters) => parameters.IsRunningOnWindows, "Artifacts-MsBuildFull-Library-Test can be tested only on Windows agents.")
+    .WithCriteria<BuildParameters>((context, parameters) => parameters.IsReleasingCI,      "Artifacts-MsBuildFull-Library-Test works only on Releasing CI.")
     .IsDependentOnWhen("Pack-Nuget", singleStageRun)
     .Does<BuildParameters>((parameters) =>
 {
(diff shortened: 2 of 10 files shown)


asbjornu:
This looks good, but as we've already learned that we have too poor test coverage for this, would it be possible to add at least one test that ensures this continues working in the future?

Sebazzz:
> This looks good, but as we've already learned that we have too poor test coverage for this, would it be possible to add at least one test that ensures this continues working in the future?

Possible yes, easy: not. Essentially you'd need to check in an empty web project, run MSBuild against it, and then verify that version information has been written. This is a test that can only run on Windows obviously. 

I'd be happy to look into that, but I need some pointers on where you think it is best to build such integration test.

asbjornu:
I believe we already have project files that are used for this sort of testing, perhaps @arturcic can let you in on the details. Here's one:

https://github.com/GitTools/GitVersion/blob/3f5392e318244bb66b0b9e0f19049f2353c70944/test/full/app.csproj#L1-L42

Sebazzz:
I've based on 47f2d4c an integration test. 

Possible complication:
- Requires ASP.NET toolset to be installed on CI

Questions:
- I don't understand why the artifacts tests are only configured to be run on release CI, with no apparent way to override it. For testing purposes locally I've temporary commented these conditions.

Things I'm not happy with:
- The current tests apparently execute the built assembly. For the built dlls this is obviously not possible. Currently I'm loading the built assembly and inspecting the right attribute. This works, but I don't like it. Possible alternative include building an additional side executable (built on netfx) which is executed and inspects the built assembly.

arturcic:
> * I don't understand why the artifacts tests are only configured to be run on release CI, with no apparent way to override it. For testing purposes locally I've temporary commented these conditions.

The reason is because we test the artifacts on different platforms than the platform we built on. We use docker containers for testing them and we need to load ~10 containers - that is taking a lot of space and time, and for a regular developer that creates a PR he does not need to set everything up, including docker.

> * The current tests apparently execute the built assembly. For the built dlls this is obviously not possible. Currently I'm loading the built assembly and inspecting the right attribute. This works, but I don't like it. Possible alternative include building an additional side executable (built on netfx) which is executed and inspects the built assembly.

That is true. There are two things I want to add.

First. Next major version (6.0) will be the last that targets Full Framework. That means we won't invest too much on this kind of integration tests

Second. Beside Asp.Net projects there are other project types that may be covered with this kind of integration tests. But the reason We have these integration tests only is that we check the assembly info is set. To be honest I'm against adding too many targets in the box.

I'd prefer to add some extension points so that anyone needs it can extend (run before or after the GitVersion targets) and perform their custom tasks.

I'm planning to remove the custom logic in the GitVersion targets that are not related to GitVersion in the next major version and leave it for others to extend (either custom nuget packages that depend on GitVersionTask or by having custom targets)

To be honest I really appreciate the time and the effort you invested in this PR. But I would move this outside the GitVersionTask when the new major version is available

arturcic:
@asbjornu  I think we should postpone this one for later times

Sebazzz:
> Next major version (6.0) will be the last that targets Full Framework. That means we won't invest too much on this kind of integration tests

I think, for a library project, this is really a bad path to go on. We're talking about - in this case - build tooling. There are many libraries still building netfx assemblies too.

> To be honest I really appreciate the time and the effort you invested in this PR. But I would move this outside the GitVersionTask when the new major version is available

I'd appreciated it to know all that *before* I invested time into this PR...

arturcic:
> I'd appreciated it to know all that _before_ I invested time into this PR...

Sorry about that, I did not have time today to follow the all messages and to respond promptly. Unfortunately I don't have much time on open source these days

arturcic:
> I think, for a library project, this is really a bad path to go on.

Agree, but we try to use our human resources as efficient as possible and focus on the current trend

PR #2197, opened 2020-03-21
Braces should be removed

teyc:
1. Make it clear that the braces around `{vcsid}` should not appear in the final string.

2. Consistently refer to `vcsid` instead of `vcsid` and `configurationId`

## Checklist:


asbjornu:
Thank you for your contributions to GitVersion, @teyc! 🙏

Issue #1994, opened 2019-12-16
WeightedPreReleaseNumber null on tag


tobiasdiez:
We use 
`assembly-versioning-format: '{major}.{minor}.{WeightedPreReleaseNumber}'`
in combination with
`mode: ContinuousDeployment`
to have an increasing build number. Everything works in a normal development workflow (PRs, direct commits etc increase the version from e.g. `1.1.1` to `1.1.2`). However, as soon as we want to release a new version, we push a tag with the new version (say `2.0-beta`). In this case, `WeightedPreReleaseNumber` is `null` (or empty) for some reason which leads to the invalid build number `2.0.` (notice the missing last bit of the version). I would have expected this to be `2.0.0` (or more generally: `2.0.x` where `x` is the weight corresponding to the beta tag).

Full GitVersion file:
https://github.com/JabRef/jabref/blob/master/GitVersion.yml/GitVersion.yml

tobiasdiez:
The issue is still present.

asbjornu:
Are you able to reproduce this problem in a `RepositoryFixture` and submit a PR with it?

jbaehr:
We're hit by the same issue. In order to have strictly increasing Assemly*File*Version (the unmanaged one) to satisfy Windows Installer's need we used to apply the weight on our own and inject it via an environment variable (cf. #1385). In order to get rid of this external custom logic we want to switch to the "WeightedPreReleaseNumber" (cf. #1433). It works great up to the point where you build from a tag.

Right now one can only configure a `pre-release-weight` per branch; I think we need a similar way for tags (i.e. a regex for matching tags, then configure how to version them. Just like the branch config works)

@ruhullahshah, any thoughts on that?

ruhullahshah:
@jbaehr From the problem description, it seems like the [TaggedCommitVersionStrategy](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/VersionCalculation/BaseVersionCalculators/TaggedCommitVersionStrategy.cs) wins during the version calculation and since there are no weights attached to tags as you described already, we get the observed behavior. What you suggest sounds like a plausible solution to me. 

I would also like to bring to your notice a [related](https://github.com/GitTools/GitVersion/pull/1600), although not exactly the same issue that might affect a potential PR for the issue we are discussing here.

jbaehr:
That the `TaggedCommitVersionStrategy` wins is expected and totally fine. We just need a way to set the weight for this case.

[...]

PR #2222, opened 2020-04-05
Mechanism to configure a pre-release-weight for the tagged commits

ruhullahshah:
## Background
This PR is motivated by issue #1994. As a user, I should be able to:
- Specify by means of a regex in the configuration as to how GitVersion should construct semantic version numbers from the Git tags.
- Specify weights for pre-release labels in the Git tags.

## Proposal
In order to achieve the stated objectives, we can extend the global configuration as:
```yaml
tag-config: 
    file-versioning-format: "v{Major}.{Minor}.{Patch}.{TagWeightedPreReleaseNumber}"
    pre-release-label-weights:
       alpha: 10000
       beta: 20000
```
A new GitVersion variable `TagWeightedPreReleaseNumber` will be introduced which can have similar semantics as the existing `WeightedPreReleaseNumber`, i.e. it can be computed as `pre-release-label-weight + CommitsSinceVersionSource`.

The `file-versioning-format` will have similar semantics as `assembly-file-versioning-format` and can specify environment variables as well.


ruhullahshah:
@jbaehr @asbjornu 

I have created an initial proposal for addressing issue #1994 and added a failing unit test for the type `TaggedCommitVersionStrategy`.  I have not set up the config for the unit test yet. Once the config is finalised, I will proceed with the feature construction and add appropriate config options to the unit test and to some integration tests. Please let me know your thoughts about the proposal.

jbaehr:
While don't fully understand the use cases of #1994 and #1600 (and, to be honest, I haven't dug deeply into the later), I can't really comment on the `tag-config` proposal. Apart from finding an additional, nested `file-versioning-format` a bit odd; What does it offer opposed to the global `assembly-file-versioning-format` with the standard `WeightedPreReleaseNumber`?

Anyway, I propose to go for a simpler approach first, just enough for the vanilla gitflow use case, where you only tag final versions. Simply defining a static default pre-release weight for tags should be enough for preventing the `WeightedPreReleaseNumber` to be *null* on builds from tagged commits. This default should be as high as possible, as there is no semver "larger" than the final build. And "commits since version source" should be always zero, per definition, else your current commit is not the tagged commit, ergo a branch config should apply. For Windows we have 16 bit for this, thus $2^16 - 1$ would give a good value. Or 60000, for those preferring human-readable numbers (programmers are not "normal" humans).

So either we can go for a new top-level config key, a simple scalar. Or, in the same spirit as the branch config something like 
```yaml
tags:
  default:
    pre-release-weight: 60000
```

Later, this can still be extend with different named tag configs, with regex matching and extracted pre-release labels, what ever...

asbjornu:
If your proposed solution is sufficient, it's definitely the simplest one to implement, maintain and document, @jbaehr. Thoughts, @ruhullahshah?

ruhullahshah:
@jbaehr, @asbjornu 
- My proposal was aimed to be a more general solution, with GitFlow being a special case, wherein under `pre-release-label-weights` one could specify `default: 60000` for the final release build. 
- `file-versioning-format: "v{Major}.{Minor}.{Patch}.{TagWeightedPreReleaseNumber}"` was aimed at instructing the `TaggedCommitVersionStrategy` as to how to construct the version numbers out of tags, for instance `1.0.0-alpha` would get mapped to `1.0.0.10000`
- However, I see @jbaehr 's point and it indeed simplifies the implementation if we are targeting GitFlow only for now. Also, it serves as a good starting point, as already mentioned by @jbaehr, for generalizing the implementation.
- I will take up the proposal made by @jbaehr and implement it.

asbjornu:
Sounds good, @ruhullahshah! Looking forward to seeing the implementation.

ruhullahshah:
@asbjornu I have added the changes that I believe are minimal and necessary to address the issues described by users in #1994.

asbjornu:
Sounds and looks great, @ruhullahshah! Could you give this a quick review too, @jbaehr? A rebase on the latest `HEAD` of `master` would probably also be a good idea. :)

jbaehr:
@ruhullahshah thanks for your work!

I only question the "0" default for the `tag-pre-release-weight`. In my eyes (as motivated in [my first comment](https://github.com/GitTools/GitVersion/pull/2222#issuecomment-624811425)) the value should be *high*, to make sense. As the `WeightedPreReleaseNumber` was previously undefined/empty/null, we don't need zero for backward compatibility, either. Instead, the default here should fit with the defaults from the branch configurations, see #1433, i.e. 60_000.

ruhullahshah:
@jbaehr  You are welcome.

I agree, a default weight of 60000 makes more sense. That way a user using the GitFlow branching model might not even need to specify the value in the configuration. I have updated the PR for the same. 

@asbjornu Could we ship this now, do not want to rebase anymore :P ?

asbjornu:
Thank you so much for your work on this @ruhullahshah, and for your valuable input @jbaehr! 🙏

ruhullahshah:
@asbjornu You are welcome!

Issue #2225, opened 2020-04-06
[Bug] Azure DevOps UseGitVersion@5 task and GitVersion 5.2.4 produces an error "AssemblyInfoFilename file not found at null" when `UpdateAssemblyInfo: true` is set


dsbut:
**Describe the bug**
I have recently been updating pipelines to use the `UseGitVersion@5` task, replacing the obsolete `GitVersion@5` task. For the most part, this has been easy, and I did not encounter any problems until I switched out the task on a pipeline which use GitVersion to update the `AssemblyInfo.cs` files for a .Net Framework solution. When run, the build fails on the `UseGitVersion@5` task, with the error message:
> ##[error]Error: AssemblyInfoFilename file not found at null

The task is defined in my pipeline as per the example in the GitVersion docs:

``` yaml
- task: UseGitVersion@5
  inputs:
    versionSpec: '5.x'
    updateAssemblyInfo: true
```

## Expected Behavior
The `UseGitVersion@5` task should update all `AssemblyInfo.cs` files under the repository

## Actual Behavior
Task execution fails with the error `AssemblyInfoFilename file not found at null`.

## Possible Fix
Unfortunately, I have nothing concrete to offer beyond wild speculation.

## Steps to Reproduce
This may be reproduced using the simple pipeline below, in a repo containing a .Net Framework solution with projects that include `AssemblyInfo.cs` files.
`Gitversion.yml` file also attached.

**See the attached log from pipeline task execution 
[UserGitVersionPipelineTaskLog.log](https://github.com/GitTools/GitVersion/files/4441157/UserGitVersionPipelineTaskLog.log)
with debugging enabled**

## Context
No longer able to automate the versioning of assemblies with `UseGitVersion@5` as ws possible with `GitVersion@5` task.

## Your Environment
Azure DevOps Pipelines running on MS hosted Windows Agent

- Version Used:
* `windows-latest` MS hosted agent
* `UseGitVersion@5` - v5.1.2
* GitVersion 5.2.4

### Sample Pipeline
``` yaml
pool:
  vmImage: 'windows-latest'

steps:
- task: UseGitVersion@5
  inputs:
    versionSpec: '5.x'
    updateAssemblyInfo: true
```

### Sample `GitVersion.yml` config used...nothing to see here really
``` yaml
branches: {}
ignore:
  sha: []
merge-message-formats: {}
```

suddenelfilio:
We are currently experiencing the same issue. No solution found yet.

[...]

PR #2240, opened 2020-04-23
GH-2225 use FileSystemGlobbing to resolve AsemblyInfo to be updated

arturcic:


arturcic:
> I really like the use of `AssemblyInfoContext` over several parameters. Would be great to roll out that design throughout the codebase.

At least for VersionConverters I have created `Context` classes

arturcic:
> Just one question: Should we perhaps add a test for the globbing, to see that it works (and doesn't regress in the future?

I will

arturcic:
> Great work, @arturcic! 👍 

Thanks

PR #2251, opened 2020-05-03
Feature/output file

### LINKED ISSUE #983: Add additional output type for saving asserted version numbers to a file

arturcic:
Resolves #983

### DIFF
--- a/src/GitVersionCore/Core/GitVersionTool.cs
+++ b/src/GitVersionCore/Core/GitVersionTool.cs
@@ -85,7 +85,7 @@ namespace GitVersion
 
             using (outputGenerator)
             {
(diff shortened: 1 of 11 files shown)


arturcic:
@asbjornu @gep13 this is ready for review

arturcic:
> This looks great! The only thing I can't seem to find is an integration test for verifying that the file is actually written to disk.

Agree, I will add a couple of them, thanks

arturcic:
@asbjornu integration tests added

PR #2259, opened 2020-05-06
Issue #2257 Ensure Valid Wix Version File

### LINKED ISSUE #2257: [Bug] /updatewixversionfile does not cleanly recreate the file
The content of the file `GitVersion_WixVersion.wxi` is not fully overwritten/recreated when the file already exists and `GitVersion /updatewixversionfile` is invoked. Instead it just overwrites the first x bytes.

## Expected Behavior
The entire content of `GitVersion_WixVersion.wxi` is overwritten and the file has a valid content afterwards.

## Actual Behavior
If the existing file is larger than the newly generated file, parts of the previously existing content remain in the file.

## Possible Fix
Delete the file prior writing/correctly write the content.

## Steps to Reproduce
1. Create a large dummy file called `GitVersion_WixVersion.wxi`in the root of a repo.
2. Run `GitVersion /updatewixversionfile`
3. After that, the file looks like this (shortend the wix related content):
```xml
<Include xmlns="http://schemas.microsoft.com/wix/2006/wi">
  <?define AssemblySemFileVer="2.0.0.0"?>
  <?define AssemblySemVer="2.0.0.0"?>
  ...
  <?define WeightedPreReleaseNumber="30001"?>
</Include>ignissim qui blandit praesent luptatum zzril delenit augue duis dolore te feugait nulla facilisi. 
```

## Your Environment
- Version Used: 5.3.2+Branch.master.Sha.438d45004092e2f6b9aa4d12b07c64becfff4031 (also in 5.2.4)
- Operating System and version (Windows 10, Ubuntu 18.04): Windows

tcs-devel:
This PR fixes #2257

Note 1: The first commit would also fix the issue but does not the IFileStream as an abstraction layer. The second commit does.

Note 2: The added unit test also succeeds without the fix :(. As far as I saw, this is because the `OpenWrite()` function of the `TestFileSystem` class does not behave the way as it documented in https://docs.microsoft.com/en-us/dotnet/api/system.io.file.openwrite?view=netcore-3.1 in regards to existing files.

### DIFF
--- a/src/GitVersionCore/VersionConverters/WixUpdater/WixVersionFileUpdater.cs
+++ b/src/GitVersionCore/VersionConverters/WixUpdater/WixVersionFileUpdater.cs
@@ -12,14 +12,12 @@ namespace GitVersion.VersionConverters.WixUpdater
     }
     public class WixVersionFileUpdater : IWixVersionFileUpdater
     {
-        private readonly IFileSystem fileSystem;
         private readonly ILog log;
         private string wixVersionFile;
         public const string WixVersionFileName = "GitVersion_WixVersion.wxi";
 
-        public WixVersionFileUpdater(IFileSystem fileSystem, ILog log)
+        public WixVersionFileUpdater(ILog log)
         {
-            this.fileSystem = fileSystem ?? throw new ArgumentNullException(nameof(fileSystem));
             this.log = log ?? throw new ArgumentNullException(nameof(log));
         }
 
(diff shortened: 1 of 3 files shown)


src/GitVersionCore/VersionConverters/WixUpdater/WixVersionFileUpdater.cs
```diff
@@ -35,6 +35,7 @@ public void Execute(VersionVariables variables, WixVersionContext context)
             var root = doc.DocumentElement;
             doc.InsertBefore(xmlDecl, root);
 
+            fileSystem.Delete(wixVersionFile);
```
asbjornu, line 38: Should this always be deleted, unconditionally?

tcs-devel: I would say yes, as the entire file is fully re-created (no update). This way it is ensured that the written file does not contain garbage after the expected content.
Maybe one could check whether the file exists, which is in my opinion useless as Delete() does not throw when the file does not exist.

asbjornu:
Thanks for your contributions, @tcs-devel! 🙏

Issue #2262, opened 2020-05-08
[Feature] - V6 Refactoring


dazinator:
## Refactoring to improve composability and allow extensibility

As discussed on slack channel gitversion 6 ideas channel:

- [ ] Move version calculation logic behind `gitversion.exe calculate` command with `-normalize` switch to do git repo normalization logic before calculating.
- [ ] Move logic to detect native supported build servers and set environment variables behind a new command which is written in the style expected for "gitversion tools" e.g it takes version variables in STDIN and writes them to STDOUT e.g `gitversion.exe calculate --normalize | gitversion buildserver`
- [ ] Refactor gitversion msbuild task to be changed to invoke `gitversion calculate` to obtain variable values - where `gitversion` will be invoked as a local or global dotnet tool.
- [ ] Documentation around how to write a tool for use with gitversion - should be simply a case of stating that the tool should take variables in STDIN and write them to STDOUT for other tools to use onwards in the pipeline as a convenience (if the tool doesn't have any useful STDOUT of it's own).

svengeance:
Was there any talk about potential refactoring for the command line parsing? It's all hand made at the moment and I must admit I struggled more with figuring out how to incorporate a new switch/arg than how to incorporate a new feature.

There are 3 main flaws with it imo
- Argument binding is done by hand by repeatedly looping through all args and manually determine if it's a switch or a file/etc
- There's a sort of 2-phase binding between a generic Arguments object, and then a more thorough Options class
- The control flow for it is confusing, with argument binding spanning multiple methods in multiple files

I'd suggest something like https://github.com/commandlineparser/commandline which I've used in the past, but the gist of it is that it supports explicit command definitions (as opposed to chaining more if-statements), property binding, validation, free helptext, you name it.

arturcic:
We're currently considering System.Commandline

gitfool:
FWIW, I've used `System.CommandLine` and like it, although progress to a stable version seems slow.

asbjornu:
One thing we should do for v6 is to make the `path` variable passed to our Docker containers obsolete. With only Linux images, this is easy by providing a smart bash script as an `ENTRYPOINT` that feeds the `gitversion` executable with `/repo`.

I'm not sure how we can provide a smart `ENTRYPOINT` in a cross-platform way, though. We can of course write a .NET Core executable as an `ENTRYPOINT`, but that feels overly complex as executing other processes from .NET isn't trivial.

Something to cook our noodles.

[...]

PR #2263, opened 2020-05-08
Spike showing new CLI usage - got one command working

dazinator:
As part of the refactoring discussed in https://github.com/GitTools/GitVersion/issues/2262

Adopting a new CLI parser library could really help simplify that stuff and give us a nice framework build / add commands around for V6.

To this end, I've got a quick spike up and running where I have got one command running: `gitversion calculate` using @arturcic 's fave CLI parser library: https://github.com/dotnet/command-line-api

## Description
- I unloaded the GitVersion.MsBuild task project. from the solution. I am not concerned with this for this spike.
- I removed logic to do with `remote repo's` - I figured users could clone down their own repo's and then run `gitversion calculate --normalize` on it. Or in a pinch we can add that command later.
- The only command implemented is `gitversion calculate` 
- You also get a free help screen and perhaps other features provided by the parsing library.

Help screen (auto created)

image

And here is the command working:

image

I've done it in a way that commands can benefit from DI:

## Outline of pattern
Each command is registered for DI in `Program.cs`

```csharp
                    services.AddSingleton<GitVersionRootCommand>();
                    services.AddSingleton<CalculateCommand>();
```

A command looks like this:

```csharp
 public class CalculateCommand : Command
    {
        private readonly IGitVersionTool gitversionTool;
        private readonly Logging.IConsole console;

        public CalculateCommand(IGitVersionTool gitversionTool, Logging.IConsole console) : base("calculate", "Calculates version information from your git repository")
        {
            this.gitversionTool = gitversionTool;
            this.console = console;
            this.AddOption(new Option<bool>(
            "--normalize",
            "Attempt to mutate your git repository so gitversion has enough information (local branches, commit history etc) to calculate."));
            this.Handler = CommandHandler.Create<bool?>(ExecuteAsync);            
        }

        private async Task ExecuteAsync(bool? normalize)
        {
            if (normalize ?? false)
            {
                await Normalize();
            }

            var variables = this.gitversionTool.CalculateVersionVariables();
            console.WriteLine(variables.ToString());
        }

        private Task Normalize()
        {
            throw new NotImplementedException();
        }
    }

```

I tried to keep as much of everything else the same, but had to make some quick adjustments, to have the command actually executed.

## Motivation and Context
Gives us a way to structure commands, and a modular way to add new commands, backed by a parsing library that we don't have to maintain ourselves and offers more powerful features beyond the primary concern of gitversion. Establishing something like this will let us build out the command structure we need for V6 more easily.

## How Has This Been Tested?
I did enough to get the solution to compile and run one command with the new parser, just as a proof of concept. The changes I made were not thought about very much. I tried to do as little as possible, and made some on the spot decisions. I unloaded projects just to get it to run. The important thing is to try and show some pattern or example and discuss from there before we have to refactor the code base.

### DIFF
--- a/src/GitVersionCore.Tests/Core/GitVersionExecutorTests.cs
+++ b/src/GitVersionCore.Tests/Core/GitVersionExecutorTests.cs
@@ -418,7 +418,7 @@ namespace GitVersionCore.Tests
 
                 sp = GetServiceProvider(gitVersionOptions);
 
-                gitVersionOptions.ProjectRootDirectory.TrimEnd('/', '\\').ShouldBe(worktreePath);
+                gitVersionOptions.GitRepositoryWorkingDirectory.TrimEnd('/', '\\').ShouldBe(worktreePath);
             }
             finally
             {
@@ -441,7 +441,7 @@ namespace GitVersionCore.Tests
             sp = GetServiceProvider(gitVersionOptions);
 
             var expectedPath = fixture.RepositoryPath.TrimEnd('/', '\\');
-            gitVersionOptions.ProjectRootDirectory.TrimEnd('/', '\\').ShouldBe(expectedPath);
+            gitVersionOptions.GitRepositoryWorkingDirectory.TrimEnd('/', '\\').ShouldBe(expectedPath);
         }
 
         [Test]
--- a/src/GitVersionCore/Core/GitPreparer.cs
+++ b/src/GitVersionCore/Core/GitPreparer.cs
@@ -35,11 +35,9 @@ namespace GitVersion
             var currentBranch = ResolveCurrentBranch();
 
             var dotGitDirectory = gitVersionOptions.DotGitDirectory;
-            var projectRoot = gitVersionOptions.ProjectRootDirectory;
 
-            log.Info($"Project root is: {projectRoot}");
             log.Info($"DotGit directory is: {dotGitDirectory}");
-            if (string.IsNullOrEmpty(dotGitDirectory) || string.IsNullOrEmpty(projectRoot))
+            if (string.IsNullOrEmpty(dotGitDirectory))
             {
                 throw new Exception($"Failed to prepare or find the .git directory in path '{gitVersionOptions.WorkingDirectory}'.");
             }
@@ -50,22 +48,17 @@ namespace GitVersion
         private void PrepareInternal(bool normalizeGitDirectory, string currentBranch, bool shouldCleanUpRemotes = false)
         {
             var gitVersionOptions = options.Value;
-            if (!string.IsNullOrWhiteSpace(gitVersionOptions.RepositoryInfo.TargetUrl))
-            {
-                CreateDynamicRepository(currentBranch);
-            }
-            else
-            {
-                if (normalizeGitDirectory)
-                {
-                    if (shouldCleanUpRemotes)
-                    {
-                        CleanupDuplicateOrigin();
-                    }
 
-                    NormalizeGitDirectory(currentBranch, gitVersionOptions.DotGitDirectory, false);
+            if (normalizeGitDirectory)
+            {
+                if (shouldCleanUpRemotes)
+                {
+                    CleanupDuplicateOrigin();
                 }
+
+                NormalizeGitDirectory(currentBranch, gitVersionOptions.DotGitDirectory, false);
             }
+
         }
 
         private string ResolveCurrentBranch()
@@ -108,90 +101,12 @@ namespace GitVersion
             }
         }
 
-        private void CreateDynamicRepository(string targetBranch)
-        {
-            var gitVersionOptions = options.Value;
-            if (string.IsNullOrWhiteSpace(targetBranch))
-            {
-                throw new Exception("Dynamic Git repositories must have a target branch (/b)");
-            }
-
-            var repositoryInfo = gitVersionOptions.RepositoryInfo;
-            var gitDirectory = gitVersionOptions.DynamicGitRepositoryPath;
-
-            using (log.IndentLog($"Creating dynamic repository at '{gitDirectory}'"))
-            {
-                var authentication = gitVersionOptions.Authentication;
-                if (!Directory.Exists(gitDirectory))
-                {
-                    CloneRepository(repositoryInfo.TargetUrl, gitDirectory, authentication);
-                }
-                else
-                {
-                    log.Info("Git repository already exists");
-                }
-                NormalizeGitDirectory(targetBranch, gitDirectory, true);
-            }
-        }
-
         private void NormalizeGitDirectory(string targetBranch, string gitDirectory, bool isDynamicRepository)
         {
             using (log.IndentLog($"Normalizing git directory for branch '{targetBranch}'"))
             {
                 // Normalize (download branches) before using the branch
                 NormalizeGitDirectory(gitDirectory, options.Value.Settings.NoFetch, targetBranch, isDynamicRepository);
-            }
-        }
-
-        private void CloneRepository(string repositoryUrl, string gitDirectory, AuthenticationInfo auth)
-        {
-            Credentials credentials = null;
-
-            if (auth != null)
-            {
-                if (!string.IsNullOrWhiteSpace(auth.Username))
-                {
-                    log.Info($"Setting up credentials using name '{auth.Username}'");
-
-                    credentials = new UsernamePasswordCredentials
-                    {
-                        Username = auth.Username,
-                        Password = auth.Password ?? string.Empty
-                    };
-                }
-            }
-
-            try
-            {
-                using (log.IndentLog($"Cloning repository from url '{repositoryUrl}'"))
-                {
-                    var cloneOptions = new CloneOptions
-                    {
-                        Checkout = false,
-                        CredentialsProvider = (url, usernameFromUrl, types) => credentials
-                    };
-
-                    var returnedPath = Repository.Clone(repositoryUrl, gitDirectory, cloneOptions);
-                    log.Info($"Returned path after repository clone: {returnedPath}");
-                }
-            }
-            catch (LibGit2SharpException ex)
-            {
-                var message = ex.Message;
-                if (message.Contains("401"))
-                {
-                    throw new Exception("Unauthorized: Incorrect username/password");
-                }
-                if (message.Contains("403"))
-                {
-                    throw new Exception("Forbidden: Possibly Incorrect username/password");
-                }
-                if (message.Contains("404"))
-                {
-                    throw new Exception("Not found: The repository was not found");
-                }
-
-                throw new Exception("There was an unknown problem with the Git repository you provided", ex);
             }
         }
 
--- a/src/GitVersionCore/Core/GitVersionTool.cs
+++ b/src/GitVersionCore/Core/GitVersionTool.cs
@@ -93,13 +93,14 @@ namespace GitVersion
         {
             var gitVersionOptions = options.Value;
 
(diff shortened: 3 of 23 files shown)


src/GitVersionCore.Tests/Core/GitVersionExecutorTests.cs
```diff
@@ -418,7 +418,7 @@ public void GetProjectRootDirectoryWorkingDirectoryWithWorktree()
 
                 sp = GetServiceProvider(gitVersionOptions);
 
-                gitVersionOptions.ProjectRootDirectory.TrimEnd('/', '\\').ShouldBe(worktreePath);
```
dazinator, line 421: Note: `ProjectRootDirectory` was actually being set to the "working directory" of the git repository. I renamed it to `GitRepositoryWorkingDirectory` to aid understanding whilst refactoring. It's not a critical change.

src/GitVersionCore/Core/GitPreparer.cs
```diff
@@ -35,11 +35,9 @@ public void Prepare()
             var currentBranch = ResolveCurrentBranch();
 
             var dotGitDirectory = gitVersionOptions.DotGitDirectory;
-            var projectRoot = gitVersionOptions.ProjectRootDirectory;
```
dazinator, line 38: This argument wasn't used, and I removed it whilst refactoring because it helped me.

src/GitVersionCore/Core/GitPreparer.cs
```diff
@@ -50,22 +48,17 @@ public void Prepare()
         private void PrepareInternal(bool normalizeGitDirectory, string currentBranch, bool shouldCleanUpRemotes = false)
         {
             var gitVersionOptions = options.Value;
-            if (!string.IsNullOrWhiteSpace(gitVersionOptions.RepositoryInfo.TargetUrl))
```
dazinator, line 53: I removed the dynamic repository stuff as per the comments on this PR. User can clone down their own repo and run gitversion normalise against it.

src/GitVersionCore/Core/GitVersionTool.cs
```diff
@@ -93,13 +93,14 @@ public void UpdateAssemblyInfo(VersionVariables variables)
         {
             var gitVersionOptions = options.Value;
 
-            if (gitVersionOptions.AssemblyInfo.ShouldUpdate)
```
dazinator, line 96: This can eventually become a seperate tool as discussed.

src/GitVersionCore/Extensions/GitVersionOptionsExtensions.cs
```diff
@@ -22,20 +20,11 @@ public static string GetDotGitDirectory(this GitVersionOptions gitVersionOptions
...
-            }
-
-            var dotGitDirectory = Repository.Discover(gitVersionOptions.WorkingDirectory);
-
-            if (string.IsNullOrEmpty(dotGitDirectory))
-                throw new DirectoryNotFoundException($"Can't find the .git directory in {dotGitDirectory}");
-
-            using var repository = new Repository(dotGitDirectory);
-            return repository.Info.WorkingDirectory;
+            //return gitVersionOptions.WorkingDirectory;
```
dazinator, line 25: Thought it made more sense to always ask git what the working directory is, rather than assume its the cli 's working directory (current directory) - given we always resolve the dotGitDirectory.

src/GitVersionCore.Tests/Core/GitVersionExecutorTests.cs
asbjornu: This has been a constant source of friction, because I think it's used to support dynamic repositories in a weird way I don't quite understand. I think it's been changed back and forth several times in different PR's, each breaking one other use case that isn't well enough tested. 🤷‍♂️

src/GitVersionCore/Core/GitPreparer.cs
asbjornu: Yes, good riddance! 😅

src/GitVersionCore/Model/GitVersionOptions.cs
```diff
@@ -9,23 +9,23 @@ namespace GitVersion
...
-            projectRootDirectory = new Lazy<string>(this.GetProjectRootDirectory);
-            dynamicGitRepositoryPath = new Lazy<string>(this.GetDynamicGitRepositoryPath);
+            gitRepositoryWorkingDirectory = new Lazy<string>(this.GetRepositoryWorkingDirectory);
         }
 
+        public string[] Args { get; set; }
         public string WorkingDirectory { get; set; }
-
         public string DotGitDirectory => dotGitDirectory.Value;
-        public string ProjectRootDirectory => projectRootDirectory.Value;
```
dazinator, line 25: ProjectRootDirectory --> renamed to GitRepositoryWorkingDirectory for selfish reasons.

dazinator:
I just noticed I got half way through trying to show an example of how a global cli option works and didn't finish. I'll push this change up tomorrow. It was basically to set up  '--logging-method Console | File' (see the GitVersionRootCommand) as a "global option" that will then have its value accessible to all sub commands so they can log accordingly (to console or file etc).  Thought we might need that in a few places.

arturcic:
Wow @dazinator it's awesome. I did not go through all the changes, but from it seems to be great. We need to fix the bugs planned for 5.3.x and we can focus on release 6.0

src/GitVersionExe/Commands/GitVersionRootCommand.cs
```diff
@@ -0,0 +1,19 @@
...
+{
+    public class GitVersionRootCommand : RootCommand
+    {
+        public GitVersionRootCommand(CalculateCommand calculateCommand, IOptions<GitVersionOptions> globalOptions) : base("Versioning for your git repository, solved!")
+        {
+            // this.AddGlobalOption()
+            //this.AddGlobalOption(new Option("--target-path") { Argument = new Argument<LoggingMethod>() });
+            //this.AddGlobalOption(new Option("--logging-method") { Argument = new Argument<LoggingMethod>() });
+            GlobalOptions = globalOptions.Value;
+            this.AddCommand(calculateCommand);
```
arturcic, line 14: So rignt now we construct the command tree by adding command like this.

Now wonder if we can somehow add these commands dynamically, I mean we can register in the DI several commands and then we can resolve them in constructor as IEnumerable<Command> that will get added to the rootCommand. This will allow us to inject commands from external dll loaded as plugin/addin?

arturcic: So the idea is you have a assembly info update tool as plugin for gitversion. You define in the tool that you have a command `assemblyInfo` with it's options/arguments. Then in the bootstrapping phase of gitversion you register that command in DI. The Root command gets the list of available commands from DI and this way you have assembly info command support. Can we add this to this PR? just as a sample?

gep13: Should we ship something "in the box", i.e. the ability to update an assemblyinfo file, but written in such a way that it can be replaced by a 3rd party plugin?  Or should we refactor it out completely, to then have the end-user configure what they want to use?

Personally, I prefer the latter, but I do worry that people will then be confused about how to go from 5.x to 6.x.

I like things being separate, which helps with the updating of those compontents.  But how do we make it so that users moving from 5.x to 6.x, have to do limited setup, configuration, to get things working how they were before?

arturcic: Well @gep13 you are right, we need to think about the migration path from 5.x to 6.0. We might want to have something in the box. Or we can have them out of the box, and somehow enable the previous behavior

arturcic: But for now we can play with the idea

gep13: I like the latter, being able to enable the previous behaviour, just not sure what the best way to do that is ☹️

dazinator: > So rignt now we construct the command tree by adding command like this.
> 
> Now wonder if we can add these commands dynamically, I mean we can register in the DI several commands and then we can resolve them in constructor as IEnumerable that will get added to the rootCommand. This will allow us to inject commands from external dll loaded as plugin/addin.

Yep, or we can also resolve addin commands separately and add them to the root command after its constructed. The only reason for me adding the explicit command types as constructor arguments is I thought it made it a bit clearer to see how the command structure looked and would error if you forgot to register a command that we saw as core, with optional or add in commands being added after construction.

dazinator:
One issue I am working through is how to use global options.
Some options will be global to all subcommands, like whether to log to console or a file for example. At the moment, in `GitVersionExecutor` we rely on already parsing the arguments, so we can set up logging, before actually executing the command. 

```
            if (gitVersionOptions.LogToConsole)
            {
                log.AddLogAppender(new ConsoleAppender());
            }
            else if (gitVersionOptions.LogFilePath != null && gitVersionOptions.LogFilePath != "console")
            {
                log.AddLogAppender(new FileAppender(gitVersionOptions.LogFilePath));
            }
```

However using the new library, there isn't an intial "parse step" that I can see,  - you just execute a command, providing the args, and it will then parse the args, then call you back by actually executing your command handler and passing in the option values you have configured.

The new pattern then (I could me missing something at this library we are using is still pretty new and not everything seems to have been documented yet), looks to be a case of defining these global options on the root command like so - here I have set up a global option called "--logging-method" which can take an Enum value `LoggingMethod` which currently has `Console` or `File`.

```
 public GitVersionRootCommand(CalculateCommand calculateCommand, IOptions<GitVersionOptions> globalOptions) : base("Versioning for your git repository, solved!")
        {             
            this.AddGlobalOption(new Option("--logging-method") { Argument = new Argument<LoggingMethod>() });         
            this.AddCommand(calculateCommand);
            this.Handler = Execute();
        }

```

And then, the first location that can take advantage of this is the handler for each `subcommand` - for example here is the `Calculate` command and you can see we can inject this global option in:

```csharp
 public class CalculateCommand : Command
    {
        private readonly IGitVersionTool gitversionTool;
        private readonly Logging.IConsole console;

        public CalculateCommand(IGitVersionTool gitversionTool, Logging.IConsole console) : base("calculate", "Calculates version information from your git repository")
        {
            this.gitversionTool = gitversionTool;
            this.console = console;
            this.AddOption(new Option<bool>(
            "--normalize",
            "Attempt to mutate your git repository so gitversion has enough information (local branches, commit history etc) to calculate."));
            this.Handler = CommandHandler.Create<bool?, LoggingMethod>(ExecuteAsync);            
        }

        private async Task ExecuteAsync(bool? normalize, LoggingMethod loggingMethod)
        {
            if (normalize ?? false)
            {
                await Normalize();
            }

            if(loggingMethod == LoggingMethod.Console)
            {
                // etc
            }

            var variables = this.gitversionTool.CalculateVersionVariables();
            console.WriteLine(variables.ToString());
        }

```

So it's a case of working out a new place to deal with the global options in a centralised mannor. Currently it looks like that would mean we'd potentially derive all of our sub commands from a common base that can be called to process the global options:

I'll push up the example of that, but thought it worth raising in case there is meant to be a different way to handle this that someone can see.

arturcic:
@dazinator do you mind if I convert the PR to draft?

dazinator:
@arturcic yep thats fine, this is dead now, I'll leave this just for reference although most of this was moved to the new branch

arturcic:
Thnaks, makes sense, let's continue on the branch

PR #2265, opened 2020-05-10
#1611 Project file updating

### LINKED ISSUE #1611: Populate new csproj format properties from Azure DevOps build pipeline task
The existing build tasks for Azure DevOps have long had the option to update AssemblyInfo.cs before running a build task (which is really sweet BTW). Is there going to be support in the future for updating the version/package metadata properties in the new csproj format from the build task?

I've previously used a powershell script to take the output of the GitVersion variables to manually go into all the csproj files and update the properties before running ```dotnet build```. It would be great if a future version of the build task could do this OOTB.

I'll try to find the powershell script I use which may better explain what I'm trying to achieve.

svengeance:
Fixes #1611 

This PR allows the ability to update csproj/vbproj/fsproj files directly, rather than forcing users to implement `<GenerateAssemblyInformation>false</GenerateAssemblyInformation>` in their project files.

## Description
I tried to keep my programming in line with how the AssemblyVersionInfoUpdater performs its tasks, so there should be no surprises as to how this works. I first acquire all project files explicitly named or discovered, verify they are in a format that I am comfortable parsing, and then simply add the version elements.

### Libraries Used
I opted to go with Linq2Xml over incorporating `Microsoft.Build` for a couple reasons; I know you guys are approaching a new rearch, and I wanted to ensure my changes didn't incorporate something large, and I also had issues out-of-the-box with the package, and was uncomfortable shipping that.

### Update Logic
There are quite a few scenarios and odds-and-ends with versioning in the project files. Users can have multiple property groups _and_ have multiple elements spread throughout (i.e. they can have two `<AssemblyFileVersion>`s in their project, where the last-in wins. To account for this, my implementation seeks the last `<PropertyGroup>` that has a version element and updates it, **or** updates the first available `<PropertyGroup>` to have the version element.

### CLI Usage
I opted to hijack the usage of `/updateassemblyinfofiles` as it was one of the least-invasive opportuinities to enter. Project file updating is triggered by adding another switch `/targetprojectfiles`. I'm totally open to suggestions on this.

### Testing
As far as testing goes I've opted to use naked XML in the test cases (and sometimes for the expected value) as I believe it offers the best visibility possible into the before-and-after changes. Note, I am missing some full integration tests (e.x. create a file, call `Program` in that directory), but I'm unsure what scope of test you guys want.

### Misc
I updated IFileSystem to Enumerate Files rather than returning the list. We were using them in contexts of `IEnumerable` and yielding results, so it'll prevent the big ol' spike of loading files and should instead parse them one at a time.

## Related Issue
https://github.com/GitTools/GitVersion/issues/1611

## Motivation and Context
By default .NET Core applications do *not* have an AssemblyInfo.cs. This change allows us to support out-of-the-box versioning without customers having to make any changes.

## How Has This Been Tested?
The core functionality can be broken down into the overall integration with two subcomponents.

- Update/replace file
  - Determine if we can update file
  - Update XML structure

I've unit tested the two pieces of business logic, and I have one test following some prior tests that I believe tests the integration of the whole.

None, but here's a picture of one of my cats for your pleasure: https://media.discordapp.net/attachments/631864419693101056/694290169691373668/20200330_170018.png?width=647&height=862
## Checklist:


src/GitVersionExe/HelpWriter.cs
```diff
@@ -53,6 +53,9 @@ public void WriteTo(Action<string> writeAction)
  # AssemblyInfo updating
     /updateassemblyinfo
                     Will recursively search for all 'AssemblyInfo.cs' files in the git repo and update them
+    /targetprojectfiles
+                    Will target project files (cs/fs/vbproj files) to update instead of AssemblyInfo class files,
```
asbjornu, line 57: Can we call the argument `updateprojectfiles` to be consistent with the rest of the `update` arguments?

```suggestion
    /updateprojectfiles
                    Will recursively search for `.csproj`, `.vbproj` and `.fsproj` files in the Git repository and update them.
```

svengeance:
@asbjornu I've made another commit to rename the switch. I suppose the only concern is that a user _may_ assume that that switch is independently capable of updating, but you do need both `/updateassemblyinfo` and `/updateprojectfiles`

Regarding the failing tasks - I'm not sure how to resolve those. The errors seemed to be out of my control, but let's wait for this latest build

asbjornu:
Why is `updateprojectfiles` dependent on `updateassemblyinfo`? Aren't they independent from one another?

> Regarding the failing tasks - I'm not sure how to resolve those. The errors seemed to be out of my control, but let's wait for this latest build

All tests pass now, so we're good. 😃

svengeance:
I mentioned that in my CLI section --

>I opted to hijack the usage of /updateassemblyinfofiles as it was one of the least-invasive opportuinities to enter. Project file updating is triggered by adding another switch /targetprojectfiles. I'm totally open to suggestions on this.

I can add additional logic to this section if need be to make it independent. I was having a difficult time _parsing_ through the arg-parsing logic and wanted to keep my changes minimal there

svengeance:
I guess -- put another way, updating AssemblyInfo compared to updating ProjectFiles differs only in 2 aspects: /ensureassemblyinfo is invalid for ProjectFiles, and we target .csproj files instead of .cs files. Given that the logic is majorly the same I thought it reasonable for this to be a modifier of the action rather than a new action unto itself.

asbjornu:
While I understand your reasoning, aren't the two options mutually exclusive anyway?

svengeance:
True enough. Like I said I'm totally open to changing how it's used. Give me a bit and I'll rewire it to be its own action.

asbjornu:
Great, @svengeance! While you're at it, will you please perform a `rebase` against the `HEAD` of `master` as well? 🙏 😃

svengeance:
@asbjornu Updated. I've left assembly/project updating loosely related in that they're in the same set of configurable options, but you can now independently target one or the other through the CLI. I imagine that the transition to System.CommandLine will mimic this approach, where you have a base command for updating files that all `/updatexyz` shares which will bind to a base set of options like the target files.

I've also explicitly disallowed updating both project and assemblyinfo files in the same run, but theoretically this should be able to be done by removing the `WarningException`s and changing the `else if` in `GitVersionTool.cs` to an `if`.

svengeance:
Oh, and I've rebased onto HEAD as requested.

src/GitVersionExe/HelpWriter.cs
```diff
@@ -53,8 +53,13 @@ public void WriteTo(Action<string> writeAction)
  # AssemblyInfo updating
     /updateassemblyinfo
                     Will recursively search for all 'AssemblyInfo.cs' files in the git repo and update them
-    /updateassemblyinfofilename
```
asbjornu, line 56: Hm, with `updateassemblyinfofilename` changed to `updatefilenames`, this is a breaking change and will have to wait for v6 anyway. Perhaps leave `updateassemblyinfofilename` as is and introduce another parameter that serves the same function, but for the `updateprojectfiles` parameter? I know it's messy, but this will all be cleaned up in v6.

svengeance: How many a developer has been fooled by the phrase _it's messy but we'll get it all cleaned up next version.._ 😄 

Will do.

svengeance: @asbjornu So uhh..I'm reverting some of the _naming_ I did, but it appears that that parameter is not implemented, and the only way to pass file names to UpdateAssemblyInfo (and now UpdateProjectInfo) is to specify filenames _after_ the update switch. Realistically that parameter can be removed without breaking anything since _including_ it seems to break things in and of itself.

In any case give it another look and see if I unbroke'd it.

svengeance: Example of a test that fails -
```cs
[Test]
public void UpdateAssemblyInfoWithFilenameFromUpdateAssemblyInfoFilenameSwitch()
{
    using var repo = new EmptyRepositoryFixture();

    var assemblyFile = Path.Combine(repo.RepositoryPath, "CommonAssemblyInfo.cs");
    using var file = File.Create(assemblyFile);

    var arguments = argumentParser.ParseArguments($"-targetpath {repo.RepositoryPath} -updateAssemblyInfo -updateassemblyinfofilename CommonAssemblyInfo.cs");
    arguments.UpdateAssemblyInfo.ShouldBe(true);
    arguments.UpdateAssemblyInfoFileName.Count.ShouldBe(1);
    arguments.UpdateAssemblyInfoFileName.ShouldContain(x => Path.GetFileName(x).Equals("CommonAssemblyInfo.cs"));
}       
```

svengeance: Provided everything I said is true - would you be comfortable mimicking `/updateAssemblyInfo` and relegating passing Filenames to just that switch, rather than implementing something for project files that isn't in assembly files?

asbjornu: Good find, @svengeance. Indeed, it looks like you're right and that `updateassemblyinfofilename` is not implemented. The actual implementation supporting filename(s) after the switch is much cleaner anyway, so yes, let's go with that for project files as well. 👍

svengeance:
@asbjornu What else would you like to see on this? Personally I'd love some sort of full integration test starting from the top of `Program.cs` but I'm not sure if that's feasible. I plan on adding at least one or two ArgParser tests, but other than that, what else?

src/GitVersionExe/HelpWriter.cs
svengeance: For reference here is the issue tracking this @asbjornu  - https://github.com/GitTools/GitVersion/issues/584
It may be good to reference this in the v6 commits as I imagine you'll end up removing this parameter as a whole.

asbjornu: Thanks for the reference. Yes, with the new CLI (#2272), I suspect a lot of the current arguments to become obsolete.

asbjornu:
A couple of argument parser tests would be good. Otherwise, I think the functionality itself is well enough tested. As this is a new feature no one uses yet, we can be more lenient on the testing and allow the community to test a bit for us. Then we add regression tests as needed if bugs pop up.

svengeance:
@asbjornu I think this may be ready. I've added a few parsing tests, but given their incoming demise, I don't see much value in being thorough. Let me know if/what else you'd like to see here 🎉

asbjornu:
Excellent, @svengeance! Thank you so much for your contributions. :pray:

Issue #358, opened 2015-01-25
Better argument parser


JakeGinnivan:
Depending on the ordering of some of the switches things do not work. We should switch to a small simple library

JakeGinnivan:
Looking into this, most decent libraries work similar to git (which actually could be good to align to).

The main difference is the single dash means a short code and must be a single letter. In git this is why you can go `git commit -am "Msg"` and that expands to `git commit --add --message "Msg"`

So the question is, should we break the existing command line usage and rethink it to make it better or do we need to stay compatible.

@andreasohlund @gep13 @SimonCropp

JakeGinnivan:
Ignore, can fix a few bugs in our parser which will make it better

PR #2272, opened 2020-05-12
New CLI

### LINKED ISSUE #2275: [Feature] New CLI
## Detailed Description

This issue is our explorations of how the new CLI of GitVersion 6 is going to look like.

### Design

The new CLI is going to have a command-based style, allowing for more laser-focused and composable execution. We want `stdin` and `stdout` to become the default way to both read and write version variables, logging instead to `stderr` so `stdout` can be kept as valid JSON throughout.

Since we want to reduce the amount of work being done for each command, GitVersion may have to be executed several times for certain builds, perhaps introducing some latency, but also making it much easier to pinpoint exactly during which step an error occurs so we can fix more easily and users of GitVersion can circumvent and monkey-patch the bugs much more efficiently.

### Architecture

The new and more laser-focused command-based CLI allows us to refactor the GitVersion codebase into smaller, more compartementalized parts. This should lead to an almost dependency-free `Core` with a few optional addon projects that can live their own lives. The API between `GitVersion.Core` and the addons will be based on the CLI infrastructure we choose. Our current bet is on [`System.CommandLine`](https://github.com/dotnet/command-line-api).

Addons will be able to add their own commands to GitVersion and besides the `Command` implementations and arguments they can add to it, they will be able to read the version variables produced by GitVersion as JSON and then perform whatever task they wish. The output from an addon should be JSON written to `stdout` so different commands can be composed with piping.

## Possible Implementation

Below is an example of some of the commands the new CLI may support, with some of their corresponding arguments.

```sh
gitversion normalize # Normalizes the repository to its required state
gitversion calculate --output:file=gitversion.json # Calculate the version number and output to gitversion.json
gitversion write --from:file=gitversion.json --to:stdout # Write the version variables to stdout (only the JSON should appear in stdout, errors will be written to stderr)
gitversion write --from:file=gitversion.json --to:assemblyinfo=./**/AssemblyInfo.cs # Read version variables from gitversion.json and write to globbed AssemblyInfo.cs files
gitversion write --from:file=gitversion.json --to:stdout | gitversion write --from:stdin --to:wix # Read version variables from stdin and write to wix (I'm fuzzy on the details of Wix, so please bear with me)
```

## Related Issue

#358, #428, #598, #572.

## Motivation and Context
The CLI of GitVersion is a road that has been built as we have walked it, with no planning and no idea of scope or feature set before implementation was started. It has proved difficult to support POSIX file systems due to forward slash `/` being chosen as the argument separator originally, the argument system doesn't allow for easy documentation generation and perhaps most importantly: Argument parsing is not a core concern of GitVersion, so we're best served to outsource this entire thing to a third-party library.

asbjornu:
## Description

This is the PR for our new CLI. Resolves #2275. Please discuss the implementation there and not in this pull request.

## Related Issue

#358, #428, #598, #572.

## Motivation and Context
The CLI of GitVersion is a road that has been built as we have walked it, with no planning and no idea of scope or feature set before implementation was started. It has proved difficult to support POSIX file systems due to forward slash `/` being chosen as the argument separator originally, the argument system doesn't allow for easy documentation generation and perhaps most importantly: Argument parsing is not a core concern of GitVersion, so we're best served to outsource this entire thing to a third-party library.

## How Has This Been Tested?

Nothing yet, but it will come.

## Checklist:

### DIFF
--- a/src/GitVersion.Cli/Commands/CalculateCommand.cs
+++ b/src/GitVersion.Cli/Commands/CalculateCommand.cs
@@ -1,0 +1,96 @@
+using GitVersion.Logging;
+using LibGit2Sharp;
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using System.Threading.Tasks;
+
+namespace GitVersion.Cli
+{
+    public class CalculateCommand : Command
+    {
+        private readonly Logging.IConsole console;
+        private readonly IGitVersionCalculator gitversionCalculator;
+        private readonly CommandWrapper executor;
+
+        public CalculateCommand(Logging.IConsole console, IGitVersionCalculator gitversionCalculator, CommandWrapper executor) : base("calculate", "Calculates version information from your git repository")
+        {
+            this.console = console;
+            this.gitversionCalculator = gitversionCalculator;
+            this.executor = executor;
+            AddOptions();
+
+            this.Handler = CommandHandler.Create<GlobalOptions, CalculateOptions>(ExecuteAsync);
+        }
+
+        private void AddOptions()
+        {
+            this.AddOption(new Option<bool>(
+                 "--normalize",
+                 "Attempt to mutate your git repository so gitversion has enough information (local branches, commit history etc) to calculate."));
+            this.AddOption(new Option<string>(
+                 "--dot-git-dir",
+                 "The path of the .git folder. If not specified, will be discovered from the envrionment current working directory."));
+            this.AddOption(new Option<string>(
+                 "--no-cache",
+                 "Whether to disable caching of the version variables. If true is specified, any previous cached calculation will be ignored and the new calculation won't be cached either."));
+
+            this.AddOption(new Option<string>(
+                "--branch",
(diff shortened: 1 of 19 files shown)


svengeance:
Two things I've noticed --

In your examples you're showing commands like `--command:arg`. I think it's more _usual_ to see `--command arg`, ex `dotnet build -v DEBUG --runtime win-x64`.

Secondly, you're sort of parsing *two* arguments in one go, and I think this will bite you going forward.

Take the following:
`gitversion write --from:file=gitversion.json --to:assemblyinfo=./**/AssemblyInfo.cs`

I think this would be much better represented as:
`gitversion write --input-type file --input gitversion.json --output-type assemblyinfo --output ./**/AssemblyInfo.cs`

It is certainly more verbose, but it gives the users and developers more flexibility when constructing their commands, and it stops you from inventing your own constructs as to how a command should look.

asbjornu:
Thanks for the feedback @svengeance! I agree with you and will adjust my examples (moved to #2275) accordingly.

PR #2277, opened 2020-05-13
Fix environment variable parsing for authentication

### LINKED ISSUE #2276: [Bug] GITVERSION_REMOTE_PASSWORD environment variable overwrites the Username field instead of setting the password
**Describe the bug**
GITVERSION_REMOTE_PASSWORD environment variable overwrites the Username field instead of setting the password

## Expected Behavior
GITVERSION_REMOTE_USERNAME sets the username for authentication.
GITVERSION_REMOTE_PASSWORD sets the password for authentication.

## Actual Behavior
GITVERSION_REMOTE_USERNAME sets the username for authentication.
GITVERSION_REMOTE_PASSWORD overwrites the value previously set for username.

## Possible Fix
I have submitted a PR with the fix, will attach once this ticket is submitted.

:Edit: Here it is: https://github.com/GitTools/GitVersion/pull/2277

bendoyon:
Fix an issue where GITVERSION_REMOTE_PASSWORD would overwrite the Username options intead of setting the Password option.

Also make sure authentication is set even if no other command line arguments are present.

## Description
Changed the one-liner where the assignation happens in ArgumentParser.cs. Pretty self-evident change. Relevant unit tests have been added.

While making this change, I realized that my tests failed when no other command line arguments were passed. I made sure to add authentication even in the case where no other command line arguments exist, and added relevant unit tests.

## Related Issue
Fixes #2276

## Motivation and Context
I need this to get GitVersion working in a GitLab CI build that needs authentication because it builds with Docker.

## How Has This Been Tested?
I have added 4 relevant unit tests.

## Checklist:

- [n/a] My change requires a change to the documentation.
- [n/a] I have updated the documentation accordingly.

### DIFF
--- a/src/GitVersionExe.Tests/ArgumentParserTests.cs
+++ b/src/GitVersionExe.Tests/ArgumentParserTests.cs
@@ -14,6 +14,7 @@ namespace GitVersionExe.Tests
     [TestFixture]
     public class ArgumentParserTests : TestBase
     {
+        private IEnvironment environment;
         private IArgumentParser argumentParser;
 
         [SetUp]
@@ -24,6 +25,7 @@ namespace GitVersionExe.Tests
                 services.AddSingleton<IArgumentParser, ArgumentParser>();
                 services.AddSingleton<IGlobbingResolver, GlobbingResolver>();
             });
+            environment = sp.GetService<IEnvironment>();
             argumentParser = sp.GetService<IArgumentParser>();
         }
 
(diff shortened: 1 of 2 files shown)


bendoyon:
@arturcic Thanks! Still some authentication on my end, I'm getting a "Could not find a 'develop' or 'master' branch, neither locally nor remotely." when in a Docker build running on GitLab CI.

This didn't end up fixing it, but we're getting closer maybe...

arturcic:
You should pass the GitLab env vars to the docker run from the GitLab pipeline

arturcic:
https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/BuildAgents/GitLabCi.cs#L10

arturcic:
https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/BuildAgents/GitLabCi.cs#L41

bendoyon:
@arturcic Yikes, getting a NullReference now:

/root/.nuget/packages/gitversiontask/5.4.0-pullrequest2277-50/build/GitVersionTask.targets(10,9): error : NullReferenceException: Object reference not set to an instance of an object. [/src/Services.Common/Services.Common.csproj]
/root/.nuget/packages/gitversiontask/5.4.0-pullrequest2277-50/build/GitVersionTask.targets(10,9): error :    at GitVersion.Extensions.RepositoryExtensions.CreateOrUpdateLocalBranchesFromRemoteTrackingOnes(IRepository repo, ILog log, String remoteName) in D:\a\GitVersion\GitVersion\src\GitVersionCore\Extensions\RepositoryExtensions.cs:line 177 [/src/Services.Common/Services.Common.csproj]

Which seems to be this line:
if (localRef.ResolveToDirectReference().TargetIdentifier == remotedirectReference.TargetIdentifier)

arturcic:
You are using GitversionTask it Gitversion.Tool? As what you fixed is for the tool, not the task

bendoyon:
Is there any way of passing parameters to the MsBuild task? I could use the dynamic repository feature to make sure it has access to everything it needs, instead of copying the .git repository into the Docker image.

bendoyon:
@arturcic  I figured the Task is an elegant way to make it work both locally, and from the CI, sort of seamlessly. Is there any way to achieve the same thing with the EXE?

For example, I have a test validating that my ASP.NET Core endpoint can return the Git version, when I run the tests on my local machine, when the tests are run in the CI, and when the application is deployed.

Is there a way to have a similar sort of workflow with the EXE?

PR #2280, opened 2020-05-20
Prevent exeution of GenerateGitVersionInformation in XAML generated intermediate-projects

### LINKED ISSUE #2228: [Bug] The item "obj\Debug\GitVersionInformation.g.cs" was specified more than once in the "Sources" parameter
Hello.

**Describe the bug**
I have C# project on .Net Framework 4.7.2.
I upgrade nuget from
`<package id="GitVersionTask" version="5.0.1" targetFramework="net472" developmentDependency="true" />`
to
`<package id="GitVersionTask" version="5.2.4" targetFramework="net472" developmentDependency="true" />`

When build I got this error:
> The item "obj\Debug\GitVersionInformation.g.cs" was specified more than once in the "Sources" parameter.  Duplicate items are not supported by the "Sources" parameter.
C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\MSBuild\Current\Bin\Roslyn\Microsoft.CSharp.Core.targets	59	

I have another 33 projects (.Net Framework and .Net Core and .Net Standard) in this solution and all other projects build OK.

Any suggesions what's happen and how fix this?

## My Environment
Operating System Windows 10
Visual Studio 16.5.3

jbaehr:
## Description

During XAML'S build magic a temporary project may get generated and
build (target "GenerateTemporaryTargetAssembly"). The generated project
file already contains compile-items with the generated files from the
original project, this `GitVerstionInformation.g.cs` is listed. When the
GitVersion task now runs again, the file is added another time resulting
in an error of CSC.
This patch tries to detect this situation and prevent the execution of
GenerateGitVersionInformation by checking for a special property
injected by GenerateTemporaryTargetAssembly. Unfortunately this is not
documented and thus may change, see https://referencesource.microsoft.com/#PresentationBuildTasks/BuildTasks/Microsoft/Build/Tasks/Windows/GenerateTemporaryTargetAssembly.cs,163

## Related Issue
This should fix #2228

## How Has This Been Tested?
This patch was tested on top of the release v5.3.3 of GitVersionTask and solved the issue on a .NET Framework-4.7.2 solution, compiled with MSBuild 15 as well as VS2019 (aka MSBuild-16).

## Checklist:

### DIFF
--- a/src/GitVersionTask/build/GitVersionTask.props
+++ b/src/GitVersionTask/build/GitVersionTask.props
@@ -27,6 +27,14 @@
 
         <!-- Property that enables GenerateGitVersionInformation -->
         <GenerateGitVersionInformation Condition=" '$(DisableGitVersionTask)' == 'true' ">false</GenerateGitVersionInformation>
+        <!--
+          During XAML's "GenerateTemporaryTargetAssembly" a new project file is generated and compiled
+          which already contains the "GeneratedCodeFiles", i.e. GitVersionInformation.g.cs.
+          Then, when GenerateGitVersionInformation is called in this temp-build, the file is added another time, which results in an error at CSC.
+          Here we try to detect this situation and prevent GenerateGitVersionInformation from running.
+          (The global property "_TargetAssemblyProjectName" is injected by XAML's above-mentioned task)
+        -->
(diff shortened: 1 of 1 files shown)


arturcic:
@jbaehr  Thanks for the investigation. I will merge it

PR #2298, opened 2020-05-29
Ensure targets run early enough for Source Link

### LINKED ISSUE #2297: [Bug] Generated source files can't be embedded by Source Link
When using [Source Link](https://github.com/dotnet/sourcelink) with `EmbedUntrackedSources` set to `true`, any generated source files that aren't committed to the repo are supposed to be embedded in the PDB instead.

However, GitVersionInformation.g.cs is not being embedded despite being a generated file.

This is happening because the target that creates it is running too late for Source Link to know that it needs to embed it.

FYI this was also a [problem](https://github.com/dotnet/sdk/issues/10614) for the `AssemblyInfo.cs` generated by the .NET Core SDK, and was fixed for that file in the 3.1.300 SDK.

## Expected Behavior
Source files generated by GitVersion should be embedded in the PDB when using Source Link.

## Actual Behavior
The files are not embedded in the PDB.

## Fix
The UpdateAssemblyInfo and GenerateGitVersionInformation targets should have `BeforeCompile` added to their `BeforeTargets` attributes.

bording:
This adds `BeforeCompile` to the `BeforeTargets` attributes of the two targets that generate source files.

Fixes #2297

### DIFF
--- a/src/GitVersionTask/build/GitVersionTask.targets
+++ b/src/GitVersionTask/build/GitVersionTask.targets
@@ -14,7 +14,7 @@
 
     </Target>
 
-    <Target Name="UpdateAssemblyInfo" BeforeTargets="CoreCompile" Condition="$(UpdateAssemblyInfo) == 'true'">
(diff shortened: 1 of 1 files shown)


asbjornu:
Thanks for your contribution, @bording! 🙏

PR #2300, opened 2020-05-31
Refactor configuration

Inok:
It's a first step towards configuration refactoring. 
All tests pass and I did my best to not introduce any breaking changes. However, the changes are quite huge, so it's a draft at this moment and I personally think it should not be released as part of 5.3.x. My intention in this PR is to show what I've done, to ask whether the changes are appropriate or not, and to discuss further steps.

@asbjornu, could you please review the changes? What do you think about them and about the next step I suggested below?

## Description
I reversed the order of configuration merging. The previous approach was "read custom configuration and add default values to what's empty". The new approach is more natural: "create default configuration, override by configuration from source 1, override by...".

Previously, several config-level settings has been written to the branch config right on merging (IncrementStrategy and VersioningMode). The new approach does that on the very last step: on the configuration finalization. Now the configuration should be built like this:
```
var config = DefaultConfigurationProvider.GetDefaultConfiguration().Apply(new Config { NextVersion = "2.0.0" }).FinalizeConfig();
```

## Next step

I guess, the next step is to do something like that:
```
Config config = new ConfigurationBuilder()
                              .AddDefault()
                              .AddYamlFile(...)
                              .AddLayer(new ConfigLayer { NextVersion = ... })
                              .Build();
```
So it will be similar to .NET Core's configuration.


asbjornu:
This is great! Everything looks good so far. Just mark this as ready whenever you are. The proposed .NET Core-compatible configuration syntax is 💯.

Inok:
The future improvements in this area seems quite complicated and I'd prefer to keep the current changes in a separate PR. So I'm marking this PR as ready for review. If the changes look good to merge at this point, I think it would be better to do that.

src/GitVersionCore/Configuration/DefaultConfigProvider.cs
```diff
@@ -0,0 +1,134 @@
...
+using GitVersion.Model.Configuration;
+using GitVersion.VersionCalculation;
+
+namespace GitVersion.Configuration
+{
+    public class DefaultConfigProvider
+    {
+        private const int DefaultTagPreReleaseWeight = 60000;
+
+        public static Config CreateDefaultConfig()
```
asbjornu, line 12: I'm not sure about this being `static`. Perhaps defaults can just be set in the `Config` (or a derived `DefaultConfig`) constructor instead? Then we can make all properties immutable so all mutations of the `Config` class instead creates a new `Config` instance. I'm not sure we need `DefaultConfigProvider` then. Thoughts?

Inok: But how can we make these properties immutable? AFAIK, there is no way to create a new `Config` instance on a property mutation in С#. And we also have to keep `Config` mutable at least for a yaml deserializer.

I agree that `DefaultConfigProvider` is not a good idea. And I personally think that the mutability of a Config (let's call it 'configuration layer') is OK. But it would be better to introduce a `ConfigurationSnapshot` (final immutable configuration based on several Configs) as a result of merging these layers. It leads us to something like this:

```csharp
// Default configuration, replaces DefaultConfigProvider.CreateDefaultConfig().Finalize();
ConfigurationShapshot defaultConfig = new ConfigurationBuilder().Build();

// Default + yaml file, replaces DefaultConfigProvider.CreateDefaultConfig().Apply(configReader.Read(yamlFile)).Finalize();
ConfigurationShapshot customConfig1 = new ConfigurationBuilder()
                                         .AddYamlLayer(/* workingDir etc */)
                                         .Build();

// Complete example: default + yaml file + /overrideconfig, 
// replaces DefaultConfigProvider.CreateDefaultConfig().Apply(configReader.Read(yamlFile)).Apply(new Config { TagPrefix = "[vV](er)?" }).Finalize();
ConfigurationShapshot customConfig2 = new ConfigurationBuilder()
                                         .AddYamlLayer(/* workingDir etc */)
                                         .AddLayer(new Config { TagPrefix = "[vV](er)?" })
                                         .Build();

customConfig2 .TagPrefix = "..." // Compilation error: property is immutable
```

What do you think about that?

asbjornu: I'm not sold on the "layer" naming, but otherwise I like using the builder pattern a lot, because then we keep things mutable until `.Build()` and from then on, everything can be immutable. If we do the following adjustments, I think this is going to be great:

```c#
IConfig customConfig = new ConfigurationBuilder()
                        .AddFile(/* workingDir etc */)
                        .Add(new Config { TagPrefix = "[vV](er)?" })
                        .Build();
```

The `IConfig` instance returned from `.Build()` can be implemented by a class private to the `ConfigurationBuilder`.

Inok: In your example, does `Config` implements `IConfig`? If so, I'd like to emphasize one problem. There are two different "concepts": configuration "layers" (default config, yaml file, overrideconfig etc.) which we merge to build a final configuration, and a final configuration itself. Now they both represented by a `Config` class. The problem is that all `Config`'s properties are nullable, but several properties of a final configuration expected to be not-null. It eventually leads to a bunch of checks:
https://github.com/GitTools/GitVersion/blob/3fccbc3b7afdbb0b38e7e9c6aa157d69fc9d8c74/src/GitVersionCore/Configuration/ConfigExtensions.cs#L224-L229

IMHO, it doesn't look natural, and the remark `this should not happen, please report an issue` in exception messages is a sign that maybe something wrong with that design. So I'd prefer a clean separation between configuration layers (`Config`) and a final configuration (I named it` ConfigurationSnapshot`, but I'm not good in naming, so let me know if you'd prefer another name).

If you agree that it would be better to introduce a new 'final configuration' with immutable non-nullable properties, I'll do that. If not, I'll gladly do what you suggested in the previous comment 😃

asbjornu: I think we should clearly separate between `ConfigurationBuilder` which will return an instance of itself for every method except `.Build()` and `IConfig`, which is the final immutable interface. None of the mutable classes should implement `IConfig`.

src/GitVersionCore/Configuration/ConfigurationBuilder.cs
```diff
@@ -0,0 +1,276 @@
+using System;
+using System.Collections.Generic;
+using GitVersion.Extensions;
+using GitVersion.Model.Configuration;
+using GitVersion.VersionCalculation;
+using JetBrains.Annotations;
+
+namespace GitVersion.Configuration
+{
+    public class ConfigurationBuilder
```
asbjornu, line 10: Nice. Whenever you're ready for a new review, click the ↺ button.

image

Inok:
I'm sorry, but I don't have enough time right now to do exactly what you've asked (immutable `IConfig`). So if the current changes look good enough to be merged - let's do that. I'll keep in mind that immutable `IConfig`, and I'll hopefully open a new PR with the `IConfig` when I have more time.

asbjornu:
Thank you so much for this, @Inok! 🙏 Whenever you find time, I would love to accept a PR with immutable `IConfig`. 👍

Issue #2305, opened 2020-06-01
[Improvement] GitVersion generated code should not participate in code coverage


odalet:
# Summary

I use GitVersion through its MSBuild task. Yesterday by running "code coverage" from Visual Studio, I noticed, the generated `GitVersionInformation` class was considered as not covered code. Well, it's generated code and therefore It'd make sense for it not be part of the coverage statistics:

![image](https://user-images.githubusercontent.com/79333/83417246-1cf6c880-a422-11ea-97e9-33d5815e1484.png)

## Detailed Description

Visual Studio (Enterprise) coverage facility uses compiled assemblies to compute code covered by unit tests statistics. Hence, even though `GitVersionInformation` wasn't written by the user, it appears in coverage results. I suspect it would be the same for other .NET coverage tools. This kind of code twists the coverage results (if only by a very tiny bit). It would be more correct to exclude it from coverage analysis.

## Context

Honestly, this is not a must-have. I can live with a little discrepancy in my coverage resutls. However, I suppose changing this would be desirable for accuracy and correctness purpose.

## Possible Implementation

The change should be rather straightforward (and I'm about to submit a PR for this): one only needs to add the `ExcludeFromCoverage` attribute on top of the generated code.

asbjornu:
Resolved in #2306.

PR #2306, opened 2020-06-01
Add ExcludeFromCodeCoverage attribute on top of generated GitVersionI…

odalet:
…nformation classes

## Description

This decorates generated code with `ExcludeFromCodeCoverage` attribute so that it does not participate in code coverage statistics.  

## Related Issue

#2305

## Motivation and Context

This prevents generated code from being detected as 'not covered'. Generated code should not be part of code coverage statistics.

## How Has This Been Tested?

In order to make sure adding this attribute resulted in 100% coverage, I crafted a simple solution with a VB.NET, a C# and a F# project + a unit test. This project is available here: <https://github.com/odalet/GitVersionTests>

I couldn't achieve 100% coverage on the F# lib and detected a potential issue on the VB project. I describe these below in subsequent comments because I don't really know what way to go with these and then am opening discussions.

## Checklist:


odalet:
So here is the problem with the VB.NET Generated code:
because it is not part of any namespace, it gets generated into the assembly's default namespace, not its 'global' namespace:

image

I suggest to enclose the generated code into this so that it behaves like the C# code:

```vb.net
Namespace Global
... existing class template
End Namespace
```

This change is not part of this PR because I wonder if I should propose another PR for this

odalet:
Concerning F#, it is a bit more tricky. Modules seem to generate plenty of stuff that is _not_ inheriting the `ExcludeFromCodeCoverage` attribute:

image

For now, I managed to achieve 100% coverage by changing the F# module into a static class. Something like that:

```fsharp
namespace global

[<AbstractClass; Sealed>]
[<global.System.Runtime.CompilerServices.CompilerGenerated>]
[<System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage>]
type GitVersionInformation() =
    static member Major = "1"
    static member Minor = "2"
    static member Patch = "3"
   ...
```

However, this being quite a big change, I didn't pushed it yet. Waiting for approval here.

odalet:
You're welcome! And what about the 2 other issues I mention above? Could you have a look at them?Should I open dedicated issues?

asbjornu:
Yes, please open one PR per issue. I think both solutions look good, I'm just a bit unsure about the backwards compatibility of them, so let's discuss that in each provided PR.

odalet:
True, both issues may raise compatibility concerns. I'll do just that and open issues when I find some time!

PR #2313, opened 2020-06-07
Modified the VB.NET Template so that GitVersionInformation is in the Global namespace

### LINKED ISSUE #2310: [Improvement] Make VB.NET generated code consistent with C# and F#
Currently, VB.NET's version of `GitVersionInformation` is part of the assembly's default namespace whereas F# and C# versions are part of the `global` namespace

## Detailed Description

Similarly to C# and F# code, VB.NET's `GitVersionInformation` is not explicitely placed in an namespace. However, VB.NET's compiler has different namespaing rules when compared to the other two ones. Given an assembly named `Foo.Bar`, when one omits to specify a type's namespace, it is placed in namespace `Foo.Bar`. This issue is raised to submit the idea that maybe it would be better to explicitely place the `GitVersionInformation` in the `Global` namespace for consistancy purpose.

## Context

It is worth mentioning here that doing so might be considered a **breaking change**: it all depends on whether and how existing users of GitVersion access this generated code: because the class is generated, I suppose any access to it is done through Reflection. Hence it is likely people try first to locate the `GitVersionInformation` class in the assembly. If they do so by explicitely providing the full name of the class (hence including its namespace), then this change will break things for these users.

On the other side, having all generated classes in the same namespace regardless of the source language would make retrieval of the `GitVersionInformation` class identical for all.

_I'll let you decide whether the gained consistency is worth the potential break._

## Possible Implementation

Would this be implemented, it'd be as simple as wrapping the existing [VB.NET template](https://github.com/GitTools/GitVersion/blob/master/src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.vb) within a `Namespace Global`. Someting like that (potentially fixing indentation and corresponding unit tests):

```vb.net
Namespace Global
<Global.System.Runtime.CompilerServices.CompilerGenerated()>
<Global.System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage()>
NotInheritable Class GitVersionInformation
    Private Sub New()
    End Sub
{0}
End Class
End Namespace
```

odalet:
## Description

* Modified the VB template so that the resulting class is part of the global namespace
* Tweaked a bit `GitVersionInfoGenerator.cs` so that it uses different tab lengths for the VB case and the F#/C# cases (this is simply a cosmetic change so that VB output formatting is correct).
* Modified `msbuild-task.md` in the documentation so that:
  * It does not try to find `GitVersionInformation` in the assembly's default namespace but in the global namespace
  * It works with F# generated classes (properties vs fields).

## Related Issue

Fixes #2310

## Motivation and Context

Now, the VB version of `GitVersionInformation` belongs to the `Global` namespace in the same way as its F# and C# counterparts.

## How Has This Been Tested?

* Fixed the units tests
* Double-checked with ILSpy that the resutling class is indeed part of the global namespace. Here is what it looks like when decompiled with ILSpy:

```csharp
// GitVersionInformation
using System.Diagnostics.CodeAnalysis;
using System.Runtime.CompilerServices;

[CompilerGenerated]
[ExcludeFromCodeCoverage]
internal sealed class GitVersionInformation
{
	public static string Major = "1";

	public static string Minor = "2";

	public static string Patch = "3";

	public static string PreReleaseTag = "unstable.4";

       ...

	public static string CommitsSinceVersionSourcePadded = "0005";

	public static string CommitDate = "2014-03-06";

	private GitVersionInformation()
	{
	}
}
```

image

## Checklist:

### DIFF
--- a/docs/input/docs/usage/msbuild-task.md
+++ b/docs/input/docs/usage/msbuild-task.md
@@ -84,34 +84,41 @@ appended to it.
 #### Other injected Variables
 
 All other [variables](../more-info/variables) will be injected into an
-internal static class:
+internal static class part of the global namespace similar to this:
 
 ```csharp
-namespace AssemblyName
+[CompilerGenerated]
+internal static class GitVersionInformation
 {
-    [CompilerGenerated]
-    internal static class GitVersionInformation
-    {
-        public static string Major = "1";
-        public static string Minor = "1";
-        public static string Patch = "0";
-        ...All other variables
-    }
+    public static string Major = "1";
+    public static string Minor = "1";
+    public static string Patch = "0";
+    ...All other variables
 }
 ```
 
 ### Accessing injected Variables
 
+**NB: depending on the source language of the assembly, the injected variables may be exposed either as fields or as properties. The examples below take care of this.**
+
 #### All variables
 
 ```csharp
 var assemblyName = assembly.GetName().Name;
-var gitVersionInformationType = assembly.GetType(assemblyName + ".GitVersionInformation");
+var gitVersionInformationType = assembly.GetType("GitVersionInformation");
 var fields = gitVersionInformationType.GetFields();
 
 foreach (var field in fields)
 {
     Trace.WriteLine(string.Format("{0}: {1}", field.Name, field.GetValue(null)));
+}
+
+// The GitVersionInformation class generated from a F# project exposes properties
+var properties = gitVersionInformationType.GetProperties();
+
+foreach (var property in properties)
+{
+    Trace.WriteLine(string.Format("{0}: {1}", property.Name, property.GetGetMethod(true).Invoke(null, null)));
 }
 ```
 
@@ -119,9 +126,21 @@ foreach (var field in fields)
 
 ```csharp
 var assemblyName = assembly.GetName().Name;
-var gitVersionInformationType = assembly.GetType(assemblyName + ".GitVersionInformation");
+var gitVersionInformationType = assembly.GetType("GitVersionInformation");
 var versionField = gitVersionInformationType.GetField("Major");
-Trace.WriteLine(versionField.GetValue(null));
+if (versionField != null)
+{
+    Trace.WriteLine(versionField.GetValue(null));
+}
+else
+{
+    // The GitVersionInformation class generated from a F# project exposes properties
+    var versionProperty = gitVersionInformationType.GetProperty("Major");
+    if (versionProperty != null)
+    {
+        Trace.WriteLine(versionProperty.GetGetMethod(true).Invoke(null, null));
+    }
+}
 ```
 
 ### Populate some MSBuild properties with version metadata
--- a/src/GitVersionCore.Tests/VersionConverters/Approved/vb/GitVersionInfoGeneratorTests.ShouldCreateFile.approved.txt
+++ b/src/GitVersionCore.Tests/VersionConverters/Approved/vb/GitVersionInfoGeneratorTests.ShouldCreateFile.approved.txt
@@ -8,40 +8,44 @@
 ' </auto-generated>
 '------------------------------------------------------------------------------
 
-<Global.System.Runtime.CompilerServices.CompilerGenerated()>
-<Global.System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage()>
-NotInheritable Class GitVersionInformation
-    Private Sub New()
-    End Sub
-    Public Shared Major As String = "1"
-    Public Shared Minor As String = "2"
-    Public Shared Patch As String = "3"
-    Public Shared PreReleaseTag As String = "unstable.4"
-    Public Shared PreReleaseTagWithDash As String = "-unstable.4"
-    Public Shared PreReleaseLabel As String = "unstable"
-    Public Shared PreReleaseNumber As String = "4"
-    Public Shared WeightedPreReleaseNumber As String = "4"
-    Public Shared BuildMetaData As String = "5"
-    Public Shared BuildMetaDataPadded As String = "0005"
-    Public Shared FullBuildMetaData As String = "5.Branch.feature1.Sha.commitSha"
-    Public Shared MajorMinorPatch As String = "1.2.3"
-    Public Shared SemVer As String = "1.2.3-unstable.4"
-    Public Shared LegacySemVer As String = "1.2.3-unstable4"
-    Public Shared LegacySemVerPadded As String = "1.2.3-unstable0004"
-    Public Shared AssemblySemVer As String = "1.2.3.0"
-    Public Shared AssemblySemFileVer As String = "1.2.3.0"
-    Public Shared FullSemVer As String = "1.2.3-unstable.4+5"
-    Public Shared InformationalVersion As String = "1.2.3-unstable.4+5.Branch.feature1.Sha.commitSha"
-    Public Shared BranchName As String = "feature1"
-    Public Shared EscapedBranchName As String = "feature1"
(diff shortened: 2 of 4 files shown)


docs/input/docs/usage/msbuild-task.md
```diff
@@ -84,44 +84,63 @@ appended to it.
 #### Other injected Variables
 
 All other [variables](../more-info/variables) will be injected into an
-internal static class:
+internal static class part of the global namespace similar to this:
```
asbjornu, line 87: I was just thinking, if `GitversionInformation` is moved to the global namespace, won't this make ILMerging difficult? If several of the merged assemblies are versioned with GitVersion, we would potentially end up with several colliding `GitversionInformation` classes?

odalet: That poses no problem because the generated classes are, and this is very important,  **internal**. An external assembly cannot see these class unless you use reflection, and when using reflection they are resolved with their full name (which happens to be only the class name) against a specific assembly, therefore, no collision.

This trick, I often use with Visual Studio's "shared projects": they are just bunches of code that get "included" into the consuming project. I always make sure classes in these shared projects are internal so that they can be used from any assembly even if these assemblies reference one another.

odalet: And if this were an issue, it would already have shown with C# or F# generated classes that are already part of the global namespace (and internal).

odalet:
Yes I noticed this _dotnet format_ issue (and same problem with PR #2314), but honestly I don't know what to do with this. It also seems to be related to files I didn't change...

docs/input/docs/usage/msbuild-task.md
odalet: Mmmh, maybe I spoke too quickly, I missed the bit about ILMerge... Everything being part of the same aszembly in the end, it may be an issue. I'll give it a try, and sorry if you thought I was teaching you how visibility worked in .NET!

odalet: Well, after a quick googling, it appears this indeeds poses problems: 

> Suppose class A is internal to App, and also to one of the lib (with same namespace). Before merging there will not be any issue. After merging it will become issue to resolve ambiguous reference.

See <https://stackoverflow.com/a/14042227/107552>

asbjornu: I see. That's an issue worth pursuingthen, because it's difficult to resolve for consumers of GitVersion. Ideas?

asbjornu:
If you run `dotnet format` locally, it should fix any problems that may be. I agree the files it complains about are unrelated to your PR, but please just try anyway.

odalet:
Will do this (but not before the WE). Week days are dedicated to software I'm paid for ;) I'll also experiment with merging assemblies.

asbjornu:
> Will do this (but not before the WE). Week days are dedicated to software I'm paid for ;) I'll also experiment with merging assemblies.

👍 

> By the way, I still have one question: do you want me to rebase/squash everything in a single commit that I'd force push or are several commits acceptable?

Several commits are fine, but if you can avoid merge commits and do `rebase` instead (to stay up to date with `master`), that would be great.

odalet:
Got it!

docs/input/docs/usage/msbuild-task.md
odalet: I experimented with IL packing using [ILRepack](https://github.com/ravibpatel/ILRepack.Lib.MSBuild.Task). The tool is rather clever: upon detection of the same type in several input assemblies, it keeps one as is (the first iot encounters I suppose) then renames the other ones:

image

Then I suppose, at least with this tool, your concern with type collision is mitigated. 
I may experiment too with [ILMerge](https://github.com/dotnet/ILMerge) or [Fody](https://github.com/Fody/Fody).

PS: As a refinement, I also looked for a way to tell **ILRepack** to favor a specific source assembly when determining what type to keep. My use case would be: merge plenty of .DLL along the main executable, and be sure the `GitVersionInformation` that remains accessible is the one in the exe. Well, I managed to achieve this (though Idon't know how robust it is) by playing with the order in which assemblies to merge are listed: the last assembly wins; therefore having the main executable merged last seems to guarantee its `GitVersionInformation` is the one that keeps its name.

odalet:
Rebasing atop master had the `dotnet format` check pass although, now the Azure DevOps macOS tests fail or rather the test runner crashes...

docs/input/docs/usage/msbuild-task.md
odalet: [ILMerge](https://github.com/dotnet/ILMerge) is similar to **ILRepack** given you pass it the correct options. Concerning the type collisions, it is slightly better as the types are renamed according to a preditable scheme:

image

odalet: And with [Fody + Costura](https://github.com/Fody/Costura), everything just works out of the box. This is because, the principle is radically different. The assembies' IL is not merged, instead the dependent assemblies are embedded as resources in the main assembly, then extracted and loaded at runtime (during module initialization). Thus, the assemblies never cease to exist as such and type resolution just works.

odalet: If you are interested in witnessing the tests I did, everything is in this repository: <https://github.com/odalet/GitVersionTests>

odalet: And now for ideas on how this whole "code generation" thing could be tweaked:

* The generated class could be put in a namespace or have a name that will guarantee non-collision. To guarantee this, the class name should be based on the assembly name (and not on some 'default' or root namespace). The class could be inside a namespace named as the assembly or be an inner class inside a class named like the assembly or be a concatenation of assembly name and `GitVersionInformation`...

* A property / command line argument could be provided to let the user choose the name of the generated class.

* I recently stumbled upon the [GitInfo](https://github.com/kzu/GitInfo). Its purpose is similar to **GitVersion**'s with the following differences:
  * Completely MSBuild-based, no C# code at all
  * Seems to have far less feature. It seems to simply expose Git information and does not attempt clever inference of versions.
  * There is however one feature that may be interesting here:

It, too, generates code containing version information. However:
* The version information is made available through constants, not properties or fields.
* The file containing this information is generated soon enough for it to be available at "design-time": it can be used without resorting to reflection.

The advantage of this strategy is that because constants are inlined in the calling code, the original type that exposes them is useless at runtime and therefore, collisions on this type are not anymore an issue. And being able to use the version information without reflection also seems more user-friendly.

So, mixing some of these ideas may be the way to go. However, in any case, it is quite some work, and most probably breaking. In the meantime, my experiments with a selection of assembly merging tools show that the collision problem can be mitigated. 

It's now up to you to reflect upon all this and decide what or what not to do :)

Hope this helped!

asbjornu: Constants are a great idea for how we can implement this in a new major version of GitVersion! Whether we'll do this for v6 or not is mainly dependent on timing; if someone submits a pull request implementing this before we go live with v6, it will be released with v6. If not, we would have to postpone it to v7.

asbjornu:
Thank you so much for your contributions, @odalet! 🙏

docs/input/docs/usage/msbuild-task.md
odalet: Generating constants should be rather straightforward. The MSBuild part that generates this soon so that the user can call into it may be a little trickier (I'm not very proficient at MSBuild ;)).

I'll probably come up with an issue that will serve as a discussion and specification of what can be done.

PR #2328, opened 2020-06-15
GH-2095 - Target .net48 instead of .net472

### LINKED ISSUE #2095: Target .net48 instead of .net472

arturcic:
Closes #2095

### DIFF
--- a/build/pack.cake
+++ b/build/pack.cake
@@ -4,7 +4,7 @@ Task("Pack-Prepare")
 {
     PackPrepareNative(Context, parameters);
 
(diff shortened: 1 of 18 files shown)


asbjornu:
Should this be labeled with `Milestone: 5.5`?

arturcic:
I prefer to put the milestone and the label on the issue and not on the PR that implements/fixes the issue, as usually the issue has the reason/ discussions

asbjornu:
Ah, that makes sense.

PR #2332, opened 2020-06-20
#if'ed the definition of ExcludeFromCodeCoverageAttribute depending o…

### LINKED ISSUE #2330: [Bug] GitVersionInformation.g.cs(12,42): error CS0234 'ExcludeFromCodeCoverageAttribute' does not exist
**Describe the bug**

Latest version 5.3.6 fails in `.Net Standard 2.0` builds.

## Expected Behavior

Expected build to not fail. Version 5.3.5 builds all frameworks correctly.

## Actual Behavior

During build of a dll in `.Net Standard 2.0` I am getting following error:

```bash
  Unity.Abstractions -> D:\a\abstractions\abstractions\src\bin\Release\netstandard2.0\Unity.Abstractions.dll
obj\Release\netstandard1.6\GitVersionInformation.g.cs(12,42): error CS0234: The type or namespace name 'ExcludeFromCodeCoverageAttribute' does not exist in the namespace 'System.Diagnostics.CodeAnalysis' (are you missing an assembly reference?) [D:\a\abstractions\abstractions\src\Unity.Abstractions.csproj]
```

The generated file fails on line 12:

```C#
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
//     GitVersion
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------

[global::System.Runtime.CompilerServices.CompilerGenerated]
[global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]   <-- This line fails in NETSTANDARD2_0
static class GitVersionInformation
{
    public static string Major = "6";
    public static string Minor = "0";
    public static string Patch = "0";
. . .
}
```

## Possible Fix

Reverted to previous version for now

## Steps to Reproduce

For more information please follow [this link](https://github.com/unitycontainer/abstractions/runs/778004700?check_suite_focus=true) and examine `Build` step.

## Context

I am using GitVerion tool in CI builds. The library is added to projects during CI steps

## Your Environment

<TargetFrameworks>netstandard2.1;netstandard2.0;netstandard1.6;netcoreapp3.0;netcoreapp2.0;netcoreapp1.0;net48;net47;net46;net45;net40</TargetFrameworks>

- Version Used: 5.3.6
- Link to your project: https://github.com/unitycontainer/abstractions
- Link to your CI build (if appropriate): https://github.com/unitycontainer/abstractions/blob/release/6.0.0/.github/workflows/cd-flow.yml

odalet:
…n the target framework

The `ExcludeFromCodeCoverageAttribute` is not supported on some older .NET framework targets:
* Classic framework < v4.0
* .NET Core < v2.0
* .NET Standard < 2.0

In order for projects targeting some of these frameworks to be able to use GitVersion,  `ExcludeFromCodeCoverageAttribute` needs to be defined conditionally when targeting these frameworks.

This is exactly what this PR does. 

This PR fixes #2330.

I used a side-project to ensure my changes were correct (see https://github.com/odalet/GitVersionTests). Unfortunately I wasn't able to enforce these tests with proper unit tests (I think it is not a trivial task to be able to ensure correct compilation of projects across all framework x language combinations).

### DIFF
--- a/src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.cs
+++ b/src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.cs
@@ -8,6 +8,22 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
+#if NET20 || NET35 || NETCOREAPP1_0 || NETCOREAPP1_1 || NETSTANDARD1_0 || NETSTANDARD1_1 || NETSTANDARD1_2 || NETSTANDARD1_3 || NETSTANDARD1_4 || NETSTANDARD1_5 || NETSTANDARD1_6
+namespace System.Diagnostics.CodeAnalysis
+{{
+    [global::System.AttributeUsage(
+        global::System.AttributeTargets.Assembly |
+        global::System.AttributeTargets.Class |
+        global::System.AttributeTargets.Struct |
+        global::System.AttributeTargets.Constructor |
+        global::System.AttributeTargets.Method |
+        global::System.AttributeTargets.Property |
+        global::System.AttributeTargets.Event,
+        Inherited = false, AllowMultiple = false)]
+    internal sealed class ExcludeFromCodeCoverageAttribute : global::System.Attribute {{ }}
+}}
+#endif
+
 [global::System.Runtime.CompilerServices.CompilerGenerated]
 [global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
 static class GitVersionInformation
--- a/src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.fs
+++ b/src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.fs
@@ -8,6 +8,22 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
+#if NET20 || NET35 || NETCOREAPP1_0 || NETCOREAPP1_1 || NETSTANDARD1_0 || NETSTANDARD1_1 || NETSTANDARD1_2 || NETSTANDARD1_3 || NETSTANDARD1_4 || NETSTANDARD1_5 || NETSTANDARD1_6
+namespace System.Diagnostics.CodeAnalysis
+
+[<Sealed>]
+[<global.System.AttributeUsage(
+    global.System.AttributeTargets.Assembly ||| 
+    global.System.AttributeTargets.Class ||| 
+    global.System.AttributeTargets.Struct ||| 
+    global.System.AttributeTargets.Constructor ||| 
+    global.System.AttributeTargets.Method ||| 
+    global.System.AttributeTargets.Property ||| 
+    global.System.AttributeTargets.Event, 
+    Inherited = false, AllowMultiple = false)>]
+type ExcludeFromCodeCoverageAttribute() = inherit global.System.Attribute()
+#endif
+
 namespace global
 
 [<AbstractClass; Sealed>]
--- a/src/GitVersionCore.Tests/VersionConverters/Approved/cs/GitVersionInfoGeneratorTests.ShouldCreateFile.approved.txt
+++ b/src/GitVersionCore.Tests/VersionConverters/Approved/cs/GitVersionInfoGeneratorTests.ShouldCreateFile.approved.txt
@@ -8,6 +8,22 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
+#if NET20 || NET35 || NETCOREAPP1_0 || NETCOREAPP1_1 || NETSTANDARD1_0 || NETSTANDARD1_1 || NETSTANDARD1_2 || NETSTANDARD1_3 || NETSTANDARD1_4 || NETSTANDARD1_5 || NETSTANDARD1_6
+namespace System.Diagnostics.CodeAnalysis
+{
+    [global::System.AttributeUsage(
+        global::System.AttributeTargets.Assembly |
+        global::System.AttributeTargets.Class |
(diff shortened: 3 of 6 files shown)


odalet:
There's still this `dotnet format` issue, but this time, it complains about `AddFormats/GitVersionInformation.cs` which is the template that generates properties. It says there are whitespace issues in locations that either look ok or are not whitespaces at all... 

Would the fact that this is not a real C# file bother the format tool?

Maybe the templates should be excluded from the analysis?

What's weird too is that `dotnet format` does not complain on my machine (Windows)

asbjornu:
> Maybe the templates should be excluded from the analysis?

Yep, that sounds like a good idea. I don't know `dotnet format` well enough to fix that just yet, though.

src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.cs
```diff
@@ -8,6 +8,30 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
+#if NET20 || NET35
+#define SHOULD_DEFINE_EXCLUDE_FROM_CODE_COVERAGE
+#elif NETCOREAPP1_0 || NETCOREAPP1_1
+#define SHOULD_DEFINE_EXCLUDE_FROM_CODE_COVERAGE
+#elif NETSTANDARD1_0 || NETSTANDARD1_1 || NETSTANDARD1_2 || NETSTANDARD1_3 || NETSTANDARD1_4 || NETSTANDARD1_5 || NETSTANDARD1_6
+#define SHOULD_DEFINE_EXCLUDE_FROM_CODE_COVERAGE
+#endif
```
asbjornu, line 17: Why all these `#elif`'s? Couldn't one big `#if` cover all targets?

src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.fs
```diff
@@ -8,6 +8,22 @@
 // </auto-generated>
 //------------------------------------------------------------------------------
 
+#if NET20 || NET35 || NETCOREAPP1_0 || NETCOREAPP1_1 || NETSTANDARD1_0 || NETSTANDARD1_1 || NETSTANDARD1_2 || NETSTANDARD1_3 || NETSTANDARD1_4 || NETSTANDARD1_5 || NETSTANDARD1_6
```
asbjornu, line 11: Shouldn't `SHOULD_DEFINE_EXCLUDE_FROM_CODE_COVERAGE` be defined in F# as well?

src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.cs
odalet: That's just an attempt to split the very long #if line... I did something similar in the VB file, but couldn't do it in F#.

src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.fs
odalet: There's no preprocessor instruction in F# that allows defining a constant (when you need one you have to pass it to the compiler with a `-d` argument). And the `SHOULD_DEFINE_EXCLUDE_FROM_CODE_COVERAGE` constant has no special meaning, it's just used to split the `#if` as I said above.

I agree this is not consistant. If you prefer, I can have everything look more like the F# version.

src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.cs
asbjornu: I see.

src/GitVersionCore/VersionConverters/GitVersionInfo/Templates/GitVersionInformation.fs
asbjornu: I would prefer consistency. Hopefully, we can replace these `#if` statements with a check before the files are generated in the first place.

odalet: >Hopefully, we can replace these #if statements with a check before the files are generated in the first place.

I suppose you are referring to a solution similar to what was suggested in the [issue](https://github.com/GitTools/GitVersion/issues/2330#issuecomment-645669350): having the support of the attribute detected by MSBuild? That would probably work but:
* I'm not that familiar (yet) with how MSBuild code works in GitVersion,
* Will you ever want to support generation of `GitVersionInformation` class from the command-line utility? If so, I think we can't rely on constants defined in a MSBuild file that would not exist in this case.

Anyway, I don't have clear ideas to how we can reliably get rid of this ugly `#if`, but in the meantime and if you're ok with this. I'll simply have everything look more like the F# version.

asbjornu: > I suppose you are referring to a solution similar to what was suggested in the [issue](https://github.com/GitTools/GitVersion/issues/2330#issuecomment-645669350): having the support of the attribute detected by MSBuild? 

Correct. I agree that the solution is difficult, though, which is why I think we should go ahead with this `#if`-based solution first.

>Anyway, I don't have clear ideas to how we can reliably get rid of this ugly #if, but in the meantime and if you're ok with this. I'll simply have everything look more like the F# version.

👍

asbjornu:
Brilliant, thank you so much @odalet! 🙏

Issue #1448, opened 2018-07-30
GitVersionTarget build is very slow, Normalisation step with many branches


danijam:
Hi, So I'm very keen to use GitVersion in our groups repo to replace some versioning scripts. When I used GitVerion in a test repo to trial it and learn how it works everything was great. Now I've added it into our teams repo I'm starting to encounter performance issues.

Namely our builds without GitVersion (master) take ~1 min with GitVersion (my topic branch) it takes ~3 mins. I was expecting maybe a small overhead but this is took much to accept. The additional time is on the MSBUILD step.

We use VSTS run the build on our established agent build servers. Our solution contains 20 projects each with the GitVersion Targets NuGet Package now installed. When I inspect the logs emitted by the build (MSBuild) I can see GitVersion is spewing lots of INFO events along these lines:

`Begin: Normalizing git directory for branch 'refs/heads/JD/GitVersion' (Count 56)`

`Creating local branch from remote tracking 'refs/remotes/origin/[DevIntial]/[BranchName' (Count 202)`

`Skipping update of 'refs/remotes/origin/[DevIntial]/[BranchName]' as it already matches the remote ref. (Count 8,327)`

`Begin: Finding merge base between '[DevIntial]/[BranchName]' and '[DevIntial]/[BranchName]' (Count 1,248)`

`End: Finding merge base between '[DevIntial]/[BranchName]' and '[DevIntial]/[BranchName]' (Count 1,248)`

 
#1327 Might be similar

I'd like to understand what is going on and why it seems to be spending so much time iterating over all the remote branches if indeed that is where the time is being lost.

I noticed this:
`No branch configuration found for branch JD/GitVersion, falling back to default configuration`

is this why it is spending so much time doing the merge base operations?

I've attached our GitVerion.yml for reference
[GitVersion.txt](https://github.com/GitTools/GitVersion/files/2241824/GitVersion.txt)

danijam:
So I've updated the GitVersion.yml to include a branch definition for my topic branch. This has removed the Merge base actions as it know I guess nows how to treat this build as it knows what type of branch it is. 

I'm still a fair way off the baseline build time without GitVersion being installed to the projects. Still have a rather large log filled with the normalising messages, creating local branch, and Skipping update of message. What are these doing?

danijam:
Also noticing a fair number of random:

[...]

PR #2333, opened 2020-06-22
A lock file implementation applied to project GitVersionCore for resolving #1031

teroneko:
A lock file implementation applied to project GitVersionCore for resolving #1031.

## Description
Inject singleton service IFileLock to ServiceCollection. Lock gets disposed when the ServiceProvider is getting disposed too.

## Related Issue
See #1031.

## Motivation and Context
There were many failing builds in CI environments due to parallelism.

## How Has This Been Tested?
Not yet tested...

## Checklist:

- [x] Receive opinions of maintainers and people of interests.


teroneko:
I ask you, which timeout is necessary. I just took 15 seconds. I am worried about the fact, that the assemblies are often cached, but the servicecollection filled and serviceprovider built is always happening once in a task, so it will always dispose correct at end of any task in GitVersionTask.MSBuild assembly.

src/GitVersionCore/Helpers/LockFile.cs
```diff
@@ -0,0 +1,208 @@
...
+            FileAccess fileAccess = DefaultFileAccess, FileShare fileShare = DefaultFileShare)
+        {
+            filePath = filePath ?? throw new ArgumentNullException(nameof(filePath));
+
+            try
+            {
+                fileStream = File.Open(filePath, fileMode, fileAccess, fileShare);
+                return true;
+            }
+            catch (Exception error) when (error.GetType() == typeof(IOException))
```
asbjornu, line 58: Can't this simply be written as:

```suggestion
            catch (IOException exception)
```

src/GitVersionCore/Helpers/LockFile.cs
```diff
@@ -0,0 +1,208 @@
...
+namespace GitVersion.Helpers
+{
+    /// <summary>
+    /// This helper class can lock files.
+    /// </summary>
+    public static class LockFile
+    {
+        public const FileMode DefaultFileMode = FileMode.OpenOrCreate;
+        public const FileAccess DefaultFileAccess = FileAccess.ReadWrite;
+        public const FileShare DefaultFileShare = FileShare.None;
```
asbjornu, line 36: Is it `FileShare.None` that is going to ensure that only one process is able to access the specified `filePath`? Can this be guaranteed across all file systems or is this file system dependent?

src/GitVersionCore/GitVersionCoreDefaults.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace GitVersion
+{
+    public static class GitVersionCoreDefaults
+    {
+        public const int LockTimeoutInMilliseconds = 1000 * 15;
+        public const string LockFileNameWithExtensions = "GitVersion.lock";
```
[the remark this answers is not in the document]
teroneko: Right, that is comprehensible.

src/GitVersionCore/Helpers/LockFile.cs
```diff
@@ -0,0 +1,208 @@
...
+using System;
+using System.IO;
+using System.Threading;
+
+namespace GitVersion.Helpers
+{
+    /// <summary>
+    /// This helper class can lock files.
+    /// </summary>
+    public static class LockFile
```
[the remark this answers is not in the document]
teroneko: This is API, it will be turned private and wrapped by FileLocker at your suggestion - ok? Like told it is only API it can be static, outside it won't be viewable, but then the FileLocker.

src/GitVersionCore/Helpers/LockFile.cs
```diff
@@ -0,0 +1,208 @@
...
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR 
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER 
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN 
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+using System;
+using System.IO;
```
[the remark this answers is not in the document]
teroneko: Yes that is advisable.

src/GitVersionCore/Helpers/LockFile.cs
teroneko: No this is a crucial part of the API. The explanation is, that when you open a FileStream that failes because it was locked, the assembly System.Private.CoreLib will throw a non-dervied System.IOException. Any other kind of exceptions also these who derives from IOException are forwarded to consumer. This is the only way to get behaviour "When you are was locked" then return false, but when you failed for any other reason, then let consumer handle this [error](https://docs.microsoft.com/en-US/dotnet/api/system.io.file.open?view=netcore-3.1).

src/GitVersionCore/Helpers/LockFile.cs
teroneko: Yes right. that's the key. So there you have your global lock over all processes . 🙂

teroneko:
At the evening I can spend some time I hope. Time is rare.

About your time consideration. So it could happen then, that one process is staling then for about 1 minute for wating to acquire the lock until it continues its IServiceCollection build process, otherwise a TimeoutException will be thrown.

src/GitVersionCore/Helpers/LockFile.cs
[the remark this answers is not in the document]
teroneko: > > I ask you, which timeout is necessary. I just took 15 seconds.
> 
> We have reports that GitVersion is taking more than a minute to complete, so I'm not sure 15 seconds is enough. [#1448 (comment)](https://github.com/GitTools/GitVersion/issues/1448#issuecomment-408936340)

See comment above.

> > I am worried about the fact, that the assemblies are often cached, but the servicecollection filled and serviceprovider built is always happening once in a task, so it will always dispose correct at end of any task in GitVersionTask.MSBuild assembly.
> 
> If I understand you correctly, that's the source of the problem and why this is so hard to fix: The lock needs to exist and work across processes. Locking inside the GitVersion process won't help, since we have many instances of GitVersion running in parallel in different processes.

The point is we have to ensure, that each Task (from MSBuild) in GitVersionTask.MSbuild assembly is correctly disposing the singleton IFileLock. (in this current state of implementation - changed later). So as you can see is IServiceProvider not implementing IDisposable, but ServiceProvider in assembly Microsoft.Extensions.DependencyInjection is doing so, we have to return a derived IServiceProvider implementation that implements IDisposable and write a wrapper like I did for IFileLock. With such exposed custom IServiceProvider we can now manually dispose [Microsoft.Extensions.DependencyInjection.ServiceProvider](https://docs.microsoft.com/de-de/dotnet/api/microsoft.extensions.dependencyinjection.servicecollectioncontainerbuilderextensions.buildserviceprovider?view=dotnet-plat-ext-3.1).

teroneko:
> > I ask you, which timeout is necessary. I just took 15 seconds.
> 
> We have reports that GitVersion is taking more than a minute to complete, so I'm not sure 15 seconds is enough. [#1448 (comment)](https://github.com/GitTools/GitVersion/issues/1448#issuecomment-408936340)

See comment above.

> > I am worried about the fact, that the assemblies are often cached, but the servicecollection filled and serviceprovider built is always happening once in a task, so it will always dispose correct at end of any task in GitVersionTask.MSBuild assembly.
> 
> If I understand you correctly, that's the source of the problem and why this is so hard to fix: The lock needs to exist and work across processes. Locking inside the GitVersion process won't help, since we have many instances of GitVersion running in parallel in different processes.

The point is we have to ensure, that each Task (from MSBuild) in GitVersionTask.MSbuild assembly is correctly disposing the singleton IFileLock. (in this current state of implementation - changed later). So as you can see is IServiceProvider not implementing IDisposable, but ServiceProvider in assembly Microsoft.Extensions.DependencyInjection is doing so, we have to return a derived IServiceProvider implementation that implements IDisposable and write a wrapper like I did for IFileLock. With such exposed custom IServiceProvider we can now manually dispose [Microsoft.Extensions.DependencyInjection.ServiceProvider](https://docs.microsoft.com/de-de/dotnet/api/microsoft.extensions.dependencyinjection.servicecollectioncontainerbuilderextensions.buildserviceprovider?view=dotnet-plat-ext-3.1).

src/GitVersionCore/Helpers/LockFile.cs
teroneko: The API needs to add `FileStream.Lock(0,0);`

See references for unix lock support:
https://github.com/dotnet/coreclr/pull/8233/
https://stackoverflow.com/questions/35444470/how-to-lock-unlock-a-file-across-process

teroneko: The API needs to add FileStream.Lock(0,0);

See references for unix lock support:
dotnet/coreclr#8233
https://stackoverflow.com/questions/35444470/how-to-lock-unlock-a-file-across-process

asbjornu:
>The point is we have to ensure, that each Task (from MSBuild) in GitVersionTask.MSbuild assembly is correctly disposing the singleton IFileLock. (in this current state of implementation - changed later). We should not rely on it, that Garbage Colelctor is doing it as some time, when built ServiceProvider is out of scope beside the fact, that the assembly can be cached and its automatical dispose is not guaranteed. 

Yes, I agree. We should explicitly dispose as soon as we are done indirectly poking the file system with LibGit2Sharp.

src/GitVersionCore.Tests/Helpers/TestFileSystem.cs
```diff
@@ -78,6 +78,11 @@ public IEnumerable<string> DirectoryEnumerateFiles(string directory, string sear
             throw new NotImplementedException();
         }
 
+        public FileStream Open(string path, FileMode mode, FileAccess access, FileShare share)
+        {
+            return new FileStream(path, mode, access, share);
```
teroneko, line 83: TestStream is deriving from Stream, but need Lock implementation of FileStream for LockFileApi. So when needed, another Stream wrapper have to be written for FileStream.

src/GitVersionCore/Core/Abstractions/IDisposableServiceProvider.cs
```diff
@@ -0,0 +1,7 @@
+using System;
+
+namespace GitVersion.Core.Abstractions
+{
+    public interface IDisposableServiceProvider : IServiceProvider, IDisposable
```
arturcic, line 5: I think ServiceProvider class that implements IServiceProvider is already implementing IDisposable. So I would not add an additional interface for that

src/GitVersionCore/FileLocking/Abstractions/IFileLocker.cs
```diff
@@ -0,0 +1,29 @@
+/*
+MIT License
+
+Copyright 2020 Teroneko
```
arturcic, line 4: @asbjornu not sure we have agreed on this, but do we allow to add custmo copyright to our code base?

src/GitVersionCore/Core/Abstractions/IDisposableServiceProvider.cs
teroneko: Why do you not confirm yourself? :) You can see [here in the documentation](https://docs.microsoft.com/en-US/dotnet/api/system.iserviceprovider?view=netcore-3.1) of Microsoft that IServiceProvider is not implementing it. When IServiceProvider is not implementing IDisposable, what's your approach? Return ServiceProvider directly?

asbjornu: `IServiceProvider` does not inherit `IDisposable`, correct. But [the `ServiceProvider` class implements `IDisposable`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.dependencyinjection.serviceprovider):

```c#
public sealed class ServiceProvider : IAsyncDisposable, IDisposable, IServiceProvider
```

teroneko: Ok I might be right, when I say, that it is okay for you to return only IDisposable, right?

teroneko: Right. So returning ServiceProvider directly should be fine.

src/GitVersionCore/FileLocking/Abstractions/IFileLocker.cs
teroneko: As I do agree with your license (that's why I enjoy to contribute to your library) the message I inserted could be shorten to:

```csharp
/* Copyright (c) 2020 Teroneko
 * Licensed under the MIT License. See LICENSE in the project root for license information.
 */
```

As I do claim copyright for this either way, it can be advisable to note its copright in the few files the copyright affects.

I really try to understand this nature of licensing and copyrights, but it is challenging me. Please let me know how you want to handle it. :)

asbjornu: We don't have a strict copyright or licensing policy in GitVersion. It would probably have been clearer if we had a [CLA](https://en.wikipedia.org/wiki/Contributor_License_Agreement), but even without making this explicit, the copyright terms are such in most countries that copyright doesn't have to be explicitly stated – it is implicit through the publicaiton of copyrighted works.

So by you contributing these source files to GitVersion, you have copyright over them. But you also share this copyright with GitVersion, so we can both do whatever we wish with the files after they are contributed. We can rework them and change them to what's best for GitVersion, but you're also free to take the files and use them in another project of yours without asking anyone for permission.

So I don't think we need any explicit copyright or license information in our source files. No other files contain it so far and I think we should stay consistent. Thus I would kindly ask you to remove this, @teroneko.

src/GitVersionCore/Core/Abstractions/IDisposableServiceProvider.cs
asbjornu: If that means the `IDisposableServiceProvider` will be deleted, I think we agree, @teroneko! 😃

src/GitVersionCore/FileLocking/Abstractions/IFileLock.cs
```diff
@@ -0,0 +1,7 @@
+using System;
+
+namespace GitVersion.FileLocking.Abstractions
```
asbjornu, line 3: We aren't using `Abstractions` anywhere else in the codebase and I don't think they provide any useful information. Please remove.

```suggestion
namespace GitVersion.FileLocking
```

src/GitVersionCore/FileLocking/Abstractions/IFileLocker.cs
```diff
@@ -0,0 +1,29 @@
...
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR 
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER 
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN 
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+namespace GitVersion.FileLocking.Abstractions
```
asbjornu, line 23:
```suggestion
namespace GitVersion.FileLocking
```

src/GitVersionCore/FileLocking/FileLockContext.cs
```diff
@@ -0,0 +1,96 @@
...
+            ErrorUnlockDone = errorUnlockDone ?? throw new ArgumentNullException(nameof(errorUnlockDone));
+        }
+
+        public void DecreaseLockUse(bool decreaseToZero, string lockId)
+        {
+            if (FileStream == null) {
+                throw new InvalidOperationException("You cannot decrease lock use when no file stream has been assgined.");
+            }
+
+            var decreaseLockUseLocker = this.decreaseLockUseLocker;
```
asbjornu, line 68: This assignment seems unnecessary and makes me think twice about what the below `lock` statement does. Does it lock the class field or the local variable?

src/GitVersionCore/FileLocking/FileLockContext.cs
```diff
@@ -0,0 +1,96 @@
...
+
+            // Why surround by lock?
+            // There is a race condition, when number of file lock uses
+            // is decrased to 0. It may not have invalidated the file
+            // stream yet. Now it can happen that the number of file lock
+            // uses is increased to 1 due to file lock, but right after another
+            // file unlock is about to decrease the number again to 0.
+            // There is the possiblity that the actual file lock gets released
+            // two times accidentally.
+            lock (decreaseLockUseLocker)
```
asbjornu, line 81:
```suggestion
            lock (this.decreaseLockUseLocker)
```

src/GitVersionCore/FileLocking/FileLockContextExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+The above copyright notice and this permission notice shall be included in all 
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR 
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER 
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN 
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
```
asbjornu, line 22:
```suggestion
```

src/GitVersionCore/FileLocking/FileLockContextExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace GitVersion.FileLocking
+{
+
+#nullable enable
+
+    internal static class FileLockContextExtensions
+    {
+        public static bool IsErroneous(this FileLockContext? fileLockContext)
```
asbjornu, line 30: Why isn't this just an instance method on the `FileLockContext` class?

src/GitVersionCore/FileLocking/FileLockUse.cs
```diff
@@ -0,0 +1,58 @@
...
+The above copyright notice and this permission notice shall be included in all 
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR 
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER 
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN 
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
```
asbjornu, line 22:
```suggestion
```

src/GitVersionCore/FileLocking/FileLockUse.cs
```diff
@@ -0,0 +1,58 @@
...
+using System;
+using System.ComponentModel;
+using System.IO;
+
+namespace GitVersion.FileLocking
+{
+
+#nullable enable
+
+    public struct FileLockUse : IDisposable
```
asbjornu, line 32: I don't quite understand what `FileLockUse` does by its name. Can you please explain so we can perhaps figure out a more self-descriptive and intuitive name for it?

src/GitVersionCore/FileLocking/ILockFileApi.cs
```diff
@@ -0,0 +1,9 @@
+using System.IO;
+
+namespace GitVersion.FileLocking
+{
+    public interface ILockFileApi
```
asbjornu, line 5: I don't quite understand the distinction between `IFileLocker` and`ILockFileApi`. Why are they separated? I'm also not a fan of using `Api` in the name of an interface. All C# interfaces (and public classes) are [application programming interfaces](https://en.wikipedia.org/wiki/Application_programming_interface). :)

src/GitVersionCore/FileLocking/LockFileApi.cs
```diff
@@ -0,0 +1,231 @@
...
+The above copyright notice and this permission notice shall be included in all 
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR 
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR 
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER 
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN 
+CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
```
asbjornu, line 22:
```suggestion
```

src/GitVersionCore/Model/FileLock.cs
```diff
@@ -0,0 +1,25 @@
+using GitVersion.FileLocking;
+using GitVersion.FileLocking.Abstractions;
+using System;
+using System.IO;
+
+namespace GitVersion.Helpers
```
asbjornu, line 6: [Please avoid using names such as `Helpers`](https://softwareengineering.stackexchange.com/a/247277/20746).

```suggestion
namespace GitVersion
```

src/GitVersionCore/Model/FileLock.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class FileLock : IFileLock
```
asbjornu, line 8: Why isn't `FileLock` placed in the `FileLocking` namespace?

src/GitVersionCore/FileLocking/Abstractions/IFileLocker.cs
teroneko: No problem. I was not sure about it. So I added it without knowledge. Now that you meantion it I know that copyright is always implicit. Thanks for your elaboration.

src/GitVersionCore/FileLocking/FileLockUse.cs
teroneko: It describes the use of a file lock. As one lock is shared by many I thought the term "use" of one file lock would be somewhat self-explanatory.

src/GitVersionCore/FileLocking/ILockFileApi.cs
teroneko: I decided to implement this in a composition flavor. When the "API" is about to have the ability to lock anything the FileLocker can lock a specific file that has the ability to lock one file but being callable by many concurrent calls to FileLocker.Lock(). So the file gets unlocked first when all FileLockUse instances are got disposed. So later you can graduate the use of the instance of FileLocker to your behave when you decide to go away from one global lock.

src/GitVersionCore/Model/FileLock.cs
teroneko: Missed it, thanks.

src/GitVersionCore/FileLocking/FileLockContext.cs
teroneko: It does lock the Unlock function. You will see that there is another lock in FileLocker in Unlock. These calls can interfer each other.

src/GitVersionCore/FileLocking/FileLockContextExtensions.cs
teroneko: Because the parameter "fileLockContext" can be null. With this "trick" you can shift the null check to this method. That's simply a shortcut.

teroneko:
When I get some free time I would set up your prepared repository. Currently I am just too busy.
The implementation itself I've tested with Xunit by myself. Perhaps they can be integrated here in GitVersion too?
So another the question is how to test GitVersionTask. The sub-projects should be in not that small. Perhaps we can take some big open source projects and include them just. :)

DerAlbertCom:
Is there way to easy get a Version of the BuildTask with this fix in? (like a alpha nuget package) so that we can test this in the wild?

teroneko:
@DerAlbertCom Unfortunately I have no time because of business work. I can repeat the words of @asbjornu that you are free to assist in this pull request. But I'm looking forward to spend time as soon as I can. I'd like to see this feature too. 

If you are really in need of a solution right now, you can use [my experimental workaround](https://www.nuget.org/packages/Teronis.MSBuild.Packaging.GitVersionCache). Use it only when you need pacakge level versioning (Nuspec file & package name), but not compile level versioning (Assembly file). It does mimic GitVersionTask behaviour and supports concurrency with the above implementation. It does so by wrapping GitVersion.exe and locking file at path "<repository-root>/.gvc/index" when talking to GitVersion.exe.
The reason why I implemented this is, that I could not wait for an official fix that long, because my project is dependant of private projects.

teroneko:
Just want to leave a message here. Can you concretely tell me what you want me to do to bring progress into this? Every time I think to this topic I push it away again because I don't know where to start or what to make. I have the feeling being left alone to this. In my current view I would remove the -Api suffix and that's it. The implementation should work. Do you want me to write a test for the implementation? You did not answer me.
@asbjornu You telled something from it is a big thing that needs big consideration and that is a start in the right direction. Can you tell me, what's the right direction now?

asbjornu:
@teroneko, I'm sorry to see this closed. Would a viable road ahead be to add this feature behind a configuration option so we can ship it for testing? Then if people report back that this actually fixes the problem, we can consider removing the configuartion in v6 and just make this the default.

riksking:
Hi all.
image

Fix this, please, and give a pre-release for test.
This is a very unnerving mistake and I want to get rid of it as soon as possible.

Tognolo:
No stale please. Any news?

arturcic:
Closing this one in favor of https://github.com/GitTools/GitVersion/pull/2410

Issue #2339, opened 2020-06-24
[Bug] assembly-informational-format doesn't support special or international characters


LazaroOnline:
The `assembly-informational-format` doesn't support special or international characters.

## Expected Behavior
Product version should just be the text from the template as-is, without replacing special characters, allowing things like:
`Some special characters: +-=#![]^&@$%:<>//\\-çñáÁ统`

*If there is any particular scenario that has problems with letting special characters in that attribute then I would like to have a configuration parameter to enable/disable special character replacement, but if there is no such case then just keep them and don't replace them.

## Actual Behavior
It replaces special characters and international unicode characters by a dash "-".
Using this config: `assembly-informational-format: 'Some special characters: +-=#![]^&@$%:<>//\\-çñáÁ统'`
Results in:
Product version: `Some-special-characters--+------------------------`
Losing all information from special chars.

## Steps to Reproduce
Create a dotnet project, add GitVersion.Task NuGet and add a GitVersion.yml config (UTF-8 encoded) file with the following text:
```yml
assembly-informational-format: 'Some special characters: +-=#![]^&@$%:<>//\\-çñáÁ统'
```

### DotNet does allow any free unicode text
You can confirm that it is not a dotnet limitation by adding special characters to the attribute:
For Full dotnet, in AssemblyInfo.cs:
```c#
[assembly: AssemblyInformationalVersion("Some special characters: +-=#![]^&@$%:<>//\\-çñáÁ统")]
```
Results in: `Some special characters: +-=#![]^&@$%:<>//\-çñáÁ统` as expected.

For dotnet Core, in the csproj file:
```xml
<InformationalVersion>Some special characters: +-=#![]^@$%://\\-çñáÁ统</InformationalVersion>
```
Results in: ` Some special characters: +-=#![]^@$%://\\-çñáÁ统` as expected.

## Context
We need to add custom free text version information to the assembly, like:
"demo to customer: Nestlé"
it can have any special character and it should be preserved.

## Your Environment
- Tested with dotnet 4.8 and dotnet core 3.1, same thing occurs in both.
- Version Used: GitVersionTask 5.3.6
- Operating System and version Windows 10

LazaroOnline:
Related to:
https://github.com/GitTools/GitVersion/issues/2199

asbjornu:
I think one viable take on this is to be able to configure the branch name replacement regex as discussed in #1674.

Issue #2199, opened 2020-03-23
[Bug] Assembly Informational Version custumization ignored


fedebn:
Using GitVersion.yml to set custumized Assembly Informational Version, any charater used to separate version number from `ShortSha` is ignored and substitued with `-`.

## Expected Behavior
verified with version 5.1.3

```yaml
assembly-informational-format: '{Major}.{Minor}.{Patch}_{ShortSha}'
```

```
1.2.3_856daf9
```

## Actual Behavior
verified with version 5.2.4

```yaml
assembly-informational-format: '{Major}.{Minor}.{Patch}_{ShortSha}'
```

```
1.2.3-856daf9
```

asbjornu:
This change might have been introduced in #2014. Perhaps we need a more fine-grained replacement strategy here. 🤔

fedebn:
I don't know wich charaters are not allowed in SemVer

asbjornu:
[The regex of characters allowed in a SemVer 2.0 compatible pre-release label](https://semver.org/#spec-item-9) is `[0-9A-Za-z-]`. So I suppose the question here is: Should we allow `InformationalVersion` to be incompatible with SemVer?

fedebn:
If the scope of GitVersion is to use InformationalVersion as SemVer, if think this bug can be rejected

asbjornu:
Well, no, it's the other way around. GitVersion generates a SemVer 2.0 compatible version number, then provides many different serializations and formats of this version numbers targeting different environments.

Since [`AssemblyInformationalVersionAttribute`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assemblyinformationalversionattribute) allows for a larger set of characters than SemVer 2.0, I think it's natural to think that `assembly-informational-format` should only affect `AssemblyInformationalVersionAttribute` and thus allow characters not allowed in SemVer 2.0. I'm just not quite sure yet how to implement it.

fedebn:
So, i'm not sure, but we can think about a warning at compile time if AssemblyInformationalVersionAttribute isn't comply with SemVer 2.0.

asbjornu:
Providing the warning at compile time is going to be difficult, but we could log a warning statement during execution.

NeilMacMullen:
[...]

PR #2341, opened 2020-06-24
Allow special characters in Assembly InformationalVersion

LazaroOnline:
## Description
Allow any special character in Assembly InformationalVersion.
Added a parameter in the common function "CheckAndFormatString" to control whether or not special chars gets replaced.

### New in 2nd commit:
Added a new config parameter: `assembly-informational-is-semver` that is true by default to make this backwards compatible.
If set to false then it allows any character in  `assembly-informational-version`.

## Related Issue
https://github.com/GitTools/GitVersion/issues/2339
https://github.com/GitTools/GitVersion/issues/2199

## Motivation and Context
Allow any special character in Assembly InformationalVersion.
It is the only place we can add free long customized text describing anything we want and readable from dll/exe file properties.

## How Has This Been Tested?
Included a test.
Run all the test and got them passing.

## Checklist:

### DIFF
--- a/docs/input/docs/configuration.md
+++ b/docs/input/docs/configuration.md
@@ -42,6 +42,7 @@ next-version: 1.0
 assembly-versioning-scheme: MajorMinorPatch
 assembly-file-versioning-scheme: MajorMinorPatchTag
 assembly-informational-format: '{InformationalVersion}'
+assembly-informational-is-semver: true
 mode: ContinuousDelivery
 increment: Inherit
 continuous-delivery-fallback-tag: ci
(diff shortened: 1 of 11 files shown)


asbjornu:
Thanks for this. But as I wrote in https://github.com/GitTools/GitVersion/issues/2199#issuecomment-630746511, I'm hesitant to stop replacing special characters as consumers of the `[AssemblyInformationalVersion]` attribute generated by GitVersion may depend on special characters being replaced. See #2014 and #1520 for more information. I'll follow up the discussion with ideas for implementation in a non-breaking way in #2199.

Issue #2336, opened 2020-06-23
[Bug] Hotfix branch name not taken as version source


dramdrung:
**Describe the bug**
In documentation: https://gitversion.net/docs/more-info/version-increments, at point **Branch name** it's mentioned that the version in **Hotfix** branch name should be taken as version source. 
It's not working. 
Was behavior changed and not documented?

## Expected Behavior
In case there is version in hotfix branch e.g. hotfix/4.21.1, the version 4.21.1 should be taken as version source.

## Actual Behavior
The version in branch name is not taken into consideration when calculating version. 
Version calculated is lower than expected.

## Possible Fix
From what I can see in **VersionInBranchNameVersionStrategy** class, the version is taken only from Release branches based on **is-release-branch** flag in GitVersion config.
It could be extended to check if a branch is hotfix with Regex, then try to get version from branch name.
## Steps to Reproduce
Submitted Test Case scenario.
https://github.com/GitTools/GitVersion/pull/2337

## Context
When creating hotfix branch for a bug in a release, I expect to build the branch with same version as release.
As I have read documentation, I could achieve it with including version in a branch name. 
Except it doesn't work. 
It calculates version lower than the Release version, cause it doesn't take version from branch name.

## Your Environment

- Version Used: 5.3.6
- Operating System and version (Windows 10, Ubuntu 18.04): Windows 10
- Link to your project:
- Link to your CI build (if appropriate):

asbjornu:
What happens if you configure hotfix branches with `is-release-branch: true`?

dramdrung:
@asbjornu 
If the **is-release-branch: true** is set for Hotfix branches, it works.
It's a solution, but I wonder if there are any other effects of setting Hotfix branches as releases?
Looking at the documentation it seemed like that behavior would be out of the box, is it intentional to work only after setting the flag?

asbjornu:
I assume this behaviour was changed in #1541, where before, accidental version numbers in branch names were causing trouble all over the place. Hotfix branches should probably be considered hotfix branches by default, though. Would you be up for submitting a PR with this default configuration change?

dramdrung:
I can take a look and try to make the change. Probably by the end of the week, when I have free time.

dramdrung:
@asbjornu I made the change to default configuration and adjusted/added the test scenarios to the it. 
PR can be seen at https://github.com/GitTools/GitVersion/pull/2346. 
[...]

PR #2346, opened 2020-07-02
[Fix/Feature] Set Hotfix as release branch by default

dramdrung:
## Description
By the definition in the documentation it seems that the Hotfix branches should by default have **is-release-branch** set to **true**.
Discussion can be followed at https://github.com/GitTools/GitVersion/issues/2336.

This change not only modifies how version in Hotfix branch name is treated but as well how **merge messages** with such branches are handled.

## Related Issue

https://github.com/GitTools/GitVersion/issues/2336
## Motivation and Context
Version in Hotfix branch names should be taken into consideration when resolving base version by default.
This change enables this behavior with default configuration.

## How Has This Been Tested?

The branch was build locally and tested manually.
- The default configuration for Hotfix branch has:  `is-release-branch: true`
- The version in branch name is taken into consideration when resolving version on hotfix branch

Additionally there were added a few tests and current test scenarios were adjusted to the change.
## Checklist:


src/GitVersionCore.Tests/IntegrationTests/HotfixBranchScenarios.cs
```diff
@@ -63,7 +63,7 @@ public void CanTakeVersionFromHotfixesBranch()
 
             // create hotfix branch
             Commands.Checkout(fixture.Repository, fixture.Repository.CreateBranch("hotfixes/1.1.1"));
-            fixture.AssertFullSemver("1.1.0"); // We are still on a tagged commit
+            fixture.AssertFullSemver("1.1.1-beta.1+0"); // Version in branch name takes precedence over Tag if it's greater
```
asbjornu, line 66: Hm. I'm torn and don't really know what to think here. My knee-jerk reaction is that tags should always take precedence over everything else, but then again the entire point with versioned `hotfix` and `release` branches is to affect the version number.

src/GitVersionCore.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
```diff
@@ -62,19 +67,19 @@ public void TakesVersionFromMergeOfReleaseBranch(string message, bool isMergeCom
...
-        [TestCase("Merge branch 'Hotfix-10.10.50'", true)]
-        [TestCase("Merge branch 'Hotfix/10.10.50'", true)]
-        [TestCase("Merge branch 'hotfix-0.1.5'", true)]
-        [TestCase("Merge branch 'hotfix-4.2.2' into support-4.2", true)]
+        [TestCase("Merge branch 'feature-4.6.6' into support-4.6", true)]
+        [TestCase("Merge branch 'Feature-10.10.50'", true)]
+        [TestCase("Merge branch 'Features-10.10.50'", true)]
+        [TestCase("Merge branch 'fix/10.10.50' into hotfix-10.0.0", true)]
+        [TestCase("Merge branch 'feature/0.1.5'", true)]
+        [TestCase("Merge branch 'feature-4.2.2' into support-4.2", true)]
```
asbjornu, line 80: I assume these test cases broke when the branches were named `hotfix*` and not `feature*`?

dramdrung: Yes, every test case scenario which is changed in this PR was failing due to the change.

dramdrung:
Yes, my first thought after running the tests were the same. 
It seems that with previous change it was probably intended mainly for hotfix branch names to not be taken into consideration, as there were a few test case scenarios which expected such behavior. 

I don't really know what to think of it as well. 
On the one hand I would expect for Hotfixes to behave like release branches in some cases by default.
But in the end it may be better to just provide appropriate documentation to previous change.

src/GitVersionCore.Tests/IntegrationTests/HotfixBranchScenarios.cs
dramdrung: I think it makes sense that even if we are on tagged commit on release branch, greater version wins between tag and branch name.
As at first Release branch for X version is just created and optionally some Hotfixes are provided. 
But the whole release process on such branch may end with no additional commits. 

In the end I just documented what I experienced when I had to update this test case scenario.

asbjornu:
I think we'll target this for version 6, since it introduces breaking changes.

asbjornu:
`main` is now primed for v6. Can you please rebase and solve conflicts so we can merge this, @dramdrung?

src/GitVersionCore.Tests/IntegrationTests/HotfixBranchScenarios.cs
HHobeck: Actually I agree with 1.1.1-beta.1+0 as a result for this scenario. If you think about it: You are going to create hotfix branch from the current tagged version from main (RTM version). Thus your plan is to fix something maybe your CI pipeline has changes or what ever. The point is: This version says here that something different is included comparing to the previous version 1.1.0 but no code changes are involved (beta.1+0). ;)

HHobeck: Maybe you have changed from x64 to x32 or used the self-contained flag in dotnet because on your production environment dotnet is not installed.

HHobeck:
I have taken a look into the source code and see that we have two IVersionStrategy implementations in the current version 5.10.3 of GitVersion which are using the IsReleaseBranch property to determine the version. It is on the one hand the VersionInBranchNameVersionStrategy and on the other hand the MergeMessageVersionStrategy class. So far so good.

I'm just wondering why we are not separating this two aspects and make it configurable. For instance we can introduce a new property with the name VersionInBranchNamePattern="{Major}\.{Minor}\.{Patch}^" or VersionInBranchNamePattern="{Major}\.{Minor}^" to specify explicit how the version can be determined.

This has the following advantages:

1. If some one want's to configure the pattern how the version is encoded in the branch name then he can do it (see [Bug] Wrong version from the release branche #2925)
2. The aspect getting the version from the branch name and the aspect getting the version from other release branches are separated and can be configured.

asbjornu:
> I'm just wondering why we are not separating this two aspects and make it configurable. For instance we can introduce a new property with the name VersionInBranchNamePattern="{Major}.{Minor}.{Patch}^" or VersionInBranchNamePattern="{Major}.{Minor}^" to specify explicit how the version can be determined.

Sure, that's even something we could introduce in v5 since it's a new configuration property.

HHobeck:
Hi @dramdrung. Please rebase your PR with the main branch and give me feedback when you are done.

HHobeck:
Has been resolved with PR https://github.com/GitTools/GitVersion/issues/2336 without the need of specifing IsRelease to true for Hotfix branches.

PR #2356, opened 2020-07-13
Add support for next-version in mainline mode

### LINKED ISSUE #2335: [Docs] How to start using GitVersion in a new project?
I'm trying to use GitVersion in a brand new project of mine, using GitHub Flow and mainline mode, and I'm having trouble configuring it to behave the way I'd like/am expecting to.

I've been able to use GitVersion in an existing project of mine, where I'm able to tag the most recent release.  For example if I've released 1.2.3, I can tag that in Git.  On changes in a new branch, GitVersion will generate 1.2.4-branch0001 versions, which is what I'm expecting and what I want.

But I can't seem to figure out how to replicate this same behavior before I have any kind of release.  What I would imagine I do is set the next-version in my config to 1.0.0, then versions will be generated like 1.0.0-branch0001, 1.0.0-branch0002, etc, until I merge into master, gitversion generates 1.0.0, and Git is tagged with that version.  At that point I can remove next-version and rely on GitVersion seeing the Git tags to determine the next version.  However when I set next-version to 1.0.0, GitVersion generates 1.0.1-branch0001 versions.  I experimented a bit, and if I run it in master with next-version set to 1.0.0 and mode:MainLine, it will generate 1.0.0.  However my build processes deploys the pre-release versions to a test environment, and I need them to be a semantically lower version than the 1.0.0 that will eventually be released.

Is this a bug in gitversion?  Is there an additional configuration setting I'm missing?  Some misunderstanding I have with how the tool operates?

FISHMANPET:
Implementing failed test for #2335

## Description
Adds test in mainline mode using next-version config

## Related Issue
Fixes #2335 
## Checklist:

### DIFF
--- a/src/GitVersionCore.Tests/VersionCalculation/NextVersionCalculatorTests.cs
+++ b/src/GitVersionCore.Tests/VersionCalculation/NextVersionCalculatorTests.cs
@@ -118,33 +118,20 @@ namespace GitVersionCore.Tests.VersionCalculation
         }
 
         [Test]
-        public void PreReleaseTagCanUseBranchNameMainline()
+        public void PreReleaseVersionMainline()
         {
             var config = new Config
             {
                 VersioningMode = VersioningMode.Mainline,
-                NextVersion = "1.0.0",
-                Branches = new Dictionary<string, BranchConfig>
-                {
-                    {
-                        "custom", new BranchConfig
-                        {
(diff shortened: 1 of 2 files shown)


asbjornu:
Thank you so much for your contributions, @FISHMANPET! 🙏

PR #2370, opened 2020-07-29
Added support for project type Microsoft.NET.Sdk.WindowsDesktop.

olivegamestudio:
I've added a check so that WindowsDesktop projects can be accepted. They are in the same format as the Microsoft.NET.Sdk.
I've added a unit test to match the existing ones as well.


arturcic:
@olivegamestudio  thank you for your contributions 👍

rkieslinger:
I'm using GitVersion since a few days and ran into this issue today. Since the fix was added to the [5.5.0](https://github.com/GitTools/GitVersion/milestone/31) milestone, when does it get released?

Issue #2371, opened 2020-07-30
[Bug] merge-message-formats gets overwritten


chrisism:
**Describe the bug**
When you set your own custom merge-message-formats it will get overwritten when applying gitversion. 
Running the gitversion /showConfig will just show a emtpy dictionary again :
``
commit-date-format: yyyy-MM-dd
merge-message-formats: {}
update-build-number: true
``
## Expected Behavior
When I insert a custom merge-message-format I expect it to end up in the combined config and applied during gitversion action.

## Actual Behavior
merge-message-formats stays { }

## Possible Fix
The default Config() that is used as override in ConfigProvider has by default an empty dictionary for mergemessageformats, which will then override the custom dictionary because it only checks for NULL value. Merge dictionaries instead of override.

## Steps to Reproduce
Alter gitversion.yml with:
``
merge-message-formats:
    tfs:'^Merged (?:PR (?<PullRequestNumber>\d+)): (M|m)erge (?<SourceBranch>.+) (in)?to (?<TargetBranch>.+)'
``
execute: gitversion /showConfig
Result:
``
ignore:
  sha: []
commit-date-format: yyyy-MM-dd
merge-message-formats: {}
update-build-number: true
``
## Context
Now I can only use tags for incrementing. All my build pipelines in azure devops are now getting incorrect versions.

## Your Environment
Azure devops, using latest gittools task as.

chrisism:
Possible fix:
Class: ConfigurationBuilder.cs
Method: ApplyOverrides(..)
Line: 63 change to
``
targetConfig.MergeMessageFormats = overrideConfig.MergeMessageFormats.Any() ? overrideConfig.MergeMessageFormats : targetConfig.MergeMessageFormats;
``

asbjornu:
Since you have a possible fix in place, do you think you could submit that change as a pull request, @chrisism?

chrisism:
Of course, here we go. #2376

arturcic:
Closed by #2376

PR #2376, opened 2020-08-03
Added simple fix for mergemessageformats override

chrisism:
Quick hotfix for issue #2371. When you set your own custom merge-message-formats it will get overwritten when applying gitversion.
Running the gitversion /showConfig will just show a emtpy dictionary again :
commit-date-format: yyyy-MM-dd merge-message-formats: {} update-build-number: true

## Description
Quick fix in ConfigurationBuilder by instead of checking only for the MergeMessageFormats to be null, actually checking if it has any values. 
Proper resolution still needed, because this will not actually merge the two dictionaries.

## Related Issue
Issue #2371

## Motivation and Context
See the issue for more details. This will now solve the situation that custom formats are not applied.

## How Has This Been Tested?
Ran all the unittests, using it in own build server environment.

## Checklist:

### DIFF
--- a/src/GitVersionCore/Configuration/ConfigurationBuilder.cs
+++ b/src/GitVersionCore/Configuration/ConfigurationBuilder.cs
@@ -60,7 +60,7 @@ namespace GitVersion.Configuration
             targetConfig.CommitMessageIncrementing = overrideConfig.CommitMessageIncrementing ?? targetConfig.CommitMessageIncrementing;
             targetConfig.Increment = overrideConfig.Increment ?? targetConfig.Increment;
             targetConfig.CommitDateFormat = overrideConfig.CommitDateFormat ?? targetConfig.CommitDateFormat;
-            targetConfig.MergeMessageFormats = overrideConfig.MergeMessageFormats ?? targetConfig.MergeMessageFormats;
+            targetConfig.MergeMessageFormats = overrideConfig.MergeMessageFormats.Any() ? overrideConfig.MergeMessageFormats : targetConfig.MergeMessageFormats;
             targetConfig.UpdateBuildNumber = overrideConfig.UpdateBuildNumber ?? targetConfig.UpdateBuildNumber;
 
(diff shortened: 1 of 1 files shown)


src/GitVersionCore/Configuration/ConfigurationBuilder.cs
```diff
@@ -60,7 +60,7 @@ private static void ApplyOverrides([NotNull] Config targetConfig, [NotNull] Conf
             targetConfig.CommitMessageIncrementing = overrideConfig.CommitMessageIncrementing ?? targetConfig.CommitMessageIncrementing;
             targetConfig.Increment = overrideConfig.Increment ?? targetConfig.Increment;
             targetConfig.CommitDateFormat = overrideConfig.CommitDateFormat ?? targetConfig.CommitDateFormat;
-            targetConfig.MergeMessageFormats = overrideConfig.MergeMessageFormats ?? targetConfig.MergeMessageFormats;
+            targetConfig.MergeMessageFormats = overrideConfig.MergeMessageFormats.Any() ? overrideConfig.MergeMessageFormats : targetConfig.MergeMessageFormats;
```
asbjornu, line 63: Is there any chance of `MergeMessageFormats` being `null`, which can cause a `NullReferenceException` on the call to `Any()`?

```suggestion
            targetConfig.MergeMessageFormats = overrideConfig.MergeMessageFormats?.Any() ? overrideConfig.MergeMessageFormats : targetConfig.MergeMessageFormats;
```

chrisism: The default value is a Dictionary actually.
``
        [YamlMember(Alias = "merge-message-formats")]
        public Dictionary<string, string> MergeMessageFormats { get; set; } = new Dictionary<string, string>();
``
So as far as I know it should be okay.

asbjornu:
Thank you for your contributions, @chrisism. 🙏

PR #2379, opened 2020-08-04
Reproduce 'Sequence contains no matching element' error

### LINKED ISSUE #2378: [Bug] Cloning develop throws InvalidOperationException: Sequence contains no matching element
**Describe the bug**
Using develop as our default branch led to GitVersion failing locally with an incomprehensible error.

## Expected Behavior
* Version should be the same as if both master and develop have been checked out locally. 
* Not having master branch should lead to a clear warning message something like "Could not find a 'master' branch, neither locally nor remotely."

## Actual Behavior
```
  System.InvalidOperationException : Sequence contains no matching element
  Stack Trace: 
    Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
    MainlineVersionCalculator.GetMainline(Commit baseVersionSource) line 133
    MainlineVersionCalculator.FindMainlineModeVersion(BaseVersion baseVersion) line 44
    NextVersionCalculator.FindVersion() line 59
```

## Possible Fix
Add origin/master to regex for master.

## Steps to Reproduce
See PR #2379 test GivenARemoteGitRepositoryWithCommitsThenClonedLocalDevelopShouldMatchRemoteVersion

## Your Environment

- Version Used: 5.3.7
- Windows 10
- Internal Azure DevOps repo

kcaswick:
## Description
Add 2 test cases to test for the expected behavior of issue #2378 

## Related Issue

Fixes #2378.

## Motivation and Context
Reproduce #2378

### DIFF
--- a/src/GitVersion.Core.Tests/IntegrationTests/MainlineDevelopmentMode.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/MainlineDevelopmentMode.cs
@@ -562,7 +562,7 @@ public class MainlineDevelopmentMode : TestBase
         fixture.Repository.Branches.Remove(fixture.Repository.Branches["main"]);
 
         var exception = Assert.Throws<WarningException>(() => fixture.AssertFullSemver("1.1.0-alpha.1", config));
-        exception.Message.ShouldMatch("No branches can be found matching the commit .* in the configured Mainline branches: main, support");
+        exception!.Message.ShouldMatch("No branches can be found matching the commit .* in the configured Mainline branches: main, support");
     }
 }
 
--- a/src/GitVersion.Core.Tests/IntegrationTests/MainlineDevelopmentMode.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/MainlineDevelopmentMode.cs
@@ -5,6 +5,7 @@ using GitVersion.Model.Configuration;
 using GitVersion.VersionCalculation;
 using LibGit2Sharp;
 using NUnit.Framework;
+using Shouldly;
 
 namespace GitVersion.Core.Tests.IntegrationTests;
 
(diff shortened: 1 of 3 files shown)


asbjornu:
I've had a look at this and the fix for the expected `WarningException` in `GivenNoMasterThrowsWarning` is quite simple to fix by adding an `if` statement and `throw WarningException` after the following line:

https://github.com/GitTools/GitVersion/blob/afbb8a4f81f3ed24dcfbec6e0b7d6b0c0561d682/src/GitVersionCore/VersionCalculation/MainlineVersionCalculator.cs#L128

Then, the `GivenARemoteGitRepositoryWithCommitsThenClonedLocalDevelopShouldMatchRemoteVersion` test can be fixed by branching to `master` before asserting the version:

```c#
fixture.BranchTo("master", config, repository: local.Repository)
fixture.AssertFullSemver("0.2.0-alpha.0", config, repository: local.Repository);
```

Otherwise, `develop` will be the checked out branch, and since `develop` is not a Mainline branch by default (only `master` and `support/*` is), no branch, and thus no commit will be found.

Thoughts, @kcaswick?

asbjornu:
I added a few changes to your PR, please pull before you do any further development. Also, please review my changes and share your thoughts about them.

asbjornu:
Sorry, scratch that. I managed to commit code that doesn't compile. 😅  Fix incoming.

asbjornu:
Ok, the stuff I was trying out didn't work. To fix this, we need to dig into the following code to figure out why `refs/remotes/origin/master` is not being considered.

https://github.com/GitTools/GitVersion/blob/afbb8a4f81f3ed24dcfbec6e0b7d6b0c0561d682/src/GitVersionCore/Core/RepositoryMetadataProvider.cs#L297-L312

kcaswick:
> ```cs
> fixture.BranchTo("master", config, repository: local.Repository)
> fixture.AssertFullSemver("0.2.0-alpha.0", config, repository: local.Repository);
> ```
> 
> Otherwise, `develop` will be the checked out branch, and since `develop` is not a Mainline branch by default (only `master` and `support/*` is), no branch, and thus no commit will be found.
> 
> Thoughts, @kcaswick?

Yes, develop being the checked out branch is the point, as that is what developers at that company will get when they clone. Changing the test case will not fix the issue, as now you are requiring every single developer who clones the repository to go through extra steps, and to know that the confusing error message means they missed it. (Remember, from the developer perspective, all they did was click clone, and click build in Visual Studio, and now it is giving them weird errors. They may not even have heard of GitVersion before that point. This is at a company with lots of contractors rotating through due to 2 year contract term limits, so it'd be easier to stop using GitVersion than deal with the constant friction.)

I'm actually no longer a contractor at the company I was when I submitted the issue, so I will not be able to test the fix against the repo I originally encountered the issue in.

asbjornu:
I improved this somewhat in #3018, but the tests and error message added in this PR are nice so I rebased and think this should be good to go.

asbjornu (approved):

arturcic:
@asbjornu tests are failing, what do you want to do with this PR?

asbjornu:
Hm, I'll have a second look at those tests, @arturcic. I thought they were going to be successful.

Issue #2340, opened 2020-06-24
GitVersion should only consider tags matching current branch config (was: GitVersion fails to return correct version when tag exists on commit)


DocZoid:
**Describe the bug**
Tags have a too strong weight on the version evaluation which I feel is a bug. I am quite new to versioning and GitVersion, so I might be on the wrong track without knowing.  I have read in some other issue that fast-forward merges for master are generally discouraged (forbidden?), and this is a part of my issue (#2). The proposed fix would as a side effect allow using fast-forward merges in my opinion.

There are two scenarios where the tag should not be used to evaluate the version:

1. When I build a snapshot from the develop branch, tag the branch with the created version (e.g. 0.1.2-snapshot.2), and then branch this commit to a new release (release/0.1.2)
2. When I build a release candidate from a release branch, tag the release branch with the new created version (e.g. 0.1.2-rc.1), and then make a fast-forward merge to master, essentially having master point to that tagged commit

## Expected Behavior
1. The resulting version should be 0.1.2-rc.1 (with the release configuration "tag: rc")
2. The resulting version should be 0.1.2

## Actual Behavior
1. The resulting version is 0.1.2-snapshot.2 based on the tag
2. The resulting version is 0.1.2-rc.1 based on the tag

## Possible Fix
Multiple branches point to the same commit. This is confusing GitVersion, which takes tags from that commit, but it does not respect the branch configuration of the current branch when analyzing the tags, e.g. when I am working on a release branch, it should only look for previous version tags which are formatted like its own tag configuration in the release branch ("with "rc" in it, so tags 0.1.2-rc.1 should be considered, but not 0.1.2-snapshot.2, because of the configuration "tag: rc")

## Steps to Reproduce

1. assume or create config: "develop -> tag: snapshot", "release -> tag: rc"
2. tag a commit in "develop" with a version tag, e.g. 0.1.2-snapshot.2
3. branch that commit to release/0.1.2
4. using release/0.1.2 as current branch, run GitVersion. Calculated version should be 0.1.2-rc.1 instead of 0.1.2-snapshot.2

## Context

We have commit hooks checking every commit message. Therefor, it seems like an easy option to use fast-forward merges for master to avoid commits, but GitVersion is missing the mentioned configuration option to be able to handle fast-forward merges. 
[...]

PR #2413, opened 2020-09-30
Git version calculating wrong version when building a tag

gerwinjansen:
Note: this PR is on top of #2411, that's why you see the changes again.

## Description
TaggedCommitVersionStrategy will try to return the tags on the current commit. If there are none, it will returns the tags from the history.

## Related Issue
Fixes #2074 GitVersion calculating wrong version when building a tag

## How Has This Been Tested?
See steps to reproduce in my comments of the related ticket.

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2416, opened 2020-10-05
Feature/dynamic tag evaluation

DocZoid:
Change tag evaluation (test only)

## Description
added test for #2340 scenario 1, branch created on tagged commit should not consider the tag if it does not match to the current branch config

## Related Issue
#2340

## Motivation and Context
There are two scenarios where the tag should not be used to evaluate the version:

When I build a snapshot from the develop branch, tag the branch with the created version (e.g. 0.1.2-snapshot.2), and then branch this commit to a new release (release/0.1.2)
When I build a release candidate from a release branch, tag the release branch with the new created version (e.g. 0.1.2-rc.1), and then make a fast-forward merge to master, essentially having master point to that tagged commit

## How Has This Been Tested?
This PR does not contain any fix yet, only a test to show the intended behavior.

## Checklist:


asbjornu:
Please have a look at the [failing `dotnet format`](https://github.com/GitTools/GitVersion/pull/2416/checks?check_run_id=1240135193#step:5:1). Also, this feels very similar to #2413, no?

Issue #1198, opened 2017-04-03
dirty flag?


Trass3r:
Couldn't find a way to get a dirty flag, like `git describe [...] --dirty`

asbjornu:
What is it you want to achieve?

Trass3r:
Well a dirty flag in the informational version when there are locally modified files.

asbjornu:
Ah, yes that would be a good thing to have, I agree. We welcome pull requests! 😄

Sam13:
I did some prototyping on that. Since the git versions are cached in the _.git_ folder, I don't know how the dirty flag should behave. Causes wrong results when caching that. @asbjornu what do you propose?

asbjornu:
Makes sense. Perhaps when calculating `--dirty`, you need to disable caching altogether?

brdmllr:
I hope this isn't a naive solution as I haven't spent a lot of time looking at the GitVersion source, but could the output of the "git status --porcelain" command act as input for this feature and if that input has a non-zero length, something is added to the informational version?

Sam13:
The thing which should be achieved by this issue, is that the informal assembly version contains an information whether the working copy has local modifications or not.

If _git status_ provides this information as well we may also use this command as source for that information.

Sam13:
Still a useful feature in my opinion

asbjornu:
@Sam13, I agree, but unless someone submits a PR that implements this, the feature won't make its way into GitVersion.

simonbuehler:
hi,

i found how this got handled in another project, could this be incooperated?
   https://github.com/desaster/gitrevsubst/commit/544462be2a60cf4ebf53543329147f40a1071302#diff-a8d7dc4a2b96ae65d37ab09ad03939dd

basicaly its testing `describe --always --dirty` if it ends with "dirty"

sbyholm:
I was searching for this, would be happy to have a dirty flag in GitVersion

asbjornu:
A PR implementing this would be merged, I'm sure.

PR #2423, opened 2020-10-13
Implemented the Dirty flag

pysco68:
Have been replacing a bunch of usages of the deprected Fody.Stamp with GitVersion lately and found that we need the "Dirty" flag as some others (see #1198). Just got ahead and implemented it.

Hope it'll meet the required project standards ;)

## Description
Added a variable ```RepositoryDirtyFlag``` which is either ```null``` or ```Dirty``` to be able to add that suffix to my version strings.

## Related Issue
As mentioned above: #1198 

## How Has This Been Tested?
I added four test cases which test the behavior of the dirty flag in ```/src/GitVersionCore.Tests/IntegrationTests/OtherScenarios.cs```. 

It should be set when there's any non-ignored change vs the HEAD of the repo (that includes staged files). 

Other than that I had to change a some other test cases which work on the outputed data to make them work again

## Checklist:
- [x] My code follows the code style of this project. (hope so :))
- [x] All new and existing tests passed. (I had a bug in a subset of the tests for the netcoreapp2.1 build which seemed to be related to some specifics on my env)

### DIFF
--- a/docs/input/docs/more-info/variables.md
+++ b/docs/input/docs/more-info/variables.md
@@ -40,7 +40,7 @@ Running `GitVersion.exe` in your repo will show you what is available. For the
   "CommitsSinceVersionSource":1,
   "CommitsSinceVersionSourcePadded":"0001",
   "CommitDate":"2014-03-06",
-  "RepositoryDirtyFlag": ""
+  "UncommittedChanges": 0
 }
 ```
 
--- a/docs/input/docs/more-info/variables.md
+++ b/docs/input/docs/more-info/variables.md
@@ -39,7 +39,8 @@ Running `GitVersion.exe` in your repo will show you what is available. For the
   "VersionSourceSha":"950d2f830f5a2af12a6779a48d20dcbb02351f25",
   "CommitsSinceVersionSource":1,
   "CommitsSinceVersionSourcePadded":"0001",
-  "CommitDate":"2014-03-06"
+  "CommitDate":"2014-03-06",
+  "RepositoryDirtyFlag": ""
 }
 ```
 
--- a/src/GitVersionCore.Tests/IntegrationTests/OtherScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/OtherScenarios.cs
@@ -97,32 +97,33 @@ namespace GitVersionCore.Tests.IntegrationTests
             fixture.AssertFullSemver("1.1.0-alpha.1");
         }
 
-        [TestCase(true, false)]
-        [TestCase(true, true)]
-        [TestCase(false, false)]
-        [TestCase(false, true)]
-        public void HasDirtyFlagIfUncommittedChangesAreInRepo(bool createTempFile, bool stageFile)
+        [TestCase(true)]
+        [TestCase(false)]
+        public void HasDirtyFlagWhenUncommittedChangesAreInRepo(bool stageFile)
+        {
+            using var fixture = new EmptyRepositoryFixture();
+            fixture.Repository.MakeACommit();
+           
+            var tempFile = Path.GetTempFileName();
+            var repoFile = Path.Combine(fixture.RepositoryPath, Path.GetFileNameWithoutExtension(tempFile) + ".txt");
+            File.Move(tempFile, repoFile);
+            File.WriteAllText(repoFile, "Hello world");
+
+            if (stageFile)
+                Commands.Stage(fixture.Repository, repoFile);
+
+            var version = fixture.GetVersion();
+            version.RepositoryDirtyFlag.ShouldBe("Dirty");
+        }
+
+        [Test]
+        public void NoDirtyFlagInCleanRepository()
         {
             using var fixture = new EmptyRepositoryFixture();
             fixture.Repository.MakeACommit();
 
-            if (createTempFile)
-            {
-                var tempFile = Path.GetTempFileName();
-                var repoFile = Path.Combine(fixture.RepositoryPath, Path.GetFileNameWithoutExtension(tempFile) + ".txt");
-                File.Move(tempFile, repoFile);
-                File.WriteAllText(repoFile, "Hello world");
-
-                if (stageFile)
-                    Commands.Stage(fixture.Repository, repoFile);
-            }
-
             var version = fixture.GetVersion();
-
-            if (createTempFile)
-                version.RepositoryDirtyFlag.ShouldBe("Dirty");
-            else
-                version.RepositoryDirtyFlag.ShouldBe(null);
+            version.RepositoryDirtyFlag.ShouldBe(null);
         }
     }
 }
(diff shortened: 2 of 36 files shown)


arturcic (approved):

arturcic:
@asbjornu can you review this one as well?

src/GitVersionCore.Tests/IntegrationTests/OtherScenarios.cs
```diff
@@ -94,5 +96,33 @@ public void DoNotBlowUpWhenDevelopAndFeatureBranchPointAtSameCommit()
             fixture.InitializeRepo();
             fixture.AssertFullSemver("1.1.0-alpha.1");
         }
+
+        [TestCase(true, false)]
+        [TestCase(true, true)]
+        [TestCase(false, false)]
+        [TestCase(false, true)]
+        public void HasDirtyFlagIfUncommittedChangesAreInRepo(bool createTempFile, bool stageFile)
```
asbjornu, line 104: This test is a bit confusing since it tests both dirty and non-dirty with a nested `if` statement. It would be clearer if you split this test in two sharply focused tests with no `if (createTempFile)` statement.

src/GitVersionCore/Core/RepositoryMetadataProvider.cs
```diff
@@ -500,6 +500,21 @@ private IEnumerable<BranchCommit> GetMergeCommitsForBranch(Branch branch, Config
             return branchMergeBases;
         }
 
+        public bool HasUncommittedChanges()
```
asbjornu, line 503: Can this be a property instead of a method?

asbjornu (changes requested):

src/GitVersionCore.Tests/IntegrationTests/OtherScenarios.cs
pysco68: Sure will do

docs/input/docs/more-info/variables.md
```diff
@@ -39,7 +39,8 @@ Running `GitVersion.exe` in your repo will show you what is available. For the
   "VersionSourceSha":"950d2f830f5a2af12a6779a48d20dcbb02351f25",
   "CommitsSinceVersionSource":1,
   "CommitsSinceVersionSourcePadded":"0001",
-  "CommitDate":"2014-03-06"
+  "CommitDate":"2014-03-06",
+  "RepositoryDirtyFlag": ""
```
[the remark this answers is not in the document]
pysco68: I think the `bool` version is most useful (at least that would be a use case for me). Which case would you make for counting the actual number of changes? I mean beyond just "can do" ? I don't see much value in seeing that the build was done with "N" uncommitted changes. Just knowing that it was "dirty" seems fine for me.

src/GitVersionCore/Core/RepositoryMetadataProvider.cs
pysco68: Sure, will do. Implemented it as method as the other aspects of that class were getter methods too. Just meant to "fit in" ^^

docs/input/docs/more-info/variables.md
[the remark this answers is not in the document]
pysco68: Perhaps a dumb idea: should I maybe implement it both? Like one `HasUncomittedChanges` boolean and `NumberUncomittedChanges` integer?

src/GitVersionCore/Core/RepositoryMetadataProvider.cs
asbjornu: Ok, I see you're right. I'm not a fan of that sort of Java-ish code, but it's better to be internally consistent than to make an half-hearted attempt at being consistent with C# conventions. So please ignore my request.

docs/input/docs/more-info/variables.md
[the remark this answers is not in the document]
asbjornu: I like `UncommitedChanges` and think it should be a number, to accompany the rest of the variables such as `CommitsSinceVersionSource` and such so it can be used directly in version numbers easily.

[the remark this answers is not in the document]
pysco68: @asbjornu gotcha! Just one small related question: how should I deal with the parsing of that in `SemanticVersionBuildMetaData.Parse()` (file: `VersionCalculation/SemanticVersioning/SemanticVersionBuildMetaData.cs`)?

I previously just looked at the `OtherMetatada` property for if it was `== "Dirty"` but that won't work as simply anymore.
I suggest I just leave out that part to have something mergeable (as we're beyond the standard SemVer stuff anyway). Or do you have any concrete idea how this should be properly serialized?

pysco68:
@asbjornu looks like the test image for the failed one couldn't fetch the repo from github.com for some reason. Name-resolution perhaps?

arturcic:
> @asbjornu looks like the test image for the failed one couldn't fetch the repo from github.com for some reason. Name-resolution perhaps?

fixed

arturcic (approved):

asbjornu (approved):

asbjornu:
Excellent, thank you so much for your contributions @pysco68! 🙏

pysco68:
Happy I could help :)

LazaroOnline:
Hello, I can't find how to use the dirty flag.
I don't find it at:
https://gitversion.net/docs/
I'm trying in the GitVersion.yml file to add something like:
```yml
assembly-informational-format: 'git-{ShortSha}-{BranchName}--{CommitDate}-{RepositoryDirtyFlag}'
```
but no luck. Any help/doc/reference?

asbjornu:
@LazaroOnline, the description in the PR does not reflect what was implemented. The property changed name to `UncommittedChanges` and its type is an integer indicating the number of uncommitted changes. A PR that updates the docs would be highly appreciated. 🙏

LazaroOnline:
Under the new name `UncommittedChanges` it is already in the API docs:
https://gitversion.net/api/GitVersion/SemanticVersionBuildMetaData/
https://gitversion.net/api/GitVersion/SemanticVersionBuildMetaData/4EEF3F34
https://gitversion.net/docs/more-info/variables

The only missing part I think would be at:
https://gitversion.net/docs/configuration
with a small description.

It is nice to have this new variable, however in my case I miss some additional features related to it,
like uncommitted file filters to exclude files from being counted in`UncommittedChanges`,
because some files may get changed during build or that doesn't affect the resulting DLL, for example:
"*.txt",
"*.md",
"*.config",
"*.json",
"*.js",
"*.css",
...

asbjornu:
> The only missing part I think would be at: https://gitversion.net/docs/configuration with a small description.

Yes, that would be a nice addition.

> It is nice to have this new variable, however in my case I miss some additional features related to it, like uncommitted file filters to exclude files from being counted in`UncommittedChanges`, because some files may get changed during build or that doesn't affect the resulting DLL,

It sounds like if `.gitignore` was taken into account, that should solve your use-case. And I can agree that would be useful. But we are basing the implementation on what LibGit2Sharp exposes and it's out of scope to do anything about `.gitignore` in GitVersion. If you figure out a way to do this with LibGit2Sharp, a PR is welcome.

Issue #2439, opened 2020-10-28
[Bug] NoFetch property not derived from buildagent when GitVersion is invoked without arguments


davikor:
Starting with 5.3.0 the `NoFetch` argument remains set false when `GitVersion` is being invoked without arguments. This is because the `ParseArguments` function checks at the beginning whether it was invoked without any arguments and then just sets some default values.

## Expected Behavior
In my opinion all arguments should be set to the same value they would be set if `GitVersion` was invoked with at least one argument. E.g. if `GitVersion` is called only with parameter `targetpath`, `NoFetch` would be inferred from the buildAgent. If called without any parameters at all `NoFetch` would be set to `false` no matter what the build agents `PreventFetch` method says. This also applies to other parameters, e.g. the authentication would not be derived from the environment vars in case `GitVersion` is invoked without any arguments

## Actual Behavior
Currently `NoFetch`remains `false` even if the buildagents `PreventFetch` function returns `true` when `GitVersion`is invoked without any arguments

## Possible Fix
Remove the zero-arguments check at the beginning of the `ParseArguments` function. I will provide a pull request if welcome

## Steps to Reproduce
invoke on a build server environment (e.g. environment variables TEAMCITY_VERSION and Git_Branch exist) one time without any arguments and the other time with unrelated argument e.g. target path and see how `NoFetch` is `false`in the one case and `true` in the other.

## Context
Updated from 5.2 to 5.5 and `GitVersion` suddenly failed because it tried to fetch data from remote

## Your Environment
TeamCity 2020.1

PR #2440, opened 2020-10-29
Derive NoFetch argument from buildAgent field when GitVersion is invoked without any arguments

davikor:
## Description
This will derive the `NoFetch` argument from the `buildAgent` field if `GitVersion` is invoked without any arguments.
It checks whether the `buildAgent` field is set and if so sets `NoFetch` to the result of its `PreventFetch` function (`false` otherwise). This is the same behavior that is applied when `GitVersion` is invoked with at least one argument (which not necessarily needs to be the `NoFetch` argument).

## Related Issue
[https://github.com/GitTools/GitVersion/issues/2439](https://github.com/GitTools/GitVersion/issues/2439)

## Motivation and Context
I think the default for the `NoFetch` argument should be the same no matter whether `GitVersion` is being invoked without any parameters at all or one or more (unrelated) parameters. An equal solution has previously been applied for the `AddAuthentication` function which in earlier versions also was called only when `GitVersion`was invoked with at least one argument.

## How Has This Been Tested?
I added a test class that is similar to the `ArgumentParserTests` class but replaced the `argumentParser` field with an instance that contains a mock build agent. The only test here calls `ParseArguments` without anys arguments and checks afterwards whether `NoFetch` is `true` (which is the return of the mocks `PreventFetch` function).

## Checklist:

- [ ] All new and existing tests passed. 16 tests in GitVersionTask.Tests fail when running as netcoreapp2.1 but these also failed before I implemented this fix so these should not be related to this pr

### DIFF
--- a/src/GitVersionExe.Tests/ArgumentParserOnBuildServerTests.cs
+++ b/src/GitVersionExe.Tests/ArgumentParserOnBuildServerTests.cs
@@ -1,0 +1,63 @@
+using System;
+using GitVersion;
+using GitVersion.OutputVariables;
+using GitVersionCore.Tests.Helpers;
+using Microsoft.Extensions.DependencyInjection;
+using NUnit.Framework;
+using Shouldly;
+
+namespace GitVersionExe.Tests
+{
+    [TestFixture]
+    public class ArgumentParserOnBuildServerTests : TestBase
+    {
+        private IArgumentParser argumentParser;
+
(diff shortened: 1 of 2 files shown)


arturcic (approved):

arturcic:
@davikor this bug was introduced when we made changes to the ArgumentParser. Thanks for fixing it 👍.

Issue #2449, opened 2020-11-11
[Bug] New GenerateBuildLogOutput for GitHub Actions in 5.5.0 doesn't export variables


DavidBoike:
**Describe the bug**
It turns out https://github.com/GitTools/GitVersion/pull/2422 was an incorrect implementation of GitHub Actions' new way of exporting environment variables from a workflow step. GitHub Actions does not parse environment variables out of the log output. It generates a temporary file, which is different for every step of a workflow, that it identifies using the $GITHUB_ENV environment variable. You need to write each key as `{NAME}={VALUE}` to that temp file. GitHub Actions then parses _that_ file and then (I believe) disposes of it before the next step starts.

## Expected Behavior
`GitVersion_{NAME}` variables should be available as environment variables for workflow steps after GitVersion runs.

## Actual Behavior
All `GitVersion_{NAME}` environment variables are missing.

## Possible Fix
See PR

## Steps to Reproduce
Run a build with GitVersion using GitHub actions, this example uses windows-latest, using a workflow including:

```
…
    steps:
      (other steps)
      - name: Build
        run: dotnet build … (project with GitVersion)
      - name: Output GitVersion value
        run: echo "GitVersion_FullSemVer: $env:GitVersion_FullSemVer" 
```

Output will not include FullSemVer value.

## Your Environment

- Version Used: 5.5.0
- Operating System and version (Windows 10, Ubuntu 18.04): `windows-latest` but would affect all runner environments
- Link to your project:
- Link to your CI build (if appropriate):

PR #2422, opened 2020-10-09
Avoid set-env environment warning during build log generation

danielmarbach:
GitHub Actions is deprecating set-env and add-path commands as outlined by

https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

Therefore starting from version 2.273.5 of the runner `GenerateBuildLogOutput` outputs

image

Environment files should be used instead

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

## Description

see above

## Related Issue

Fixes #2421 and GitTools/actions#213.
## How Has This Been Tested?

No, I edited the file online and did not try to build and intermediate package to a private feed and then use that in my code. We should be able though to see the impact on the build on this project. I'll update the screenshots as soon as this compiles

## Screenshots (if appropriate):

Also happens on the build of this project, see

image

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #2450, opened 2020-11-11
For GitHub Actions, outgoing env vars need to be written to $GITHUV_ENV file

DavidBoike:
It turns out https://github.com/GitTools/GitVersion/pull/2422 was an incorrect implementation of GitHub Actions' new way of exporting environment variables from a workflow step. GitHub Actions does not parse environment variables out of the log output. It generates a temporary file, which is different for every step of a workflow, that it identifies using the $GITHUB_ENV environment variable. You need to write each key as `{NAME}={VALUE}` to that temp file. GitHub Actions then parses _that_ file and then (I believe) disposes of it before the next step starts.

So while https://github.com/GitTools/GitVersion/pull/2422 still resulted in a correctly built binary, the exported environment variables essentially disappeared from any later step, making it impossible to do things like deploy prerelease builds to one environment and production builds to a different one.

## Description
* Identify file path from GITHUB_ENV value
* Output key=value pairs to file
* Change log output to indicate values are being written to file, to erase confusion that they are parsed from log output
* If GITHUB_ACTIONS is set but GITHUB_ENV is missing, log a warning message

## Related Issue
* Fixes https://github.com/GitTools/GitVersion/issues/2449
* Bug introduced in https://github.com/GitTools/GitVersion/pull/2422

## Motivation and Context
Regression in behavior previously working in GitHub Actions

## How Has This Been Tested?
* Updated unit tests for GitHub Actions to provide a temporary file in GITHUB_ENV and assert its contents are written correctly.
* All unit tests run successfully.
* Though I'd like to generate a package and test on my target project, I'm getting errors executing build.ps1.

N/A

## Checklist:

### DIFF
--- a/src/GitVersionCore/BuildAgents/GitHubActions.cs
+++ b/src/GitVersionCore/BuildAgents/GitHubActions.cs
@@ -1,7 +1,6 @@
 using GitVersion.Logging;
 using GitVersion.OutputVariables;
 using System.IO;
-using System.Text;
 
 namespace GitVersion.BuildAgents
 {
@@ -11,12 +10,10 @@ namespace GitVersion.BuildAgents
 
         public GitHubActions(IEnvironment environment, ILog log) : base(environment, log)
         {
-            this.environment = environment;
         }
 
         public const string EnvironmentVariableName = "GITHUB_ACTIONS";
         public const string GitHubSetEnvTempFileEnvironmentVariableName = "GITHUB_ENV";
-        private readonly IEnvironment environment;
 
         protected override string EnvironmentVariable { get; } = EnvironmentVariableName;
 
@@ -47,7 +44,7 @@ namespace GitVersion.BuildAgents
                 return;
             }
 
-            var gitHubSetEnvFilePath = environment.GetEnvironmentVariable(GitHubSetEnvTempFileEnvironmentVariableName);
+            var gitHubSetEnvFilePath = this.Environment.GetEnvironmentVariable(GitHubSetEnvTempFileEnvironmentVariableName);
 
             if (gitHubSetEnvFilePath != null)
             {
--- a/src/GitVersionCore/BuildAgents/GitHubActions.cs
+++ b/src/GitVersionCore/BuildAgents/GitHubActions.cs
@@ -1,5 +1,7 @@
 using GitVersion.Logging;
 using GitVersion.OutputVariables;
+using System.IO;
+using System.Text;
 
 namespace GitVersion.BuildAgents
 {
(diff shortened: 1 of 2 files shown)


src/GitVersionCore/BuildAgents/GitHubActions.cs
```diff
@@ -24,16 +30,31 @@ public override string GenerateSetVersionMessage(VersionVariables variables)
...
 
             if (!string.IsNullOrWhiteSpace(value))
             {
-                var key = $"GitVersion_{name}";
+                var gitHubSetEnvFilePath = environment.GetEnvironmentVariable(GitHubSetEnvTempFileEnvironmentVariableName);
+                var assignment = $"GitVersion_{name}={value}";
+
+                if (gitHubSetEnvFilePath != null)
+                {
+                    using (var streamWriter = File.AppendText(gitHubSetEnvFilePath)) // Already uses UTF-8 as required by GitHub
```
danielmarbach, line 44: I still wonder if it would be better to use the base WriteIntegration instead. Now because this is called by the base for each key value pair we end up open/closing file handles N times. With WriteIntegration we could do a select projection of all key value pairs and write all those lines in one go to the file. This would eliminate the N+1 problem

DavidBoike: Good call @danielmarbach - pushed up a new implementation.

DavidBoike:
Downloaded the NuGet artifact and tested it in my project - now all GitVersion variables are availble when dumping all env vars with `Get-ChildItem Env:*`.

src/GitVersionCore/BuildAgents/GitHubActions.cs
```diff
@@ -9,9 +11,13 @@ public class GitHubActions : BuildAgentBase
 
         public GitHubActions(IEnvironment environment, ILog log) : base(environment, log)
         {
+            this.environment = environment;
         }
 
         public const string EnvironmentVariableName = "GITHUB_ACTIONS";
+        public const string GitHubSetEnvTempFileEnvironmentVariableName = "GITHUB_ENV";
+        private readonly IEnvironment environment;
```
asbjornu, line 19: Perhaps `BuildAgentBase` can expose a protected `Environment` property instead?

asbjornu (changes requested):
This looks great! Just a tiny nitpick and this should be good to go.

src/GitVersionCore/BuildAgents/GitHubActions.cs
DavidBoike: It already does! Missed that. Pushing a commit now.

asbjornu (approved):

asbjornu:
Excellent, thank you so much @DavidBoike & @danielmarbach! 🙏

JesperMonsted:
Looking forward to having your fix in a release 👍

PR #2457, opened 2020-11-17
Fix for the wrong updateBuildNumber behavior

kirkone:
When `updateBuildNumber` is set to false the Build Number is still be overwritten.

## Description
As discussed in #1770 the switch should result in a different behavior.

## Related Issue
#1457
#1770 

## Motivation and Context
The behavior was not implemented in the way it was intended. This should fix this.

## How Has This Been Tested?
Tested on my local machine and in my own Azure DevOps instance.

## Checklist:


kirkone:
@arturcic I am not sure about the tests. Is it ok to have them that way?

arturcic (approved):

arturcic:
@kirkone thank you so much for your contributions 👍

PR #2458, opened 2020-11-17
Update bamboo.md

dejoost:
changed the sed arguments, the example was deleting the wrong line and the proposed alternative will always delete the last line, regardless of the number of lines in the file


asbjornu (approved):
Indeed, that's much better!

asbjornu:
Thanks for your contribution, @dejoost! 🙏

PR #2468, opened 2020-11-28
Adds CodeBuild support for builds that are not initiated via a webhook

carcer:
The current CodeBuild support only supports retrieving the branch when a build has been initiated via a webhook, mainly GitHub.  There are other use cases when a this environment variable will not be populated, and branch name is available via the `CODEBUILD_SOURCE_VERSION` variable.  This PR updates the CodeBuild agent to also check this environment variable.

This does result in a difference from the other build agents, as this involves 2 possible environment variables.

## Description
Updates the CodeBuild agent to check the `CODEBUILD_SOURCE_VERSION` if the value from `CODEBUILD_WEBHOOK_BASE_REF` is null or empty.  Overloads the base class so it no longer uses the `EnvironmentVariable` property, as there are 2 variables to be checked.  Throws an exception if the property is accessed

## Motivation and Context
We want to use GItVersion within CodeBuild from builds initiated from Jenkins that are initiated via webhook

## How Has This Been Tested?
Added additional unit tests and confirmed they pass

## Checklist:

### DIFF
--- a/src/GitVersionCore/BuildAgents/CodeBuild.cs
+++ b/src/GitVersionCore/BuildAgents/CodeBuild.cs
@@ -8,7 +8,8 @@ namespace GitVersion.BuildAgents
     public sealed class CodeBuild : BuildAgentBase
     {
         private string file;
-        public const string EnvironmentVariableName = "CODEBUILD_WEBHOOK_HEAD_REF";
+        public const string WebHookEnvironmentVariableName = "CODEBUILD_WEBHOOK_HEAD_REF";
+        public const string SourceVersionEnvironmentVariableName = "CODEBUILD_SOURCE_VERSION";
 
         public CodeBuild(IEnvironment environment, ILog log) : base(environment, log)
         {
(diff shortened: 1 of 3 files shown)


src/GitVersionCore/BuildAgents/CodeBuild.cs
```diff
@@ -20,7 +21,7 @@ public void WithPropertyFile(string propertiesFileName)
             file = propertiesFileName;
         }
 
-        protected override string EnvironmentVariable { get; } = EnvironmentVariableName;
+        protected override string EnvironmentVariable => throw new InvalidOperationException();
```
asbjornu, line 24: Can you please provide the exception with a message explaining why accessing `EnvironmentVariable` is an invalid operation?

asbjornu (changes requested):

src/GitVersionCore/BuildAgents/CodeBuild.cs
carcer: I have updated the PR based on your suggestion.  While considering an appropriate error message, I came to conclusion that an NotSupportedException may be clearer exception to use.

asbjornu (approved):

asbjornu:
Thanks for your contributions, @carcer! 🙏

Issue #2479, opened 2020-12-16
[Bug] Mainline with feature branches throws exception


andyfurniss4:
I'm trying to achieve the following:

Versioning NuGet packages using GitVersion. I do not intend to use develop or release branches because it's just overkill for me and my team and would frankly just be a waste of time and effort.

We want to be able to use master as our released branch. Versions on the master branch will always be format {major}.{minor}.{patch}. We would like to use feature branches for new functionality and would like feature branches to also trigger package builds but of a pre-release format such as {major}.{minor}.{patch}-{tag}.{number}. Feature branches merged into master would increment the minor version automatically. Finally, we would like to use hotfix branches for bug fixes. Hotfixes builds would have the same pre-release format as feature branches but when they're merged into master, they'd increment the patch version.

I have tried using both Mainline and Continuous Deployment modes to achieve this but both have failed me so far. I might just be doing something wrong so please advise as to whether this is possible and how.

The [documentation](https://gitversion.readthedocs.io/en/latest/input/docs/reference/versioning-modes/mainline-development/) shows that my desired outcome should indeed be possible with mainline mode, however this does not seem to be the case in reality.

The documentation contains a graphic which begins like this:

![image](https://user-images.githubusercontent.com/16683916/102403486-3b976800-3fde-11eb-88ff-19ea3a3fd6a4.png)

Firstly, this shows that a feature branch has been tagged with **1.0.1-foo.1+1.** The current mainline functionality creates a tag with name **1.0.1-foo.1** (missing the +1). Is this documentation just old? Has the +1 part been removed now?

Secondly, the diagram shows that once the feature branch is merged into master, master is then tagged with 1.0.1 (same issue with +2). This is not the case, GitVersion outputs version **1.0.1-foo.2** for me.

### Attempt 1
The best I have managed to achieve is by using the following config:

```
mode: mainline
branches:
  master:
    tag: ''
    regex: ^master$|^main$
    increment: minor
  feature:
    increment: minor
    tag: 'alpha'
ignore:
  sha: []
merge-message-formats: {}
```

This results in something like the following:

- master has 1.0.0
- feature branch from master
- commit to feature branch -> 1.1.0-alpha.1 (correct)
- merge feature into master -> 1.1.0 (correct)
- branch hotfix from master
- commit to hotfix branch -> 1.2.0-alpha.1 (incorrect, should be 1.1.1-alpha.1)
- merge hotfix into master -> 1.2.0 (incorrect, should be 1.1.1)

[...]

PR #2480, opened 2020-12-16
Allow use of features and hotfixes with mainline mode

andyfurniss4:
See #2479 for detailed description of issue.

This PR allows the use of feature/hotfix branch pre-release builds when using mainline mode. Previously this was throwing an exception with message "Mainline development mode doesn't yet support pre-release tags on master". This would occur after the following steps:

- master has 1.0.0
- feature branch from master
- commit to feature branch -> 1.1.0-alpha.1 (correct)
- merge feature into master -> Exception "Mainline development mode doesn't yet support pre-release tags on master" as it's trying to tag master with 1.1.0-alpha.2 rather than 1.1.0.

## Description
After getting the base versions in `BaseVersionCalculator.GetBaseVersion()`, check if the configured versioning mode is mainline, if it is then filter out pre-release versions from `baseVersions` list before proceeding.

## Related Issue
#2479

## Motivation and Context
I'm trying to achieve the following:

Versioning NuGet packages using GitVersion. I do not intend to use develop or release branches because it's just overkill for me and my team and would frankly just be a waste of time and effort.

We want to be able to use master as our released branch. Versions on the master branch will always be format {major}.{minor}.{patch}. We would like to use feature branches for new functionality and would like feature branches to also trigger package builds but of a pre-release format such as {major}.{minor}.{patch}-{tag}.{number}. Feature branches merged into master would increment the minor version automatically. Finally, we would like to use hotfix branches for bug fixes. Hotfixes builds would have the same pre-release format as feature branches but when they're merged into master, they'd increment the patch version.

I have tried using both Mainline and Continuous Deployment modes to achieve this but both have failed me so far. I might just be doing something wrong so please advise as to whether this is possible and how.

The [documentation](https://gitversion.readthedocs.io/en/latest/input/docs/reference/versioning-modes/mainline-development/) shows that my desired outcome should indeed be possible with mainline mode, however this does not seem to be the case in reality.

The documentation contains a graphic which begins like this:

image

Firstly, this shows that a feature branch has been tagged with **1.0.1-foo.1+1.** The current mainline functionality creates a tag with name **1.0.1-foo.1** (missing the +1). Is this documentation just old? Has the +1 part been removed now?

Secondly, the diagram shows that once the feature branch is merged into master, master is then tagged with 1.0.1 (same issue with +2). This is not the case, GitVersion outputs version **1.0.1-foo.2** for me.

## How Has This Been Tested?
Added new tests to cover new scenarios and also ensured all existing tests still pass successfully.

## Checklist:

### DIFF
--- a/src/GitVersionCore.Tests/VersionCalculation/BaseVersionCalculatorTests.cs
+++ b/src/GitVersionCore.Tests/VersionCalculation/BaseVersionCalculatorTests.cs
@@ -133,6 +133,45 @@ namespace GitVersionCore.Tests.VersionCalculation
             baseVersion.SemanticVersion.ShouldBe(lowerVersion.SemanticVersion);
         }
 
+        [Test]
+        public void ShouldIgnorePreReleaseVersionInMainlineMode()
+        {
+            var fakeIgnoreConfig = new TestIgnoreConfig(new ExcludeSourcesContainingExclude());
+
+            var lowerVersion = new BaseVersion("dummy", false, new SemanticVersion(1), new MockCommit(), null);
+            var preReleaseVersion = new BaseVersion(
+                "prerelease",
+                false,
+                new SemanticVersion(1, 0, 1)
+                {
+                    PreReleaseTag = new SemanticVersionPreReleaseTag
+                    {
+                        Name = "alpha",
+                        Number = 1
+                    }
+                },
+                new MockCommit(),
+                null
+            );
+
+            var versionCalculator = GetBaseVersionCalculator(contextBuilder =>
+            {
+                contextBuilder
+                    .WithConfig(new Config { VersioningMode = VersioningMode.Mainline, Ignore = fakeIgnoreConfig })
+                    .OverrideServices(services =>
+                    {
+                        services.RemoveAll<IVersionStrategy>();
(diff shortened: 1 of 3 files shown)


asbjornu (approved):

asbjornu:
>Firstly, this shows that a feature branch has been tagged with 1.0.1-foo.1+1.

It's not a git tag, it's the resulting version number from running GitVersion that yields `1.0.1-foo.1+1`.

> The current mainline functionality creates a tag with name 1.0.1-foo.1 (missing the +1). Is this documentation just old? Has the +1 part been removed now?

The `+1` part is the build metadata. More on that [here](https://gitversion.net/docs/more-info/incrementing-per-commit).

> Secondly, the diagram shows that once the feature branch is merged into master, master is then tagged with 1.0.1 (same issue with +2). This is not the case, GitVersion outputs version 1.0.1-foo.2 for me.

I agree that sounds wrong and I think the changes made by you look fine. There's 3 tests failing, though, which we need to figure out before merging. I've tried re-running them, but there's no reaction from AzDO yet.

arturcic:
@asbjornu I rebased the PR on top of master, now build passes

asbjornu:
Thank you for your contributions, @andyfurniss4! 🙏🏼

andyfurniss4:
Brilliant, thanks for finishing this up for me guys. I'm glad it's all working and accepted!

PR #2452, opened 2020-11-12
targetframework net5.0

vip32:
## Checklist:

- [x ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ x] All new and existing tests passed.

PR #2486, opened 2020-12-18
Target .net 5.0

arturcic:
Continuation of #2452 

This one has support for .net 5.0. Added additional artifacts tests on new linux distros (alpine and ubuntu 20.04) - thanks to the latest libgit2sharp preview package.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -103,7 +103,7 @@ jobs:
     strategy:
       matrix:
         os: [linux]
-        targetFramework: [ 3.1 ]
+        targetFramework: [ 3.1, 5.0 ]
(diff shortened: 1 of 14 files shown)


faina09:
so `<TargetFramework>net5.0</TargetFramework>` should be supported, but in my webapp compiled with **Visual Studio Professional 2019 ver 16.8.4** I get the error:


> Error	NU1202	Package GitVersion.Tool 5.6.4 is not compatible with net5.0 (.NETCoreApp,Version=v5.0). Package GitVersion.Tool 5.6.4 supports:
>   - net5.0 (.NETCoreApp,Version=v5.0) / any
>   - netcoreapp3.1 (.NETCoreApp,Version=v3.1) / any

PR #2494, opened 2020-12-30
Allow configuring the branch name extraction per branch.

### LINKED ISSUE #2201: [Bug] Tag: useBranchName isn't working in some cases
**Describe the bug**
I faced issue with property "tag: useBranchName", I'm trying to use it, but it's not working in some cases, for example even if I use tag "tag.{BranchName}" I got "FullSemVer":"3.0.0-tag..1+8".

I'm not sure is it bug or not, but all details below

Configuration:

```yaml
assembly-versioning-scheme: MajorMinorPatchTag
assembly-file-versioning-scheme: MajorMinorPatchTag
assembly-informational-format: '{InformationalVersion}'
tag-prefix: '[vV]'
legacy-semver-padding: 4
build-metadata-padding: 4
commits-since-version-source-padding: 4
commit-message-incrementing: Disabled
continuous-delivery-fallback-tag: prerelease
branches:
  master:
    regex: ^master
    mode: ContinuousDelivery
    tag: useBranchName
    increment: None
    prevent-increment-of-merged-branch-version: true
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
  feature:
    regex: .*
    mode: ContinuousDelivery
    tag: useBranchName
    increment: None
    prevent-increment-of-merged-branch-version: false
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
```

## Expected Behavior
master - tag is "master"
custom/foo - tag is "foo"
foo - tag is "foo"
feature/name - tag is "name"
feature - tag is "feature"

## Actual Behavior
Tag: useBranchName isn't working for master branch and some other branches:
master, custom/foo, foo, feature/name, feature - no tag 
(
useBranchName: "FullSemVer":"3.0.0+16", 
tag.{BranchName}: "FullSemVer":"3.0.0+16",
tag.{BranchName}: "FullSemVer":"3.0.0-tag..1+16" (if we set tag: tag.{BranchName})
)

But, if we change regex in feature branch config to this "^features?[/-]" it will work because of this:
http://regexstorm.net/tester?p=%5efeatures%3f%5b%2f-%5d&i=features%2ffoo
regex cut all matched part, and remnants are used for tag, but if we configure regex to match all branches (like in case above) it will not work correctly

## Context
I would like to add branch name tag to any branch (except master)

## Possible Fix
BaseVersionCalculator.GetBaseVersion return incorrect version

## Steps to Reproduce
Just try to use "useBranchName" or "{BranchName}" in master or configure to match all branch names

## Your Environment
Azure devops (pipelines, nuget/artifacts), nect core 3.1

- Version Used: 5.2.4
- Operating System and version (Windows 10, Ubuntu 18.04): Ubuntu 18.04

bddckr:
## Description

With this change the branch name used for version tags can be configured via a regex.

## Related Issue

This change resolves #2201.

## Motivation and Context

As outlined in #2201 the branch name extraction was hardcoded to just trim everything a branch configuration's `regex` matched. With this change that behavior stays as is by default, but users can override the existing behavior by using a new setting called `trim-regex`. This regex is used to extract the part of a branch name to use.

## How Has This Been Tested?

`OtherBranchScenarios` has two new tests. One ensures the user can use the whole branch name altogether by specifying an empty regex, the other one confirms that only the captured part is used for the branch name for version tags.

This change is backwards compatible, as such all existing tests have been untouched and are still green.

## Checklist:

### DIFF
--- a/docs/input/docs/configuration.md
+++ b/docs/input/docs/configuration.md
@@ -360,6 +360,14 @@ values, but here they are if you need to:
 This is the regex which is used to match the current branch to the correct
 branch configuration.
 
+### trim-regex
+
+This regex is used to capture the name to use for the branch when calculating
+the version tag. This regex is matched on the branch name. The first capture
+group is what defines the resulting name. By default (i.e. when not set), this
+is set to extract by using `{regex}(.*)` where `{regex}` is set from
+[above](#regex).
+
 ### source-branches
 
 Because git commits only refer to parent commits (not branches) GitVersion
--- a/src/GitVersionCore.Tests/IntegrationTests/OtherBranchScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/OtherBranchScenarios.cs
@@ -1,4 +1,7 @@
+using System.Collections.Generic;
 using GitTools.Testing;
+using GitVersion;
+using GitVersion.Model.Configuration;
 using GitVersionCore.Tests.Helpers;
 using LibGit2Sharp;
 using NUnit.Framework;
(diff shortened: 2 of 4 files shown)


asbjornu:
I'm sorry that this went unreviewed before being closed as stale, but doesn't #2649 resolve this?

Issue #2505, opened 2021-01-04
[Bug] When PreventIncrementOfMergedBranchVersion is set to true and a release branch is merged back to develop, the CommitsSinceVersionSource goes down


ruhullahshah:
**Description**
Upon setting `prevent-increment-of-merged-branch-version` to `true` in `GitVersion.yml` and merging a release branch into develop followed by the deletion of the release branch, the `CommitsSinceVersionSource` decreases and is not equal to what is expected.

## Expected Behavior
The `CommitsSinceVersionSource` should not decrease on `develop` upon merging a release branch and deleting it.

## Actual Behavior
The `CommitsSinceVersionSource` decreases on `develop` upon merging a release branch and deleting it.

ruhullahshah:
PR #2506 aims to discuss and address this issue.

PR #2506, opened 2021-01-04
CommitsSinceVersionSource should not decrease upon merging a release branch into develop

ruhullahshah:
`CommitsSinceVersionSource` decreases when a release branch is merged to develop and `PreventIncrementOfMergedBranchVersion` is set to `true`.

For now I have added a minimalistic testcase, exhibiting the problem defined in issue #2505.

### DIFF
--- a/src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
@@ -329,7 +329,6 @@ namespace GitVersionCore.Tests.IntegrationTests
             fixture.AssertFullSemver("1.2.0-alpha.6");
             fixture.AssertFullSemver("1.2.0-alpha.6", config);
 
-            
             config.Branches = new Dictionary<string, BranchConfig>
             {
                 { "develop", new BranchConfig { PreventIncrementOfMergedBranchVersion = true } },
--- a/src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
@@ -303,9 +303,7 @@ namespace GitVersionCore.Tests.IntegrationTests
             fixture.MakeACommit();
             fixture.BranchTo("develop");
             fixture.MakeATaggedCommit("1.0.0");
-            fixture.Repository.MakeCommits(10);
-
-            fixture.AssertFullSemver("1.1.0-alpha.10", config);
+            fixture.Repository.MakeCommits(1);
 
             // Create a release branch and make some commits
             fixture.BranchTo(ReleaseBranch);
@@ -317,12 +315,12 @@ namespace GitVersionCore.Tests.IntegrationTests
             fixture.ApplyTag("v1.1.0");
             fixture.Checkout("develop");
             // Simulate some work done on develop while the release branch was open.
-            fixture.Repository.MakeCommits(10);
+            fixture.Repository.MakeCommits(2);
             fixture.MergeNoFF(ReleaseBranch);
 
             // Version numbers will still be correct when the release branch is around.
-            fixture.AssertFullSemver("1.2.0-alpha.14");
-            fixture.AssertFullSemver("1.2.0-alpha.14", config);
+            fixture.AssertFullSemver("1.2.0-alpha.6");
+            fixture.AssertFullSemver("1.2.0-alpha.6", config);
 
             var versionSourceBeforeReleaseBranchIsRemoved = fixture.GetVersion(config).Sha;
 
@@ -330,8 +328,8 @@ namespace GitVersionCore.Tests.IntegrationTests
             fixture.Repository.Branches.Remove(ReleaseBranch);
             var versionSourceAfterReleaseBranchIsRemoved = fixture.GetVersion(config).Sha;
             Assert.AreEqual(versionSourceBeforeReleaseBranchIsRemoved, versionSourceAfterReleaseBranchIsRemoved);
-            fixture.AssertFullSemver("1.2.0-alpha.14");
-            fixture.AssertFullSemver("1.2.0-alpha.14", config);
+            fixture.AssertFullSemver("1.2.0-alpha.6");
+            fixture.AssertFullSemver("1.2.0-alpha.6", config);
         }
     }
 }
--- a/src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
+++ b/src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
@@ -286,7 +286,7 @@ namespace GitVersionCore.Tests.IntegrationTests
         }
 
         [Test]
-        public void WhenPreventIncrementOfMergedBranchVersionIsSetToFalseCommitsSinceVersionSourceShouldNotGoDownWhenMergingReleaseToDevelop()
+        public void WhenPreventIncrementOfMergedBranchVersionIsSetToFalseForDevelopCommitsSinceVersionSourceShouldNotGoDownWhenMergingReleaseToDevelop()
         {
             var config = new Config
             {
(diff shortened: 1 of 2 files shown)


ruhullahshah:
@asbjornu @arturcic Happy New Year!
Is this expected behavior with the config presented in the testcase? In my opinion, no. I find the documentation on `PreventIncrementOfMergedBranchVersion` to be not so informative to decide this completely on my own.

src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
```diff
@@ -284,5 +284,50 @@ public void PreviousPreReleaseTagShouldBeRespectedWhenCountingCommits()
 
             fixture.AssertFullSemver("1.0.0-alpha.5");
         }
+
+        [Test]
+        public void WhenPreventIncrementOfMergedBranchVersionIsSetToTrueCommitsSinceVersionSourceShouldNotGoDownWhenMergingReleaseToDevelop()
```
asbjornu, line 289:
```suggestion
        public void WhenPreventIncrementOfMergedBranchVersionIsSetToTrue_AndMergingReleaseToDevelop_CommitsSinceVersionSourceShouldNotDecrement()
```

src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
```diff
@@ -284,5 +284,50 @@ public void PreviousPreReleaseTagShouldBeRespectedWhenCountingCommits()
...
+            // Simulate some work done on develop while the release branch was open.
+            fixture.Repository.MakeCommits(10);
+            fixture.MergeNoFF(ReleaseBranch);
+
+            // Version numbers will still be correct when the release branch is around.
+            fixture.AssertFullSemver("1.2.0-alpha.14");
+            fixture.AssertFullSemver("1.2.0-alpha.14", config);
+
+            // Removing the release branch and calculating the version with the config exhibits the problem.
+            fixture.Repository.Branches.Remove(ReleaseBranch);
```
asbjornu, line 328: I agree that removal of the release branch should not change the version number. Can you add an assert of the version source as well, just to ensure that it doesn't change when the release branch is removed? If the version source is the same, the version number should also stay the same.

asbjornu (changes requested):
Happy new year, @ruhullahshah! :tada: I think the test looks fine and that it should be successful. It would be great if the PR could also contain a clarification of the docs on `PreventIncrementOfMergedBranchVersion`.

src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
ruhullahshah: @asbjornu Thanks for the feedback. I would love to enhance the documentation, however, to be honest, I am not so thorough with the use cases that `PreventIncrementOfMergedBranchVersion` targets. Having said that, I am sure constructing the fix for this PR would unearth some of the intended use cases, but please feel free to add your thoughts that you think should be added to the relevant docs.

src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
ruhullahshah: @asbjornu I have extracted the commit hashes before and after the deletion of the release branch and asserted equality. I hope that is what you meant.

asbjornu: Ah yes, I see now the comparison on line 332. Sorry.

ruhullahshah:
### Analysis
|CommitGraph|
|:--:| 
| *Commit Graph* |

The commit graph corresponds to the repository state after merging the release branch to develop (Line 319 of `DevelopScenario.cs`) and commits from this graph will be referred to, for explaining the observed behavior

### 1. Why does the `CommitsSinceVersionSource` decrease in this scenario?
- The base version calculation strategies yield the following list of candidate versions:

```
baseVersions
Count = 5
    [0]: {Fallback base version: 0.1.0 with commit count source 2afdc59845ce1c76975a5d5c09e708d0e2c31de3 | 0.1.0}
    [1]: {Merge message 'Merge branch 'release/1.1.0'': 1.1.0 with commit count source 3dd090ad5f5cdc406c4afe3b13dba4e26d641f7a | 1.1.0}
    [2]: {Git tag '1.0.0': 1.0.0 with commit count source 1d54558c0c51d19286b908749c3f20ed9cf454a8 | 1.1.0}
    [3]: {Git tag 'v1.1.0': 1.1.0 with commit count source 2e7a6e89e58423a1678e651c062555fcd6847174 | 1.2.0}
    [4]: {Git tag '1.0.0': 1.0.0 with commit count source 1d54558c0c51d19286b908749c3f20ed9cf454a8 | 1.1.0}
```
-  Each candidate should be interpreted as {Strategy type with commit count source | IncrementedVersion}. So this means that 
  [3] was calculated using the `TaggedCommitVersionStrategy` with the commit count source being `2e7a6e8` and the incremented version would be `1.2.0`.
- If `PreventIncrementOfMergedBranchVersion` was not set to `true`, the `IncrementedVersion` of [1] would have been `1.2.0` as well. [This](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/VersionCalculation/BaseVersionCalculators/MergeMessageVersionStrategy.cs#L38) line of code disallows the version increment of [1].
- Due to this, [3] is the only candidate selected [here](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/VersionCalculation/BaseVersionCalculator.cs#L71).
- The `CommitsSinceVersionSource` for the selected base version is calculated [here](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/VersionCalculation/MainlineVersionCalculator.cs#L89) with `baseVersionSource` being `2e7a6e8` and `currentCommit` being `fe0c9f2`. Please refer to the `Commit Graph` above.
- [GetCommitLog](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/Core/RepositoryMetadataProvider.cs#L409) returns the following commit log (please refer to `Commit Graph` to verify the commits returned):
```
[2]	fe0c9f2 Merge branch 'release/1.1.0'	LibGit2Sharp.Commit
[1]	9443ab2 Test Commit for file '7059d6d4-dfb8-4fea-bac0-aa4858867ff2' -	LibGit2Sharp.Commit
[0]	ad246ca Test Commit for file '1fbd1f78-7453-488d-b0b8-0ea3c53ef48f' -	LibGit2Sharp.Commit
```
- The count, i.e. 3, of the returned commit log above is used as the `CommitsSinceVersionSource` causing the observed decrease and a deviation from the expected value of 6.
- Please note that the commits on the release branch are completely ignored in the commit log extraction.
- Also, had `PreventIncrementOfMergedBranchVersion` been set to `false` for `develop`, [1] would have been [selected](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/VersionCalculation/BaseVersionCalculator.cs#L80) and with its version source(`3dd090a`) the `CommitsSinceVersionSource` would have been as expected, i.e. 6

### 2. Open questions:
2.1) Is [GetCommitLog](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/Core/RepositoryMetadataProvider.cs#L409) doing the right thing for the scenario mentioned above?
2.2) Should we just avoid setting `PreventIncrementOfMergedBranchVersion`, atleast for develop, update the unit test and the documentation in order to close this PR?

@asbjornu Could you please have a look?

asbjornu:
> 2.1) Is [GetCommitLog](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/Core/RepositoryMetadataProvider.cs#L409) doing the right thing for the scenario mentioned above?

I don't think that looks correct, no. The commits from the release branch should definitely be in there, imho.

> 2.2) Should we just avoid setting `PreventIncrementOfMergedBranchVersion`, atleast for develop, update the unit test and the documentation in order to close this PR?

Which consequences does that have? I assume that will break other tests?

ruhullahshah:
@asbjornu You are welcome.

Here are my further findings:

1) Upon giving it more thought, I think [GetCommitLog](https://github.com/GitTools/GitVersion/blob/da8713ebc1e5ff018e3592cb9a3f8e0b5b9bce02/src/GitVersionCore/Core/RepositoryMetadataProvider.cs#L409) is doing the right thing with the input passed to it, i.e.  with a `baseVersionSource` of `2e7a6e8` and `currentCommit` being `fe0c9f2`, the commit filter should return 3 commits if one looks up the field documentation provided by [LibGit2Sharp](https://github.com/libgit2/libgit2sharp/blob/master/LibGit2Sharp/CommitFilter.cs#L55).
2) Another question that one should be answered is, if the version source of the `TaggedCommitVersionStrategy` is correct or not? In my opinion, it is correct because a tagged commit is a version source.
3) I have updated the documentation for `prevent-increment-of-merged-branch-version` and also updated the unit test to illustrate the behavior of GitVersion in this scenario.

### Conclusion:
I do not think any code changes are necessary, the behavior should be documented (done already) and a unit test documenting this behavior should be added (done already)

src/GitVersionCore.Tests/IntegrationTests/DevelopScenarios.cs
ruhullahshah: @asbjornu Updated the docs

ruhullahshah:
@asbjornu Sorry, missed to answer this:
> Which consequences does that have? I assume that will break other tests?
The consequences of setting `prevent-increment-of-merged-branch-version` to false for `develop` are that we give the version sources provided by the `MergeMessageBaseVersionStrategy` a chance as well and that as exhibited by the updated unit test yields the expected correct `CommitsSinceVersionSource` and no other unit tests break as there are no code changes.

asbjornu (approved):
Perfect! Great work, @ruhullahshah!

ruhullahshah:
@asbjornu Thanks. I think in a Gitflow-based setting, we might also need to check the implications of merging a hotfix/major.minor.patch to `develop` with `prevent-increment-of-merged-branch-version` set to false. I will investigate this and we could merge these changes afterwards.

asbjornu:
> @asbjornu Thanks. I think in a Gitflow-based setting, we might also need to check the implications of merging a hotfix/major.minor.patch to `develop` with `prevent-increment-of-merged-branch-version` set to false. Concretely, if we have `hotfix/1.0.1`, but I still want `develop` to build `1.1.0` after it gets merged to `develop`

Indeed, a test for that would be nice.

> I will investigate this and we could merge these changes afterwards.

Since the build is failing due to what seems to be AzDO outage, please feel free to add to this PR.

ruhullahshah:
@asbjornu I have validated that setting `prevent-increment-of-merged-branch-version` to `false` for `develop` does not cause problems while merging a versioned `hotfix` into `develop` and added a test case for the same.

asbjornu:
Thank you so much for your contributions, @ruhullahshah! ❤️ 🙏🏼

ruhullahshah:
You are welcome @asbjornu

PR #2507, opened 2021-01-04
Add fallback values for GitVersionFileExe and GitVersionAssemblyFile

### LINKED ISSUE #2502: [Bug] Build with GitVersion.MsBuild fails because GitVersionFileExe is not set
Because _GitVersionTask_ seems to be deprecated, I have switched to the _GitVersion.MsBuild_ package. However, the build of my netstandard2.1 library fails with the message

> C:\Users\Thomas\.nuget\packages\gitversion.msbuild\5.6.1\tools\GitVersion.MsBuild.targets(3,38): error MSB4022: The result "" of evaluating the value "$(GitVersionAssemblyFile)" of the "AssemblyFile" attribute in element <UsingTask> is not valid. [C:\priv\GitVersionMsBuildIssue\src\DemoLib\DemoLib.csproj]

I have noticed that in `GitVersion.MsBuild.props`, `GitVersionFileExe` and `GitVersionAssemblyFile` are not set when the target framework is _netstandard2.1_. See https://github.com/GitTools/GitVersion/blob/69431b40003b07c1ad0b679eabeb5d1ada531901/src/GitVersion.MsBuild/msbuild/tools/GitVersion.MsBuild.props#L19-L32

I've fixed it on my local machine by adding following fallback:

```xml
<PropertyGroup Condition="'$(GitVersionFileExe)' == ''">
    <GitVersionFileExe>dotnet $(MSBuildThisFileDirectory)netcoreapp3.1/gitversion.dll</GitVersionFileExe>
    <GitVersionAssemblyFile>$(MSBuildThisFileDirectory)netcoreapp3.1/GitVersion.MsBuild.dll</GitVersionAssemblyFile>
</PropertyGroup>
```

I have created a repo to reproduce the issue at https://github.com/thoemmi/GitVersionMsBuildIssue

This fixed the build. However, on every build a _gitversion.json_ file is generated in each project folder, which is really annoying. Is there another MsBuild property required to supress that?

Regarding the deprecation of _GitVersionTask_: The documentation for [MSBuild Task](https://gitversion.net/docs/usage/msbuild-task) still mentions the _GitVersionTask_ package.

ermshiperete:
This fixes #2502.

### DIFF
--- a/src/GitVersion.MsBuild/msbuild/tools/GitVersion.MsBuild.props
+++ b/src/GitVersion.MsBuild/msbuild/tools/GitVersion.MsBuild.props
@@ -32,6 +32,9 @@
     </PropertyGroup>
 
     <PropertyGroup>
+        <GitVersionFileExe Condition="'$(GitVersionFileExe)' == ''">dotnet $(MSBuildThisFileDirectory)netcoreapp3.1/gitversion.dll</GitVersionFileExe>
+        <GitVersionAssemblyFile Condition="'$(GitVersionAssemblyFile)' == ''">$(MSBuildThisFileDirectory)netcoreapp3.1/GitVersion.MsBuild.dll</GitVersionAssemblyFile>
+
         <DisableGitVersionTask Condition=" '$(DisableGitVersionTask)' == '' ">false</DisableGitVersionTask>
 
         <!-- Property that enables WriteVersionInfoToBuildLog -->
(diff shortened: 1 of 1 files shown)


arturcic:
will this work for net5.0-windows?

ermshiperete:
> will this work for net5.0-windows?

Good question. It will work if .NET Core 3.1 is installed, but if it isn't installed (and you run into the bug in #2502) it probably won't work.

arturcic:
Closing in favor of #2521

Issue #2485, opened 2020-12-18
Support 'assembly-versioning-format' option as a part of '/overrideconfig' command line parameter


nenadvicentic:
It would be very useful to support syntax overriding **'assembly-versioning-format'** as a part of command-line arguments for `/overrideconfig`. At the moment of writing this, only `tag-prefix` option is supported.

**Reason for request:**

Currently, only way to override **'assembly-versioning-format'** option is via configuration files. For example `GitVersion.yml` file in the root of repository. This means that to use custom version format on the company level, one has to commit new file in each repository. It's not possible to control formatting this via CI build-configuration alone.

Example syntax:
```
> gitversion /overrideconfig assembly-versioning-format='{Major}.{Minor}.{Patch}.{env:CI_JOB_ID ?? 0}'
```

Or perhaps double-quotes would be more consistent for the value-part?
`"{Major}.{Minor}.{Patch}.{env:CI_JOB_ID ?? 0}"`

asbjornu:
Extending `/overrideconfig` to support every property supported in `GitVersion.yml` is something I support. I'm kind of surprised only `tag-prefix` is the only supported property. Pull requests are welcome.

nenadvicentic:
After checking the code, I see multiple problems related to full `GitVersion.yml` implementation:

- Some of the properties of `GitVersion.yml` are complex types. For example `Branches` are dictionary of `BranchConfigs, `MergeMessageFormats` or `Ignore`, having multiple properties of their own. 
- Different parsing implementations. `GitVersion.yml` relies on `YamlDotNet` and it's attributes to parse `GitVersion.yml` file and deserialize it in `GitVersion.Model.Configuration.Config` class. Separate implementation would be required for line arguments parsing.
- Keeping all properties of `Config` class in sync with command line arguments would require maintenance two separate lists of properties. Second option would be `Reflection` on `YamlMemberAttribute`s of `Config` class, but that would create other issues.

I would suggest to start smaller: implement only a few properties that make sense to be used from command line and are usually not project specific.

For example: `assembly-informational-format`, `assembly-versioning-format` and `assembly-file-versioning-format`, as versioning schema makes a lot of sense to be applied globally, across all company's projects.

asbjornu:
Thanks for diving into it, @nenadvicentic. That sounds reasonable. Perhaps it makes sense to only allow simple typed properties that can be represented by strings on the command line?

phitattoo:
[...]

PR #2524, opened 2021-01-09
Additional options for command-line switch `/overrideconfig`.

nenadvicentic:
All 'Gitversion.yml` options with simple property type (string, enum, int, bool) are now supported - 21 in total.

## Description
As discussed in issue #2485 I've added support for all writeable properties from `GitVersion.Model.Configuration.Config` class (effectively `Gitversion.yml`) which have simple type (string, Enum, int, bool and their nullable counterparts).

Implementation uses Reflection to parse which properties fit criteria and uses "aliases" for property names provided by `YamlMemberAttribute`'s `Alias` value.

Following `/overrideconfig` options are supported:
1. `assembly-file-versioning-format`
1. `assembly-file-versioning-scheme`
1. `assembly-informational-format`
1. `assembly-versioning-format`
1. `assembly-versioning-scheme`
1. `build-metadata-padding`
1. `commit-date-format`
1. `commit-message-incrementing`
1. `commits-since-version-source-padding`
1. `continuous-delivery-fallback-tag`
1. `increment`
1. `legacy-semver-padding`
1. `major-version-bump-message`
1. `minor-version-bump-message`
1. `mode`
1. `next-version`
1. `no-bump-message`
1. `patch-version-bump-message`
1. `tag-prefix`
1. `tag-pre-release-weight`
1. `update-build-number`

## Related Issue

Support 'assembly-versioning-format' option as a part of '/overrideconfig' command line parameter #2485

## Motivation and Context

In CI environment (GitLab, TeamCity), it makes sense that some of the customizations to the gitversion configuration can be overridden via command line. That allows for applying global changes (for example different assembly-versioning style) on company level, via shared CI build script template, instead of committing same configuration option to every single repository separately in `Gitversion.yml` files.

Since `tag-prefix` option has already been supported, I requested `assembly-versioning-format`, but instead agreed with @asbjornu that it is better make `/overrideconfig` option-parity with `Gitversion.yml` - to the extent that it is possible/reasonable , due to differences in (de)serialization fidelity between `yml` files and command-line arguments.

## How Has This Been Tested?

Code is unit-tested for all currently available `/overrideconfig` options, common errors (wrong key, wrong value, format errors) and setting of multiple options at once.
In addition, I manually tested in windows console as well for couple of common options.

## Checklist:

### DIFF
--- a/docs/input/docs/usage/cli/arguments.md
+++ b/docs/input/docs/usage/cli/arguments.md
@@ -134,7 +134,7 @@ Following options are supported:
 
 Read more about [Configuration](docs/reference/configuration).
 
-It will not change config file 'GitVersion.yml'.
+Using `override-config` on the command line will not change the contents of the config file `GitVersion.yml`.
 
 ### Example: How to override configuration option 'tag-prefix' to use prefix 'custom'
 
--- a/docs/input/docs/usage/cli/arguments.md
+++ b/docs/input/docs/usage/cli/arguments.md
@@ -99,15 +99,14 @@ gitversion init     Configuration utility for gitversion
 
 ## Override config
 
-`/overrideconfig [key=value]` will override appropriate key from 'GitVersion.yml'.
+`/overrideconfig [key=value]` will override appropriate `key` from 'GitVersion.yml'.
 
-Multiple options are separated by semicolon: `key1=value1;key2=value2`.
+To specify multiple options add multiple `/overrideconfig [key=value]` entries:
+`/overrideconfig key1=value1 /overrideconfig key2=value2`.
 
 To have **space characters** as a part of `value`, `value` has be enclosed with double quotes - `key="My value"`.
 
 Double quote character inside of the double quoted `value` has to be be escaped with a backslash '\\' - `key="My \"escaped-quotes\""`.
-
-If closing double-qoute character is missing, it will be implicitily added at the very end of the command line argument - `key="My Value;key2=2(")`.
 
 Following options are supported:
 1. `assembly-file-versioning-format`
@@ -151,6 +150,10 @@ Will pickup up environment variable `BUILD_NUMBER` or fallback to zero for assem
 `GitVersion.exe /output json /overrideconfig assembly-versioning-scheme=MajorMinor`
 
 Will use only major and minor version numbers for assembly version. Assembly build and revision numbers will be 0 (e.g. `1.2.0.0`)
+
+### Example: How to override multiple configuration options
+
+`GitVersion.exe /output json /overrideconfig tag-prefix=custom /overrideconfig assembly-versioning-scheme=MajorMinor`
 
 ### Example: How to override configuration option 'update-build-number'
 
--- a/docs/input/docs/usage/cli/arguments.md
+++ b/docs/input/docs/usage/cli/arguments.md
@@ -131,7 +131,7 @@ Following options are supported:
 1. `tag-pre-release-weight`
 1. `update-build-number`
 
-Read more about [Configuration](docs/reference/configuration).
+Read more about [Configuration](/docs/reference/configuration).
 
 Using `override-config` on the command line will not change the contents of the config file `GitVersion.yml`.
 
--- a/docs/input/docs/usage/cli/arguments.md
+++ b/docs/input/docs/usage/cli/arguments.md
@@ -99,9 +99,63 @@ gitversion init     Configuration utility for gitversion
 
 ## Override config
 
-`/overrideconfig [key=value]` will override appropriate key from
-`GitVersion.yml`.
+`/overrideconfig [key=value]` will override appropriate key from 'GitVersion.yml'.
 
-At the moment only `tag-prefix` option is supported. Read more about
-[Configuration](/docs/reference/configuration). It will not change config file
-`GitVersion.yml`.
+Multiple options are separated by semicolon: `key1=value1;key2=value2`.
+
+To have **space characters** as a part of `value`, `value` has be enclosed with double quotes - `key="My value"`.
+
+Double quote character inside of the double quoted `value` has to be be escaped with a backslash '\\' - `key="My \"escaped-quotes\""`.
+
+If closing double-qoute character is missing, it will be implicitily added at the very end of the command line argument - `key="My Value;key2=2(")`.
+
+Following options are supported:
+1. `assembly-file-versioning-format`
+1. `assembly-file-versioning-scheme`
+1. `assembly-informational-format`
+1. `assembly-versioning-format`
+1. `assembly-versioning-scheme`
+1. `build-metadata-padding`
+1. `commit-date-format`
+1. `commit-message-incrementing`
+1. `commits-since-version-source-padding`
+1. `continuous-delivery-fallback-tag`
+1. `increment`
+1. `legacy-semver-padding`
+1. `major-version-bump-message`
+1. `minor-version-bump-message`
+1. `mode`
+1. `next-version`
+1. `no-bump-message`
+1. `patch-version-bump-message`
+1. `tag-prefix`
+1. `tag-pre-release-weight`
+1. `update-build-number`
+
+Read more about [Configuration](docs/reference/configuration).
+
+It will not change config file 'GitVersion.yml'.
+
+### Example: How to override configuration option 'tag-prefix' to use prefix 'custom'
+
+`GitVersion.exe /output json /overrideconfig tag-prefix=custom`
+
+### Example: How to override configuration option 'assembly-versioning-format'
+
+`GitVersion.exe /output json /overrideconfig assembly-versioning-format="{Major}.{Minor}.{Patch}.{env:BUILD_NUMBER ?? 0}"`
+
+Will pickup up environment variable `BUILD_NUMBER` or fallback to zero for assembly revision number.
+
+### Example: How to override configuration option 'assembly-versioning-scheme'
+
+`GitVersion.exe /output json /overrideconfig assembly-versioning-scheme=MajorMinor`
+
+Will use only major and minor version numbers for assembly version. Assembly build and revision numbers will be 0 (e.g. `1.2.0.0`)
+
+### Example: How to override configuration option 'update-build-number'
+
+`GitVersion.exe /output json /overrideconfig update-build-number=true`
+
+### Example: How to override configuration option 'next-version'
+
+`GitVersion.exe /output json /overrideconfig next-version=6`
--- a/src/GitVersion.App.Tests/ArgumentParserTests.cs
+++ b/src/GitVersion.App.Tests/ArgumentParserTests.cs
@@ -394,7 +394,12 @@ namespace GitVersion.App.Tests
             };
             yield return new TestCaseData("assembly-versioning-scheme=WrongEnumValue")
             {
-                ExpectedResult = $"Could not parse /overrideconfig option: assembly-versioning-scheme=WrongEnumValue. Ensure that 'value' is valid for specified 'key' enumeration: {Environment.NewLine}MajorMinorPatchTag{Environment.NewLine}MajorMinorPatch{Environment.NewLine}MajorMinor{Environment.NewLine}Major{Environment.NewLine}None{Environment.NewLine}"
+                ExpectedResult = $"Could not parse /overrideconfig option: assembly-versioning-scheme=WrongEnumValue. Ensure that 'value' is valid for specified 'key' enumeration: {System.Environment.NewLine}" +
+                    $"MajorMinorPatchTag{System.Environment.NewLine}" +
+                    $"MajorMinorPatch{System.Environment.NewLine}" +
+                    $"MajorMinor{System.Environment.NewLine}" +
+                    $"Major{System.Environment.NewLine}" +
+                    $"None{System.Environment.NewLine}"
             };
         }
 
--- a/src/GitVersion.App.Tests/ArgumentParserTests.cs
+++ b/src/GitVersion.App.Tests/ArgumentParserTests.cs
@@ -559,22 +559,22 @@ namespace GitVersion.App.Tests
         [TestCaseSource(nameof(OverrideconfigWithMultipleOptionsTestData))]
         public void OverrideconfigWithMultipleOptions(string options, Config expected)
         {
-            var arguments = argumentParser.ParseArguments($"/overrideconfig {options}");
+            var arguments = argumentParser.ParseArguments(options);
             arguments.OverrideConfig.ShouldBeEquivalentTo(expected);
         }
 
         private static IEnumerable<TestCaseData> OverrideconfigWithMultipleOptionsTestData()
         {
             yield return new TestCaseData(
-               "tag-prefix=sample;assembly-versioning-scheme=MajorMinor",
+               "/overrideconfig tag-prefix=sample /overrideconfig assembly-versioning-scheme=MajorMinor",
                new Config
                {
                    TagPrefix = "sample",
                    AssemblyVersioningScheme = AssemblyVersioningScheme.MajorMinor,
                }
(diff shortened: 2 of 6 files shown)


nenadvicentic:
@arturcic, @asbjornu  Guys, do you have any estimation when can review for this feature be done?

docs/input/docs/usage/command-line.md
```diff
@@ -119,14 +119,65 @@ Will iterate through each file and update known attributes (`AssemblyVersion`,
...
 `/overrideconfig [key=value]` will override appropriate key from 'GitVersion.yml'.
 
-At the moment only `tag-prefix` option is supported. Read more about [Configuration](../configuration).
+Multiple options are separated by semicolon: `key1=value1;key2=value2`.
+
+To have **space characters** as a part of `value`, `value` has be enclosed with double quotes - `key="My value"`.
+
+Double quote character inside of the double quoted `value` has to be be escaped with a backslash '\\' - `key="My \"escaped-quotes\""`.
+
+If closing double-qoute character is missing, it will be implicitily added at the very end of the command line argument - `key="My Value;key2=2(")`.
```
asbjornu, line 128: Let's just throw an exception when parsing fails instead. Autocorrection just leads to more errors and confusion, imho.

docs/input/docs/usage/command-line.md
```diff
@@ -119,14 +119,65 @@ Will iterate through each file and update known attributes (`AssemblyVersion`,
...
+1. `next-version`
+1. `no-bump-message`
+1. `patch-version-bump-message`
+1. `tag-prefix`
+1. `tag-pre-release-weight`
+1. `update-build-number`
+
+Read more about [Configuration](../configuration).
 
 It will not change config file 'GitVersion.yml'.
```
asbjornu, line 155:
```suggestion
Using `override-config` on the command line will not change the contents of the config file `GitVersion.yml`.
```

asbjornu (changes requested):
Excellent stuff!

docs/input/docs/usage/command-line.md
```diff
@@ -119,14 +119,65 @@ Will iterate through each file and update known attributes (`AssemblyVersion`,
 
 `/overrideconfig [key=value]` will override appropriate key from 'GitVersion.yml'.
 
-At the moment only `tag-prefix` option is supported. Read more about [Configuration](../configuration).
+Multiple options are separated by semicolon: `key1=value1;key2=value2`.
```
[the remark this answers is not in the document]
nenadvicentic: This is well-known aproach used by MSBuild command line, to accept multiple key-value pairs with `;`. For example:

> msbuild /target:build /property:Configuration=Release;TargetFramework=net5.0

Multiple `/property` entries are possible in MSBuild as well. So since your current command-line syntax reassembles MSBuild syntax quite and a lot of developers will use it in the same build scripts, I thought it's good idea to follow already established syntax.

Regarding your concrete suggestion, space character cannot work as a separator for key-values, because key would be confused for next command line switch.

Adding multiple occurrences of `/overrideconfig` argument would be possible, there is an issue with current implementation as well.

My parsing of additional overrideconfig keys hooked into existing implementation, at the point where command line switches are already parsed. Namely inside of the method: `ParseOverrideConfig` of `ArgumentParser`. As far as I can tell, accepting multiple occurrences of `/overrideconfig` would require rewrite of complete command line parsing logic. Perhaps someone who is familiar with current implementation can give us 2nd opinion on this change?

docs/input/docs/usage/command-line.md
nenadvicentic: For this behavior I also consulted MSBuild implementation. Specifically `[QuotingUtilities.cs](https://github.com/dotnet/msbuild/blob/e3eac4c45510dcb85a4a2aea323dae81189f60bb/src/Shared/QuotingUtilities.cs)` class:

```
        /*
         * Quoting Rules:
         * 
         * A string is considered quoted if it is enclosed in double-quotes. A double-quote can be escaped with a backslash, or it
         * is automatically escaped if it is the last character in an explicitly terminated quoted string. A backslash itself can
         * be escaped with another backslash IFF it precedes a double-quote, otherwise it is interpreted literally.
         * 
         * e.g.
         *      abc"cde"xyz         --> "cde" is quoted
         *      abc"xyz             --> "xyz" is quoted (the terminal double-quote is assumed)
         *      abc"xyz"            --> "xyz" is quoted (the terminal double-quote is explicit)
         * 
         *      abc\"cde"xyz        --> "xyz" is quoted (the terminal double-quote is assumed)
         *      abc\\"cde"xyz       --> "cde" is quoted
         *      abc\\\"cde"xyz      --> "xyz" is quoted (the terminal double-quote is assumed)
         * 
         *      abc"""xyz           --> """ is quoted
         *      abc""""xyz          --> """ and "xyz" are quoted (the terminal double-quote is assumed)
         *      abc"""""xyz         --> """ is quoted
         *      abc""""""xyz        --> """ and """ are quoted
         *      abc"cde""xyz        --> "cde"" is quoted
         *      abc"xyz""           --> "xyz"" is quoted (the terminal double-quote is explicit)
         * 
         *      abc""xyz            --> nothing is quoted
         *      abc""cde""xyz       --> nothing is quoted
         */
```

I assumed this is expected behavior when parsing command line quotes.

docs/input/docs/usage/command-line.md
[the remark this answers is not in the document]
asbjornu: > This is well-known aproach used by MSBuild command line, to accept multiple key-value pairs with `;`

Yea, I think MSBuild is a horrendous example of command line interface and GitVersion should not be modelled by it. The more different from MSBuild we can make GtiVersion's CLI, the better. 😄 

> Multiple `/property` entries are possible in MSBuild as well.

Good. 👍🏼 

> Regarding your concrete suggestion, space character cannot work as a separator for key-values, because key would be confused for next command line switch.

Keys and values should be separated with `=`. But key-value-pairs should be separated by space.

> Perhaps someone who is familiar with current implementation can give us 2nd opinion on this change?

If it's too involved, let's postpone this to GitVersion 6 where we will redo the entire CLI (see #2275).

docs/input/docs/usage/command-line.md
asbjornu: It may be expected when working with MSBuild, but GitVersion is not an MSBuild utility (although it has an MSBuild abstraction), but its own self-sustained tool. It's more important that GitVersion provides a good, clean interface familiar to developers used to a wide variety of CLI tools and not just Microsoft developers only familiar with MSBuild.

asbjornu (changes requested):

docs/input/docs/usage/command-line.md
nenadvicentic: I think MSBuild did try not invent new way of parsing, just modeled their implementation on most common scenario, but this is no problem. We can easily throw the exception.

docs/input/docs/usage/command-line.md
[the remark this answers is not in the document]
nenadvicentic: > Yea, I think MSBuild is a horrendous example of command line interface and GitVersion should not be modelled by it. The more different from MSBuild we can make GtiVersion's CLI, the better. 😄

Wherever I found command line switch that accept multiple values, they are splitted with `;` or `,` or some other funky character. For example `:` in `--resolve`:

`curl --resolve 'yada.com:443:127.0.0.1' https://yada.com/something`

> Keys and values should be separated with =. But key-value-pairs should be separated by space.

That's what I was referring too. Normally, arguments are split by space characters and we get array:

`switch value switch value switch value (etc)`

If we split key-value pairs with spaces, we suddenly get ambiguity, which argument is command line switch, which one is value of command line switch, which one is key value pair.

`switch value1 value2 switch value switch value`

It creates complexities for parser. Add on top of that that some switches might repeat and it becomes a mess. And I also think, maybe it's on the surface prettier, but it quickly becomes less human-readable to figure what is switch and what is key-value pair in a long command.

> If it's too involved, let's postpone this to GitVersion 6 where we will redo the entire CLI

Let me explain why I think it's this is a big change. Currently, repeated parameter `/output` will just be ignored:

`gitversion /output json /output buildserver`

In implementation where multiple occurrences are valid, you will have 2 cases now. One case where, like `/output` one value of parameter overrides another value. And second case where you want to merge key-value pairs from multiple occurrences. If key-value appear multiple times, then they should be overwritten/thrown away.

Regardless where this option is introduced, in GitVersion 5 or 6, it will introduce additional complexity that is currently not required.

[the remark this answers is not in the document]
asbjornu: > `switch value switch value switch value (etc)`

No, I was thinking like this:

```shell
gitversion /overrideconfig key1=value1 key2=value2 key3=value3
```

Each key-value pair is separated with space, key and value is separated by `=`.

docs/input/docs/usage/command-line.md
asbjornu: Exception please 👍🏼

docs/input/docs/usage/command-line.md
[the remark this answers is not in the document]
nenadvicentic: Yes, that we are talking about the same thing. This is 2nd example I made:

`switch value1 value2 switch value switch value`

Or translated to your example:
`gitversion /overrideconfig key1=value1 key2=value2 /ouput json /output buildserver`

The way how arguments are parsed universally, it's the array: `/overrideconfig`, `key1=value`, `key2=value`, `/output`, etc... Further, order is assumed/convention, option is followed by the value. Following that logic, `key2=value` ends up in a position of command line option and `/output` in the place of value.

What you are proposing introduces variable positions of options, values and child-values. And that introduces all kinds of combinatorial issues and need for top level parser to know all valid values of child-options to do parse this correctly You end up with mixing different concerns on the top level.

docs/input/docs/usage/command-line.md
nenadvicentic: While implementing throwing of exception, I discovered why MSBuild parsing of quotes was implemented with implicit termination at the end of string. It has same behavior as .NET Command Line applications. I suspect this to be convention and that other command-line apps have similar behavior (eg. C++ command line apps).

If I pass following, real-life parameters to the GitVersion.exe with double-quote missing at the very end:

`/overrideconfig assembly-versioning-format="{Major}.{Minor}.{Patch}.{env:BUILD_NUMBER ?? 0}`

`Main(string[] args)` method receives `args` where open double-quote is implicitly terminated and removed. More importantly, last parameter has spaces ` ?? 0`, but was not split, proving that the double-quote was indeed assumed at the very end of the line. Screenshot will make it more clear:

image

So, once I have implemented throwing of the exception, what happens is that unit-tests are throwing, but same parameters missing closing double-quotes in the command line just work normally. Therefore we just introduced inconsistency.

Additionally, now that I looked up the source code again after a few weeks - reason why I looked at MSBuild command-line parsing logic in the first place was that you have MSBuildTask implementation of `GitVersion`. This one can be called via `msbuild`, something like:

`msbuild /target:GitVersion /property:key=value;key=value`

So, my thinking was that more similarities in format would help users.

docs/input/docs/usage/command-line.md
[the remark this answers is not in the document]
asbjornu: I see. It's not correct that the top-level parser needs to know what the underlying switches or values are, it just needs to know the difference between switches (arguments prefixed by `/`, `-` or `--`) and values (one or more arguments following a switch).

The way to implement this is to first split by space, then loop through each argument and stack switches with values before the values are passed to the switch handler. In the case of `/overrideconfig`, the values passed would be the array `["key1=value1", "key2=value2", "key3=value3]` and then it would be up to the switch handler to split the keys and values.

 understand that the current implementation is too naïve to take this into consideration. I still want to investigate alternatives to semicolon, though. Ideas?

docs/input/docs/usage/command-line.md
asbjornu: If it's difficult keeping the behavior consistent, then don't throw an exception. But I'm adamant we should not use `;` as a separator if we can avoid it.

arturcic: > If it's difficult keeping the behavior consistent, then don't throw an exception. But I'm adamant we should not use `;` as a separator if we can avoid it.

In any case we need to have a solution that will be easy to migrate in v6. There we use the System.CommandLine

nenadvicentic: Guys,

Requirements you put forward do not add up for me, from what I see looking at the current code.

> But I'm adamant we should not use ; as a separator if we can avoid it.

You can use other special character, instead of `;` for the sake of being different from MSBuild, but I ';' is probably the cleanest of all the special separator characters.

In order to use space character complete parsing logic of `ArgumentParser` would have to be rewritten.

Current `ArgumentParser` implementation has a lot of explicit branching decisions per switch value. Each switch value has it's own parsing logic and there is very little general-purpose parsing code. Check `ParseArguments` method. Further, there is even explicit check `EnsureArgumentValueCount` to ensure each switch accepts single argument, which is in contradiction with space character being used to separate child-options for a switch. We would screw up existing validation logic.

Regarding portability of this pull request to v6 command line parsing:

> In any case we need to have a solution that will be easy to migrate in v6. There we use the [System.CommandLine](https://github.com/dotnet/command-line-api)

All that would have to be done to make current `ParseOverrideConfig(Arguments arguments, string value)` method work in version 6 would be to make it accept array `string[] values` instead of value, and remove line inside of method that splits `value` based on `;` character.

> If it's too involved, let's postpone this to GitVersion 6 where we will redo the entire CLI (see #2275).

I'm afraid that postponing this change to GitVersion 6 would mean throwing current implementation away (code + unit-tests) and start everything from scratch.  This pull request started from idea to add single additional `/overrideconfig` option, extended to add 26 extra options, which based on my free time, I could do. For re-implementing all command line parsing in v5 or re-implementing pull request in v6, I cannot commit myself.

My proposal would be to integrate this as is. It will be breaking change to switch from `key1=value1;key2=value2` to space characters `key1=value1 key2=value2` in v6, but you will have breaking changes in any case, because complete parsing logic will be different.

asbjornu: Sorry for the late reply. I wonder, is it also impossible to repeat the `/overrideconfig` argument?

nenadvicentic: I commented on the possibility of repeated options here:
https://github.com/GitTools/GitVersion/pull/2524#discussion_r565340553

The way parsing is implemented at the moment, second `/output` is not taken into consideration:

```
gitversion /output json /output buildserver
```

As far as I investigated the code, allowing multiple instances of the same option would also be global parser change.

asbjornu: `-output json -output buildserver` was implemented in #2128 and must be working as the following test (and many more) is successful?

https://github.com/GitTools/GitVersion/blob/1fd097957a0160ab0558867d36fe89917468b17b/src/GitVersion.App.Tests/ArgumentParserTests.cs#L173-L180

nenadvicentic: @asbjornu This was very helpful example. After investigation implementation of this example, I managed to implement specifying multiple `overrideconfig` options using repeated `/overrideconfig` arguments, with only a few changes.

asbjornu (approved):
I think this looks good now, thanks for adjusting the argument format. Rebase against `main`, resolve the conflicts and this should be good to go! 👏🏼

nenadvicentic:
> I think this looks good now, thanks for adjusting the argument format. Rebase against `main`, resolve the conflicts and this should be good to go! 👏🏼

I re-based the branch. It should be fine now. The unstable unit-tests (the once that failed now) seem to be something I've picked up during rebase from the main branch, unrelated code changes in this pull request.

arturcic (approved):

Issue #2456, opened 2020-11-16
[Bug] 5.5.1 Continue failing to get version when building from tag


farlop:
**Describe the bug**
As fix for bug #2074 was released in v5.5.1, I've checked that maybe I'm running in a different scenario:

We are using TrunkBasedDevelopment (TBD) with a configuration file like:
```
mode: ContinuousDeployment
assembly-informational-format: '{FullSemVer}'
branches:
  develop: 
    increment: Minor
    regex: master    
    tracks-release-branches: true
    prevent-increment-of-merged-branch-version: true
    track-merge-target: false
    tag: 'dev'
  master:
    regex: none
  hotfix:
    regex: ^fix(es)?[/-]
    source-branches: ['develop']
  feature:
    regex: ^features?[/-]
    source-branches: ['develop']
  release:
    regex: ^releases?[/-]    
    source-branches: ['develop']
    increment: Patch
    tag: 'qa'
```

## Expected Behavior
The build for a tagged commit shoud be calculated as **1.0.4**

## Actual Behavior
The build number is being calculated as **1.1.0-tags-1-0-4.1+15**
```

--------------------------
Installing GitVersion.Tool version 5.x
--------------------------
Herramienta encontrada en caché: GitVersion.Tool 5.5.1 x64
Anteponiendo la variable de entorno de la ruta con el directorio: C:\agents\A\_work\_tool\GitVersion.Tool\5.5.1\x64
[command]C:\agents\A\_work\_tool\GitVersion.Tool\5.5.1\x64\dotnet-gitversion.exe C:/agents/A/_work/56/s /output buildserver /nofetch /config C:\agents\A\_work\56\s\gitversion.yml
INFO [11/16/20 13:17:40:44] Working directory: C:/agents/A/_work/56/s
INFO [11/16/20 13:17:40:50] Branch from build environment: refs/tags/1.0.4
INFO [11/16/20 13:17:40:50] Project root is: C:\agents\A\_work\56\s\
INFO [11/16/20 13:17:40:50] DotGit directory is: C:\agents\A\_work\56\s\.git
INFO [11/16/20 13:17:40:50] Begin: Normalizing git directory for branch 'refs/tags/1.0.4'
  INFO [11/16/20 13:17:40:58] One remote found (origin -> 'https://xxxxxx.xxx.es/tfs/xxxx/xxxx/_git/VC-xxxx').
[...]

PR #2577, opened 2021-02-01
(GH-2456) Correctly assert version number for tagged commit

gep13:
## Description

When running on a CI system, like GitHub Actions, when building a tagged commit, rather than asserting something like `0.2.0`, which is expected, GitVersion asserts a version number like `0.3.0-tags-0-2-0-0001`.  An example of this can be seen here:

https://github.com/cake-contrib/Cake.StrongNameSigner/runs/1779759425?check_suite_focus=true#step:5:217

A fix has been made to the code to only increment the semantic version number when on a tagged commit, if the current sha of the repository is different to the sha of the tagged semantic version.

This was discussed on a Twitch stream with @arturcic 

## Related Issue

#2456 

## How Has This Been Tested?

I have tested this code locally on my machine, and added unit tests to cover the code that has been changed.

## Checklist:

### DIFF
--- a/src/GitVersion.App.Tests/TagCheckoutInBuildAgentTests.cs
+++ b/src/GitVersion.App.Tests/TagCheckoutInBuildAgentTests.cs
@@ -15,6 +15,18 @@ namespace GitVersion.App.Tests
     public class TagCheckoutInBuildAgentTests
     {
         [Test]
+        public async Task VerifyTagCheckoutOnAzurePipelines()
+        {
+            var env = new Dictionary<string, string>
+            {
+                { AzurePipelines.EnvironmentVariableName, "true" },
+                { "BUILD_SOURCEBRANCH", "refs/tags/0.2.0" },
+            };
+
+            await VerifyTagCheckoutVersionIsCalculatedProperly(env);
+        }
+
+        [Test]
         public async Task VerifyTagCheckoutOnGitHubActions()
         {
             var env = new Dictionary<string, string>
@@ -23,6 +35,11 @@ namespace GitVersion.App.Tests
                 { "GITHUB_REF", "ref/tags/0.2.0" },
             };
 
+            await VerifyTagCheckoutVersionIsCalculatedProperly(env);
+        }
+
+        private static async Task VerifyTagCheckoutVersionIsCalculatedProperly(Dictionary<string, string> env)
+        {
             using var fixture = new EmptyRepositoryFixture();
             var remoteRepositoryPath = PathHelper.GetTempPath();
             RepositoryFixtureBase.Init(remoteRepositoryPath);
--- a/src/GitVersion.App.Tests/TagCheckoutInBuildAgentTests.cs
+++ b/src/GitVersion.App.Tests/TagCheckoutInBuildAgentTests.cs
@@ -1,0 +1,60 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using GitTools.Testing;
+using GitVersion.BuildAgents;
+using GitVersion.Core.Tests.Helpers;
+using LibGit2Sharp;
+using NUnit.Framework;
+using Shouldly;
+
+namespace GitVersion.App.Tests
+{
+    [TestFixture]
+    public class TagCheckoutInBuildAgentTests
+    {
+        [Test]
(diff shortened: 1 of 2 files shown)


src/GitVersion.App.Tests/TagCheckoutInBuildAgentTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                remoteRepository.Config.Set("user.name", "Test");
+                remoteRepository.Config.Set("user.email", "test@email.com");
+                fixture.Repository.Network.Remotes.Add("origin", remoteRepositoryPath);
+                Console.WriteLine("Created git repository at {0}", remoteRepositoryPath);
+                remoteRepository.MakeATaggedCommit("0.1.0");
+                Commands.Checkout(remoteRepository, remoteRepository.CreateBranch("develop"));
+                remoteRepository.MakeACommit();
+                Commands.Checkout(remoteRepository, remoteRepository.CreateBranch("release/0.2.0"));
+                remoteRepository.MakeACommit();
+                Commands.Checkout(remoteRepository, "main");
```
arturcic, line 40:
```suggestion
                Commands.Checkout(remoteRepository, TestBase.MainBranch);
```

arturcic (changes requested):

src/GitVersion.App.Tests/TagCheckoutInBuildAgentTests.cs
gep13: Doh! Missed that one, thanks for catching that!  Fixed up.

arturcic:
@gep13 I wonder if we need to add a similar test for AzurePipelines

gep13:
I actually had the same thought. I can break the test up in a similar way to the PullRequestBuildAgentTewt class.

Just away to make the kids supper just now, but should be able to get this updated later tonight.

Does the single test, and code change make sense?

gep13:
In general I mean? Is this a sensible change?

arturcic:
> In general I mean? Is this a sensible change?

I think it makes sense. Would be great to add an additional test in the `Core.Test` project as the change is made in the `Core`

This one is more like an integration test.

arturcic:
> Would be great to add an additional test in the `Core.Test`

in `TagCheckoutScenarios`

gep13:
I started looking there, but the problem only occurs when the environment variables come into play, which only seemed to be in the integration tests.

If the environment variables aren't passed in, the correct version number is asserted without the code change.

arturcic:
> I started looking there, but the problem only occurs when the environment variables come into play, which only seemed to be in the integration tests.
> 
> If the environment variables aren't passed in, the correct version number is asserted without the code change.

But is the environment variable is the issue or it's value? I tend to think it's the value, so in that case we might be able to reproduce the value in a test

gep13:
I am not sure that I follow 😢 

This test: https://github.com/GitTools/GitVersion/blob/6c8e199897e1693c4d148ccf15a4c426b3a13cb6/src/GitVersion.Core.Tests/IntegrationTests/TagCheckoutScenarios.cs#L34-L49 essentially tests the same thing as I am testing with my new test.  This test works when running GitVersion locally, without any CI's involved.

As soon as you add a CI into the mix, the wrong version number is asserted.

I believe this is due to the addition work that it done to normalize the repository prior to asserting the version number.

arturcic:
> I am not sure that I follow 😢
> 
> This test:
> 
> https://github.com/GitTools/GitVersion/blob/6c8e199897e1693c4d148ccf15a4c426b3a13cb6/src/GitVersion.Core.Tests/IntegrationTests/TagCheckoutScenarios.cs#L34-L49
> 
> essentially tests the same thing as I am testing with my new test. This test works when running GitVersion locally, without any CI's involved.
> As soon as you add a CI into the mix, the wrong version number is asserted.
> 
> I believe this is due to the addition work that it done to normalize the repository prior to asserting the version number.

Ok, you're right. Will you add the AzurePipelines test as well?

gep13:
>@arturcic said...
Will you add the AzurePipelines test as well?

Yip, I can do that, but it won't be until later this evening at the earliest.

gep13:
@arturcic an additional test has been added to cover Azure Pipelines.

arturcic (approved):

arturcic:
@gep13  thank you so much for your contributions 👍

Issue #1381, opened 2018-03-11
GitVersionTask Parallel restore issue when multi targeting


dazinator:
I am noticing this issue in my NETSTANDARD PR branch, but I think it might impact master as well (although not tested master directly I am inferring it is a problem)

The issue is, the following target runs when restoring:

```
  <Target Name="WriteVersionInfoToBuildLog" BeforeTargets="DispatchToInnerBuilds;GenerateNuspec;_GenerateRestoreProjectSpec" Condition="$(WriteVersionInfoToBuildLog) == 'true'">
    <WriteVersionInfoToBuildLog SolutionDirectory="$(GitVersionPath)" NoFetch="$(GitVersion_NoFetchEnabled)"/>
  </Target>
```

The `WriteVersionInfoToBuildLog` msbuild task ends up calling `ExecuteGitVersion` which ends up using `libgit2sharp`.

If you are restoring a solution with dotnet restore (I am not sure about /T:Restore but I assume it's the same?) restoration happens in parallel unless you opt out via `dotnet restore --disable-parallel` switch.

Because of this, I am seeing libgit2sharp throw errors about locking as it's being used in parallel to inspect the same repo:

https://ci.appveyor.com/project/dazinator/gitversion-pr1269/build/0.1.1+5.build.1#L32

dazinator:
This might be down to an error in my PR branch, I'll reopen once I know for sure it's an issue in master.

dazinator:
This issue seems only relevant to dotnet build, if you use msbuild /t:build or msbuild /t:Restore its fine.

I beleive this is because dotnet build now implictly does a dotnet restore. And dotnet restore is done in parralel unless you opt out using --disable-parallel. 

This causes a problem for our build task, libgit2sharp keeps locks and doesn't like to be used concurrently.

Therefore in [my sample repo](https://ci.appveyor.com/project/dazinator/gitversion-pr1269/build/0.1.1+9.build.5) in order to work around this I have had to use `dotnet restore --disable-parallel`.

dazinator:
This should probably result in a docs change - i.e gitversiontask doesn't support building projects in parallel, which is a build switch that should be disabled

ivan-danilov:
Not supporting parallel build is often a show-stopper for big solutions. Turning parallel build off easily makes 3-minutes build into 15-minutes. Do you take votes on what is the most important? Count mine here ;)

ivan-danilov:
[...]

PR #2410, opened 2020-09-23
Lock file

asbjornu:
A rebase and `dotnet format` fixed version of #2333. Hopefully the tests turn green so we can make this available in a beta for testing.

Resolves #2333. Hopefully fixes #1031. Hopefully fixes #1381.

PR #2669, opened 2021-04-21
Add global named mutex to prevent concurrent use

bording:
When using GitVersion.MsBuild in multiple projects, you can get concurrent invocations of the GitVersion app. This can result in exceptions when normalization is run. This PR attempts to fix this problem by introducing a global mutex to ensure only a single instance of GitVersion is running at a time, per working directory.

## Description
I have added a named mutex in the `GitVersionExecutor.RunGitVersionTool` method, using the `WorkingDirectory` value as the name. Giving it a name along with the `Global\` prefix means that the mutex will be shared across all processes on the machine. 
This ensures that for a given working directory, only one instance of GitVersion will be executing at a time.

NOTE: Mutex names cannot have a backslash in them, so that is why I am removing them before using it as the name.

## Related Issues

https://github.com/GitTools/GitVersion/issues/1381
https://github.com/GitTools/GitVersion/issues/1031

## Motivation and Context
This change is required to prevent the various exceptions that can occur when the normalization process is run concurrently, which is likely to happen when more than one project in a solution is using GitVersion.MsBuild.

These errors include:
- `LibGit2Sharp.LockedFileException: failed to lock file`
- `LibGit2Sharp.NameConflictException: failed to write reference `

## Checklist:

### DIFF
--- a/src/GitVersion.App/GitVersionExecutor.cs
+++ b/src/GitVersion.App/GitVersionExecutor.cs
@@ -57,7 +57,7 @@ namespace GitVersion
 
         private int RunGitVersionTool(GitVersionOptions gitVersionOptions)
         {
-            var mutexName = gitVersionOptions.WorkingDirectory.Replace("\\", "");
+            var mutexName = gitVersionOptions.WorkingDirectory.Replace(Path.DirectorySeparatorChar.ToString(), "");
             using var mutex = new Mutex(true, $@"Global\{mutexName}", out var acquired);
 
             try
--- a/src/GitVersion.App/GitVersionExecutor.cs
+++ b/src/GitVersion.App/GitVersionExecutor.cs
@@ -1,6 +1,7 @@
 using System;
 using System.IO;
 using System.Reflection;
+using System.Threading;
 using GitVersion.Configuration;
 using GitVersion.Extensions;
 using GitVersion.Logging;
@@ -56,8 +57,16 @@ namespace GitVersion
 
         private int RunGitVersionTool(GitVersionOptions gitVersionOptions)
         {
+            var mutexName = gitVersionOptions.WorkingDirectory.Replace("\\", "");
+            using var mutex = new Mutex(true, $@"Global\{mutexName}", out var acquired);
+
             try
             {
+                if (!acquired)
+                {
+                    mutex.WaitOne();
+                }
+
                 var variables = gitVersionCalculateTool.CalculateVersionVariables();
 
                 var configuration = configProvider.Provide(overrideConfig: gitVersionOptions.ConfigInfo.OverrideConfig);
@@ -91,6 +100,10 @@ namespace GitVersion
                     log.Error("Couldn't dump the git graph due to the following error: " + dumpGraphException);
                 }
                 return 1;
+            }
+            finally
+            {
+                mutex.ReleaseMutex();
             }
 
             return 0;


jetersen:
@bording a previous PR already did a lot of work: https://github.com/GitTools/GitVersion/pull/2410

bording:
@jetersen I saw that, but what I'm proposing is much simpler. There's no need for lock files, just need a mutex to prevent concurrent usage.

bording:
I see that there are some build failures, but I think that they are unrelated to my changes. The appear to be related to some sort of cake problem?

I have built the code locally with `dotnet build` on Windows and Linux, and it works just fine.

dazinator:
Thanks @bording - a named mutex would also be my preferred solution to this, over lock files. File locking is difficult to do correctly with certain file systems throwing edge cases etc. Named mutexes are simpler (specifically designed for this use case) and should work on all supported platforms.

teroneko:
I would not recommend to use named mutex. What I've heard it is not well supported on linux. The file lock is the most native lock mechanism you can use. The universal file lock is already written in 30 lines (compare #2333):
```
        /// <summary>
        /// Try to acquire lock on file but only as long the file stream is opened.
        /// </summary>
        /// <param name="filePath">The path to file that is tried to get locked..</param>
        /// <param name="fileStream">The locked file as file stream.</param>
        /// <param name="fileMode">The file mode when opening file.</param>
        /// <param name="fileAccess">The file access when opening file.</param>
        /// <param name="fileShare">The file share when opening file</param>
        /// <returns>If true the lock acquirement was successful.</returns>
        public bool TryAcquire(string filePath, [MaybeNullWhen(false)] out FileStream fileStream, FileMode fileMode = DefaultFileMode,
            FileAccess fileAccess = DefaultFileAccess, FileShare fileShare = DefaultFileShare)
        {
            filePath = filePath ?? throw new ArgumentNullException(nameof(filePath));

            try {
                fileStream = fileSystem.Open(filePath, fileMode, fileAccess, fileShare);
                // Add UNIX support (reference https://github.com/dotnet/coreclr/pull/8233).
                fileStream.Lock(0, 0);
                return true;
            }
            // The IOException does specify that the file could not been accessed because 
            // it was partially locked. All other exception have to be handled by consumer.
            // 
            // See references:
            // https://docs.microsoft.com/en-US/dotnet/api/system.io.file.open?view=netcore-3.1 (exceptions)
            // https://docs.microsoft.com/en-US/dotnet/api/system.io.filestream.lock?view=netcore-3.1#exceptions
            catch (Exception error) when (error.GetType() == typeof(IOException)) {
                fileStream = null;
                return false;
            }
        }
```
It is simple too. I wonder, why no one can or could finished my tiny implementation. The code around the implementation of the PR #2333 is just fancy thread stuff to allow thread-safety and nested locks of one file when such lock is used in complex scenario. In the case of GitVersion.MSBuild you just need to lock the same file at program start and release at program end. Ah and some logic to wait for lock release (e.g. `SpinWait`) and timeout (that `SpinWait` also support). You are to take the part I contributed via #2333  to solve the problem. 🙂

bording:
> I would not recommend to use named mutex. What I've heard it is not well supported on linux.

On the other hand, I have used these before on both macOS and Linux, and they've worked just fine. If there is a problem with them, then that's a bug to file with Microsoft to fix in .NET.

I don't really see the point of reinventing synchronization primitives when the framework and OS provides what we need here.

src/GitVersion.App/GitVersionExecutor.cs
```diff
@@ -56,8 +57,16 @@ public int Execute(GitVersionOptions gitVersionOptions)
 
         private int RunGitVersionTool(GitVersionOptions gitVersionOptions)
         {
+            var mutexName = gitVersionOptions.WorkingDirectory.Replace("\\", "");
```
danielmarbach, line 60: If there is no need to make them readable of some sort you could also just hash the path into a guid

dazinator (approved):

bording:
> Since the build is failing on the `Clean` step, I wonder whether the mutex may be at fault, causing some sort of file system lock preventing stuff from being deleted?

@asbjornu I don't really understand Cake, so I can't really follow what the build log failures mean. To me, it just looks like some unrelated failure.

Can you point me to something that shows what's actually failing and maybe I can try to figure it out?

asbjornu:
@gep13 and @arturcic are you able to assist @bording with debugging this?

arturcic:
> @gep13 and @arturcic are you able to assist @bording with debugging this?

I will try to repro locally

bording:
I took another look at the failing build logs, and finally found the error message lurking in there. 

I have pushed up another commit the fixes the problem, and everything is green now!

arturcic:
No, Azure Pipelines are failing pretty often.  I was considering to switch to only GitHub Actions, but for now keeping Azure Pipelines

bording:
@arturcic Thanks for merging this. Do you have an ETA for this being released? We're in the process of converting our repos to GitHub Actions and have been hitting the problem this PR solves quite a lot!

arturcic:
@bording version 5.6.9 was released

joaorosado:
Hi,

This looks like it is still causing concurrency problems. Now creating a lock file.

> 2021-04-30T20:06:29.2006077Z     ERROR [04/30/21 20:06:27:40] An unexpected error occurred:
2021-04-30T20:06:29.2006843Z   LibGit2Sharp.LockedFileException: failed to create locked file 'C:/agent/_work/1/s/.git/refs/heads/release/11.11.x.lock': The file exists.
2021-04-30T20:06:29.2007433Z   
2021-04-30T20:06:29.2007765Z      at LibGit2Sharp.Core.Ensure.HandleError(Int32 result)
2021-04-30T20:06:29.2008240Z      at LibGit2Sharp.Core.Ensure.ZeroResult(Int32 result)
2021-04-30T20:06:29.2008920Z      at LibGit2Sharp.Core.Proxy.git_reference_create(RepositoryHandle repo, String name, ObjectId targetId, Boolean allowOverwrite, String logMessage)
2021-04-30T20:06:29.2009702Z      at LibGit2Sharp.ReferenceCollection.Add(String name, ObjectId targetId, String logMessage, Boolean allowOverwrite)
2021-04-30T20:06:29.2010443Z      at LibGit2Sharp.ReferenceCollection.Add(String name, String canonicalRefNameOrObjectish, String logMessage, Boolean allowOverwrite)
2021-04-30T20:06:29.2011197Z      at LibGit2Sharp.ReferenceCollection.Add(String name, String canonicalRefNameOrObjectish, Boolean allowOverwrite)
2021-04-30T20:06:29.2012217Z      at GitVersion.ReferenceCollection.Add(String name, String canonicalRefNameOrObjectish, Boolean allowOverwrite) in D:\a\GitVersion\GitVersion\src\GitVersion.LibGit2Sharp\Git\ReferenceCollection.cs:line 20
2021-04-30T20:06:29.2013470Z      at GitVersion.GitPreparer.CreateOrUpdateLocalBranchesFromRemoteTrackingOnes(String remoteName) in D:\a\GitVersion\GitVersion\src\GitVersion.Core\Core\GitPreparer.cs:line 343
2021-04-30T20:06:29.2015907Z      at GitVersion.GitPreparer.NormalizeGitDirectory(Boolean noFetch, String currentBranchName, Boolean isDynamicRepository) in D:\a\GitVersion\GitVersion\src\GitVersion.Core\Core\GitPreparer.cs:line 191
2021-04-30T20:06:29.2017103Z      at GitVersion.GitPreparer.NormalizeGitDirectory(String targetBranch, Boolean isDynamicRepository) in D:\a\GitVersion\GitVersion\src\GitVersion.Core\Core\GitPreparer.cs:line 147
2021-04-30T20:06:29.2018320Z      at GitVersion.GitPreparer.PrepareInternal(Boolean normalizeGitDirectory, String currentBranch, Boolean shouldCleanUpRemotes) in D:\a\GitVersion\GitVersion\src\GitVersion.Core\Core\GitPreparer.cs:line 79
2021-04-30T20:06:29.2019373Z      at GitVersion.GitPreparer.Prepare() in D:\a\GitVersion\GitVersion\src\GitVersion.Core\Core\GitPreparer.cs:line 60
2021-04-30T20:06:29.2020347Z      at GitVersion.GitVersionCalculateTool.CalculateVersionVariables() in D:\a\GitVersion\GitVersion\src\GitVersion.Core\Core\GitVersionCalculateTool.cs:line 45
2021-04-30T20:06:29.2021892Z      at GitVersion.GitVersionExecutor.RunGitVersionTool(GitVersionOptions gitVersionOptions) in D:\a\GitVersion\GitVersion\src\GitVersion.App\GitVersionExecutor.cs:line 70

I updated to 5.6.9 and removed the workaround to enabled the parallelism in msbuild.
My builds are running on Azure DevOps on a Windows agent.

bording:
@joaorosado Hmm, that is a bit surprising, and I'm not quite sure how that might be possible. I've rolled out the update on all of my repos, and it has indeed fixed the problem as I was expecting it to.

I think I would need some way to repro it myself before I could understand how it's not working for you.

joaorosado:
What value does the gitVersionOptions.WorkingDirectory have?

The command that fails looks like this, but the actual project ("MyProjectName") is not always the same:
> C:\Users\VssAdministrator\.nuget\packages\gitversion.msbuild\5.6.9\tools\GitVersion.MsBuild.targets(9,9): error MSB3073: The command "dotnet --roll-forward Major "C:\Users\VssAdministrator\.nuget\packages\gitversion.msbuild\5.6.9\tools\netcoreapp3.1/gitversion.dll" "D:\a\1\s\src\MyProjectName" -output file -outputfile obj\gitversion.json" exited with code 1. [D:\a\1\s\src\MyProjectName\MyProjectName.csproj]

Could it be that gitVersionOptions.WorkingDirectory is different for each project in the solution?
I did a search and in `GitVersionTaskExecutor.cs` it seems that it comes from `task.IntermediateOutputPath.GetFileWriteInfo` that sets it from a temp path `Path.Combine(Path.GetTempPath(), "GitVersionTask")`. and could make it different on multiple executions. 

(I don't have much context on these tasks, just did a quick search ..so can be totally off target)

bording:
`WorkingDirectory` should be the root folder of the repo where the .git folder is. If you've got something configured in your solution in a way that is causing that value to be different per project, then yes I would expect to see the behavior you're experiencing because that would mean each instance is acquiring a different mutex.

I'm not sure what you'd be doing in your project to make that happen though.

joaorosado:
I don't see anything special on my solution.
The only thing similar would be this line on the Directory.Build.props, but since it's only one at the top level it wouldn't make a different file: `<SourceRoot Include="$(MSBuildThisFileDirectory)/"/>`

I'll try to build a version of GitVersion locally to attempt to debug.

joaorosado:
@bording just confirmed and the gitVersionOptions.WorkingDirectory is being called for each project with the folder of the project.
It is read directly from the arguments of that command line I posted above.
WorkingDirectory is the project folder (passed as first argument)
and then OutputFile is "obj\gitversion.json", os it writes to "\ProjectFolder\obj\gitversion.json" ..so it looks correct, but breaks the assumption that the WorkingDirectory is always the top level folder.

bording:
Something sounds wrong there. Per the [documentation](https://gitversion.net/docs/usage/cli/arguments), the path parameter, which is what the WorkingDirectory value supposed to be:

>The directory containing .git. If not defined current directory is used. (Must be first argument)

It should not be getting a per-project value for that. 

@arturcic Any insight on this? Why would the wrong value be passed in there? Is the documentation wrong?

joaorosado:
By the way the repositoryInfo object does have the correct information at this point:

image

bording:
@joaorosado The thing I want to make sure we avoid is actually creating any LibGit2Sharp types too early, outside of the mutex, to ensure there's no chance of any stale info being used when repository normalization occurs.

I'll take a look at the code path you're showing to see if that looks safe to use. 

Otherwise, if WorkingDirectory actually isn't guaranteed to be the working directory of the repo from a git perspective, then another option would be to just make the mutex name something like `Global\GitVersion` and ensure that there are not two instances running concurrently anywhere on the machine.

I originally chose to use the WorkingDirectory because it seemed like a good value to scope the mutex down, since two instances running against different repos wouldn't conflict.

dazinator:
> Otherwise, if WorkingDirectory actually isn't guaranteed to be the working directory of the repo from a git perspective, then another option would be to just make the mutex name something like Global\GitVersion and ensure that there are not two instances running concurrently anywhere on the machine.

If you have to go that route it might be nicer to use branch name as that value (we commonly have multiple builds of the same repo happening concurrently on the same build agent - but different feature branches being built for example).

bording:
Looks like `IGitRepositoryInfo.DotGitDirectory` should do what I want, and what I thought `WorkingDirectory` was giving me, without introducing any too-early initialization of the LibGit2Sharp `Repository`, so here's a follow-up PR: #2676

Issue #2675, opened 2021-04-29
[Bug] '.' not accepted as PATH argument (in 5.6.9)


stevesea:
**Describe the bug**
My gitlab pipelines started failing after the 5.6.9 release -- failing with 
```
The system cannot open the device or file specified. : 'Global\.'
```

My gitlab pipeline job looks like:

```yaml

version:
  stage: version
  cache: {}
  image:
    name: gittools/gitversion:latest
    entrypoint: [""]
  rules:
    - if: $CI_MERGE_REQUEST_IID
    - if: $CI_COMMIT_BRANCH == $CI_DEFAULT_BRANCH
  artifacts:
    name: "$CI_JOB_NAME-$CI_COMMIT_REF_SLUG"
    paths:
      - gitversion.properties
  script:
    - /tools/dotnet-gitversion . /output buildserver
    - cat gitversion.properties
```

later jobs in the pipeline depend on the version job and import the env vars from gitversion.properties. 

## Expected Behavior

With 5.6.8 and earlier, using '.' as the path works:
```shell
❯ docker run --rm -v "$(pwd):/repo" -it --entrypoint /bin/bash gittools/gitversion:5.6.8       
root@7e5bc4da4ee2:/tools# cd /repo
root@7e5bc4da4ee2:/repo# /tools/dotnet-gitversion . /output json 
{
  "Major": 3,
  "Minor": 3,
  "Patch": 1,
```

## Actual Behavior
With 5.6.9, '.' as the path fails:

```shell
❯ docker run --rm -v "$(pwd):/repo" -it --entrypoint /bin/bash gittools/gitversion:5.6.9           
root@b4e01d80db52:/tools# cd /repo
root@b4e01d80db52:/repo# /tools/dotnet-gitversion . /output json  
The system cannot open the device or file specified. : 'Global\.'

# however, I can use $(pwd) instead and it's happy
root@b4e01d80db52:/repo# /tools/dotnet-gitversion $(pwd) /output json  
{
  "Major": 3,
  "Minor": 3,
  "Patch": 1,
```

## Possible Fix
to work-around it on gitlab, I'm probably going to replace my pipeline job definition with

```
    - /tools/dotnet-gitversion $(pwd) /output buildserver
```

or 
```
    - /tools/dotnet-gitversion ${CI_PROJECT _PATH} /output buildserver
```

arturcic:
Closed by #2676

PR #2676, opened 2021-05-01
Use .git directory for mutex name

bording:
Follow-up PR to #2669 to fix the mutex name. Based on the [command line documentation](https://gitversion.net/docs/usage/cli/arguments) and my initial investigation of the code, I thought `WorkingDirectory` would be  the actual git working directory, but that turns out to not be true.

Instead, I've switched to using `IGitRepositoryInfo.DotGitDirectory`, which should always be pointing at the same location for all projects in a solution.

If there are projects that have different `DotGitDirectory` values in a solution, then that would mean they are operating on completely different repos, but that likely means you have something misconfigured in your solution anyway. 

## Related Issue

https://github.com/GitTools/GitVersion/pull/2669

## Checklist:

### DIFF
--- a/src/GitVersion.App/GitVersionExecutor.cs
+++ b/src/GitVersion.App/GitVersionExecutor.cs
@@ -58,7 +58,7 @@ namespace GitVersion
         private int RunGitVersionTool(GitVersionOptions gitVersionOptions)
         {
             var mutexName = repositoryInfo.DotGitDirectory.Replace(Path.DirectorySeparatorChar.ToString(), "");
-            using var mutex = new Mutex(true, $@"Global\{mutexName}", out var acquired);
+            using var mutex = new Mutex(true, $@"Global\gitversion{mutexName}", out var acquired);
 
             try
             {
--- a/src/GitVersion.App/GitVersionExecutor.cs
+++ b/src/GitVersion.App/GitVersionExecutor.cs
@@ -57,7 +57,7 @@ namespace GitVersion
 
         private int RunGitVersionTool(GitVersionOptions gitVersionOptions)
         {
-            var mutexName = gitVersionOptions.WorkingDirectory.Replace(Path.DirectorySeparatorChar.ToString(), "");
+            var mutexName = repositoryInfo.DotGitDirectory.Replace(Path.DirectorySeparatorChar.ToString(), "");
             using var mutex = new Mutex(true, $@"Global\{mutexName}", out var acquired);
 
             try


bording:
I see one test has failed on ubuntu, but there's no real information to go on to understand why it might have failed. That same test seems to have passed on Windows and mac. Any chance that test is flaky and just needs another run? I'm thinking it might be since it looks like it did pass on the Azure run.

arturcic (approved):

arturcic:
@bording do you think this one is related to the previous fix https://github.com/GitTools/GitVersion/issues/2675

src/GitVersion.App/GitVersionExecutor.cs
```diff
@@ -57,7 +57,7 @@ public int Execute(GitVersionOptions gitVersionOptions)
 
         private int RunGitVersionTool(GitVersionOptions gitVersionOptions)
         {
-            var mutexName = gitVersionOptions.WorkingDirectory.Replace(Path.DirectorySeparatorChar.ToString(), "");
+            var mutexName = repositoryInfo.DotGitDirectory.Replace(Path.DirectorySeparatorChar.ToString(), "");
```
jetersen, line 60: Perhaps prefixing `gitversion` to the mutex to avoid anyone else using the same mutex?

bording:
> @bording do you think this one is related to the previous fix #2675

Yes, I suspect it's likely if the path value being passed in from the command line.

src/GitVersion.App/GitVersionExecutor.cs
bording: It seems a bit unlikely that another mutex with the exact name of the path of the .git folder would exist, but it probably wouldn't hurt to be a bit more defensive.

Looking at the mutex documentation, there is some vague stuff about potential maximum lengths of mutex names, but there are no actual values given, so I guess we'll have to see if there are any bug reports about names being too long and determine what do to if that happens.

bording:
@arturcic Looks like that test failed again, this time on Windows. Is this a known flaky test, or is this something I need to be trying to figure out if the PR is causing it?

asbjornu:
> Looks like that test failed again, this time on Windows. Is this a known flaky test, or is this something I need to be trying to figure out if the PR is causing it?

That error seems to have been with us for some time, IIRC. I think it's a bug in Cake or at least in how we are using Cake, somehow. See https://github.com/cake-build/cake/discussions/3164#discussioncomment-681775 for details.

arturcic (approved):

src/GitVersion.App/GitVersionExecutor.cs
thoemmi: The maximum length of mutex names is `MAX_PATH`, which is defined as 260 characters. So you should be fine here. (Source: [CreateMutexW documentation](https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-createmutexw#parameters))
Please note that mutex names are case sensitive. Should be ok as long as `repositoryInfo.DotGitDirectoryrepositoryInfo.DotGitDirectory` has always the same casing.

Issue #2390, opened 2020-08-21
System.OverflowException for ParseInt32[Bug]


OatmealOgre:
**Describe the bug**
When gitversion tries to parse a prerelease tag followed by a 10+ digit number an exception is thrown.

## Expected Behavior
Gitversion should catch the exception and handle it some other way

## Actual Behavior
Gitversion doesn't handle any exceptions inside a method called TryParse

## Possible Fix
Make TryParse method of `GitVersionCore/VersionCalculation/SemanticVersioning/SemanticVersion.cs` actually use a try-catch as the name would suggest.
Alternatively have `GitVersionCore/VersionCalculation/SemanticVersioning/SemanticVersionPreReleaseTag.cs` Parse method use a TryParse for parsing the Int

## Steps to Reproduce
Use a tag with a long number in it. Ours has been automated.
Here's the exception along with the tag that likely caused the issue:
```
System.OverflowException: Value was either too large or too small for an Int32.
   at System.Number.ParseInt32(String s, NumberStyles style, NumberFormatInfo info)
   at GitVersion.SemanticVersionPreReleaseTag.Parse(String preReleaseTag)
   at GitVersion.SemanticVersion.TryParse(String version, String tagPrefixRegex, SemanticVersion& semanticVersion)
   at GitVersion.GitRepoMetadataProvider.GetValidVersionTags(IRepository repository, String tagPrefixRegex, Nullable`1 olderThan)
   at GitVersion.VersionCalculation.BaseVersionCalculators.TaggedCommitVersionStrategy.GetTaggedVersions(GitVersionContext context, Branch currentBranch, Nullable`1 olderThan)
   at GitVersion.VersionCalculation.BaseVersionCalculators.TaggedCommitVersionStrategy.GetVersions(GitVersionContext context)
   at System.Linq.Enumerable.<SelectManyIterator>d__17`2.MoveNext()
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at GitVersion.VersionCalculation.BaseVersionCalculator.GetBaseVersion(GitVersionContext context)
   at GitVersion.VersionCalculation.NextVersionCalculator.FindVersion(GitVersionContext context)
   at GitVersion.GitVersionFinder.FindVersion(GitVersionContext context)
[...]

PR #2703, opened 2021-05-22
Replace int.Parse() with int.TryParse() to avoid exceptions caused by overflowing values in tags

james-bjss:
Replace int.Parse() with int.TryParse() This is to prevent the tool failing with an overflowing value.

## Description
If a tag has a suffix which is a numeric value larger than a 32bit int then gitversion fails with an Overflow error.
Whilst the tag can be ignore by using the `tag-prefix` config It would be preferable if it didn't fall over when it encounters such a tag.

## Related Issue
 #2390

## Motivation and Context
We have some tags in our repository which are suffixed with a timestamp value (created by another build process).
We have used the `tag-prefix` config to ignore these, but without it gitversion falls over with an Int32 Overflow exception.

## How Has This Been Tested?
<currently testing>

## Checklist:

- [X ] My code follows the code style of this project.
- [ ] I have added tests to cover my changes. ( Will add if this change is deemed acceptable)
- [X ] All new and existing tests passed.

### DIFF
--- a/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
+++ b/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
@@ -191,7 +191,7 @@ namespace GitVersion
                 Major = major,
                 Minor = minor,
                 Patch = patch,
-                PreReleaseTag = SemanticVersionPreReleaseTag.Parse(parsed.Groups["Tag"].Value),
+                PreReleaseTag = SemanticVersionPreReleaseTag.TryParse(parsed.Groups["Tag"].Value),
                 BuildMetaData = semanticVersionBuildMetaData
             };
 
--- a/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
+++ b/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
@@ -6,6 +6,7 @@ namespace GitVersion
     public class SemanticVersion : IFormattable, IComparable<SemanticVersion>
     {
         private static SemanticVersion Empty = new SemanticVersion();
+
         private static readonly Regex ParseSemVer = new Regex(
             @"^(?<SemVer>(?<Major>\d+)?(\.(?<Minor>\d+))?(\.(?<Patch>\d+))?)(\.(?<FourthPart>\d+))?(-(?<Tag>[^\+]*))?(\+(?<BuildMetaData>.*))?$",
             RegexOptions.Compiled);
@@ -196,8 +197,6 @@ namespace GitVersion
 
             return true;
         }
-
-
 
         public int CompareTo(SemanticVersion value)
         {
--- a/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
+++ b/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
@@ -6,7 +6,6 @@ namespace GitVersion
     public class SemanticVersion : IFormattable, IComparable<SemanticVersion>
     {
         private static SemanticVersion Empty = new SemanticVersion();
-
         private static readonly Regex ParseSemVer = new Regex(
             @"^(?<SemVer>(?<Major>\d+)?(\.(?<Minor>\d+))?(\.(?<Patch>\d+))?)(\.(?<FourthPart>\d+))?(-(?<Tag>[^\+]*))?(\+(?<BuildMetaData>.*))?$",
             RegexOptions.Compiled);
@@ -147,10 +146,10 @@ namespace GitVersion
         public static bool TryParse(string version, string tagPrefixRegex, out SemanticVersion semanticVersion)
         {
             var match = Regex.Match(version, $"^({tagPrefixRegex})?(?<version>.*)$");
+            semanticVersion = null;
 
             if (!match.Success)
             {
-                semanticVersion = null;
                 return false;
             }
 
@@ -159,7 +158,6 @@ namespace GitVersion
 
             if (!parsed.Success)
             {
-                semanticVersion = null;
                 return false;
             }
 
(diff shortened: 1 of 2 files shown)


asbjornu (changes requested):
Great stuff, I just have some design feedback: Please add the method `SemanticVersion.TryParse()` instead to deal with this and then use `Int32.TryParse()` within it to avoid the exceptions.

james-bjss:
> Great stuff, I just have some design feedback: Please add the method `SemanticVersion.TryParse()` instead to deal with this and then use `Int32.TryParse()` within it to avoid the exceptions.

Sure, will update this eve and add tests. Thanks!

asbjornu:
> Sure, will update this evening and add tests. Thanks!

👍🏼 

> Should we swallow the errors silently or would you like it to print to STDERR?

If you're able to inject an instance of `ILog` somewhere natural, you can always do a `log.Error()` statement when parsing fails. For more verbosity you can add `log.Warning()´ or `log.Debug()` statements as well, perhaps on the individual field version part level.

james-bjss:
Updated to use `TryParse()`, the code is already within a method called `SemanticVersion.TryParse()` unless you meant to create a private method for this?

Defaulted the our param to null, the code returns if major version is not matched (there was no default 0 value set for this previously) or if the `int.TryParse()s` fail.

I looked into the logging but could't see a decent way to provide a logger. I'm wondering how useful it is now to check each field if we can't provide details of the failure to the user?

I noticed the `SemanticVersion.Parse()` method wraps `SemanticVersion.TryParse()` and this throws a ` WarningException`. Perhaps it could throw an exception with information about the parsing failure?


```c#
   public static SemanticVersion Parse(string version, string tagPrefixRegex)
        {
            if (!TryParse(version, tagPrefixRegex, out var semanticVersion))
                throw new WarningException($"Failed to parse {version} into a Semantic Version");

            return semanticVersion;
        }
```

src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
```diff
@@ -159,29 +158,47 @@ public static bool TryParse(string version, string tagPrefixRegex, out SemanticV
...
 
             semanticVersion = new SemanticVersion
             {
-                Major = int.Parse(parsed.Groups["Major"].Value),
-                Minor = parsed.Groups["Minor"].Success ? int.Parse(parsed.Groups["Minor"].Value) : 0,
-                Patch = parsed.Groups["Patch"].Success ? int.Parse(parsed.Groups["Patch"].Value) : 0,
+                Major = major,
+                Minor = minor,
+                Patch = patch,
                 PreReleaseTag = SemanticVersionPreReleaseTag.Parse(parsed.Groups["Tag"].Value),
```
asbjornu, line 193: We should probably change `SemanticVersionPreReleaseTag.Parse` to `SemanticVersionPreReleaseTag.TryParse` as well to have a consistent no-throwing semantics inside the `SemanticVersion.TryParse` method.

asbjornu (changes requested):
> Updated to use `TryParse()`

👍🏼 

> the code is already within a method called `SemanticVersion.TryParse()`

Sorry, I did not explore the surrounding code and context fully.

> unless you meant to create a private method for this?

No, this is perfect.

> Defaulted the out param to null, the code returns if major version is not matched (there was no default 0 value set for this previously) or if the `int.TryParse()s` fail.

Sounds reasonable to me.

> I looked into the logging but could't see a decent way to provide a logger. I'm wondering how useful it is now to check each field if we can't provide details of the failure to the user?

If it's difficult it's not worth it. If possible, we can log on the outside (when `SemanticVersion.TryParse` returns `false`) instead.

> I noticed the `SemanticVersion.Parse()` method wraps `SemanticVersion.TryParse()` and this throws a ` WarningException`. Perhaps it could bubble-up/throw an exception with information about the parsing failure?

I'm not sure why we would need the `Parse` method. Our own codebase should be confidently and consistently using `TryParse` everywhere, imho.

> ```cs
>    public static SemanticVersion Parse(string version, string tagPrefixRegex)
>         {
>             if (!TryParse(version, tagPrefixRegex, out var semanticVersion))
>                 throw new WarningException($"Failed to parse {version} into a Semantic Version");
> 
>             return semanticVersion;
>         }
> ```

This works if we actually want exceptions, but if this again is just caught and written to `ILog`, it would be better to just use `TryParse` and log if it returns `false`.

src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs
james-bjss: Sure, will finish this up this evening. Noticed the last build failed, but not entirely sure why as couldn't see a failing test. Will debug this later.

arturcic:
@james-bjss any updates?

james-bjss:
> @james-bjss any updates?

Sorry, have been away on leave and busy with day job. Will try and take a look at this again this week.

james-bjss:
@asbjornu @arturcic - Have updated the PR with the requested changes. 

The `WriteVersionInfoTaskShouldLogOutputVariablesToBuildOutputInAzurePipeline`  is failing though, I noticed these build pipeline tests seem intermittent and seem to fail on other PRs. Not 100% sure this is failing due to my change as previous builds this passed and the GitHub Actions test failed.

Is this a flaky test or do I need to dig deeper?

arturcic:
> @asbjornu @arturcic - Have updated the PR with the requested changes.
> 
> The `WriteVersionInfoTaskShouldLogOutputVariablesToBuildOutputInAzurePipeline` is failing though, I noticed these build pipeline tests seem intermittent and seem to fail on other PRs. Not 100% sure this is failing due to my change as previous builds this passed and the GitHub Actions test failed.
> 
> Is this a flaky test or do I need to dig deeper?

It's a flaky test unfortunately.

src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersionPreReleaseTag.cs
```diff
@@ -84,10 +84,10 @@ public override int GetHashCode()
 
         public static implicit operator SemanticVersionPreReleaseTag(string preReleaseTag)
         {
-            return Parse(preReleaseTag);
+            return TryParse(preReleaseTag);
         }
 
-        public static SemanticVersionPreReleaseTag Parse(string preReleaseTag)
+        public static SemanticVersionPreReleaseTag TryParse(string preReleaseTag)
```
asbjornu, line 90: Since this method does not have the `bool TryParse(string, out T)` signature, I'm not sure it should be named `TryParse`. The method signature indicates `Parse` while its behaviour indicates `TryParse`. How much breakage would it cause to change the method signature to `bool TryParse(string, out SemanticVersionPreReleaseTag)`?

```suggestion
        public static bool TryParse(string preReleaseTag, out SemanticVersionPreReleaseTag)
```

asbjornu (changes requested):

src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersionPreReleaseTag.cs
james-bjss: The `Int.TryParse()` in the `SemanticVersionPreReleaseTag` defaults the value to null if it's not parseable.

Will let you be the guide here as to whether you would prefer to rename the method it back to Parse, return false on failure and modify the signature as you have suggested above or take the same approach as in `SemanticVersion.cs`

In terms of breakage the only place it seems to be directly called is within `SemanticVersion.cs` when it constructs and returns a new `SemanticVersion` and assigns a pre-release tag. 

The tests do not seem to directly reference this method directly either.

From there there we can decide if we want to do anything about it eg. returning false from `SemanticVersion.TryParse()` which would result in the calling method `Parse()` throw a new Warning exception.

NB. All of the code calls `SemanticVersion.Parse()` nothing seems to call the `TryParse()` directly. I could follow the same pattern in `SemanticVersionPreReleaseTag` and wrap the method?

Calling Code in `TryParse()` 

```c#
            semanticVersion = new SemanticVersion
            {
                Major = major,
                Minor = minor,
                Patch = patch,
                PreReleaseTag = SemanticVersionPreReleaseTag.TryParse(parsed.Groups["Tag"].Value),
                BuildMetaData = semanticVersionBuildMetaData
            };`

asbjornu: I would prefer consistent use of `TryParse` returning `bool` everywhere, if possible.

asbjornu:
@james-bjss, could you please go over my feedback and rebase this onto the `HEAD` of `main`? 🙏🏼

asbjornu:
While I think I've solved this by changing `int` to `long` in #3028, I like the changes made in this PR and would love to see them rebased and merged.

james-bjss:
> While I think I've solved this by changing `int` to `long` in #3028, I like the changes made in this PR and would love to see them rebased and merged.

Hi, Sorry for the late reply. I have had quite a lot on work-wise. Will try to grab some time to revisit this and put this PR to bed.

Just to confirm the outstanding change is to replace all usages with 'TryParse()`?

I suppose that then means adding logic in all the calling code to check the repose and throw an error?

asbjornu:
I believe a `bool TryParse()` pattern is better, but I'm not 100% certain it's needed anymore after changing from `int` to `long` in #3028. Please have a look at the code and let me know whether you think a `TryParse` pattern is going to provide an improvement or not still. If it is, I'd love to see this PR rebased and completed.

arturcic:
This one is not needed anymore due to recent changes in the `main` branch

PR #2716, opened 2021-06-04
msbuild.md: update for rename to GitVersion.MsBuild from GitVersionTask

fowl2:
as above. I'm not sure if subtask has been renamed as well

PS. a note on the home page might be nice, as the "package url" of the old package goes there and it's a bit confusing.

### DIFF
--- a/docs/input/docs/usage/msbuild.md
+++ b/docs/input/docs/usage/msbuild.md
@@ -25,7 +25,7 @@ is coming soon.
 
 ### Install the MSTask targets
 
-Add the [GitVersion.MsBuild](https://www.nuget.org/packages/GitVersionTask/) NuGet
+Add the [GitVersion.MsBuild](https://www.nuget.org/packages/GitVersion.MsBuild/) NuGet
 Package into the project you want to be versioned by GitVersion.
 
 From the Package Manager Console:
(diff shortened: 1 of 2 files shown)


docs/input/docs/usage/msbuild.md
```diff
@@ -17,38 +17,42 @@ version information that is compiled into the resulting artifact.
...
+> **Note**  
+  The nuget package was "*[GitVersionTask](https://www.nuget.org/packages/GitVersionTask/)*" up until version 5.5.1.   
+  From version 5.6.0 it has been called "*[GitVersion.MsBuild](https://www.nuget.org/packages/GitVersionTask/)*"
+
 ## TL;DR
 
 ### Install the MSTask targets
 
-Add the [GitVersionTask](https://www.nuget.org/packages/GitVersionTask/) NuGet
+Add the [GitVersion.MsBuild](https://www.nuget.org/packages/GitVersionTask/) NuGet
```
asbjornu, line 28:
```suggestion
Add the [GitVersion.MsBuild](https://www.nuget.org/packages/GitVersion.MsBuild/) NuGet
```

asbjornu (changes requested):
Good catch! Thanks for fixing this.

> PS. a note on the home page might be nice, as the "package url" of the old package goes there and it's a bit confusing.

A note saying that the MSBuild task has been renamed? Good idea, please just add that in this very PR.

asbjornu (approved):

PR #2717, opened 2021-06-08
README.md: note GitVersionTask name change

fowl2:
follow up to #2716. Somehow it worked in the preview but not once it was committed.


asbjornu (changes requested):
Good catch. Can you please change the inline link of GitVersion task to a reference link as well? I can't suggest the change in this review, but just change this:

```md
[GitVersionTask](https://www.nuget.org/packages/GitVersionTask/)
```

to this:

```md
[GitVersionTask][gitversiontask]
```

And then add this to the footer:

```md
[gitversiontask]: https://www.nuget.org/packages/GitVersionTask/
```

asbjornu (approved):

PR #2738, opened 2021-06-28
Update misleading error and give resolution hints

### LINKED ISSUE #2598: [Improvement] Confusing Exception message: "Could not find a 'develop' or 'main' branch, neither locally nor remotely."
When I read this error message, I thought my 'master' branch was not recognized by GitVersion whereas the problem was completely unrelated.

## Detailed Description

I was trying to have GitVersion work on a private GitLab instance of ours (I'm still a beginner concerning GitLab) and was presented with this exception. I wrongly assumed there was a problem with GitVersion not finding my 'master' branch... Even tried to declare it in `GitVersion.yml` but that was for naught. It appears the problem was on the GitLab side due to how we had configured the way it retrieves a git clone (a detached head only listing a unique (non-master) remote branch, no wonder GitVersion didn't work).

I understand, by reading GitVersion code, that 'develop' and 'main' are *not* real branch names, but rather monikers to whatever the development and the main/principal branch are, but the quotation and the fact they can be real branch names as well is, I suppose, the source of my confusion.

I think that the message either:

* should list 'master' along 'main' as one of the branches that could not be found:

> Could not find a 'develop' or 'main' or 'master' branch, neither locally nor remotely.

* or avoid the use of the quotes and try to have the reader understand we are talking about branch roles rather than branch names. Something along this line:

> Could not find any development or main branch, neither locally nor remotely.

## Context

I think this (obvioulsy simple and dispendable) change may help diagnose issues like the one I had.

## Possible Implementation

If one or the other of my propositions is accepted, I'll gladly modify the relevant string in a PR

DarwinJS:
Fixes #2598

Previous message sounded like literal branches rather than Gitversion branch "types".  Also gave hints on checking local repo meets Gitversion requirements or to use dynamic repositories.

## Description
Previous message sounded like literal branches rather than Gitversion branch "types".  Also gave hints on checking local repo meets Gitversion requirements or to use dynamic repositories.

## Motivation and Context
The existing error message tries to use quotes to indicate these are not actual branch names - but this is easily misleading as the quotes can be taken to mean literalness rather than "Gitversion develop".  The problem can also occur when the local repository clone or checkout do not follow the requirements, but the literal branches "develop" and "main" are actually  present.

## How Has This Been Tested?
It's just a message text change.

## Checklist:

### DIFF
--- a/src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
+++ b/src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
@@ -132,7 +132,7 @@ namespace GitVersion.Configuration
                 {
                     // TODO We should call the build server to generate this exception, each build server works differently
                     // for fetch issues and we could give better warnings.
-                    throw new InvalidOperationException("Gitversion could not determine which branch to treat as the development branch (default = develop) nor release branch (default = main or master), neither locally nor remotely. Ensure the local clone and checkout match the requirements or considering using 'Gitversion Dynamic Repositories'");
+                    throw new InvalidOperationException("Gitversion could not determine which branch to treat as the development branch (default is 'develop') nor releaseable branch (default is 'main' or 'master'), either locally or remotely. Ensure the local clone and checkout match the requirements or considering using 'GitVersion Dynamic Repositories'");
                 }
 
                 log.Warning($"{errorMessage}{System.Environment.NewLine}Falling back to {chosenBranch} branch config");
--- a/src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
+++ b/src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
@@ -132,7 +132,7 @@ namespace GitVersion.Configuration
                 {
                     // TODO We should call the build server to generate this exception, each build server works differently
                     // for fetch issues and we could give better warnings.
-                    throw new InvalidOperationException("Could not find a 'develop' or 'main' branch, neither locally nor remotely.");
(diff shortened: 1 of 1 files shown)


src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
```diff
@@ -132,7 +132,7 @@ private BranchConfig InheritBranchConfiguration(IBranch targetBranch, BranchConf
                 {
                     // TODO We should call the build server to generate this exception, each build server works differently
                     // for fetch issues and we could give better warnings.
-                    throw new InvalidOperationException("Could not find a 'develop' or 'main' branch, neither locally nor remotely.");
+                    throw new InvalidOperationException("Gitversion could not determine which branch to treat as the development branch (default = develop) nor release branch (default = main or master), neither locally nor remotely. Ensure the local clone and checkout match the requirements or considering using 'Gitversion Dynamic Repositories'");
```
asbjornu, line 135:
```suggestion
                    throw new InvalidOperationException("Gitversion could not determine which branch to treat as the development branch (default 'develop') nor releasable branch (default 'main' or 'master'), neither locally nor remotely. Ensure the local clone and checkout match the requirements or considering using 'GitVersion Dynamic Repositories'");
```

asbjornu (changes requested):

src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
DarwinJS: I also didn't like the "neither" "nor" structure so I did a commit that incorporated your changes exactly, but also resolved the neither nor - which was a double negative I think.

asbjornu (approved):

PR #2743, opened 2021-06-30
Add test for the version argument

asbjornu:
The GitVersion executable currently returns `1.0.0` when invoked with the `-version` argument. We have no tests for this, so this PR adds such a test. We need to figure out how to return the proper version number, though.

## Description

The version numbering of our assemblies has suddenly stopped working and we need to fix it. This PR adds a test to ensure that `-version` does not return `1.0.0`, but we still need to figure out how to fix the problem.

## Motivation and Context

Returning `1.0.0` as GitVersion's version number is a bug and needs to be fixed.

## How Has This Been Tested?

I added a test that provokes the problem, but I have not yet made any attempt fixing the problem.

## Checklist:

### DIFF
--- a/src/GitVersion.App.Tests/ExecCmdLineArgumentTest.cs
+++ b/src/GitVersion.App.Tests/ExecCmdLineArgumentTest.cs
@@ -53,6 +53,18 @@ namespace GitVersion.App.Tests
         }
 
         [Test]
+        public void VersionArgumentReturnsVersionNumber()
+        {
+            using var fixture = new EmptyRepositoryFixture();
+            fixture.MakeACommit();
+
+            var result = GitVersionHelper.ExecuteIn(fixture.RepositoryPath, arguments: " -version", logToFile: false);
+
+            result.ExitCode.ShouldBe(0);
+            result.Output.ShouldNotBe("1.0.0");
+        }
+
+        [Test]
         public void WorkingDirectoryWithoutGitFolderFailsWithInformativeMessage()
         {
             var result = GitVersionHelper.ExecuteIn(System.Environment.SystemDirectory, arguments: null, logToFile: false);


arturcic (approved):

asbjornu:
@arturcic, do you have any idea why `gitversion -version` has started returning `1.0.0`?

arturcic:
I suspect the `InformationalVersion` that the info is read from is not set, and that might be because the executable used in the unit tests is not built with versioning information

asbjornu:
Yeah, that may explain the problem in the unit test, but not in the released executable. I have a Homebrew-installed GitVersion version `5.6.10` that returns `1.0.0`:

```sh
$ gitversion -version
1.0.0
```

`dotnet-gitversion` is installed as version `5.6.7` and for some reason, it works:

```sh
$ dotnet gitversion -version
5.6.7+Branch.main.Sha.e991fcfe958eb104c37f4a0fd021ac1d1e0dce49
```

So either Homebrew GitVersion is broken or something happened between `5.6.7` and `5.6.10` that broke `-version`.

asbjornu:
I just tested with GitVersion `5.8.2` and it has the same discrepancy between the Homebrew and .NET Tool version:

```sh
$ gitversion -version
1.0.0
```

```sh
$ dotnet gitversion -version
5.8.2+Branch.main.Sha.90dc2bedb08ac54498bddcc279cde5c4596a18d3
```

There has to be something about how the two binaries are built that is different, somehow. Ideas, @arturcic?

arturcic:
> I just tested with GitVersion `5.8.2` and it has the same discrepancy between the Homebrew and .NET Tool version:
> 
> ```shell
> $ gitversion -version
> 1.0.0
> ```
> 
> ```shell
> $ dotnet gitversion -version
> 5.8.2+Branch.main.Sha.90dc2bedb08ac54498bddcc279cde5c4596a18d3
> ```
> 
> There has to be something about how the two binaries are built that is different, somehow. Ideas, @arturcic?

@asbjornu 
I think the Homebrew version does not include the Version information in the Formulae. 

These are the properties we set when building the GitVersion.Tool
https://github.com/GitTools/GitVersion/blob/support/5.x/build/build/BuildLifetime.cs#L30-L38.

We do not set any of those when publishing in homebrew.

What I suggest is to run the publish 2 times - once to generate a temporary binary of GitVersion to run and obtain the version information, and the second time to run publish with the version information.

Then we can fix this issue.

asbjornu:
Resolved by https://github.com/Homebrew/homebrew-core/pull/113622.

PR #2756, opened 2021-07-16
Split throw part from the yielding function

### LINKED ISSUE #2755: [Improvement] Throw in yield function could result in unexpected behavior
I would like to suggest a small change in the [GetBranchesContainingCommit](https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/Core/RepositoryStore.cs#L240) yielding function. Its current form makes it fully "asynchronous" causing the exception to be thrown only when the collection will be iterated. That's currently not a problem in the codebase because all call sites are doing direct iteration but this could cause some issue in the future if new call sites are added without respecting this condition.

## Detailed Description

Please use the following [Sharplab repro](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRj1UwAJUCAWAbjwEgGHiBmc9cggdjwG96N09PrgYUAnAAoARAAkApgBsFAe1IB1ZWAVwpASlq5SR40ahyA7h04TdAOgDic4ADFlyiVAhL9pAPS/SADlVOQAPAGM5AAdgAEtlKFJoOIVSWOA5MABDDLg8ExMzSwprO0cXN3QPLwUff1IAUQjouITSYAALMGVzRLhYsDlw4AUAT3oAX3p+VgpmAB5YqGAAPlJy13dlACMAKyHgUmVdIXo6WIAzUglVAF5b0k9vM7pO7ssi0gBBMABzCAAtnJloEak1IjF4lAPFkgcoLjddPozmcKARyJxSJgDKICOjUJiCDipoZjPljKhZgQFktVusnJsqjt9sMjicRMIGJdrncHk9ai83j1HhZvn9AcDgKClOCWlCYXCEcdkSI6OSChjSABJAFRBQ2AwaozqgpzRbLNY6vU2E0mTl0ExozXYs6OvGaom24wkugkiZAA===) to see the problem live.

## Context

As said in the description, there is no direct problem but this is an issue to become if people are unaware of this behavior and so do not respect the condition use.

## Possible Implementation

The sharplab link contains a suggestion of implementation (with a local function), the name has to be defined. I usually go for `Iterator`, `Impl` or function name + suffix (e.g. `XXXIterator` or `XXXImpl`). It's also probably worth adding a comment to explain the reasoning of the split.

Evangelink:
Fixes #2755

### DIFF
--- a/src/GitVersion.Core/Core/RepositoryStore.cs
+++ b/src/GitVersion.Core/Core/RepositoryStore.cs
@@ -239,52 +239,58 @@ namespace GitVersion
         // TODO Should we cache this?
         public IEnumerable<IBranch> GetBranchesContainingCommit(ICommit commit, IEnumerable<IBranch> branches = null, bool onlyTrackedBranches = false)
         {
-            branches ??= repository.Branches.ToList();
-            static bool IncludeTrackedBranches(IBranch branch, bool includeOnlyTracked) => includeOnlyTracked && branch.IsTracking || !includeOnlyTracked;
-
             if (commit == null)
             {
                 throw new ArgumentNullException(nameof(commit));
             }
 
-            using (log.IndentLog($"Getting branches containing the commit '{commit.Id}'."))
+
+            branches ??= repository.Branches.ToList();
+            static bool IncludeTrackedBranches(IBranch branch, bool includeOnlyTracked) => includeOnlyTracked && branch.IsTracking || !includeOnlyTracked;
+
+            return Impl();
+
+            IEnumerable<IBranch> Impl()
             {
-                var directBranchHasBeenFound = false;
-                log.Info("Trying to find direct branches.");
-                // TODO: It looks wasteful looping through the branches twice. Can't these loops be merged somehow? @asbjornu
-                var branchList = branches.ToList();
-                foreach (var branch in branchList)
+                using (log.IndentLog($"Getting branches containing the commit '{commit.Id}'."))
                 {
-                    if (branch.Tip != null && branch.Tip.Sha != commit.Sha || IncludeTrackedBranches(branch, onlyTrackedBranches))
+                    var directBranchHasBeenFound = false;
+                    log.Info("Trying to find direct branches.");
+                    // TODO: It looks wasteful looping through the branches twice. Can't these loops be merged somehow? @asbjornu
+                    var branchList = branches.ToList();
+                    foreach (var branch in branchList)
                     {
-                        continue;
+                        if (branch.Tip != null && branch.Tip.Sha != commit.Sha || IncludeTrackedBranches(branch, onlyTrackedBranches))
+                        {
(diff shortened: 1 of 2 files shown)


src/GitVersion.Core/Core/RepositoryStore.cs
```diff
@@ -239,52 +239,58 @@ public IEnumerable<IBranch> ExcludingBranches(IEnumerable<IBranch> branchesToExc
...
             }
 
-            using (log.IndentLog($"Getting branches containing the commit '{commit.Id}'."))
+
+            branches ??= repository.Branches.ToList();
+            static bool IncludeTrackedBranches(IBranch branch, bool includeOnlyTracked) => includeOnlyTracked && branch.IsTracking || !includeOnlyTracked;
+
+            return Impl();
+
+            IEnumerable<IBranch> Impl()
```
Evangelink, line 253: Let me know what's your preference for the name of the local function.

Evangelink, line 253: Do you want to include a comment with the reasoning of the change? Do you want to add any specific test?

asbjornu: >Let me know what's your preference for the name of the local function.

How about `InnerGetBranchesContainingCommit`?

```suggestion
            IEnumerable<IBranch> InnerGetBranchesContainingCommit()
```

> Do you want to include a comment with the reasoning of the change?

Yes, that would be nice.

> Do you want to add any specific test?

Yes, please.

asbjornu (changes requested):

src/GitVersion.Core/Core/RepositoryStore.cs
```diff
@@ -239,52 +239,60 @@ public IEnumerable<IBranch> ExcludingBranches(IEnumerable<IBranch> branchesToExc
...
             }
 
-            using (log.IndentLog($"Getting branches containing the commit '{commit.Id}'."))
+            return InnerGetBranchesContainingCommit();
+
+            static bool IncludeTrackedBranches(IBranch branch, bool includeOnlyTracked) => includeOnlyTracked && branch.IsTracking || !includeOnlyTracked;
+
+            // Yielding part is split from the main part of the method to avoid having the exception check performed lazily.
+            // Details at https://github.com/GitTools/GitVersion/issues/2755
+            IEnumerable<IBranch> InnerGetBranchesContainingCommit()
```
Evangelink, line 253: I could also make this local function static and pass-in required arguments. The performance diff is minimal enough to be more of a stylistic choice than anything else.

asbjornu: Yes, that would be less exotic and more approachable, I think. 👍🏼

asbjornu (approved):

PR #2765, opened 2021-07-19
Enable nullable feature on GitVersion.Core

### LINKED ISSUE #2764: [Improvement] Enable nullable on GitVersion.Core

Enable the `nullable` functionality on `GitVersion.Core` project.

## Detailed Description

This will help to understand (and also help analyzers) when a piece of code is expected to be always not-null or could be null.

## Context

See #2763

## Possible Implementation

Enable nullable on the project and fix warnings.

Evangelink:
## Description

Enable the `nullable` feature on `GitVersion.Core` project.

## Related Issue

Fixes #2764

### DIFF
--- a/src/GitVersion.Core/Configuration/Init/Wizard/ConfigInitStepFactory.cs
+++ b/src/GitVersion.Core/Configuration/Init/Wizard/ConfigInitStepFactory.cs
@@ -5,7 +5,7 @@ namespace GitVersion.Configuration.Init.Wizard
 {
     public class ConfigInitStepFactory : IConfigInitStepFactory
     {
-        private readonly IServiceProvider sp;
+        private readonly IServiceProvider? sp;
 
         public ConfigInitStepFactory()
         {
@@ -16,6 +16,6 @@ namespace GitVersion.Configuration.Init.Wizard
             this.sp = sp ?? throw new ArgumentNullException(nameof(sp));
         }
 
-        public T CreateStep<T>() => sp.GetService<T>();
+        public T? CreateStep<T>() => sp!.GetService<T>();
     }
 }
--- a/src/GitVersion.Core/Extensions/NullableAttributes.cs
+++ b/src/GitVersion.Core/Extensions/NullableAttributes.cs
@@ -1,0 +1,149 @@
+// This was copied from https://github.com/dotnet/coreclr/blob/60f1e6265bd1039f023a82e0643b524d6aaf7845/src/System.Private.CoreLib/shared/System/Diagnostics/CodeAnalysis/NullableAttributes.cs
+// and updated to have the scope of the attributes be internal.
+
+#pragma warning disable CA1019 // Define accessors for attribute arguments
+
+namespace System.Diagnostics.CodeAnalysis
+{
+#if !NETCOREAPP
+
+    /// <summary>Specifies that null is allowed as an input even if the corresponding type disallows it.</summary>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
+    internal sealed class AllowNullAttribute : Attribute { }
+
+    /// <summary>Specifies that null is disallowed as an input even if the corresponding type allows it.</summary>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property, Inherited = false)]
+    internal sealed class DisallowNullAttribute : Attribute { }
+
+    /// <summary>Specifies that an output may be null even if the corresponding type disallows it.</summary>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Parameter | AttributeTargets.Property | AttributeTargets.ReturnValue, Inherited = false)]
+    internal sealed class MaybeNullAttribute : Attribute { }
+
+    /// <summary>Specifies that an output will not be null even if the corresponding type allows it.</summary>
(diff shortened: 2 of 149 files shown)


src/GitVersion.Core/Extensions/StringExtensions.cs
```diff
@@ -130,9 +131,15 @@ public static string RegexReplace(this string input, string pattern, string repl
...
         }
 
-        public static bool IsEquivalentTo(this string self, string other)
+        public static bool IsEquivalentTo(this string self, string? other)
         {
             return string.Equals(self, other, StringComparison.OrdinalIgnoreCase);
         }
+
+        /// <inheritdoc cref="StringExtensions.IsNullOrEmpty(string)"/>
+        public static bool IsNullOrEmpty([NotNullWhen(returnValue: false)] string? value) => string.IsNullOrEmpty(value);
```
Evangelink, line 140: Simple wrappers with some annotation to give hints to the analyzers/compiler about nullability states.

Let me know if you want these somewhere else.

src/GitVersion.Core/Extensions/NullableAttributes.cs
```diff
@@ -0,0 +1,149 @@
+// This was copied from https://github.com/dotnet/coreclr/blob/60f1e6265bd1039f023a82e0643b524d6aaf7845/src/System.Private.CoreLib/shared/System/Diagnostics/CodeAnalysis/NullableAttributes.cs
+// and updated to have the scope of the attributes be internal.
+
+#pragma warning disable CA1019 // Define accessors for attribute arguments
+
+namespace System.Diagnostics.CodeAnalysis
```
Evangelink, line 6: Copy-paste all the attributes that will be available in new versions of dotnet. This class could be removed when dropping support to non-netcore fwk.

Evangelink:
I will rebase this PR once the `readonly` and the `analyzers` got merged.

src/GitVersion.Core/Extensions/StringExtensions.cs
asbjornu: Can we make these into proper extension methods?

```suggestion
        public static bool IsNullOrEmpty([NotNullWhen(returnValue: false)] this string? value) => string.IsNullOrEmpty(value);
```

src/GitVersion.Core/Extensions/StringExtensions.cs
```diff
@@ -130,9 +131,15 @@ public static string RegexReplace(this string input, string pattern, string repl
...
+
+        /// <inheritdoc cref="StringExtensions.IsNullOrWhiteSpace(string)"/>
+        public static bool IsNullOrWhiteSpace([NotNullWhen(returnValue: false)] string? value) => string.IsNullOrWhiteSpace(value);
```
asbjornu, line 143:
```suggestion
        public static bool IsNullOrWhiteSpace([NotNullWhen(returnValue: false)] this string? value) => string.IsNullOrWhiteSpace(value);
```

asbjornu (changes requested):
Great stuff!

src/GitVersion.Core/Extensions/StringExtensions.cs
Evangelink: Absolutely! I wasn't sure what was your preference.

FYI, in roslyn they went with `RoslynString.XXX`.

src/GitVersion.Core/Extensions/StringExtensions.cs
Evangelink: Done

src/GitVersion.Core/Extensions/StringExtensions.cs
Evangelink: Done.

asbjornu (changes requested):
This is great! I love that it's now visible how many opportunities for `NullReferenceException` we have in the codebase. 😅 That's something I would very much like to remedy, though. I'm not saying all `?` can or should be removed, but I do think there's way too many in the current codebase.

If you see any places where `null` guards would alleviate the need for `null` values, please sprinkle them throughout the codebase as you see fit. As long as the tests are green; the more `null` guards and less `?` we have, the better.

src/GitVersion.Core/Configuration/Init/Wizard/ConfigInitStepFactory.cs
```diff
@@ -5,7 +5,7 @@ namespace GitVersion.Configuration.Init.Wizard
 {
     public class ConfigInitStepFactory : IConfigInitStepFactory
     {
-        private readonly IServiceProvider sp;
+        private readonly IServiceProvider? sp;
```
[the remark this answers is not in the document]
Evangelink: There is a pattern in your codebase where a lot of classes have an empty ctor. Thus, it's possible to have some fields initialized to their default value. That's the case here.

The problem for me is that I don't own/master enough the codebase to see if something is there for some specific reason or not. 

On this specific example what's the purpose of the empty ctor? Can it be made private/protected/internal? Can it be removed?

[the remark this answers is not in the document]
asbjornu: Yes, I've noticed. It's not a pattern I condone. I've created #2775 so we can try to fix that in a future PR.

asbjornu (approved):

arturcic:
@Evangelink thank you for the effort. I know what this kind of refactoring is so many thanks. Hope we can improve the codebase even more so that we can prepare it for version 6.0 (and target net 6.0 as it's LTS and drop support for net48 and netcore3.1)

PR #2780, opened 2021-07-26
CA1816: correct idisposable implementation

Evangelink:
Fix implementation of the disposable pattern.

## Description
See [Implementing Dispose](https://docs.microsoft.com//dotnet/standard/garbage-collection/implementing-dispose) and [CA1816](https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1816)

### DIFF
--- a/src/GitTools.Testing/Fixtures/RepositoryFixtureBase.cs
+++ b/src/GitTools.Testing/Fixtures/RepositoryFixtureBase.cs
@@ -33,8 +33,19 @@ namespace GitTools.Testing
         /// <summary>
         ///     Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
         /// </summary>
-        public virtual void Dispose()
+        public void Dispose()
         {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposing)
+            {
+                return;
+            }
+
             Repository.Dispose();
 
             try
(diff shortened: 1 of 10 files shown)


src/GitTools.Testing/Fixtures/RepositoryFixtureBase.cs
```diff
@@ -33,8 +33,19 @@ protected RepositoryFixtureBase(IRepository repository)
         /// <summary>
         ///     Performs application-defined tasks associated with freeing, releasing, or resetting unmanaged resources.
         /// </summary>
-        public virtual void Dispose()
+        public void Dispose()
```
Evangelink, line 36: "Proper" implementation of the disposable pattern.

src/GitVersion.Core/VersionConverters/AssemblyInfo/AssemblyInfoFileUpdater.cs
```diff
@@ -13,7 +13,7 @@ public interface IAssemblyInfoFileUpdater : IVersionConverter<AssemblyInfoContex
     {
     }
 
-    public class AssemblyInfoFileUpdater : IAssemblyInfoFileUpdater
+    public sealed class AssemblyInfoFileUpdater : IAssemblyInfoFileUpdater
```
Evangelink, line 16: When class wasn't inherited I went with the option of sealing the class but we could also go for the disposable pattern implementation.

Note that this change could be considered as breaking change if you consider this was part of your API.

asbjornu (approved):
I think this is fine. I can't imagine anyone using `GitTools.Testing` beside us, so I don't think there are any consumers to care for. If there are, perhaps with this change we'll find out.

PR #2781, opened 2021-07-26
Ensure IEquatable takes nullable type

Evangelink:
Ensure that `IEquatable<T>` targets nullable type and update related implementation.

## Related Issue

Follow-up of #2765

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -178,3 +178,6 @@ dotnet_diagnostic.CA1810.severity = warning
 
 # CA1816: Dispose methods should call SuppressFinalize
 dotnet_diagnostic.CA1816.severity = warning
+
+# IDE0005: Using directive is unnecessary.
+dotnet_diagnostic.IDE0005.severity = warning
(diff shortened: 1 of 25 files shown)


asbjornu (commented):
Can you please provide a bit more context and information regarding this change? I hoped we would get fewer `?` in the codebase going forward, not more?

Evangelink:
I just assumed that as we were doing null checks in the equals, it was better to explicitly allow nulls but I understand your point. We can probably close the PR.

asbjornu:
Ok, I see. I suppose it's difficult to ensure these types aren't `null` earlier in the codebase?

Evangelink:
At this point yes. I assume/hope it will be easier after a couple of null-check iterations :)

asbjornu:
Indeed, let's merge this then. Can you please rebase and resolve conflicts?

Evangelink:
@asbjornu Done.

asbjornu (approved):

PR #2784, opened 2021-07-26
Ensure parameter names matches interfaces definitions

Evangelink:
It is generally a good practice to keep the parent (interface or class) parameter name rather than changing them in a child implementation.


asbjornu:
Hm, I'm unable to get the AzDO tests to pass. I think I've re-run them 20 times now. Thoughts, @arturcic?

arturcic:
If the GitHub actions pass you can merge @asbjornu. The AzDo should be disable for now

PR #2800, opened 2021-08-07
Fix TaggedCommitVersionStrategy performance problems

ron-macneil-youi:
The problem was a branch commits x version tags nested loop that quickly
grew into the millions of iterations as the number of commits and/or
tags in a repo increased, even modestly.

Each iteration performed Git tag-to-commit resolution and GitVersion
commit object-to-object comparison, both of which take a little bit of
time and aren't particularly slow in isolation but which add up fast
when performed millions of times.

This is solved by doing a single pass over the tags, resolving each to a
commit only once; then using that to build a commit-to-tag lookup
(hashtable) keyed by commit SHA, eliminating the object comparisons; and
finally doing a single pass over the branch commits, looking up
associated tags from the hashtable by SHA, eliminating the nested loop.

Testing on a repo with ~20,000 commits and ~5,000 tags showed an overall
GitVersion run time reduction from over 2 minutes down to around 5
seconds.

## Checklist:

### DIFF
--- a/src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/TaggedCommitVersionStrategy.cs
+++ b/src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/TaggedCommitVersionStrategy.cs
@@ -21,22 +21,13 @@ namespace GitVersion.VersionCalculation
 
         internal IEnumerable<BaseVersion> GetTaggedVersions(IBranch? currentBranch, DateTimeOffset? olderThan)
         {
-            var allTags = this.repositoryStore.GetValidVersionTags(Context.Configuration?.GitTagPrefix, olderThan);
-
-            var taggedCommits = currentBranch
-                ?.Commits
-                .SelectMany(commit => allTags.Where(t => IsValidTag(t.Item1, commit))).ToList();
-
-            var taggedVersions = taggedCommits
-                .Select(t =>
-                {
-                    var commit = t.Item1?.PeeledTargetCommit();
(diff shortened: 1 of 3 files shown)


src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/TaggedCommitVersionStrategy.cs
```diff
@@ -21,22 +21,13 @@ public class TaggedCommitVersionStrategy : VersionStrategyBase
...
-                .Select(t =>
-                {
-                    var commit = t.Item1?.PeeledTargetCommit();
-                    return commit != null ? new VersionTaggedCommit(commit, t.Item2!, t.Item1!.Name.Friendly) : null;
-                })
-                .Where(versionTaggedCommit => versionTaggedCommit != null)
-                .Select(versionTaggedCommit => CreateBaseVersion(Context, versionTaggedCommit!))
-                .ToList();
-
+            if (currentBranch is null) return Enumerable.Empty<BaseVersion>();
```
asbjornu, line 24:
```suggestion
            if (currentBranch is null) 
                return Enumerable.Empty<BaseVersion>();
```

asbjornu (changes requested):
This looks great!

asbjornu:
Can you please run `dotnet format ./ --folder --exclude **/AddFormats/`, commit the changes and push?

ron-macneil-youi:
Hi @asbjornu I ran the `dotnet format` command and it hasn't changed anything.  Were you expecting a particular change?

asbjornu (approved):

Issue #1967, opened 2019-12-04
Gitversion crashes on teamcity when there is no local master branch


jakubsuchybio:
Info:
We are using gitlab and teamcity. Both are not publicly visible from internet. They are only visible inside our company network. Therefore they only have self-signed certificates to operate. We had to disable ssl to use them with git. And because of this, we can't use dynamic repository with GitVersion.
We are already using ```teamcity.git.fetchAllHeads=true```

The problem is like this:
When we run teamcity build with option to clear working directory (empty the build folder before building) teamcity then fetches all remote branches and also the current feature branch that is building.
Because of clean working directory, there is no local master branch, only origin/master.
And this results in crash of gitversion. (more info in log below)
The hot fix for this is to run that build configuration for master branch and then not deleting working directory => then there exists local master branch => gitversion works.
But everytime someone would want to clear working directory when building any non-master branch, it will crash.

I'm not quite sure, why gitversion can't infer master branch from that origin/master. I really don't understand why this is happening.

Thanks

Build logs:
```
[Step 1/1] Running: C:\ProgramData\chocolatey\bin\gitversion.exe "C:\BuildAgent\work\da7b98356013a858" /output "buildserver" /UpdateAssemblyInfo /EnsureAssemblyInfo /nonormalize
[21:29:01]	[Step 1/1] INFO [12/04/19 21:29:01:49] Working directory: C:\BuildAgent\work\da7b98356013a858
[21:29:01]	[Step 1/1] INFO [12/04/19 21:29:01:50] IsDynamicGitRepository: False
[21:29:01]	[Step 1/1] INFO [12/04/19 21:29:01:54] Returning Project Root from DotGitDirectory: C:\BuildAgent\work\da7b98356013a858\.git\ - C:\BuildAgent\work\da7b98356013a858\
[21:29:01]	[Step 1/1] INFO [12/04/19 21:29:01:54] Running on Windows.
[21:29:01]	[Step 1/1] INFO [12/04/19 21:29:01:54] Applicable build agent found: 'TeamCity'.
[21:29:01]	[Step 1/1] INFO [12/04/19 21:29:01:54] Branch from build environment: refs/heads/feature/test-build
[21:29:01]	[Step 1/1] INFO [12/04/19 21:29:01:54] Project root is: C:\BuildAgent\work\da7b98356013a858\
[...]

PR #2831, opened 2021-08-31
Feature/fix nonormalize with azurepipeline

mtren:
## Related Issue
Maybe https://github.com/GitTools/GitVersion/issues/1967

## Motivation and Context

The combination of repos with detached head (like on buildservers) with the /nonormalize parameter did not work.

## How Has This Been Tested?

I testet it by executing gitversion within an azure devops buildserver pipeline by using the /nonormalize parameter.

## Checklist:

### DIFF
--- a/src/GitVersion.Core/Core/GitPreparer.cs
+++ b/src/GitVersion.Core/Core/GitPreparer.cs
@@ -68,7 +68,12 @@ namespace GitVersion
             }
             else
             {
-                if (!normalizeGitDirectory) return;
+                if (!normalizeGitDirectory)
+                {
+                    EnsureLocalBranchExistsForCurrentBranch(EnsureOnlyOneRemoteIsDefined(), currentBranch);
+                    return;
+                }
+
                 if (shouldCleanUpRemotes)
                 {
                     CleanupDuplicateOrigin();
--- a/src/GitVersion.Core/Core/GitPreparer.cs
+++ b/src/GitVersion.Core/Core/GitPreparer.cs
@@ -70,6 +70,7 @@ namespace GitVersion
             {
                 if (!normalizeGitDirectory)
                 {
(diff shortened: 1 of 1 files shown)


src/GitVersion.Core/Core/GitPreparer.cs
```diff
@@ -68,7 +68,13 @@ private void PrepareInternal(bool normalizeGitDirectory, string? currentBranch,
             }
             else
             {
-                if (!normalizeGitDirectory) return;
+                if (!normalizeGitDirectory)
+                {
+                    if (currentBranch == null) return;
+                    EnsureLocalBranchExistsForCurrentBranch(EnsureOnlyOneRemoteIsDefined(), currentBranch);
```
asbjornu, line 74: It seems weird to me to perform normalization even to `/nonormalize` is being passed as an argument. Isn't it better to ensure that the Git repository is normalized before GitVersion is executed?

asbjornu (changes requested):
Thanks for submitting this PR.

src/GitVersion.Core/Core/GitPreparer.cs
mtren: GitVersion has low performance with many branches. I found out, that the normalization was the reason why it takes minutes.
But with the /nonormalize parameter GitVersion did not work when it is used on a buildserver.

asbjornu: `/nonormalize` works on build servers if you normalize the repository yourself. [GitVersion's operation requirements are found in the documentation](https://gitversion.net/docs/reference/requirements). When the repository is normalized to meet those requirements, it should execute successfully in any environment.

arturcic:
Closing as it's not by design

Issue #2454, opened 2020-11-13
Hotfix branch is not increasing Patch


roofiq:
Hello,

I am using version 5.5. I want to implement semantic versioning to my GitFlow. I got quite a lot of errors but have managed to work it out - the documentation is really poor.
Still I can't manage to achieve one thing. On bugfixing situation, when I am creating new branch from master, it should have increased patch number - i.e. `master` is 1.0.0, so the `hotfix/test` should be 1.0.1 but it is 1.0.0.

I can achieve above when I am adding comment '+semver: patch'.
Then, when I want to create 2nd patch, the version is not increasing at all (even with the comment).
How it should work in this scenario?
Here is mine GitVersion file.

```yaml
next-version: 2.0.0
branches:
  master:        
    regex: master
    mode: ContinuousDelivery
    tag: ''    
    increment: Patch
    prevent-increment-of-merged-branch-version: true
    track-merge-target: false
    tracks-release-branches: false
    is-release-branch: false
   hotfix:
    tag: beta
    mode: ContinuousDelivery
    regex: ^hotfix(es)?[/-]
    increment: Patch
    prevent-increment-of-merged-branch-version: false
    track-merge-target: false
    tracks-release-branches: false
ignore:
  sha: []
merge-message-formats: {}
```

Thanks,
Rafal

asbjornu:
As described in #2441, `next-version` doesn't seem to work well with commit message incrementing anymore. If you remove `next-version` and instead add a `git tag 2.0.0` on the appropriate commit, what happens?

senritsu:
We have a similar issue, but without using `next-version` (in fact even with an empty GitVersion.yml).

For some reason GitVersion completely ignores all hotfix merges. It does find one of 3 existing merges (and 3 times at that), but it does not increment at all:

```
INFO [04/01/21 10:48:06:28] Running against branch: master (fa43b09 Merge branch 'hotfix/gitversion-config')
INFO [04/01/21 10:48:06:28] Begin: Calculating base versions
  INFO [04/01/21 10:48:06:33] Fallback base version: 0.1.0 with commit count source 9a8dbe6583dff424ec8a6dc5a94c27f7f5e1498f
  INFO [04/01/21 10:48:06:37] Found commit [fa43b09 Merge branch 'hotfix/gitversion-config'] matching merge message format: Default
[...]

PR #2847, opened 2021-09-13
Add unit test to investigate and hopefully verify #2454

senritsu:
## Description
Added a unit test for investigating branch versioning issues, as prompted by @asbjornu in the discussion of #2454.

## Related Issue
#2454

## Motivation and Context

The test was initially created to reproduce #2454. The test failed when it was initially created, but passes on current `main` (with some modifications to the version in the issue discussion). There are a few details that I am still not completely clear on:

- Should hotfix branches get a `beta` prerelease tag? In the new test they don't, but according to [this line in another test](https://github.com/senritsu/GitVersion/blob/48b0e9dadd97344aac3142dc5cdcb0f1f3a2a48c/src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs#L95) maybe they should?
- Build metadata does not appear, but that is probably because of the mode CD in the `Config` used?
- Some of the versioning in the different tests in the file seemed inconsistent when first writing the test, I did not have the time to fully check that this time around, so it could use some more scrutiny
- Some comments need to still be removed before the PR should be merged

Disregarding those details, it seems like #2454 is not an issue anymore (?).

## How Has This Been Tested?
The change is a unit test, it was executed

N/A
## Checklist:
- [ ] ~~My change requires a change to the documentation.~~
- [ ] ~~I have updated the documentation accordingly.~~
- [ ] ~~I have added tests to cover my changes.~~


asbjornu:
The tests are successful?

senritsu:
Seems like it, yes. I made a few adjustments because apparently some things changed since the original discussion. 

Disregarding the few comments the test may actually validate that #2454 was unintentionally fixed in the meantime, but I would not hold my breath on it just yet.

asbjornu:
The successful test may be a false positive as discovered in https://github.com/GitTools/GitVersion/pull/2830#issuecomment-918809996.

asbjornu:
Can you please rebase against `main` to see whether the test is still successful?

senritsu:
I can confirm that the test still passes, even with the current changes from `main`.

I am still unsure if a hotfix branch should have a `beta` prerelease tag, similar to a `release` branch. Currently it only has (with versioning mode CD) the same tag like `main`/`master`, `-ci.n`.

When that is clarified I would remove the corresponding comments from the unit test, and the test would seem to confirm that at least one scenario of #2454 appears to work now. Of course that only goes for `mode: ContinuousDeployment`, and there may be edge cases that are missing.

senritsu:
Curiously, when trying out 5.7.0 with one of our internal library repositories, the hotfix branch *does* in fact get a `beta` tag, albeit without `mode: ContinuousDeployment`. 

This adds another datapoint and confirms that in at least one of our productive scenarios, #2454 now seems to be fixed.

senritsu:
Probably an unrelated issue, but the same package that works perfectly fine now locally still fails to properly account for the hotfix on our buildserver (TeamCity). Locally it versions to `0.5.1`, on TC it ignores the hotfix and stays on `0.5.0`.

senritsu:
Additional scenarios have been added to the test, both new tests now fail. In both modes the second hotfix is ignored for versioning purposes.

asbjornu:
As the accompanying issue #2454 is closed with a workaround, is this PR relevant anymore?

arturcic:
Closes in favor of #3445

PR #2863, opened 2021-09-29
Linux Arm64 support

arturcic:
In this PR I added build infrastructure support for building Arm64 docker images and to test the artifacts with docker images based on Arm64.

Also extended the list of distributions we test with and docker images we build 

Here is the list 

alpine.3.12, alpine.3.13, alpine.3.14, centos.7, centos.8, debian.9, debian.10, debian.11, fedora.33, ubuntu.18.04, ubuntu.20.04

All of them except Alpine have Arm64 support, alpine in our case is not supported and the LibGit2Sharp.Native does not yet 
 include support for `linux-musl-arm64`.

Also part of the docker images we build I added docker manifests.

For example for debian 11 we have these images:

`docker.io/gittools/gitversion:5.7.1-debian.11-5.0-amd64`
`docker.io/gittools/gitversion:5.7.1-debian.11-5.0-arm64`

and a manifest 
`docker.io/gittools/gitversion:5.7.1-debian.11-5.0` that bundles those 2 toghether.

That means you can either pull the docker image by full name and tag:

`docker pull docker.io/gittools/gitversion:5.7.1-debian.11-5.0-amd64`

or by specifying the platform:

`docker pull --platform linux/amd64 docker.io/gittools/gitversion:5.7.1-debian.11-5.0 `

The exact same images are pushed to GitHub Package Repository

`ghcr.io/gittools/gitversion:5.7.1-debian.11-5.0-amd64`
`ghcr.io/gittools/gitversion:5.7.1-debian.11-5.0-arm64`

and a manifest 
`ghcr.io/gittools/gitversion:5.7.1-debian.11-5.0`
https://github.com/GitTools/GitVersion/pkgs/container/gitversion


arturcic:
@asbjornu Question. Now that we also have arm64 support we basically have double the images in dockerhub/github packages.

I wonder if the current approach of publishing every time we commit to `main` we publish docker images is a good thing. I suggest we publish docker images only when we have release (that will decrease the amount of docker images we have) and will make it easier for the consumers to choose which docker image to use. Thoughts?

asbjornu:
@arturcic, perhaps we can only publish to the GitHub Package Repository on non-stable builds and to Docker Hub on tagged commits?

arturcic:
yes makes sense, I will adapt the code

asbjornu:
I just noticed that we seem to be doing the same on NuGet since version 5.2.

arturcic:
Right, so we keep the same approach then for docker images

Issue #2864, opened 2021-09-29
[Bug] Azure Execute hangs for ~25 minutes and then fails with generic node error.


scphantm:
**Describe the bug**
i reported this here https://github.com/GitTools/actions/issues/511 but havn't gotten any response, my logs and such are over there. 

Heres the basic, when we are seeding a repo, the azure pipeline will hang on the GitVersion Execute task.  we don't know why, but when it does happen, it will hang for about 30 minutes and then fail with a generic node error.  What fixes it is simply creating a new PR with just a white space commit to it.

the bug is 2 fold

1 - why is it failing
2 - why is it failing in such a way that we can't debug.

## Expected Behavior
It works

## Actual Behavior
It hangs on the execute task until Node or Azure times out.

## Possible Fix
putting in a white space commit makes it work again.

## Steps to Reproduce
when we are seeding our repos, the initial commits cause the thing to hang.  Simply adding a blank white space commit will clear the error.

## Context
we have hundreds of repos and pipelines.  This is killing us.

## Your Environment
Azure Pipelines

- Version Used:
- Operating System and version (Windows 10, Ubuntu 18.04):
- Link to your project:
- Link to your CI build (if appropriate):

scphantm:
I just posted this in the other bug:

we are really having a problem with this.  its bad enough where im getting pressure to pull the tool.  we show about 15% of our pipelines fail because of this error.  The problem case has expanded to include pipelines that have already worked correctly.  meaning we did the empty commit mentioned above and got it to work, then the next PR, it fails.

Any ideas here?  It really seems like node is either hanging because it doesn't like something in the repo, or, what i really think is happening is gitversion is failing for some reason and node isn't picking up the exit code or logs.  possibly because the action has it always parsing json?  if there is an error, is the return 

https://github.com/GitTools/actions/blob/d957a15d75c7fc44255d91da910f7a2c0b0be42f/src/tasks/gitversion/main.ts#L49

here, able to capture and return an error?  I know in testing locally, if gitversion generates an error, the output to the screen is not json.  Its only json when things go correctly.

scphantm:
i submitted a patch, but @asbjornu is going to have to refactor it into the current code base as im not able to run dot net 6 in any of my environments yet.

InKahootz:
Do you have a minimal repro? You said you were able to get the issue to occur locally finally.

PR #2883, opened 2021-10-18
Added Infinite loop protection

scphantm:
I added recursion breakout if it goes 50 levels deep


scphantm:
addresses #2864

scphantm:
My dev machine is a Mac so I'm unable to run .net6 in either rider or VStudioMac yet.  Since the code was already refactored for net6, i had to checkout the 5.7.0 tag and apply my changes there.  Thats the latest version i was able to run on my machine.  Someone with Windows is going to have to resolve the merge conflicts as i can't run that code at all.

scphantm:
you are going to have to fix whatever is wrong.  its well beyond me until rider releases .net6 support which they said they won't until .net6 goes GA

arturcic:
@scphantm I can understand that. I will see if we can find some time to check the changes you've done.
One thing to note though is you can have multiple .net sdk installed on a PC, and you can control which one is used for a specific project using global.json -> that means you can have .net 6 rc2 installed side by side with those sdk you already have for the other projects. Second thing. I use Jetrains ToolBox and am able to install the Rider version that I work for other projects that does not support .net6 (same as you have), and side by side I installed the EAP version that does support .net 6. So it's possible

mrt-135 (changes requested):

Issue #2668, opened 2021-04-19
[Feature] Gitversion support for apple silicon


iamNoah1:
I would like to use gitversion on my apple silicon. 

Installing using brew gives me: `Error: dotnet: no bottle available!`

Try to use the docker image gives me:
```
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
Failed to create CoreCLR, HRESULT: 0x8007FF06
```

arturcic:
Unfortunately we currently do not support the new silicon. Even though there is some level of support for ARM provided by Libgit2sharp, but we do not support it official

iamNoah1:
Any plans on supporting it in the future?

arturcic:
I think we should support it in long term, the issue is we are not able to build it in GitHubAction or AzurePipelines at the moment. When that is available we could add support for the M1

jshinevar:
I have not tried this yet, but here is the full error I see from homebrew:

`Error: gitversion: no bottle available!
You can try to install from source with:
  brew install --build-from-source gitversion
Please note building from source is unsupported. You will encounter build
failures with some formulae. If you experience any issues please create pull
requests instead of asking for help on Homebrew's GitHub, Twitter or any other
official channels.`

Has anyone tried using the --build-from-source?  Were you successful?

asbjornu:
[Support for M1 in .NET is underway](https://github.com/orgs/dotnet/projects/18), but still far from completed. Until both .NET and LibGit2Sharp has full support for M1, GitVersion won't be able to provide full support for M1.

andymac4182:
For me I need to run the arm64 version with docker which is supported in dotnet as far as I am aware. 

It would be great if there could be a arm64 docker image. https://docs.docker.com/desktop/multi-arch/

MuhKuh7:
any updates on this one?

arturcic:
@MuhKuh7 actually it should already support osx-arm64 starting with 5.8.1 https://github.com/GitTools/GitVersion/releases/tag/5.8.1

MuhKuh7:
@arturcic thx, will 5.8.1 be available on homebrew (https://formulae.brew.sh/formula/gitversion) soon?

arturcic:
unfortunately for the homebrew we ned to have the dotnet formula for version 6.0, that's something @asbjornu was working on. So for now no, you can only use the tar.gz version for now

PR #2916, opened 2021-11-11
(deps) update LibGit2Sharp 0.27.0-preview-0158

arturcic:


asbjornu (approved):
👍🏼 Do we have any existing issues that may be resolved by this upgrade?

arturcic:
There is a fix in building of the native libs for macos arm64

asbjornu:
This should resolve #2668, then?

arturcic:
In theory yes. I'm not able to test this, but I think we can ask those with macos m1 to give it a try

PR #2990, opened 2022-02-08
Sped up the Normalization by removing an unnecessary O(n^2) loop

ni-fgenois:
This should speedup the normalization when there are a lot of branches.
The `.Any(x => x.Name.Equals(referenceName)` was causing an unnecessary O(n^2) loop, as `Refs` is already an indexed structure (Dict).

## Checklist:

### DIFF
--- a/src/GitVersion.Core/Core/GitPreparer.cs
+++ b/src/GitVersion.Core/Core/GitPreparer.cs
@@ -314,15 +314,14 @@ Please run `git {GitExtensions.CreateGitLogArgs(100)}` and submit it along with 
         {
             var remoteTrackingReferenceName = remoteTrackingReference.Name.Canonical;
             var branchName = remoteTrackingReferenceName.Substring(prefix.Length);
-            var localCanonicalName = "refs/heads/" + branchName;
+            var localReferenceName = ReferenceName.FromBranchName(branchName);
 
-            var referenceName = ReferenceName.Parse(localCanonicalName);
             // We do not want to touch our current branch
             if (this.repository.Head.Name.EquivalentTo(branchName)) continue;
 
-            if (this.repository.Refs.Any(x => x.Name.Equals(referenceName)))
+            if (this.repository.Refs[localReferenceName] is not null)
             {
-                var localRef = this.repository.Refs[localCanonicalName]!;
+                var localRef = this.repository.Refs[localReferenceName]!;
                 if (localRef.TargetIdentifier == remoteTrackingReference.TargetIdentifier)
                 {
                     this.log.Info($"Skipping update of '{remoteTrackingReference.Name.Canonical}' as it already matches the remote ref.");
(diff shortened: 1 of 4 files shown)


arturcic (approved):

src/GitVersion.Core/Core/GitPreparer.cs
```diff
@@ -316,11 +316,10 @@ private void CreateOrUpdateLocalBranchesFromRemoteTrackingOnes(string remoteName
             var branchName = remoteTrackingReferenceName.Substring(prefix.Length);
             var localCanonicalName = "refs/heads/" + branchName;
 
-            var referenceName = ReferenceName.Parse(localCanonicalName);
```
arturcic, line 319: I think this line is needed - we need to parse the localCanonicalName

ni-fgenois: Is the `ReferenceName.Parse` doing any validation? Because we are not using `referenceName` anywhere

arturcic: Yes it does https://github.com/GitTools/GitVersion/blob/3626ce5126e1375549224193d039820d521d3f77/src/GitVersion.Core/Git/ReferenceName.cs#L29

ni-fgenois: The following validation snippet from `ReferenceName.Parse` : `IsPrefixedBy(canonicalName, LocalBranchPrefix)`
will match the constructed string `"refs/heads/" + branchName;`

What do you think about changing the construction of the string to `ReferenceName.LocalBranchPrefix + branchName;` instead?

ni-fgenois: Well, the problem here is actually using the `IComparable` interface for searching a matching `IReference`. Doing such a linear lookupleads to O(n) algorithmic complexity, and as we are already within a loop, this leads into an O(n^2) algorithm complexity.

We should instead use an data structure with logarithmic lookups directly in order to take this first algorithmic complexity down to a O(log(n)), and the whole algorithm to a O(n log(n)) algorithmic complexity.

Lucky for us, the `IReferenceCollection` class does provide such an access (`IReference? this[string name] { get; }`), but the key of that dictionnary seems to be a `CanonicalName`.

I've modified my code so we use a `ReferenceName` to query the `IReferenceCollection` structure.

asbjornu (approved):
I think this now looks like a decent improvement. Thoughts @arturcic?

arturcic:
> I think this now looks like a decent improvement. Thoughts @arturcic?

Agree. It will need a rebase on top of `main` and it should be good to go

ni-fgenois:
Thanks to you two! I've just rebased my changes :)

ni-fgenois:
@arturcic @asbjornu what is the process to get a new release?

ni-fgenois:
Ok! Is there one planned soon? (as the last one was Nov 15th)

ni-fgenois:
Also, is there anything I can do to help?

arturcic:
I think we can have a release now.

PR #3002, opened 2022-02-17
Sunset .net framework 4.8 and .net5.0

arturcic:
This PR removes the support for .net50 and .net48 in the source code


arturcic:
Yes that was my plan as well. For now I will not merge, need to check first that build infrastructure allows releasing from a  different branch.

PR #3022, opened 2022-03-02
Add GitVersion_CommandLineArguments parameter to MSBuild task

### LINKED ISSUE #3021: [Feature] Allow to pass extra command line parameters to MSBuild task
**Is your feature request related to a problem? Please describe.**
GitVersion has some command line options that can't be passed via MSBuild task, making it impossible to e.g. use non-default config file with stock nuget tasks.

## Detailed Description
Current there is no way to insert into $(GitVersion_ToolArgments) anything besides three predefined options. 

## Context
In my particular case I want to use different configsfor different builds, which is easy with commandline version, but impossible with MSBuild task. There are more options not available via MSBuild too, though not all of them makes sense to expose (e.g. init or /?). Yet I don't see any reason why user wouldn't want to run msbuild with /diag, /overrideconfig or /l + /verbosity. 

## Possible Implementation
There are two ways it can be acieved. 

1. MsBuild can receive raw piece of command line and pass it directly to Exec. 
   - pros: 
     - easy to implement
     - absolute flexibility
     - unlimited forward compatibility for future parameters
   - cons:
     - somewhat higher chance to shoot own leg
     - not that elegant
2. Add separate MsBuild parameter for every option. 
    - pros: 
      - more elegant
      - harder to do something wrong
    - cons:
      - more code
      - more documentation
      - no automatic support for future keys
     
In fact, those approaches are not mutually exclusive, they can easily coexist.

professor-k:
Implemented generic way to pass arbitrary command line parameter for GetVersion.

## Description
This parameter allows to pass arbitrary string which is then literally passed down to Exec.

## Related Issue
Fixes #3021.

## Motivation and Context
Allows to pass in arbitrary extra parameters with just one parameter

## How Has This Been Tested?
Couple manual runs with or without this parameter, both via dotnet/msbuild command line /p key, and via variable declaration in the csproj file. I looked into the code and it doesn't look like there is anything to test msbuild scripts per se. As for C# code I didn't touch it (to be fair, I wasn't even able to open solution, as VS 2022 was hanging every time I was trying to do so). 

As for documentation, I got `Error: The target 'Preview-Documentation' was not found.`. Something must be broken. Anyway, I wrote couple lines section in documentation for MsBuild GetVersion configuration, and it looks okay when viewed via GitHub. Just make sure to proofread, as I'm not a native speaker.

## Checklist:


asbjornu (changes requested):
I think this is a welcome change, especially the added documentation, but the title of the PR seems to be wrong (should probably say `Added GitVersion_CommandLineArguments parameter to MSBuild task`) and I would love to have a test added for this. You can see some tests exercising the MSBuild Task here:

https://github.com/GitTools/GitVersion/blob/32977368a9f29bb215273ca7e205fc7a96b42270/src/GitVersion.MsBuild.Tests/Tasks/GetVersionTaskTests.cs#L27-L41

professor-k:
@asbjornu yes, you are correct about the pull request name. I had two branches, and was considering making pull request for Config override too. 

As for MSBuild script tests. Looked into code once again, and still don't see easy way to do it. Test you pointed out tests MSBuild Task directly without even touching MSBuild scripts. Yet my changes are in .props file and even more, they are not affecting GetVersion task, they are affecting parameters of the Exec task.

Well, I have some idea how to do this: there are other tests that create .csproj and run msbuild on it. I would probably build on that by injecting both .props and .targets into blank .csproj in order to mimic what nuget reference does. But then I would need to find compiled gitversion.dll, which would probably involve adding dependency on GitVersion.App to tests. And it would test way more than just adding one parameter that is then literally passed down to Exec. Nothing impossible, but 1) feels way out of scope of this pull request. 2) Changing architecture of solution is not something I like to do on my own initiative in my first pull request on project I'm just getting myself familiarized with :)

asbjornu:
I agree that the amount of heavy lifting you're describing is a bit much to ask for in a first contributor PR, @professor-k. I do have a feeling we have or at least had something along the lines you describe in place, though. Do you know, @arturcic?

phitattoo:
> Implemented generic way to pass arbitrary command line parameter for GetVersion.
> 
> 
> 
> ## Description
> 
> This parameter allows to pass arbitrary string which is then literally passed down to Exec.
> 
> 
> 
> ## Related Issue
> 
> Fixes #3021.
> 
> 
> 
> ## Motivation and Context
> 
> Allows to pass in arbitrary extra parameters with just one parameter
> 
> 
> 
> ## How Has This Been Tested?
> 
> Couple manual runs with or without this parameter, both via dotnet/msbuild command line /p key, and via variable declaration in the csproj file. I looked into the code and it doesn't look like there is anything to test msbuild scripts per se. As for C# code I didn't touch it (to be fair, I wasn't even able to open solution, as VS 2022 was hanging every time I was trying to do so). 
> 
> 
> 
> As for documentation, I got `Error: The target 'Preview-Documentation' was not found.`. Something must be broken. Anyway, I wrote couple lines section in documentation for MsBuild GetVersion configuration, and it looks okay when viewed via GitHub. Just make sure to proofread, as I'm not a native speaker.
> 
> 
> 
> ## Screenshots (if appropriate):
> 
> 
> 
> ## Checklist:
> 
> 
> 
> 
> 
> 
> 
> - [x] My code follows the code style of this project.
> 
> - [x] My change requires a change to the documentation.
> 
> - [x] I have updated the documentation accordingly.
> 
> - [ ] I have added tests to cover my changes.
> 
> - [ ] All new and existing tests passed.

PR #3036, opened 2022-03-08
Remove NoMono

### LINKED ISSUE #2316: [Improvement] Remove NoMono categories
I think we can remove the `NoMono` category from our tests now, see https://github.com/shouldly/shouldly/issues/376#issuecomment-640912945.

asbjornu:
Resolve #2316 by removing the `NoMono` test category.

## Description
It may be that we can do without the `NoMono` test category now. This PR tests that assumption.

## Related Issue
Resolves #2316.

## Motivation and Context
If we don't need to filter out tests to run, we shouldn't.

## Checklist:

### DIFF
--- a/build/build/Tasks/Test/UnitTest.cs
+++ b/build/build/Tasks/Test/UnitTest.cs
@@ -89,7 +89,7 @@ public class UnitTest : FrostingTask<BuildContext>
 
         if (string.Equals(framework, Constants.FullFxVersion48))
         {
-            settings.Filter = context.IsRunningOnUnix() ? $"TestCategory!={Constants.NoMono}" : $"TestCategory!={Constants.NoNet48}";
+            settings.Filter = context.IsRunningOnUnix() ? string.Empty : $"TestCategory!={Constants.NoNet48}";
         }
 
         context.DotNetTest(project.FullPath, settings, coverletSettings);
--- a/src/GitVersion.Core.Tests/Configuration/ConfigProviderTests.cs
+++ b/src/GitVersion.Core.Tests/Configuration/ConfigProviderTests.cs
@@ -228,8 +228,6 @@ branches:
 
     [Test]
     [MethodImpl(MethodImplOptions.NoInlining)]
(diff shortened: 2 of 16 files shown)


asbjornu:
Ok, it seems like we need to await #3002 for this to be successful as a few former `NoMono` categorized tests are still failing on `net48` on macOS.

arturcic:
@asbjornu I think we could release a new version 5.9.0 that contains the nullable fixes I did and then we can merge the PR with the sunset of net48 and net50 - thoughts?

asbjornu:
Are you suggesting to sunset `net48` and `net50` in 5.9?

arturcic:
No, there are some changes that were merged related to nullability, that one should get a minor version - 5.9. Right after that we can merge the sunset `net48` and `net50` into main and that will be version 6.0 as it's a breaking change.

asbjornu:
Ok, agreed. We need to get a speed on the new CLI that we also want to break in v6, then. 😅

arturcic:
> Ok, agreed. We need to get a speed on the new CLI that we also want to break in v6, then. 😅

Or we can push the CLI to v7.0 to be more in line with the .net version :D

asbjornu:
True, that's also an option. Do we have any other, lower-hanging fruits of breaking changes we want to roll into v6? Or said reversely, which of the issues in the [v6 milestone](https://github.com/GitTools/GitVersion/milestone/19) besides new CLI should we postpone to v7?

arturcic:
> True, that's also an option. Do we have any other, lower-hanging fruits of breaking changes we want to roll into v6? Or said reversely, which of the issues in the [v6 milestone](https://github.com/GitTools/GitVersion/milestone/19) besides new CLI should we postpone to v7?

I moved those related to CLI to v7, the others should stay in v6 I think, can you check as well?

asbjornu:
Yeah, I agree.

asbjornu:
This was resolved in #3002.

Issue #3095, opened 2022-04-20
Change GitVersionInformation C# template ('const string' instead of 'static string')


kvasnevskiy:
Similar request was already open #1376 and then closed. But I see that it's still `static string`. Can template be changed to `const string`? Because I want to use GitVersionInformation in attributes and value there should be constant,

arturcic:
@kvasnevskiy  do you mind creating a PR for this? 
As a starting point here https://github.com/GitTools/GitVersion/blob/support/5.x/src/GitVersion.Core/VersionConverters/GitVersionInfo/AddFormats/GitVersionInformation.cs, You might need to update some of the tests as well

kvasnevskiy:
@arturcic ok, I will create a PR for this.

asbjornu:
As mentioned by @jbaehr in https://github.com/GitTools/GitVersion/issues/1376#issuecomment-412799435, the change from `static` to `const` is a breaking change. For more insight into the reflection side of things, please see https://github.com/GitTools/GitVersion/issues/895#issuecomment-226118785, the linked issues and the surrounding discussion.

arturcic:
Closed by #3113

PR #3102, opened 2022-04-27
Change GitVersionInformation C# template ('const string' instead of 'static string')

kvasnevskiy:
Changed GitVersionInformation C# template from `static string` to `const string` as discussed in #3095. Also I tested this with reflection and I didn't find any problems, I am able to get GitVersionInformation through reflection, maybe I missed smth. But as was mentioned in #3095 this change should be breaking, so  if it's breaking please let me know what exactly stopped working and how I can test it.


asbjornu:
@kvasnevskiy, the breaking change is **how** existing reflection code works. If we change from `static string` to `const string`, existing code that performs reflection to find the `GitVersionInformation` class and its data, will need to change.

asbjornu:
So I'm not opposed to implement this, I just think we need to do it on `main` and fold it into v6 and not release it as a part of v5.

asbjornu (changes requested):
I'm not opposed to implement this, I just think we need to do it on `main` and fold it into v6 and not release it as a part of v5.

arturcic:
@kvasnevskiy right

kvasnevskiy:
Closed and will reopen on `main` branch.

PR #3130, opened 2022-05-28
#2325 - remove legacy variables

### LINKED ISSUE #2325: [Improvement] Remove legacy version variables and related configuration options in v6
We currently have quite a few version variables and related configuration options that are tailored towards sytems that don't support SemVer 2.0, such as `LegacySemVerPadding`, `BuildMetaDataPadding`, `CommitsSinceVersionSourcePadding`, etc.

We should remove these in v6, at least from Core. If pepole still need these, we should look into adding them through an extension to the Core somehow.

arturcic:
FIxes #2325 

This PR introduces breaking changes by removing some legacy variables and configurations. Here is the list of the variables that are removed:

`BuildMetaDataPadded`, `LegacySemVer`, `LegacySemVerPadded`, `NuGetVersionV2`, `NuGetVersion`, `NuGetPreReleaseTagV2`, `NuGetPreReleaseTag`, `CommitsSinceVersionSourcePadded`.

And these are the configurations that are also removed:
`legacy-semver-padding`, `build-metadata-padding`, `commits-since-version-source-padding`


asbjornu (approved):
👏🏼

PR #3131, opened 2022-05-30
Implement SemVer strict parsing

arturcic:
This PR adds 2 ways of parsing the semver string into SemanticVersion object - `Loose` - the old way we used so far, and `Strict` using the regex from https://semver.org.

Updated the tests to use the strict mode by default.


docs/input/docs/reference/configuration.md
```diff
@@ -572,6 +572,18 @@ is set, it would be added to the `PreReleaseNumber` to get a final
 `pre-release-weight` will be used in the calculation. Related Issues [1145]
 and [1366].
 
+### semver-format
+
+Specifies the semver format that is used when parsing the string.
+Can be `Strict` - using the [regex](https://regex101.com/r/Ly7O1x/3/)
+or `Loose` the old way of parsing. Example of invalid `Strict`, but valid `Loose`
```
asbjornu, line 579: We should specify whether `Strict` or `Loose` is the default.

asbjornu (approved):
Looks good, only a little nit on the documentation.

PR #3179, opened 2022-08-26
Add local project file location to exception when parsing fails

### LINKED ISSUE #3178: [Improvement] Report xml file location when xml parsing fails
I was confused when in a solution with 20+ projects I ran gitversion and got an XmlException printed that didn't mention the exact file that was causing problems. And 100 lines of git log were printed in addition.

## Detailed Description
I was trying to update csproject filesin a solution with 20+ projects. I ran gitversion and got an XmlException printed that didn't mention the exact file that was causing problems. At this point I had no idea what files gitversion was even trying to parse that made it fail.

## Context
Updating version info in multiple csproject files in a large solution.

## Possible Implementation
Simply wrap xml reading call in try/catch and add file location in XmlException message.

chhh:
## Description
Just wrap xml parsing call in try/catch.
Helps in solutions with multiple projects, when one of csproj files becomes corrupt. User knows where to apply fixes.

## Related Issue
Fixes #3178.

## Motivation and Context
Updating csproj files in a large solution with multiple projects.

## How Has This Been Tested?
- Ran it in a well-formed solutio with well formed projects - ran fine as usual.
- Corrupted one of `csproj` files by inserting a few random characters before the starting `<Project` element. This causes `System.Xml.Linq.XElement.Parse()` method call to fail parsing. The printed error log after applying the change contained the location of the file and all the same information as before as well (location within the file where the parsing error occured) as well as git log lines.

## Checklist:

### DIFF
--- a/src/GitVersion.Core/VersionConverters/AssemblyInfo/ProjectFileUpdater.cs
+++ b/src/GitVersion.Core/VersionConverters/AssemblyInfo/ProjectFileUpdater.cs
@@ -47,7 +47,15 @@ public sealed class ProjectFileUpdater : IProjectFileUpdater
             var localProjectFile = projectFile.FullName;
 
             var originalFileContents = this.fileSystem.ReadAllText(localProjectFile);
-            var fileXml = XElement.Parse(originalFileContents);
+            XElement fileXml;
+            try
+            {
+                fileXml = XElement.Parse(originalFileContents);
+            }
+            catch (XmlException e)
+            {
+                throw new XmlException($"Unable to parse file as xml: {localProjectFile}", e);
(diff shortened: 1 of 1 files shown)


asbjornu (approved):

asbjornu:
This seems simple enough and something we should be able to roll into v5 as it isn't a breaking change. Do you mind changing your target branch to `support/5.x` so we can get this into production more quickly than if we have to wait for v6 (`main` represents v6) to be done?

chhh:
@asbjornu Done

asbjornu:
Meh, unfortunately a lot of tests are failing.

Issue #3060, opened 2022-03-29
Cannot use empty tag in branch specific configuration


jws305:
The dotnet tool and the Azure Pipeline task are producing different outputs given the same input.

Given a Git graph of:
```
commit (HEAD)
commit (tag: v2.0.11)
...
```

And a GitVersion.yml of:
```yml
mode: ContinuousDeployment
continuous-delivery-fallback-tag: ci
branches:
  main:
    tag: ''
```

I get the following output:
```json
{
  "Major": 2,
  "Minor": 0,
  "Patch": 12,
  "PreReleaseTag": "ci.1",
  "PreReleaseTagWithDash": "-ci.1",
  "PreReleaseLabel": "ci",
  "PreReleaseLabelWithDash": "-ci",
  "PreReleaseNumber": 1,
  "WeightedPreReleaseNumber": 55001,
  "BuildMetaData": null,
  "BuildMetaDataPadded": "",
  "FullBuildMetaData": "Branch.main.Sha.cba365c7e2fa53dbf7935b74d780d542b35eca04",
  "MajorMinorPatch": "2.0.12",
  "SemVer": "2.0.12-ci.1",
  "LegacySemVer": "2.0.12-ci1",
  "LegacySemVerPadded": "2.0.12-ci0001",
  "AssemblySemVer": "2.0.12.0",
  "AssemblySemFileVer": "2.0.12.0",
  "FullSemVer": "2.0.12-ci.1",
  "InformationalVersion": "2.0.12-ci.1+Branch.main.Sha.cba365c7e2fa53dbf7935b74d780d542b35eca04",
  "BranchName": "main",
  "EscapedBranchName": "main",
  "Sha": "cba365c7e2fa53dbf7935b74d780d542b35eca04",
  "ShortSha": "cba365c",
  "NuGetVersionV2": "2.0.12-ci0001",
  "NuGetVersion": "2.0.12-ci0001",
  "NuGetPreReleaseTagV2": "ci0001",
  "NuGetPreReleaseTag": "ci0001",
  "VersionSourceSha": "b69e8de1075fd57dbf79ec4a971fb469cc5e7a55",
  "CommitsSinceVersionSource": 1,
  "CommitsSinceVersionSourcePadded": "0001",
  "UncommittedChanges": 1,
  "CommitDate": "2022-03-29"
}
```
But I would expect to get:
```json
{
  "Major": 2,
  "Minor": 0,
  "Patch": 12,
[...]

PR #3208, opened 2022-09-26
Don't consider "tag-prefix" as optional if it's explicitly specified

### LINKED ISSUE #3216: [Improvement] Don't consider `tag-prefix` as optional
## Detailed Description

Currently `tag-prefix` is always considered as optional which makes calculation of the next version incorrect. It mixes prefixed and unprefixed versions and takes the largest of both.

## Context

To correctly calculate the next version with the prefix. For example, we develop the main product and some its plugin in the same repository and have the following git tags list:
- 0.1.0
- 0.1.1
- 0.2.0
- 0.2.1
- plugin_0.1.0
- plugin_0.1.1
The expected next plugin version is 0.1.2 not 0.2.2 (plugin_ prefix will be prepended before setting the tag).

AlexPykavy:
## Description

To correctly calculate the next version with the prefix.

For example, we develop the main product and some its plugin in the same repository and have the following git tags list:
- 0.1.0
- 0.1.1
- 0.2.0
- 0.2.1
- plugin_0.1.0
- plugin_0.1.1

The expected next plugin version is `0.1.2` not `0.2.2` (`plugin_` prefix will be prepended before setting the tag).

## Related Issue

Resolves #3216.

## Motivation and Context

Makes possible to have 2 independent versioning sequences in the same repository.

## How Has This Been Tested?

Unit test case is added.

## Checklist:


asbjornu:
I'm not sure whether to consider this a breaking change or not. If it isn't, we should re-target and rebase it on `support/5.x` instead of `main` and ship it in a v5 minor release. Also please have a look at #3218 and #3060 which seems related.

AlexPykavy:
@asbjornu , I've reviewed https://github.com/GitTools/GitVersion/issues/3218 (and the corresponding PR https://github.com/GitTools/GitVersion/pull/3224) and https://github.com/GitTools/GitVersion/issues/3060. I agree with https://github.com/GitTools/GitVersion/pull/3224#issuecomment-1277296403 answer that we are talking about different tags.

I think that the only reason why this change can be considered as a breaking if `gitversion init` command puts the default `tag-prefix`. I checked it and looks like it doesn't. Then, may I just retarget the PR if there are no other comments?

asbjornu:
@AlexPykavy, yes, I think you can retarget this against `support/5.x`.

AlexPykavy:
> @AlexPykavy, yes, I think you can retarget this against `support/5.x`.

Done.

Issue #3101, opened 2022-04-27
[Bug] Version not generated correct when creating a feature branch from a release branch


HHobeck:
**Describe the bug**
Hi there. I'm using the GitFlow branching strategy descripted on this page https://gitversion.net/docs/learn/branching-strategies/gitflow/examples. Because I want to ensure the stability of the develop and release branch direct committing is not allowed. Thus I need to go via feature branches and got an unexpected version generation on the following scenario:

```csharp
[Test]
public void __Just_A_Test__()
{
    using var fixture = new EmptyRepositoryFixture();
    fixture.Repository.MakeATaggedCommit("1.0.0");
    fixture.BranchTo("develop");
    fixture.Repository.MakeACommit();
    fixture.BranchTo("release/1.1.0");
    fixture.Repository.MakeACommit();
    fixture.BranchTo("feature/just-a-test");
    fixture.Checkout("develop");
    fixture.Repository.MakeACommit();
    fixture.Checkout("feature/just-a-test");
    fixture.MergeNoFF("develop");
    fixture.Repository.MakeACommit();
    var configuration = new Config()
    {
        Branches = new Dictionary<string, BranchConfig>()
        {
            { "release", new BranchConfig() { TracksReleaseBranches = true } }
        }
    };
    fixture.AssertFullSemver("1.2.0-just-a-test.1+3", configuration); // 1.1.0 expected
}
```
That produces the following repository:
```
    * 8df9b9f 52 minutes ago  (HEAD -> feature/just-a-test)
    *   999e22a 54 minutes ago 
    |\  
    | * 26d759f 55 minutes ago  (develop)
    * | b5addb0 57 minutes ago  (release/1.1.0)
    |/  
    * 065c6eb 59 minutes ago 
    * 73835c8 61 minutes ago  (tag: 1.0.0, main)
```

with following commit messages:
![image](https://user-images.githubusercontent.com/56404113/165685619-c3833e11-dcb6-479f-b093-f0dc4aba294c.png)

## Expected Behavior
Git Version should generate the semantic version 1.1.0 on feature branch just-a-test in this scenario.

## Actual Behavior
The semantic version 1.2.0 will be generated.

## Possible Fix
I'm not sure if it is a bug or I have misconfigured something.

## Steps to Reproduce
Please take look to the test steps above

## Context
I'm using the latest version of GitVersion 5.10.1. My build and deployment pipelines are in AzureDevOps but I can reproduce it locally.

## Your Environment
[...]

PR #3215, opened 2022-10-01
Refactoring/remove the EffectiveConfiguration property in the GitVersionContext class

HHobeck:
## Description
Refactoring/remove the EffectiveConfiguration property in the GitVersionContext class. This change is a preparation for fixing the issue #3101 in PR #3190 and targets the v5 version. No business logic changed.

## Related Issue
[Bug] Version not generated correct when creating a feature branch from a release branch #3101 

## Motivation and Context
This PR is a preparation for PR [#3190](https://github.com/GitTools/GitVersion/pull/3190) and should be integrated in v5.

## How Has This Been Tested?
All unit and integration tests have been executed.

## Checklist:


HHobeck:
Hi @asbjornu and @arturcic. Do you have an idea why the CI pipeline is not passing?

CodeQL / Analyze (scharp) (pull_request):
```
  ========================================
  CodeFormat
  ========================================
  Code format...
  Unrecognized command or argument '-p:UseSharedCompilation=false'.
  Unrecognized command or argument '-p:UseSharedCompilation=false'.
  
  Description:
    Formats code to match editorconfig settings.
  
  Usage:
  format [<The project or solution file to operate on. If a file is not specified, the command will search the current directory for one.>] [command] [options]
```

CI / Unit Test code (windows-latest, net5.0) (pull_request):
```
Error: unknown test file type for 'artifacts/test-results/GitVersion.MsBuild.Tests.net5.0.results.xml'
```

arturcic:
I will check this one

arturcic:
@HHobeck  seems like recently codeql changed a bit the way they run the build so I had to fix that. You will need to rebase on top of support/5.x to have this fixed

asbjornu:
@arturcic, we also have a problem with the `Test Summary` step:

> Error: unknown test file type for 'artifacts/test-results/GitVersion.App.Tests.net48.results.xml'

I think the issue may be that that the `test-summary` action does not support NUnit test reports. Perhaps we should just remove the `Test Summary` step until https://github.com/test-summary/action/issues/12 is resolved?

src/GitVersion.Core/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,48 @@
+abstract GitVersion.VersionCalculation.VersionStrategyBase.GetBaseVersions(GitVersion.Model.Configuration.EffectiveBranchConfiguration! configuration) -> System.Collections.Generic.IEnumerable<GitVersion.VersionCalculation.BaseVersion!>!
```
asbjornu, line 1: Most of these should probably be moved to `.Shipped`, no?

src/GitVersion.Core.Tests/VersionCalculation/BaseVersionCalculatorTests.cs
```diff
@@ -26,10 +26,10 @@ public void ChoosesHighestVersionReturnedFromStrategies()
 
         var baseVersion = versionCalculator.GetBaseVersion();
```
asbjornu, line 27: I don't mind using tuples if we consistently use the tuple syntax introduced in C# 7, so we avoid having non-descriptive properties like `Item1` in the codebase.

```suggestion
        var (baseVersion, _) = versionCalculator.GetBaseVersion();
```

src/GitVersion.Core.Tests/VersionCalculation/BaseVersionCalculatorTests.cs
```diff
@@ -26,10 +26,10 @@ public void ChoosesHighestVersionReturnedFromStrategies()
...
 
-        baseVersion.SemanticVersion.ToString().ShouldBe("2.0.0");
-        baseVersion.ShouldIncrement.ShouldBe(true);
-        baseVersion.BaseVersionSource.ShouldNotBeNull();
-        baseVersion.BaseVersionSource.When.ShouldBe(dateTimeOffset);
+        baseVersion.Item1.SemanticVersion.ToString().ShouldBe("2.0.0");
+        baseVersion.Item1.ShouldIncrement.ShouldBe(true);
+        baseVersion.Item1.BaseVersionSource.ShouldNotBeNull();
+        baseVersion.Item1.BaseVersionSource.When.ShouldBe(dateTimeOffset);
```
asbjornu, line 32:
```suggestion
        baseVersion.SemanticVersion.ToString().ShouldBe("2.0.0");
        baseVersion.ShouldIncrement.ShouldBe(true);
        baseVersion.BaseVersionSource.ShouldNotBeNull();
        baseVersion.BaseVersionSource.When.ShouldBe(dateTimeOffset);
```

src/GitVersion.Core/VersionCalculation/Abstractions/IBaseVersionCalculator.cs
```diff
@@ -1,6 +1,8 @@
+using GitVersion.Model.Configuration;
+
 namespace GitVersion.VersionCalculation;
 
 public interface IBaseVersionCalculator
 {
-    BaseVersion GetBaseVersion();
+    (BaseVersion, EffectiveBranchConfiguration) GetBaseVersion();
```
asbjornu, line 7: An alternative to using tuple here would be to create a class holding the `BaseVersion` and `EffectiveBranchConfiguration` properties. I think I prefer that over tuple, tbh. Especially if you can find a good, descriptive name for it. :)

src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/TaggedCommitVersionStrategy.cs
```diff
@@ -10,18 +11,19 @@ namespace GitVersion.VersionCalculation;
 /// </summary>
 public class TaggedCommitVersionStrategy : VersionStrategyBase
 {
-    private readonly IRepositoryStore repositoryStore;
+    private IRepositoryStore RepositoryStore { get; }
```
asbjornu, line 14: Why are you changing `RepositoryStore` from `readonly` field to property? As it's still `private`, it can just remain a field, no? However, since `IRepositoryStore` seems to be in use in all `VersionStrategy` implementations, perhaps we can just pass it down into `VersionStrategyBase` and expose it as a `protected` property there?

src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/VersionStrategyBase.cs
```diff
@@ -1,12 +1,15 @@
 using GitVersion.Extensions;
+using GitVersion.Model.Configuration;
 
 namespace GitVersion.VersionCalculation;
 
-public class VersionStrategyBase : IVersionStrategy
+public abstract class VersionStrategyBase : IVersionStrategy
```
asbjornu, line 6: 👍🏼

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -98,35 +106,35 @@ public class IncrementStrategyFinder : IIncrementStrategyFinder
...
     /// </summary>
-    private IEnumerable<ICommit> GetIntermediateCommits(IGitRepository repo, IGitObject baseCommit, ICommit? headCommit)
+    private IEnumerable<ICommit> GetIntermediateCommits(IGitRepository repository, IGitObject baseCommit, ICommit? headCommit)
     {
-        var map = GetHeadCommitsMap(repo, headCommit);
+        var map = GetHeadCommitsMap(repository, headCommit);
         if (!map.TryGetValue(baseCommit.Sha, out var baseIndex)) return Enumerable.Empty<ICommit>();
         var commitAfterBaseIndex = baseIndex + 1;
-        var headCommits = GetHeadCommits(repo, headCommit);
+        var headCommits = GetHeadCommits(repository, headCommit);
```
asbjornu, line 117: 👍🏼

src/GitVersion.Core/VersionCalculation/MainlineVersionCalculator.cs
```diff
@@ -291,9 +291,9 @@ private VersionField TryFindIncrementFromMergeMessage(ICommit? mergeCommit)
         return FindDefaultIncrementForBranch(context);
     }
 
-    private static VersionField FindDefaultIncrementForBranch(GitVersionContext context, string? branch = null)
+    private static VersionField FindDefaultIncrementForBranch(GitVersionContext context, string? branchName = null)
     {
-        var config = context.FullConfiguration.GetConfigForBranch(branch ?? context.CurrentBranch.Name.WithoutRemote);
+        var config = context.FullConfiguration.GetBranchConfiguration(branchName ?? context.CurrentBranch.Name.WithoutRemote);
```
asbjornu, line 296: 👍🏼

src/GitVersion.Core/VersionCalculation/NextVersion.cs
```diff
@@ -0,0 +1,30 @@
+using GitVersion.Extensions;
+using GitVersion.Model.Configuration;
+
+namespace GitVersion.VersionCalculation;
+
+public class NextVersion
```
asbjornu, line 6: I like how you encapsulate everything related to the "next version" into a class. 👍🏼

src/GitVersion.Core/VersionCalculation/NextVersionCalculator.cs
```diff
@@ -11,21 +12,23 @@ public class NextVersionCalculator : INextVersionCalculator
...
     private readonly IMainlineVersionCalculator mainlineVersionCalculator;
     private readonly IRepositoryStore repositoryStore;
+    private readonly IIncrementStrategyFinder incrementStrategyFinder;
     private readonly Lazy<GitVersionContext> versionContext;
     private GitVersionContext context => this.versionContext.Value;
 
     public NextVersionCalculator(ILog log, IBaseVersionCalculator baseVersionCalculator,
         IMainlineVersionCalculator mainlineVersionCalculator, IRepositoryStore repositoryStore,
-        Lazy<GitVersionContext> versionContext)
+        IIncrementStrategyFinder incrementStrategyFinder, Lazy<GitVersionContext> versionContext)
```
asbjornu, line 21: Can you please just format this constructor so every argument sits on its own line? 

We should also do something about the number of arguments here, as everything more than 3 is too much, imho. Not in this PR, though. :)

src/GitVersion.Core/VersionCalculation/SemanticVersioning/VersionField.cs
```diff
@@ -0,0 +1,9 @@
+namespace GitVersion;
+
+public enum VersionField
```
asbjornu, line 3: Good stuff moving `VersionField` to its own file. 👍🏼

asbjornu (changes requested):
Thank you so much for your hard work on this @HHobeck! It is much appreciated. 🙏🏼 Please have a look at my review comments. As soon as they are resolved and the tests are fixed, I think this should be merged. Then we'll do a merge of `support/5.x` into `main` so you can rebase #3190 on top of the great work you've done here.

src/GitVersion.Core.Tests/Configuration/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
```diff
@@ -57,10 +57,8 @@ branches:
     pre-release-weight: 30000
   feature:
     mode: ContinuousDelivery
-    tag: useBranchName
+    tag: '{BranchName}'
```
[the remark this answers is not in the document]
HHobeck: Yep that's true. Here are some aspects I need to explain. Please let me know if you are not agree with my argumentation or if you have any better solutions:

1. I have removed the default initialization behavior of BranchConfig and the usage of BranchConfig.CreateDefaultBranchConfig. This was necessary because otherwise we cannot inherit the properties Tag, PreventIncrementOfMergedBranchVersion, TrackMergeTarget, TracksReleaseBranches, IsReleaseBranch, IsMainline when specifying the IncrementStrategy to Increment.Inherit. You see that the changes only effects this branch configuration where the IncrementStrategy flag is set to inherit.
2. The tag useBranchName name is equivalent to {BranchName}. In my opinion useBranchName is a magic string and not so intuitive. Better using templates like {BranchName} which is clear that this will be replaced with the name of the branch. I can think of that we have other properties in future.
3. Hotfix branches can be created from release (new), main, support, hotfix (new) but not from develop.
4. The default behavior if you not defining IncrementStrategy is Inherit

src/GitVersion.Core/Configuration/ConfigExtensions.cs
```diff
@@ -7,13 +7,55 @@ namespace GitVersion.Configuration;
...
         {
-            throw new ArgumentNullException(nameof(branchName));
+            branchConfiguration = GetUnknownBranchConfiguration(configuration);
+            branchConfiguration.Name = branchName;
         }
+        return branchConfiguration;
+    }
+
+    // TODO: Please make the unknown settings also configurable in the yaml.
+    public static BranchConfig GetUnknownBranchConfiguration(this Config configuration) => new()
```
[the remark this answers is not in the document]
HHobeck: Of course we can do it. What is your recommendation?

src/GitVersion.Core/Configuration/ConfigExtensions.cs
```diff
@@ -7,13 +7,55 @@ namespace GitVersion.Configuration;
...
+    {
+        Name = "Unknown",
+        Regex = "",
+        Tag = "{BranchName}",
+        VersioningMode = configuration.VersioningMode,
+        Increment = IncrementStrategy.Inherit
+    };
 
+    // TODO: Please make the fallback settings also configurable in the yaml.
+    public static BranchConfig GetFallbackBranchConfiguration(this Config configuration)
```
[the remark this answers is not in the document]
HHobeck: Of course we can do it. What is your recommendation? Do you mean to place it in the yaml file? This will be a breaking change or at least an additional change. I would say we are doing this in v6.

src/GitVersion.Core/Configuration/ConfigExtensions.cs
```diff
@@ -7,13 +7,55 @@ namespace GitVersion.Configuration;
 
 public static class ConfigExtensions
 {
-    public static BranchConfig? GetConfigForBranch(this Config config, string? branchName)
+    public static BranchConfig GetBranchConfiguration(this Config configuration, IBranch branch)
```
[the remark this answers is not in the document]
HHobeck: I totally agree with you. An extensions class only makes sense if you are not owner or you want to place common logic on the top of an e.g. interface. But like the Domain Driven Design recommended it: It is cleaner to have business entities with the explicit business logic.

src/GitVersion.Core/PublicAPI.Unshipped.txt
HHobeck: Actually I'm not certain what the pattern is behind it. So I just try to fix the warnings. ;) If you like I can move it to Shipped never mind.

src/GitVersion.Core/Core/RepositoryStore.cs
```diff
@@ -158,7 +155,7 @@ public IEnumerable<IBranch> GetBranchesForCommit(ICommit commit)
     public IEnumerable<IBranch> GetExcludedInheritBranches(Config configuration)
         => this.repository.Branches.Where(b =>
         {
-            var branchConfig = configuration.GetConfigForBranch(b.Name.WithoutRemote);
+            var branchConfig = configuration.FindConfigForBranch(b.Name.WithoutRemote);
```
arturcic, line 158: I saw there are some renamed methods called _sometingConfig_ some _somethingConfiguration_. I think we need to agree to a naming convention for these methods, and I'd suggest the Configuration name to be used

src/GitVersion.Core/Core/RepositoryStore.cs
```diff
@@ -4,23 +4,20 @@
...
 public class RepositoryStore : IRepositoryStore
 {
-    private readonly IIncrementStrategyFinder incrementStrategyFinder;
     private readonly ILog log;
     private readonly IGitRepository repository;
     private readonly Dictionary<IBranch, List<SemanticVersion>> semanticVersionTagsOnBranchCache = new();
     private readonly MergeBaseFinder mergeBaseFinder;
 
-    public RepositoryStore(ILog log, IGitRepository repository, IIncrementStrategyFinder incrementStrategyFinder)
+    public RepositoryStore(ILog log, IGitRepository repository)
```
arturcic, line 17: 👍 happy that the RepositoryStore is finally having dependency only on Log and Repository interfaces

src/GitVersion.Core.Tests/VersionCalculation/BaseVersionCalculatorTests.cs
HHobeck: Jep good point. I'm going to change it. Actually it's not so ideal to use Tuples as a result for GetBaseVersion. But here I used is because it will be fused with NextVersionCaluclator in the v6 refactoring.

src/GitVersion.Core/Configuration/ConfigExtensions.cs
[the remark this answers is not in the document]
asbjornu: Ah, no, I meant it purely as a property available in C#. I didn't think about the fact that `Config` is serialized to disk. I think that should be a separate concern, to be honest. But not something we'll tackle in this PR. Ideally, our `Config` class should be an immutable C# class offering the rest of the codebase abstractions related to the configuration. How the configuration is serialized and deserialized is a concern I would like to separate to the perimeter of the application. But as mentioned, not in this PR.

src/GitVersion.Core.Tests/VersionCalculation/BaseVersionCalculatorTests.cs
HHobeck: Done

src/GitVersion.Core/Configuration/ConfigExtensions.cs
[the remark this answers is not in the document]
asbjornu: Exactly. We should try to tackle this in the future so our core `Config` class doesn't have to be fitted for serialization and deserialization.

src/GitVersion.Core/VersionCalculation/Abstractions/IBaseVersionCalculator.cs
HHobeck: I haven't introduce a class because I would like to fuse the BaseVersionCalculator with NextVersionCalculator in the v6 refactoring. But if you disagree than I can do it no problem. Please let me know.

src/GitVersion.Core/PublicAPI.Unshipped.txt
asbjornu: If I'm not mistaken, these files are artifacts from `Microsoft.CodeAnalysis.PublicApiAnalyzers`. [You can read more about it here](https://www.drunkonmonads.com/tracking-breaking-changes/).

src/GitVersion.Core/Core/RepositoryStore.cs
asbjornu: I agree.

src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/TaggedCommitVersionStrategy.cs
HHobeck: Good idea I'm doing this and move it to VersionStrategyBase.

src/GitVersion.Core/VersionCalculation/NextVersion.cs
HHobeck: Thank you very much for the positive comments.

src/GitVersion.Core/VersionCalculation/NextVersionCalculator.cs
HHobeck: Totally agree with you. In my refactoring of v6 I already removed some calls to IRepositoryStore. In my opinion this dependency needs to be removed:

1. IRepositoryStore need to be removed because it is to low level and not necessary
2. The logic of IBaseVersionCalculator will be merged to NextVersionCalculator.
3. GitVersionContext should be not created in the DI factory. This class holds context related properties with a dedicated scope in the NextVersionCalculator class.
4.  IIncrementStrategyFinder is also obsolete in long-term in my opinion. It makes no sense because we have more than one increment strategy in v6.
5. IMainlineVersionCalculator is also obsolete in long-term in my opinion. It makes no sense to have a business logic for a dedicated mode. This gives to much if then else. Better would be to write dedicated IVersionStrategy implementation and make it configurable via the configuration file.

src/GitVersion.Core/Core/RepositoryStore.cs
HHobeck: I agree with you. In my opinion abbreviation should be avoided. Better to name it configuration.

But we have some conflicts here if I think about the ConfigBuilder and the ConfigurationBuilder. Both have different usage:

1. The ConfigurationBuilder is used to create the Config class for productive.
2. The ConfigBuilder is used for unit and integration tests.

src/GitVersion.Core/Core/RepositoryStore.cs
HHobeck: Yep that's a good sign

src/GitVersion.Core/Core/RepositoryStore.cs
HHobeck: Also I use Find for a method if the result can be null. If not then I use normally Get

HHobeck: In the v6 refactoring the GetExcludedInheritBranches method will be removed and the FindConfigForBranch (at this moment internal) will be marked as private. The new GetBranchConfiguration should be used instead.

src/GitVersion.Core/VersionCalculation/NextVersionCalculator.cs
arturcic: > 5\. IMainlineVersionCalculator is also obsolete in long-term in my opinion. It makes no sense to have a business logic for a dedicated mode. This gives to much if then else. Better would be to write dedicated IVersionStrategy implementation and make it configurable via the configuration file.

I agree with this

src/GitTools.Testing/Fixtures/RepositoryFixtureBase.cs
```diff
@@ -65,7 +65,7 @@ protected virtual void Dispose(bool disposing)
 
     public void Checkout(string branch) => Commands.Checkout(Repository, branch);
 
-    public static void Init(string path) => GitTestExtensions.ExecuteGitCmd($"init {path} -b main");
+    public static void Init(string path, string branchName) => GitTestExtensions.ExecuteGitCmd($"init {path} -b {branchName}");
```
arturcic, line 68: 👍

src/GitVersion.Core/PublicAPI.Unshipped.txt
HHobeck: Done.

src/GitVersion.Core/VersionCalculation/NextVersionCalculator.cs
asbjornu: I think I agree with everything. Especially point 5 where I think each `mode` should give a different `Strategy` that all use the same `Calculator` implementations in different ways depending on how stuff is configured. So a `Strategy` basically just serves as a default template of configuration and then every key can be overridden in `GitVersion.yml`, resulting in a custom strategy based on the selected strategy as defined with `mode`. I hope that makes sense. :)

HHobeck: Done

src/GitVersion.Core/Core/RepositoryStore.cs
HHobeck: Done

src/GitVersion.Core.Tests/Helpers/ConfigBuilder.cs
```diff
@@ -0,0 +1,161 @@
+using GitVersion.Configuration;
+using GitVersion.Model.Configuration;
+using GitVersion.VersionCalculation;
+
+namespace GitVersion.Core.Tests.Helpers;
+
+public sealed class ConfigBuilder
```
arturcic, line 7: just rename this one to TestConfigurationBuilder

arturcic, line 7: just rename this one to TestConfigurationBuilder

arturcic (changes requested):

src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
```diff
@@ -34,7 +34,7 @@ private BranchConfig GetBranchConfigurationInternal(int recursions, IBranch targ
             throw new InfiniteLoopProtectionException($"Inherited branch configuration caused {recursions} recursions. Aborting!");
         }
 
-        var matchingBranches = configuration.GetConfigForBranch(targetBranch.Name.WithoutRemote);
+        var matchingBranches = configuration.FindConfigurationForBranch(targetBranch.Name.WithoutRemote);
```
asbjornu, line 37: Would it be better if we just passed `targetBranch` directly to the method and allowed it to resolve `.Name.WithoutRemote` itself? [I don't like passing around `string` and other primitives if we can avoid it](https://refactoring.guru/smells/primitive-obsession).

Also, since the method sits on the `Config` object, perhaps we can just shorten it to `.ForBranch()`? I think it is clear that what we're asking for is the configuration for a specific branch that way.


```suggestion
        var matchingBranches = configuration.ForBranch(targetBranch);
```

asbjornu (changes requested):

src/GitVersion.Core.Tests/Helpers/ConfigBuilder.cs
HHobeck: Done

src/GitVersion.Core/Configuration/BranchConfigurationCalculator.cs
HHobeck: Done

asbjornu (approved):
👏🏼 Now we just need to figure out what to do with the failing `Test Summary` task. Thoughts @arturcic?

arturcic:
> 👏🏼 Now we just need to figure out what to do with the failing `Test Summary` task. Thoughts @arturcic?

The interesting thing is on main the Test summary works just fine, sometimes the macos is failing but on rerun it's ok

asbjornu:
> The interesting thing is on main the Test summary works just fine, sometimes the macos is failing but on rerun it's ok

Is it really fine, or does it work when `artifacts/test-results/*.results.xml` returns no files, while it fails when it does?

src/GitVersion.Core.Tests/Configuration/ConfigProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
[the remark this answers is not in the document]
asbjornu: Ok, sounds good.

arturcic:
> > The interesting thing is on main the Test summary works just fine, sometimes the macos is failing but on rerun it's ok
> 
> Is it really fine, or does it work when `artifacts/test-results/*.results.xml` returns no files, while it fails when it does?

if you check this run for example https://github.com/GitTools/GitVersion/actions/runs/3173292935 you will see 5 test summaries for every platform excluding macos + netcore3.1

HHobeck:
> > > The interesting thing is on main the Test summary works just fine, sometimes the macos is failing but on rerun it's ok
> > 
> > 
> > Is it really fine, or does it work when `artifacts/test-results/*.results.xml` returns no files, while it fails when it does?
> 
> if you check this run for example https://github.com/GitTools/GitVersion/actions/runs/3173292935 you will see 5 test summaries for every platform excluding macos + netcore3.1

What does this mean? I can see the following error:

```
Run test-summary/action@v1
Error: unknown test file type for 'artifacts/test-results/GitVersion.MsBuild.Tests.net5.0.results.xml'
```

On the main branch this error seems to be not existing.

arturcic:
@HHobeck I think on support branch we can disable the test-summary action, so please rebase again on support branch, I disabled it there

HHobeck:
> @HHobeck I think on support branch we can disable the test-summary action, so please rebase again on support branch, I disabled it there

No problem I can do this. Please let me know if you have changed it.

image

arturcic:
@HHobeck you can rebase it now

arturcic (approved):

arturcic:
@HHobeck you don't need to push new commits to re-trigger the failed unit test step. I can retrigger it

arturcic:
Yes, that is the plan

Issue #2388, opened 2020-08-19
[Docs] Use of GitVersionTask package


Tognolo:
Hello, I'd like to use the GitVersionTask package to version my code, but I can't get information about it's configuration. I see a lot of configuration about the usage of the MsBuild setup, but it's not clear if the GitVersionTask package is enough or I've to install other packages. Any demo configuration is blessed.

asbjornu:
You just install the NuGet package. Nothing else should be required.

Tognolo:
I did it. But I can't get the right version in my libraries. E.g., I'm getting a version 2.4.0 for mi DLLs, even if I didn't wrote that tag in any of my projects, neither in the code. Also, it's not clear which branch is using to work. 

How can I run it in debug mode?

asbjornu:
If you want more insight into why and how GitVersion calculates a given version number, I recommend you install the command line version and run it in the solution root. Its logs should give you an idea into what it bases the version number on. The command line is packaged in many different ways, so which one you should install is up to you. You can find the different variants detailed in #2326.

Tognolo:
Running the command line gave me a better comprehension of how it works. Now I can ask a more specific question. I have the current tags sequences (and relative incremented versions):

* 2.3.9 (current tag) >> 2.3.9
* 2.3.8 >> 2.4.0
* <etc>
* 2.3.3 (the oldest, the previous ones are gone...) >> 2.4.0

In `BaseVersionCalculator.GetBaseVersion` it gets all the base versions and for each one calcs the the incremented version, so for every commit it will be 2.4.0, except the current one. At this point there is:

```cs
var maxVersion = baseVersions
    .Aggregate((v1, v2) => v1.IncrementedVersion > v2.IncrementedVersion ? v1 : v2); // it is 2.4.0
var matchingVersionsOnceIncremented = baseVersions
    .Where(b => b.Version.BaseVersionSource != null && b.IncrementedVersion == maxVersion.IncrementedVersion)
    .ToList();
```

`matchingVersionsOnceIncremented` has every commit except the current one, because 2.3.9 < 2.4.0. At this point I need to understand how to:

* bump patch instead of minor value;
* set versioning mode to MainLIne.

asbjornu:
I'm happy you've figured out so much by yourself, @Tognolo. Can you please paste your `GitVersion.yml` file and explain which branches you have and version(s) you expect?

Tognolo:
This is my `GitVersion.yml`:

```yaml
[...]

PR #3226, opened 2022-10-12
Feature/3101 rename config to configuration

HHobeck:
- Rename the class Config to Configuration. Change namespace Configuration to Configurations.
- Rename the class BranchConfig to BranchConfiguration. Change parameter name config to configuration..

## Description
This is a refactoring step without any business logic change.

## Related Issue
no related issue.

## Motivation and Context
Refactoring of Configuration to have clear naming.

## How Has This Been Tested?
All unit and integration tests have been executed.

## Checklist:


asbjornu:
This is great, but please let's discuss changes like this in an issue before you invest time in a PR. I think the namespace should remain `GitVersion.Configuration` and the `Config` class can be renamed to `GitVersionConfiguration`. Thoughts @HHobeck @arturcic?

HHobeck:
> This is great, but please let's discuss changes like this in an issue before you invest time in a PR. I think the namespace should remain `GitVersion.Configuration` and the `Config` class can be renamed to `GitVersionConfiguration`. Thoughts @HHobeck @arturcic?

Okay no problem. I can change it back. I did the namespace change because otherwise the namespace conflicts with the class name. Actually I think we are in the domain of GitVersion to name the class GitVersionConfiguration or GitVersionContext seem for me technically motivated. What about OverallConfiguration!? ;)

asbjornu:
> I think we are in the domain of GitVersion to name the class GitVersionConfiguration or GitVersionContext seem for me technically motivated. 

As the configuration file is named `GitVersion.yml`, I think the name `GitVersionConfiguration` is quite fitting.

> What about OverallConfiguration!? ;)

Not a fan, sorry.

src/GitVersion.Core/Configuration/ConfigFileLocator.cs
```diff
@@ -39,7 +39,7 @@ public void Verify(string? workingDirectory, string? projectRootDirectory)
         return GetConfigFilePath(HasConfigFileAt(workingDirectory) ? workingDirectory : projectRootDirectory);
     }
 
-    public Config ReadConfig(string workingDirectory)
+    public GitVersionConfiguration ReadConfig(string workingDirectory)
```
asbjornu, line 42: [`ConfigFileLocator`](https://github.com/GitTools/GitVersion/pull/3226/files#diff-6491018f0aac7f0c026e4566f551df102c3e2070267f4cfaebcedcf737604d9a) can probably be renamed to `ConfigurationFileLocator`.

src/GitVersion.Core/Configuration/ConfigSerializer.cs
```diff
@@ -6,21 +6,21 @@ namespace GitVersion.Configuration;
 
 public class ConfigSerializer
```
asbjornu, line 7: I would like everything related to serialization gone from Core. Not sure that's something we should do in this PR, but just FYI. Ideally, `GitVersion.Core` should not have any external dependencies.

```suggestion
public class ConfigurationSerializer
```

src/GitVersion.Core/Configuration/Init/BuildServer/AppVeyorSetup.cs
```diff
@@ -5,12 +5,6 @@
 
 namespace GitVersion.Configuration.Init.BuildServer;
 
-internal enum ProjectVisibility
-{
-    Public = 0,
-    Private = 1
-}
-
 internal class AppVeyorSetup : ConfigInitWizardStep
```
asbjornu, line 8: FYI, I don't think `AppVeyorSetup` belongs in `GitVersion.Core`. The entire `Init` directory should probably be moved into the `GitVersion.App` project. Not something we need to deal with in this PR, but perhaps in a future one.

src/GitVersion.Core/Configuration/Init/BuildServer/ProjectVisibility.cs
```diff
@@ -0,0 +1,7 @@
+namespace GitVersion.Configuration.Init.BuildServer;
+
+internal enum ProjectVisibility
```
asbjornu, line 3: 👍🏼

asbjornu (changes requested):
Awesome! 👏🏼

src/GitVersion.Core/Configuration/Init/BuildServer/AppVeyorSetup.cs
arturcic: I agree with that, and I wonder if we can deprecate the 'init' in v6 as it's not maintained and it does not cover all the options we currently have implemented. What do you think @asbjornu

asbjornu: I agree that if we are to keep it, it needs a major revamp. So let's say that if no one invests time to revamp it completely until the release of `6.0.0`, we'll scrap it entirely.

src/GitVersion.Core/Configuration/ConfigSerializer.cs
HHobeck: > I would like everything related to serialization gone from Core. Not sure that's something we should do in this PR, but just FYI. Ideally, `GitVersion.Core` should not have any external dependencies.

Good point. I agree the core dll should not have so many dependencies. I think we should do this in a separate PR.

arturcic:
There are a couple of other classes to be renamed:
OverrideConfigOptionParser
ConfigExtensionsTests
ConfigProviderTests
DefaultConfigFileLocatorTests 
IgnoreConfigTests

src/GitVersion.Core.Tests/IntegrationTests/DevelopScenarios.cs
```diff
@@ -54,13 +54,13 @@ public void WhenDevelopBranchedFromTaggedCommitOnMainVersionDoesNotChange()
     [Test]
     public void CanChangeDevelopTagViaConfig()
     {
-        var config = new Config
+        var configuration = new GitVersionConfiguration
```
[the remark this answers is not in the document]
HHobeck: I agree we need to change the behavior of how to passing in the configuration (change partial to complete). At the moment the TestGitVersionConfigurationBuilder creates per default the git flow configuration you can override. When you define additional branches it gives an error. In my opinion we need to change it and support other branches as well. But before invest time I would like to point out some problems in a productive scenario:

If you want to override e.g. the BranchName of a feature branch with the value null then this is not possible because in the code we are not distinguish a property which hasn't been set and a property which is null.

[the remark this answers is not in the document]
HHobeck: But before I'm going to invest time I would like to point out some problems in a productive scenario:

If you want to override e.g. the Tag of a feature branch (defaults to {BranchName}) with the value null then this is not possible because in the code we are not distinguish a property which hasn't been set and a property which is null.

```csharp
var initalConfiguration = TestConfigurationBuilder.New.WithoutTag("feature").Build();
initalConfiguration.Branches["feature"].Tag.ShouldBe(null); // okay
var configurationBuilder = new ConfigurationBuilder();
configurationBuilder.Add(initalConfiguration);
var configuration = configurationBuilder.Build();
configuration.Branches["feature"].Tag.ShouldBe(null); // exception because Tag
```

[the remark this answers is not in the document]
HHobeck: I would say not to do anything in this PR and pick it up in this issue https://github.com/GitTools/GitVersion/issues/2388. What do you think?

src/GitVersion.Core.Tests/Configuration/IgnoreConfigurationTests.cs
```diff
@@ -20,12 +19,12 @@ public void CanDeserialize()
...
         using var reader = new StringReader(yaml);
-        var config = ConfigSerializer.Read(reader);
+        var configuration = ConfigurationSerializer.Read(reader);
 
-        config.Ignore.ShouldNotBeNull();
-        config.Ignore.ShAs.ShouldNotBeEmpty();
-        config.Ignore.ShAs.ShouldBe(new[] { "b6c0c9fda88830ebcd563e500a5a7da5a1658e98" });
-        config.Ignore.Before.ShouldBe(DateTimeOffset.Parse("2015-10-23T12:23:15"));
+        configuration.Ignore.ShouldNotBeNull();
+        configuration.Ignore.ShAs.ShouldNotBeEmpty();
```
asbjornu, line 25: Huh, I've never noticed this before. Why is `ShAs` written with an uppercase `A`? 🧐

src/GitVersion.Core.Tests/IntegrationTests/DevelopScenarios.cs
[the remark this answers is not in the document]
asbjornu: Agreed.

src/GitVersion.Core/Configuration/Init/BuildServer/AppVeyorSetup.cs
asbjornu: I created #3229 to track this.

src/GitVersion.Core/Model/GitVersionContext.cs
```diff
@@ -9,9 +8,9 @@ namespace GitVersion;
 public class GitVersionContext
```
asbjornu, line 8: Feel free to move every other file inside the `Model` directory, such as `GitVErsionContext.cs`, as well. It would be good to get rid of the `GitVersion.Model` namespace entirely.

asbjornu (changes requested):
Looking good! 👍🏼

HHobeck:
Okay I'm done please take a look and give me feedback. I would prefer to hold on and not to refactor anything else in the PR.

HHobeck:
Please also notice the implementation of ScratchConfigurationBuilder (I'm not sure if EmptyConfigurationBuilder is maybe better!?) and GitFlowConfigurationBuilder. In my opinion this should be used for unit unit / integration tests.

src/GitVersion.Core.Tests/Configuration/IgnoreConfigurationTests.cs
HHobeck: Done

arturcic (changes requested):

arturcic:
@asbjornu when you're done with the review please merge this one

asbjornu (approved):
Marvellous work @HHobeck! 👏🏼

src/GitVersion.Core/VersionCalculation/Cache/GitVersionCacheKey.cs
```diff
@@ -1,15 +1,13 @@
 using GitVersion.Extensions;
 
-namespace GitVersion.Cache;
+namespace GitVersion.VersionCalculation.Cache;
```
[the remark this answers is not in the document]
HHobeck: Really nice talk. I like it very much how you explain the SOLID pattern and moving at the end to a domain driven design. :)

So that means we need to name the Configuration namespace to Configuring!?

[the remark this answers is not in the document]
asbjornu: Thanks! Haha! While I think using gerunds is a nice general rule, the word "configuring" doesn't work that well, imho. 😄

Issue #1627, opened 2019-03-06
GitVersion has a bug, your HEAD has moved after repo normalisation.


LordMike:
We have a peculiar issue that keeps happening on one specific commit, that is not at the tip of our branch. The latest commit in this branch does not exhibit the issue (so we're past it). I'm creating this issue because it might show some other problem you guys have (and because the exception told me to).

```
$ --- equivalent of dotnet-gitversion ---
GitVersion has a bug, your HEAD has moved after repo normalisation.

To disable this error set an environmental variable called IGNORE_NORMALISATION_GIT_HEAD_MOVE to 1

Please run `git log --graph --format="%h %cr %d" --decorate --date=relative --all --remotes=* -n 100` and submit it along with your build log (with personal info removed) in a new issue at https://github.com/GitTools/GitVersion
   at GitVersion.GitRepositoryHelper.NormalizeGitDirectory(String gitDirectory, AuthenticationInfo authentication, Boolean noFetch, String currentBranch)
   at GitVersion.GitPreparer.Initialise(Boolean normaliseGitDirectory, String currentBranch, Boolean shouldCleanUpRemotes)
   at GitVersion.ExecuteCore.ExecuteGitVersion(String targetUrl, String dynamicRepositoryLocation, Authentication authentication, String targetBranch, Boolean noFetch, String workingDirectory, String commitId, Config overrideConfig, Boolean noCache)
   at BuildLib.Tasks.ArtBuildTasks.DetermineVersionInfo(ChronosBuildBase buildBase) in C:\Project\Builder\BuildTasks.cs:line 27
   at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
   at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
   at System.Lazy`1.CreateValue()
   at BuildLib.Bases.ChronosBuildBase.<get_GenerateVersionInfo>b__40_1() in C:\Project\Builder\BuildBase.cs:line 118
   at System.Collections.Generic.List`1.ForEach(Action`1 action)
   at Nuke.Common.Execution.BuildExecutor.Execute(NukeBuild build, IEnumerable`1 executionList)
   at Nuke.Common.Execution.BuildExecutor.Execute[T](Expression`1 defaultTargetExpression)
```

[...]

PR #3245, opened 2022-10-25
Adds trouble shooting info for #1627

rominator1983:
## Description
Adds trouble shooting information for the case discussed in https://github.com/GitTools/GitVersion/issues/1627

## Related Issue
https://github.com/GitTools/GitVersion/issues/1627

## Motivation and Context
Get more information when this case happens

## How Has This Been Tested?
Unit tests have been run. It is not clear how this scenario can be created so it's not possible at the moment to create a unit test

## Checklist:

- [x] My change requires a change to the documentation. => no
- [x] I have updated the documentation accordingly. => no
- [x] I have added tests to cover my changes. => no. See above
- [x] All new and existing tests passed. => yes

### DIFF
--- a/src/GitVersion.Core/Core/GitPreparer.cs
+++ b/src/GitVersion.Core/Core/GitPreparer.cs
@@ -156,47 +156,49 @@ public class GitPreparer : IGitPreparer
         var expectedSha = this.repository.Head.Tip?.Sha;
         var expectedBranchName = this.repository.Head.Name.Canonical;
 
-        try
+        var remote = EnsureOnlyOneRemoteIsDefined();
+        EnsureRepositoryHeadDuringNormalisation(nameof(EnsureOnlyOneRemoteIsDefined), expectedSha);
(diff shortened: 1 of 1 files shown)


asbjornu:
I can't really see why the build is failing – I can't find the test that fails. Are you able to locate it?

rominator1983:
I don't think that there are really single tests that fail. There are complete configurations like MacOS that fail while most configurations just work. I can check later.

Can you run the build on the current main? I think that this won't work either at the moment.

arturcic:
I noticed the build recently got pretty unstable, especially the macOs agents, seems like there were some build agents updates that might affect the stability of the build, for now re-run of the failed jobs helps (sometimes couple of reruns)

rominator1983:
Yeah tell me about it. We experience the same in other repositories. I think you restarted the checks? Thanks for that.

arturcic (approved):

PR #3028, opened 2022-03-03
Change `Int32` to `Int64`

asbjornu:
Change the following properties from `Int32` to to `Int64` to avoid `System.OverflowException` on `int.Parse()` of very large numbers:

- `SemanticVersion.Major`
- `SemanticVersion.Minor`
- `SemanticVersion.Patch`
- `SemanticVersionBuildMetaData.CommitsSinceTag`
- `SemanticVersionBuildMetaData.CommitsSinceVersionSource`
- `SemanticVersionBuildMetaData.UncommittedChanges`
- `SemanticVersionPreReleaseTag.Number`

## Related Issue
Fixes #2715.
Fixes #2915.
Fixes #2390.

## Motivation and Context
Large numbers in a version number may cause the following lines to throw a `System.OverflowException`:

https://github.com/GitTools/GitVersion/blob/32977368a9f29bb215273ca7e205fc7a96b42270/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersion.cs#L171-L173

https://github.com/GitTools/GitVersion/blob/32977368a9f29bb215273ca7e205fc7a96b42270/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersionBuildMetaData.cs#L127

https://github.com/GitTools/GitVersion/blob/32977368a9f29bb215273ca7e205fc7a96b42270/src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersionPreReleaseTag.cs#L79

## How Has This Been Tested?
I've added test cases that failed before the code changes were implemented.

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #3248, opened 2022-10-27
Handle int64 numbers when writing version variables

RatserX:
## Description

Long numeric values for certain version variable fields (e.g. PreReleaseNumber) cause the number parsing to fail and throw the InvalidOperationException.
[WriteNumberValue](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.utf8jsonwriter.writenumbervalue?view=net-6.0#system-text-json-utf8jsonwriter-writenumbervalue(system-int64)) supports int64 so this should be a pretty safe update.

## Checklist:

### DIFF
--- a/src/GitVersion.Core/Model/VersionVariablesJsonNumberConverter.cs
+++ b/src/GitVersion.Core/Model/VersionVariablesJsonNumberConverter.cs
@@ -34,7 +34,7 @@ public class VersionVariablesJsonNumberConverter : JsonConverter<string>
         {
             writer.WriteNullValue();
         }
-        else if (int.TryParse(value, out var number))
+        else if (long.TryParse(value, out var number))
         {
             writer.WriteNumberValue(number);
         }


asbjornu:
This seems like something we should release in v5. It also feels like something that should be roll into @james-bjss' #2703 which also deals with integer parsing. Perhaps you can pick up his PR and rebase it against `support/5.x`, @RatserX?

RatserX:
> This seems like something we should release in v5. It also feels like something that should be roll into @james-bjss' #2703 which also deals with integer parsing. Perhaps you can pick up his PR and rebase it against `support/5.x`, @RatserX?

I just rebased against `support/5.x`. I checked that PR but it seems that we're already handling int64 numbers, so we shouldn't run into overflow exceptions in that case. Do we still need to do a TryParse?

asbjornu (approved):
True, #3028 did most of that but you've found a missed or new occurrence, which is great. I still think there's value in consistently having a `TryParse()` pattern, but as it's going to fail a lot less with `long` than `int`, I think we can merge this and leave #2703 open as a avenue for future improvement.

RatserX:
Is there any ETA as to when is a new 5.x version is going to be released? This change would actually help us unblock a lot of our projects.

PR #3314, opened 2022-12-19
Implement #1054 - renamed tag => label

### LINKED ISSUE #1054: Rename tag to label

arturcic:
Closes #1054 

Renamed 
tag-pre-release-weight => label-pre-release-weight, 
tag-prefix => label-prefix, 
continuous-delivery-fallback-tag => continuous-delivery-fallback-label, 
tag-number-pattern => label-number-pattern, 
tag => label


asbjornu (approved):
Awesome! 🎉 ❤️ 

Can you please note the change in `BREAKING_CHANGES.md` as well? 🙏🏼

arturcic:
Sure, I will

ni-jsuchocki:
Hello! Maybe it's too late but since 6.0.0 is still in beta, I figured I will ask - I understand that this change was done to avoid confusing "tag" with "git tag", right? In this case, shouldn't "tag-pre-release-weight" and "tag-prefix" stay the way they were?

asbjornu:
I believe you are quite right there, @ni-jsuchocki. The rename may have been a bit too greedy. Would you like to create an issue for it, or perhaps even fix it in a PR?

HHobeck:
Actually I was also struggling sometimes when to use tag or when to use label. Do you change it back completely to `tag`?

If you consider the following sentences:

- On the release branch I have defined the label name to beta. That means the next version is 6.0.0-beta.2.
- On the develop branch I have defined an empty label name. That means the next version is 6.0.0-2.
- On the main branch I have defined no label name. That means the next version is 6.0.0
- The commit has been tagged with the value 6.0.0. The next version on release branch is 6.1.0-beta.1. The label name is still beta and the label is beta.1 and the tag for this version might be 6.1.0-beta.1.

What I'm tring to say is that in GitVersion `tag` and `label` is not the same and we should make a distinguish the terms.

HHobeck:
{label} = {label-name}.{label-number/pre-release-number}
{semantic-version} = {major}.{minor}.{patch}-{label}+{meta-data}
{tag} = v{major}.{minor}

asbjornu:
> Actually I was also struggling sometimes when to use tag or when to use label. 

@HHobeck, it may become clearer if you read #1054. `tag` refers to `git tag` while `label` refers to "labels" as defined by [SemVer](https://semver.org/): 

> Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

> Do you change it back completely to `tag`?

No, we need to have both, but with a clear separation between what a `tag` and `label` is.

> What I'm trying to say is that in GitVersion `tag` and `label` is not the same and we should distinguish between the terms.

Yes, absolutely. 👍🏼

ni-jsuchocki:
> I believe you are quite right there, @ni-jsuchocki. The rename may have been a bit too greedy. Would you like to create an issue for it, or perhaps even fix it in a PR?

Yes, I will try and provide a PR!

arturcic:
@ni-jsuchocki I've created an issue you can reference https://github.com/GitTools/GitVersion/issues/3480

HHobeck:
I have taken a look and think it's not a good idea to rename label-pre-release-weight to tag-pre-release-weight.

We have already a branch related property with name pre-release-weight. Why not using just this property and remove label-pre-release-weight?

Anyway I think the business logic can be simplified:
image

ni-jsuchocki:
The documentation says:
```
The pre-release weight in case of tagged commits.
```

So I think having `tag` in the name is better than `label`. Can we continue this discussion in a PR I opened? https://github.com/GitTools/GitVersion/pull/3481 I think changing the logic to merge those two properties is out of scope for this change

PR #3335, opened 2023-01-09
#3334 - re-target to .net6.0 and .net7.0 in place of netstandard2.0

### LINKED ISSUE #3334: [Improvement] Target .net6.0 and .net7.0  instead of netstandard2.0
See https://github.com/GitTools/GitVersion/pull/3244#issuecomment-1374425045 for context

arturcic:
Fixes #3334


src/GitVersion.Core/Extensions/ServiceCollectionExtensions.cs
```diff
@@ -10,6 +10,6 @@ public static IServiceCollection AddModule(this IServiceCollection serviceCollec
         return serviceCollection;
     }
 
-    public static TService GetServiceForType<TService, TType>(this IServiceProvider serviceProvider) =>
+    public static TService? GetServiceForType<TService, TType>(this IServiceProvider serviceProvider) =>
         serviceProvider.GetServices<TService>().SingleOrDefault(t => t?.GetType() == typeof(TType));
```
HHobeck, line 14: May I ask you to change the implementation from SingleOrDefault to just Single and returning always a non-null type? it makes it easier to use IMO.

src/GitVersion.Core.Tests/VersionCalculation/Strategies/VersionInBranchNameBaseVersionStrategyTests.cs
```diff
@@ -90,12 +97,15 @@ public void CanTakeVersionFromNameOfRemoteReleaseBranch(string branchName, strin
...
+
+        strategy.ShouldNotBeNull();
+
         var baseVersion = strategy.GetBaseVersions(new(gitRepository.FindBranch(branchName)!, effectiveConfiguration)).Single();
 
         baseVersion.SemanticVersion.ToString().ShouldBe(expectedBaseVersion);
     }
 
-    private static IVersionStrategy GetVersionStrategy(string workingDirectory, IGitRepository repository, string branch, GitVersionConfiguration? configuration = null)
+    private static IVersionStrategy? GetVersionStrategy(string workingDirectory, IGitRepository repository, string branch, GitVersionConfiguration? configuration = null)
```
HHobeck, line 108: It think we should change the implementation in ServiceCollectionExtensions class and ensure that a non-null instance will be returned.

HHobeck (changes requested):

src/GitVersion.Core/Extensions/ServiceCollectionExtensions.cs
asbjornu: I agree that we should try to return a non-nullable here, but I would prefer doing the `null` check ourselves as that makes it possible for us to throw a more meaningful exception than `Single()` would do if no `TType` is found.

arturcic:
@HHobeck can you tell us how we can reach you by twitter/ email?

PR #3398, opened 2023-02-24
Move output (VersionConverters) to its own module

arturcic:


src/GitVersion.Core/Output/IVersionConverter.cs
```diff
@@ -1,6 +1,6 @@
 using GitVersion.OutputVariables;
 
-namespace GitVersion.VersionConverters;
+namespace GitVersion.Output;
 
 public interface IVersionConverter<in T> : IDisposable where T : IConverterContext
```
asbjornu, line 5: Shouldn't the interface remain in `Core`?

arturcic: Yeah, could be, for now focused on moving out of GitVersion.Core, but we will need to review the namespaces in the Core

asbjornu (approved):

PR #3399, opened 2023-02-26
Generate json schemas for the configuration and output files

### LINKED ISSUE #2874: [Feature Request] Add JSON schema to help configure 'GitVersion.yml'
## Detailed Description
GitVersion only has a JSON schema for the *output* JSON, but it lacks one for the *input* YAML/JSON config.

## Context
I've found that going back and forth between the documentation and the config file I'm editing takes up more time than it could if my editor was able to detect a schema and use that to suggest properties and values. It wouldn't be perfect, but nothing is.

arturcic:
Resolves #2874


arturcic:
And example on how the schema is used in rider:

image
image

Currently it's done with a local schema, but when we publish the next stable version it will be also published on [gitversion.net](https:/gitversion.net) and then we can add to the  https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json and then every IDE will be able to use it.

Same for the output format

Issue #3306, opened 2022-12-09
[Improvement] Move serialization and deserialization process of configuration out of the scope of core domain


HHobeck:
**Is your improvement request related to a problem? Please describe.**
In the discussion of pull-request https://github.com/GitTools/GitVersion/pull/3235 (Create new fallback and unknown section in GitVersionConfiguration) we came to the conclusion that it might be a good idea to move the serialization and deserialization process out of the scope of the core domain (see https://github.com/GitTools/GitVersion/pull/3235#discussion_r1030138326).

## Detailed Description
The core domain should have only the business logic without the dependency to the Yaml library.

## Possible Implementation
Move the logic in GitVersionContextFactory from core to another place and only provide the final GitVersionConfiguration instance to the core domain. The idea is to either make the GitVersionConfiguration immutable or provide the entity via interface or abstract class.

arturcic:
I was thinking we need to move `Configuration` related code into a separate module similarly I did for BuildAgents and Output, I just wonder what should we keep in your opinion in the `GitVersion.Core`?

The idea is to keep in the `GitVersion.Core`  mainly the interfaces, and the classes related to version calculation.

What do you think?

HHobeck:
I'm not sure what exactly the target architecture might be: Maybe onion architecture or 3-layer architecture with dedicated business entities. Nevermind independent of that the plan is to have some where a place (GitVersion.Core?) with business logic of GitVersion with no dependencies to other services or infrastructure like de-/serialization components. Thus this classes and intefaces which giving the additional value to GitVersion (without dependencies) needs to be present in GitVersion.Core.

HHobeck:
Maybe we need to create a concept and write a component diagram to illustrate the dependencies first? @asbjornu: What do you think?

arturcic:
@HHobeck @asbjornu - please check this one https://github.com/GitTools/GitVersion/discussions/3407

PR #3433, opened 2023-03-10
Change GitVersionConfiguration and BranchConfiguration to record type and extract interfaces

HHobeck:
Change GitVersionConfiguration and BranchConfiguration to record type and extract interfaces

## Description
Preparation for https://github.com/GitTools/GitVersion/issues/3306

## Checklist:

### DIFF
--- a/src/GitVersion.Core/Configuration/BranchConfiguration.cs
+++ b/src/GitVersion.Core/Configuration/BranchConfiguration.cs
@@ -1,127 +1,101 @@
 using GitVersion.Attributes;
+using GitVersion.Extensions;
 using GitVersion.VersionCalculation;
 
 namespace GitVersion.Configuration;
 
-public class BranchConfiguration
+public record BranchConfiguration : IBranchConfiguration
 {
-    public BranchConfiguration()
-    {
-    }
-
-    public BranchConfiguration(BranchConfiguration branchConfiguration)
-    {
-        VersioningMode = branchConfiguration.VersioningMode;
-        Label = branchConfiguration.Label;
-        Increment = branchConfiguration.Increment;
-        PreventIncrementOfMergedBranchVersion = branchConfiguration.PreventIncrementOfMergedBranchVersion;
-        LabelNumberPattern = branchConfiguration.LabelNumberPattern;
-        TrackMergeTarget = branchConfiguration.TrackMergeTarget;
-        TrackMergeMessage = branchConfiguration.TrackMergeMessage;
-        CommitMessageIncrementing = branchConfiguration.CommitMessageIncrementing;
-        TracksReleaseBranches = branchConfiguration.TracksReleaseBranches;
-        Regex = branchConfiguration.Regex;
-        IsReleaseBranch = branchConfiguration.IsReleaseBranch;
-        IsMainline = branchConfiguration.IsMainline;
-        Name = branchConfiguration.Name;
-        SourceBranches = branchConfiguration.SourceBranches;
-        IsSourceBranchFor = branchConfiguration.IsSourceBranchFor;
-        PreReleaseWeight = branchConfiguration.PreReleaseWeight;
-    }
-
     [JsonPropertyName("mode")]
     [JsonPropertyDescription("The versioning mode for this branch. Can be 'ContinuousDelivery', 'ContinuousDeployment', 'Mainline'.")]
-    public VersioningMode? VersioningMode { get; set; }
+    public VersioningMode? VersioningMode { get; internal set; }
 
-    /// <summary>
-    /// Special value 'useBranchName' will extract the tag from the branch name
-    /// </summary>
     [JsonPropertyName("label")]
     [JsonPropertyDescription("The label to use for this branch. Can be 'useBranchName' to extract the label from the branch name. Use the value {BranchName} as a placeholder to insert the branch name.")]
-    public string? Label { get; set; }
+    public string? Label { get; internal set; }
 
     [JsonPropertyName("increment")]
     [JsonPropertyDescription("The increment strategy for this branch. Can be 'Inherit', 'Patch', 'Minor', 'Major', 'None'.")]
-    public IncrementStrategy Increment { get; set; }
+    public IncrementStrategy Increment { get; internal set; }
 
     [JsonPropertyName("prevent-increment-of-merged-branch-version")]
     [JsonPropertyDescription("Prevent increment of merged branch version.")]
-    public bool? PreventIncrementOfMergedBranchVersion { get; set; }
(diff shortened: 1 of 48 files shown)


src/GitVersion.Core/Core/RepositoryStore.cs
```diff
@@ -104,8 +104,9 @@ public IBranch GetTargetBranch(string? targetBranchName)
 
     public IBranch? FindMainBranch(GitVersionConfiguration configuration)
     {
+        var branches = configuration.Branches;
         var mainBranchRegex = configuration.Branches[ConfigurationConstants.MainBranchKey].Regex
```
arturcic, line 108:
```suggestion
        var mainBranchRegex = branches[ConfigurationConstants.MainBranchKey].Regex
```

src/GitVersion.Core/Configuration/BranchConfiguration.cs
```diff
@@ -1,127 +1,101 @@
...
     [JsonPropertyName("commit-message-incrementing")]
     [JsonPropertyDescription("Sets whether it should be possible to increment the version with special syntax in the commit message. Can be 'Disabled', 'Enabled' or 'MergeMessageOnly'.")]
-    public CommitMessageIncrementMode? CommitMessageIncrementing { get; set; }
+    public CommitMessageIncrementMode? CommitMessageIncrementing { get; internal set; }
 
     [JsonPropertyName("regex")]
-    [JsonPropertyDescription("The regex pattern to use to match this branch.")]
-    public string? Regex { get; set; }
+    [JsonPropertyDescription("The regular expression pattern to use to match this branch.")]
+    public string? Regex { get; internal set; }
```
arturcic, line 44:
```suggestion
    public string? RegularExpression { get; internal set; }
```

src/GitVersion.Core/Configuration/GitVersionConfiguration.cs
```diff
@@ -4,103 +4,107 @@
...
-    public bool UpdateBuildNumber { get; set; } = true;
+    public bool UpdateBuildNumber { get; internal set; } = true;
 
     [JsonPropertyName("semantic-version-format")]
     [JsonPropertyDescription("Specifies the semantic version format that is used when parsing the string. Can be 'Strict' or 'Loose'.")]
-    public SemanticVersionFormat SemanticVersionFormat { get; set; }
+    public SemanticVersionFormat SemanticVersionFormat { get; internal set; }
+
+    [JsonIgnore]
+    IReadOnlyDictionary<string, BranchConfiguration> IGitVersionConfiguration.Branches => Branches;
```
arturcic, line 94: I wonder if we can use the newly interface here
```suggestion
    IReadOnlyDictionary<string, IBranchConfiguration> IGitVersionConfiguration.Branches => Branches;
```
and the next line as well?

arturcic:
@HHobeck  will this follow up with a PR that will start using the new interfaces in the core instead of the classes? Or we can include that in the current PR?

arturcic:
Please also run from the root 
```pwsh
dotnet run --project .\src\GitVersion.Schema --framework net7.0 -- --outputdirectory schemas --version 6.0
```
to generate the updated schema as you've changed the descriptions

src/GitVersion.Core/Configuration/GitVersionConfiguration.cs
HHobeck: Only the first line because the secon line is used for initialiization.

src/GitVersion.Core/Configuration/BranchConfiguration.cs
HHobeck: Done

src/GitVersion.Core/Core/RepositoryStore.cs
HHobeck: Done

HHobeck:
> @HHobeck will this follow up with a PR that will start using the new interfaces in the core instead of the classes? Or we can include that in the current PR?

I think it might be better to seperate it and create another PR.

src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
```diff
@@ -0,0 +1,15 @@
...
+
+namespace GitVersion.Configuration
+{
+    public interface IIgnoreConfiguration
+    {
+        DateTimeOffset? Before { get; }
+
+        IReadOnlyList<string> Shas { get; }
+
+        IEnumerable<IVersionFilter> ToFilters();
```
arturcic, line 11: Question, can we move this method out of the interface/record, to an extension maybe or private method at the place of usage, Currently there are only 2 usages.

arturcic (approved):

src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
arturcic: Could be done in the other PR

arturcic:
TODOs for the next PR:
- make `IgnoreConfiguration`, `BranchConfiguration` and `GitVersionConfiguration` records internal
- make usage of the respective interfaces in the code (maybe with the exception of unit tests)
- consider making the properties init instead of private set for the above records, and maybe readonly were possible

Please add more to this list what you have in mind, and then you can use this list as description for the new PR

arturcic:
LGTM! It moves in the direction that will allow us to move the implementations details of Configuration into its own module

src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
HHobeck: Yep good idea. :) done

HHobeck:
> TODOs for the next PR:
> 
> * make `IgnoreConfiguration`, `BranchConfiguration` and `GitVersionConfiguration` records internal
> * make usage of the respective interfaces in the code (maybe with the exception of unit tests)
> * consider making the properties init instead of private set for the above records, and maybe readonly were possible
> * consider moving the [JsonProperty*] attributes from the record to the interface and adapt the schema generation to use the interfaces instead

Beside the fact that it might be a lot of work and tricky doing it I think it is not feasible (last point) moving the JsonProperty to the interface. The reason is that we are using mutable data types in the interface and YamlDotnet cannot handle it.

> LGTM! It moves in the direction that will allow us to move the implementations details of Configuration into its own module

Yes that's the goal to seperate the aspect of loading/merging the configuration from using it. I'm done please merge to main.

arturcic:
> Beside the fact that it might be a lot of work and tricky doing it I think it is not feasible (last point) moving the JsonProperty to the interface. The reason is that we are using interfaces and mutable data types in the interface and YamlDotnet cannot handle it.

That's the reason I put `consider`. I will have a look if it's feasible

PR #3447, opened 2023-03-22
Implement when PreleaseLabel is empty, the PreleaseTag is generated correctly

### LINKED ISSUE #2347: [Bug] When PreleaseLabel is empty, the PreleaseTag is not correctly generated
**Describe the bug**

In **ContinousDeployment** or **Mainline mode** I would like to use the **CommitsSinceVersionSource** as **PreleaseTag** without any additional **PrereleaseLabel** before it.
In order to achieve that I tried to set the **PrereleaseTag** to **empty** but in such case the **CommitsSinceVersionSource** is not properly processed to the **Prerelease** properties.

After looking through the code it seems that such behavior might be intended for **Mainline mode** as there are a lot of tests where by default for **FullSemVer** is expected format **MajorMinorPatch** instead of **MajorMinorPatch-PrereleaseTag** (PrereleaseLabel.PrereleaseNumber).

_____

Is this expected behavior? When there is no PrereleaseLabel, there won't be PrereleaseTag generated from just Number?

Shouldn't **CommitsSinceVersionSource** taken as **PrereleaseNumber** generate **PrereleaseTag** at least for **ContinuousDeployment** mode, even if PrereleaseLabel is empty? 
As it is mentioned in source code comment (VariableProvider.cs, Line 29):

> // Continuous Deployment always requires a pre-release tag unless the commit is tagged
____

Example of the generated properties for ContinuousDeployment:
image

There is only **PrereleaseTag** set to the value of **CommitsSinceVersionSource**, without any formatting. 
Even the **PrereleaseNumber** is not properly returned.
The **FullSemVer** include only MajorMinorPatch.

## Expected Behavior
With GitVersion.yml:
```
continuous-delivery-fallback-tag: ''
branches:
  master:
    mode: ContinuousDeployment
```

The Prelease properties generated should look like this:
```
"PreReleaseTag": "3",
"PreReleaseTagWithDash":"-3",
"PreReleaseLabel":"",
"PreReleaseNumber":"3",
"FullSemVer":"5.3.8-3",
```
In relation to Properties shown below:
image

## Actual Behavior
With GitVersion.yml:
```
continuous-delivery-fallback-tag: ''
branches:
  master:
    mode: ContinuousDeployment
```
The output generated looks like this:
image

## Possible reason
The **CommitsSinceVersionSource** are set as **Number** for **PrereleaseTag** properly but in turn the **PromotedFromCommits** flag is set.  
Then the Properties are not properly handled by **SemanticVersionFormatValues** due to **HasTag** method in **SemanticVersionPreReleaseTag** class.
The **HasTag** returns false due to **PromotedFromCommits** flag being true and **Name** being null/empty, in result **SemanticVersionFormatValues** doesn't return proper values.

## Steps to Reproduce
Use GitVersion.yml config file:
```
continuous-delivery-fallback-tag: ''
branches:
  master:
    mode: ContinuousDeployment
```
Make several commits, and then check the generated output of GitVersion.
The Prerelease properties should not be generated properly.
The FullSemVer doesn't include the PrereleaseTag as well.

## Context
In ContinousDeployment or Mainline mode I would like to use the **CommitsSinceVersionSource** as **PreleaseTag** without any additional **PrereleaseLabel** before it. The number of commits is enough without any additional string. 

## Your Environment

- Version Used: 5.3.7+Branch.master.Sha.b22e8726683a0997484b53c5efd54394e93b5306
- Operating System and version (Windows 10, Ubuntu 18.04): Windows 10

HHobeck:
Fix [Bug] When PreleaseLabel is empty, the PreleaseTag is not correctly generated  https://github.com/GitTools/GitVersion/issues/2347

Close #2347

## Checklist:


HHobeck:
I'm done please review and merge to main.

arturcic:
Is there any reason to have 4 new interfaces `IContinuousDeliveryVersionCalculator`, `IContinuousDeploymentVersionCalculator`, `IManualDeploymentVersionCalculator`, and `ITrunkBasedVersionCalculator` that have the same method with the same signature and to not share a `IVersionCalculator`? 

If we take in consideration your suggestion regarding 

```csharp
public enum VersioningMode
{
    ManuallyDeployment, // previously ContinuousDelivery
    ContinuousDelivery, // previously ContinuousDeployment
    ContinuousDeployment, // new
    Trunkbased // previously Mainline??
}
``` 
I would rather have a  `IVersionCalculator` that has the `VersioningMode` field and each implementation of this interfaces will set the appropriate `VersioningMode` andd then in the `NextVersionCalculator` I would inject all implementations and depending on the `VersioningMode` it will be selected the appropriate one. Thoughts?

HHobeck:
> Is there any reason to have 4 new interfaces `IContinuousDeliveryVersionCalculator`, `IContinuousDeploymentVersionCalculator`, `IManualDeploymentVersionCalculator`, and `ITrunkBasedVersionCalculator` that have the same method with the same signature and to not share a `IVersionCalculator`?
> 
> If we take in consideration your suggestion regarding
> 
> ```cs
> public enum VersioningMode
> {
>     ManuallyDeployment, // previously ContinuousDelivery
>     ContinuousDelivery, // previously ContinuousDeployment
>     ContinuousDeployment, // new
>     Trunkbased // previously Mainline??
> }
> ```
> 
> I would rather have a `IVersionCalculator` that has the `VersioningMode` field and each implementation of this interfaces will set the appropriate `VersioningMode` andd then in the `NextVersionCalculator` I would inject all implementations and depending on the `VersioningMode` it will be selected the appropriate one. Thoughts?

Yep good point. At this moment I have separated the behavior which not aligns with the `VersioningMode` enumerator. But if we are going to decide to change it so we have a one to one relation then yes I think your idea is great.

Do you have something like this in mind?

```csharp
public interface IVersionCalculator
{
    VersioningMode VersioningMode { get; }

    SemanticVersion Calculate(NextVersion nextVersion);
}

public IVersionCalculator GetSpecificVersionCalculator(VersioningMode versioningMode)
{
    return _serviceProvider.GetServices<IVersionCalculator>().Single(
        element => element.VersioningMode == versioningMode
    );
}
```

arturcic:
> > Is there any reason to have 4 new interfaces `IContinuousDeliveryVersionCalculator`, `IContinuousDeploymentVersionCalculator`, `IManualDeploymentVersionCalculator`, and `ITrunkBasedVersionCalculator` that have the same method with the same signature and to not share a `IVersionCalculator`?
> > If we take in consideration your suggestion regarding
> > ```cs
> > public enum VersioningMode
> > {
> >     ManuallyDeployment, // previously ContinuousDelivery
> >     ContinuousDelivery, // previously ContinuousDeployment
> >     ContinuousDeployment, // new
> >     Trunkbased // previously Mainline??
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I would rather have a `IVersionCalculator` that has the `VersioningMode` field and each implementation of this interfaces will set the appropriate `VersioningMode` andd then in the `NextVersionCalculator` I would inject all implementations and depending on the `VersioningMode` it will be selected the appropriate one. Thoughts?
> 
> Yep good point. At this moment I have separated the behavior which not aligns with the `VersioningMode` enumerator. But if we are going to decide to change it so we have a one to one relation then yes I think your idea is great.
> 
> Do you have something like this in mind?
> 
> ```cs
> public interface IVersionCalculator
> {
>     VersioningMode VersioningMode { get; }
> 
>     SemanticVersion Calculate(NextVersion nextVersion);
> }
> 
> public IVersionCalculator GetSpecificVersionCalculator(VersioningMode versioningMode)
> {
>     return _serviceProvider.GetServices<IVersionCalculator>().Single(
>         element => element.VersioningMode == versioningMode
>     );
> }
> ```

Exactly, in this case we don't need that many interfaces for the all the modes, and we have only the implementations and one interface

asbjornu:
> Exactly, in this case we don't need that many interfaces for the all the modes, and we have only the implementations and one interface

Agreed.

HHobeck:
> > Exactly, in this case we don't need that many interfaces for the all the modes, and we have only the implementations and one interface
> 
> So what do you think @HHobeck should we make those changes to the VersionMode, or put that change into another PR?

I think it is better to do this in a separate PR. You mentioned already to do this after the beta.2 release. Which is fine with me.

HHobeck:
Also we need to discuss the naming. IMO the naming of VersioningMode is missleading. Maybe a better name might be DeploymentMode. What do you think? In addition what do you like more Trunkbased or Mainline?

```csharp
public enum DeploymentMode
{
    ManuallyDeployment, // previously ContinuousDelivery
    ContinuousDelivery, // previously ContinuousDeployment
    ContinuousDeployment, // new
    Trunkbased // previously Mainline??
}
```

arturcic:
@asbjornu do you agree with this PR to be merged?

arturcic (approved):

src/GitVersion.Core/VersionCalculation/Abstractions/IContinuousDeliveryVersionCalculator.cs
```diff
@@ -0,0 +1,6 @@
+namespace GitVersion.VersionCalculation;
+
+public interface IContinuousDeliveryVersionCalculator
```
asbjornu, line 3: Didn't we agree to reduce these interfaces down to just one, as they all have the same method signature `SematicVersion Calculate(NextVersion nextVersion)`?

src/GitVersion.Core/VersionCalculation/ContinuousDeliveryVersionCalculator.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+    }
+
+    public SemanticVersion Calculate(NextVersion nextVersion)
+    {
+        using (this.log.IndentLog("Using continues delivery typology to calculate the incremented version!!"))
+        {
+            var preReleaseTag = nextVersion.IncrementedVersion.PreReleaseTag;
+            if (!preReleaseTag.HasTag() || !preReleaseTag.Number.HasValue)
+                throw new WarningException("--PRE--CONDITION--FAILED--");
```
asbjornu, line 20: Why the strange warning message `--PRE--CONDITION--FAILED—`? I like that exceptions convey a user-friendly, descriptive error message. Can we say something like this instead?

```suggestion
                throw new WarningException("No pre-release tag found");
```

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
```diff
@@ -0,0 +1,61 @@
...
+        : base(log, repositoryStore, versionContext)
+    {
+    }
+
+    public SemanticVersion Calculate(NextVersion nextVersion)
+    {
+        using (this.log.IndentLog("Using continues deployment typology to calculate the incremented version!!"))
+        {
+            if (!nextVersion.Configuration.IsMainline || nextVersion.Configuration.Label is not null)
+                throw new WarningException("--PRE--CONDITION--FAILED--");
```
asbjornu, line 19: I'm not entirely sure what the warning message should be here, but please find something more user friendly.

```suggestion
                throw new WarningException("...");
```

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
```diff
@@ -0,0 +1,61 @@
...
+internal sealed class ContinuousDeploymentVersionCalculator : NonTrunkBasedVersionCalculatorBase, IContinuousDeploymentVersionCalculator
+{
+    public ContinuousDeploymentVersionCalculator(ILog log, IRepositoryStore repositoryStore, Lazy<GitVersionContext> versionContext)
+        : base(log, repositoryStore, versionContext)
+    {
+    }
+
+    public SemanticVersion Calculate(NextVersion nextVersion)
+    {
+        using (this.log.IndentLog("Using continues deployment typology to calculate the incremented version!!"))
```
asbjornu, line 16:
```suggestion
        using (this.log.IndentLog("Using continuous deployment workflow to calculate the incremented version."))
```

src/GitVersion.Core/VersionCalculation/ManualDeploymentVersionCalculator.cs
```diff
@@ -0,0 +1,39 @@
...
+    {
+    }
+
+    public SemanticVersion Calculate(NextVersion nextVersion)
+    {
+        using (this.log.IndentLog("Using manual deployment typology to calculate the incremented version!!"))
+        {
+            var preReleaseTag = nextVersion.IncrementedVersion.PreReleaseTag;
+            if (!preReleaseTag.HasTag() || !preReleaseTag.Number.HasValue)
+                throw new WarningException("--PRE--CONDITION--FAILED--");
```
asbjornu, line 19:
```suggestion
                throw new WarningException("...");
```

src/GitVersion.Core/VersionCalculation/ManualDeploymentVersionCalculator.cs
```diff
@@ -0,0 +1,39 @@
...
+internal sealed class ManualDeploymentVersionCalculator : NonTrunkBasedVersionCalculatorBase, IManualDeploymentVersionCalculator
+{
+    public ManualDeploymentVersionCalculator(ILog log, IRepositoryStore repositoryStore, Lazy<GitVersionContext> versionContext)
+        : base(log, repositoryStore, versionContext)
+    {
+    }
+
+    public SemanticVersion Calculate(NextVersion nextVersion)
+    {
+        using (this.log.IndentLog("Using manual deployment typology to calculate the incremented version!!"))
```
asbjornu, line 15:
```suggestion
        using (this.log.IndentLog("Using manual deployment workflow to calculate the incremented version."))
```

src/GitVersion.Core/VersionCalculation/NonTrunkBasedVersionCalculatorBase.cs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    protected bool ShouldTakeIncrementedVersion(NextVersion nextVersion)
+    {
+        var label = nextVersion.Configuration.GetBranchSpecificLabel(
+            Context.CurrentBranch.Name, nextVersion.BaseVersion.BranchNameOverride
+        );
+
+        ////
+        // TODO: We need to decide whether or not to calculate the upcoming semantic version or the previous one because of tagging. Actually this should be part of the branch configuration system.
```
asbjornu, line 30: Agreed. Can you please add this as an issue we can follow up on?

src/GitVersion.Core/VersionCalculation/NonTrunkBasedVersionCalculatorBase.cs
```diff
@@ -0,0 +1,93 @@
...
+        return Context.CurrentCommit?.Sha != nextVersion.BaseVersion.BaseVersionSource?.Sha
+            || Context.CurrentCommitTaggedVersion == null
+            || !Context.CurrentCommitTaggedVersion.IsMatchForBranchSpecificLabel(label);
+        //
+    }
+
+    protected SemanticVersion CalculateIncrementedVersion(NextVersion nextVersion)
+    {
+        ////
+        // TODO: We need to consider somehow the IGitVersionConfiguration::Ignore property here!!
```
asbjornu, line 40: Agreed. Can you please add this as an issue we can follow up on?

src/GitVersion.Core/VersionCalculation/TrunkBasedVersionCalculator.cs
```diff
@@ -0,0 +1,7 @@
+namespace GitVersion.VersionCalculation;
+
+internal sealed class TrunkBasedVersionCalculator : ITrunkBasedVersionCalculator
+{
+    // TODO: Please implement trunk based version here and remove MainlineVersionCalculator.
+    public SemanticVersion Calculate(NextVersion nextVersion) => throw new NotImplementedException();
```
asbjornu, line 6: In which circumstance may this exception be thrown? Is it possible to exercise the code in a way that this exception is invoked? Either way I would like the exception to have a user-friendly message.

```suggestion
    public SemanticVersion Calculate(NextVersion nextVersion) =>
        throw new NotImplementedException("Trunk based version calculation is not yet implemented. Use Mainline mode instead.");
```

asbjornu (changes requested):
This is great stuff. The only thing I'm a bit wary about is the change from `1.0.1+46` to `1.0.1-46` unless `label` is set to `null`. Not because the change doesn't make sense, I think it's a good default, but because it's a breaking change. However, I think we should test this change out in the community and see what feedback we get.

src/GitVersion.Core/VersionCalculation/Abstractions/IContinuousDeliveryVersionCalculator.cs
arturcic: Well we discussed about that, but agreed to do the change after beta2, as it will require changing the `VersioningMode`

asbjornu: Can you please explain why merging these interfaces will require changing `VersioningMode`?

arturcic: Please check this changes

https://github.com/GitTools/GitVersion/pull/3447/files#diff-d1bb8310a1a6168b70d713420c6ecc455c3bd3311b4735c3811fb8e249db8da4R61-R69

We currently have 3 `VersionMode`s but we have 4 interfaces, and @HHobeck proposed to change the `VersionMode` to 

```csharp
public enum VersioningMode
{
    ManuallyDeployment, // previously ContinuousDelivery
    ContinuousDelivery, // previously ContinuousDeployment
    ContinuousDeployment, // new
    Trunkbased // previously Mainline??
}
```

Or 

```csharp
public enum DeploymentMode
{
    ManuallyDeployment, // previously ContinuousDelivery
    ContinuousDelivery, // previously ContinuousDeployment
    ContinuousDeployment, // new
    Trunkbased // previously Mainline??
}
```

Then we have 4 modes and 4 interfaces. In that case if we define a property in a single interface like

```csharp
public interface IVersionCalculator
{
    VersioningMode VersioningMode { get; }

    SemanticVersion Calculate(NextVersion nextVersion);
}
```

each implementation of this interface will set the mode accordingly.

src/GitVersion.Core/VersionCalculation/TrunkBasedVersionCalculator.cs
HHobeck: This code will never be called but it is a reminder to go ahead with the development. I have placed a TODO description as well.

src/GitVersion.Core/VersionCalculation/ManualDeploymentVersionCalculator.cs
HHobeck: Done

src/GitVersion.Core/VersionCalculation/ManualDeploymentVersionCalculator.cs
HHobeck: Done

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
HHobeck: Done

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
HHobeck: Done

src/GitVersion.Core/VersionCalculation/ContinuousDeliveryVersionCalculator.cs
HHobeck: Done

BREAKING_CHANGES.md
```diff
@@ -27,6 +27,7 @@
...
 *   A new configuration property with name `version-in-branch-pattern` has been introduced. This setting only applies on branches where the option `is-release-branch` is set to `true`. Please notice that the branch name needs to be defined after the version number by default (instead of `support/lts-2.0.0` please name the branch like `support/2.0.0-lts`).
 *   The `is-release-branch` property of the `hotfix` branch setting has been changed from `false` to `true`. If present the hotfix number will be considered now by default.
+*   In the git hub and the git flow workflow the `label` property is by default set to an empty string on the `main` branch. This yields to a pre-release version on `main` with an empty tag. Instead of for instance `1.0.1+46` git-version generates the full semantic version `1.0.1-46` instead. This behavior can be changed to generate only main versions (no pre-release version) with setting the label to `null` (Please keep in mind that the `label` property on root needs to be set to `null` as well otherwise the fallback applies). This change is caused by issue #2347.
```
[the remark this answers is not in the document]
HHobeck: Done

src/GitVersion.Core/VersionCalculation/NonTrunkBasedVersionCalculatorBase.cs
HHobeck: I think the issue is already existent: https://github.com/GitTools/GitVersion/issues/3453

src/GitVersion.Core/VersionCalculation/NonTrunkBasedVersionCalculatorBase.cs
HHobeck: I have created an issue for that: https://github.com/GitTools/GitVersion/issues/3461

src/GitVersion.Core/VersionCalculation/TrunkBasedVersionCalculator.cs
HHobeck: Done

src/GitVersion.Core/VersionCalculation/ContinuousDeliveryVersionCalculator.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    public SemanticVersion Calculate(NextVersion nextVersion)
+    {
+        using (this.log.IndentLog("Using continuous delivery workflow to calculate the incremented version."))
+        {
+            var preReleaseTag = nextVersion.IncrementedVersion.PreReleaseTag;
+            if (!preReleaseTag.HasTag() || !preReleaseTag.Number.HasValue)
+            {
+                throw new WarningException("Continues delivery deployment requires a pre-release tag.");
```
asbjornu, line 21:
```suggestion
                throw new WarningException("Continuous delivery requires a pre-release tag.");
```

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+    }
+
+    public SemanticVersion Calculate(NextVersion nextVersion)
+    {
+        using (this.log.IndentLog("Using continuous deployment workflow to calculate the incremented version."))
+        {
+            if (nextVersion.Configuration.Label is not null)
+            {
+                throw new WarningException("Continues deployment requires no pre-release tag.");
```
asbjornu, line 20:
```suggestion
                throw new WarningException("Continuous deployment requires no pre-release tag.");
```

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+            if (!nextVersion.Configuration.IsMainline)
+            {
+                throw new WarningException("Continues deployment only supported for mainline branches.");
```
asbjornu, line 24:
```suggestion
                throw new WarningException("Continuous deployment is only supported for mainline branches.");
```

asbjornu (changes requested):
Just a couple of typos, then I think this is good.

src/GitVersion.Core/VersionCalculation/Abstractions/IContinuousDeliveryVersionCalculator.cs
asbjornu: I still don't understand why we can't have one interface with 4 different implementations? I would also prefer discussing renaming and other large changes in separate issues. If we are to keep a `mode` for what is basically mainline, I think we can just keep the name `Mainline`.

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
HHobeck: done

src/GitVersion.Core/VersionCalculation/ContinuousDeliveryVersionCalculator.cs
HHobeck: done

src/GitVersion.Core/VersionCalculation/ContinuousDeploymentVersionCalculator.cs
HHobeck: done

asbjornu (approved):
Superb! 👏🏼

src/GitVersion.Core/VersionCalculation/Abstractions/IContinuousDeliveryVersionCalculator.cs
HHobeck: Mainline is confisuing IMO because we have a `IsMainline` property as a branch configuration and a `Mainline` value on the version mode. If you say the branch needs to be mainline then it's confusing. It was just a proposal from my side because if we are touching this lines of code it's easy to do it right away. But if you don't like it nevermind.

ddprince-pro:
> This is great stuff. The only thing I'm a bit wary about is the change from `1.0.1+46` to `1.0.1-46` unless `label` is set to `null`. Not because the change doesn't make sense, I think it's a good default, but because it's a breaking change. However, I think we should test this change out in the community and see what feedback we get.

@asbjornu I've been trying to make it so it doesn't output a prerelease number on my side, but setting "label" to null under GitVersion.yml doesn't work. This markdown file is also mentioning a similar solution, but still I can't figure it out: https://github.com/GitTools/GitVersion/blob/main/BREAKING_CHANGES.md. 

image
Go that when applying the change:
image
"{BranchName}" is appended inside the "FullSemVer" variable. 

When removing the label, I got that: 
image

My configuration so far: 
image

Would it be possible to help me so that I have for the variable FullSemVer, the value '1.0.0+37' instead of '1.0.0-37'? 

Thanks in advance!

PR #3451, opened 2023-03-28
Added GitVersion.Abstractions module

arturcic:
Moving most of the interfaces from Core to Abstractions, made modules depend on Abstractions instead of Core


asbjornu:
Sorry, but I don’t agree with this change. We don’t need to move interfaces out of Core, that makes it not be the core anymore.

arturcic:
Core will change to Calculation, it will contain only the version calculation logic, eventually

arturcic:
> Sorry, but I don’t agree with this change. We don’t need to move interfaces out of Core, that makes it not be the core anymore.

The `Abstractions` module has the contracts defined, that means the modules we have (output and agents for now) will depend on the contracts. Now the implementations of those interfaces are defined in the specific module.

arturcic:
@asbjornu please have a look here https://github.com/GitTools/GitVersion/discussions/3407#discussioncomment-5151131

asbjornu:
> Core will change to Calculation, it will contain only the version calculation logic, eventually

Don't you consider version calculation to be "core" to what GitVersion does?

> The `Abstractions` module has the contracts defined, that means the modules we have (output and agents for now) will depend on the contracts. Now the implementations of those interfaces are defined in the specific module.

Sounds like the separation we already had with `Core`?

arturcic:
> Don't you consider version calculation to be "core" to what GitVersion does?

Yes, and only version calculation should be in core, all the other modules - like output, agents, maybe configuration, should not belong to core in my opinion. The version calculation code should know about the Options and the EffectiveConfiguration as inputs, and those should be provided/computed by the 'external modules', same for the output - GitVersionVariables, that is used by 'external modules'

Basically the `Core` takes input and gives back the output. It should not be aware how the input/output is prepared/used.

Now if we keep those interfaces and implementations in the `Core` then it has too many responsibilities. That's why I was saying that `Core` will eventually become `Calculation`. That's why I introduced the Abstractions (can be renamed to Contracts).

arturcic:
I can revert this PR if you want to have a broader discussion on this refactoring.

arturcic:
The PR is moving to that direction, not yet there

Issue #3453, opened 2023-03-29
[Feature] Same version computed on different branches


xpicio:
**Describe the bug**
Hello, to be honest i'm not sure if it is a gitversion issue or i'm not using gitflow right. But i hope you could help me.

I created this repository https://github.com/xpicio/compute-version to test incremental version using conventional commits, the repo is based on gitflow model and i'm using git tower with the gitflow plugin.

In the actual state (after i closed an hotfix, pushed both branches develop and main, and finally added a tag) when i compute the version on develop or main branch, the version is the same: 0.2.4. This behaviour is causing an issue in CI/CD pipelines, when i need to restart a pipeline (should happen to fix a bug on the pipeline) the artifacts tagged from the develop branch will assume the same version of the artifacts from the main branch.

## Expected Behavior
| action         | branch  | commit message | sha                                      | version       |
| -------------- | ------- | -------------- | ---------------------------------------- | ------------- |
| close hotfix (random) |               |                                   |                                          |                |
|                | develop |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.3.0-alpha.9 |
|                | main    |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.2.4         |
| new tag v0.2.4 |         |                | 53720f5d878cbc9eab27fbb127615c3e85971409 |               |
|                | develop |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.x.x-alpha.x         |
|                | main    |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.2.4         |

The version on develop branch is different to main

## Actual Behavior
| action         | branch  | commit message | sha                                      | version       |
| -------------- | ------- | -------------- | ---------------------------------------- | ------------- |
| close hotfix (random) |               |                                   |                                          |                |
|                | develop |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.3.0-alpha.9 |
|                | main    |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.2.4         |
| new tag v0.2.4 |         |                | 53720f5d878cbc9eab27fbb127615c3e85971409 |               |
|                | develop |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.2.4         |
|                | main    |                | 53720f5d878cbc9eab27fbb127615c3e85971409 | 0.2.4         |

The version on develop branch is equal to main

## Steps to Reproduce
```
[...]

Issue #2394, opened 2020-08-26
[Bug] GitVersionTask: VersionSource is set to wrong tag


fdiewald:
**Describe the bug**
Using the default configuration of GitVersionTask v5.3.7 the VersionSource ends up being not the latest tagged commit on master but rather the one before that.

Image of the test repository branch history: (blue: develop, red: master, green: release branches)
![image](https://user-images.githubusercontent.com/29732058/91311348-46ba5400-e7b3-11ea-9581-7fc05e43d79c.png)

Output of `gitversion /diag`:

````
INFO [08/26/20 16:21:25:50] Dumping commit graph: 
INFO [08/26/20 16:21:25:62] *   4c6828d 2 hours ago  (HEAD -> develop, origin/develop)
|\  
| *   27f83ef 2 hours ago  (tag: 0.4.0, origin/master, master)
| |\  
| | * 925521e 2 hours ago 
| |/  
|/|   
* | ac89157 2 hours ago 
* | b6e93d6 2 hours ago 
|\| 
| *   b6df2d3 2 hours ago  (tag: 0.3.0)
| |\  
| | * 27277e7 3 hours ago 
| |/  
|/|   
* | ae5d80a 3 hours ago 
* | 73ffb7d 3 hours ago 
|\| 
| *   a3fe19c 3 hours ago  (tag: 0.2.0)
| |\  
| | * 9a40911 3 hours ago 
| |/  
|/|   
* | c074d61 4 hours ago 
* | 00cdcbe 5 hours ago 
|/  
* 18dbaa2 5 hours ago  (tag: 0.1.0)

INFO [08/26/20 16:21:25:62] Working directory: D:\Workspace\GitVersionTest
INFO [08/26/20 16:21:25:67] Project root is: D:\Workspace\GitVersionTest\
INFO [08/26/20 16:21:25:67] DotGit directory is: D:\Workspace\GitVersionTest\.git
INFO [08/26/20 16:21:25:76] Using latest commit on specified branch
INFO [08/26/20 16:21:25:79] Running against branch: develop (4c6828daf637c00e93f7e1d7b47e029a0d867bd2)
INFO [08/26/20 16:21:25:79] Begin: Calculating base versions
  INFO [08/26/20 16:21:25:87] Found commit [4c6828daf637c00e93f7e1d7b47e029a0d867bd2] matching merge message format: Default
[...]

PR #3455, opened 2023-03-31
Fix bug: VersionInBranchNameVersionStrategy only considers the release branch

### LINKED ISSUE #2693: [Bug] VersionInBranchNameVersionStrategy only considers the release branch
**Description**

In order to calculate the next version, GitVersion evaluates the version numbers produced by various [candidates](https://github.com/GitTools/GitVersion/tree/main/src/GitVersion.Core/VersionCalculation/BaseVersionCalculators) and selects the one that yields the highest version number on a give branch. [VersionInBranchNameVersionStrategy](https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/VersionInBranchNameVersionStrategy.cs) is one of the candidates and from the name of it, should only look at the current branch. For instance, `hotfix/17.1.0`. However, this strategy seems to be [biased](https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/VersionInBranchNameVersionStrategy.cs#L32) and only considers release branches. This approach probably violates the single responsibility principle and can be a source of bugs.

Additionally, another [strategy](https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/VersionCalculation/BaseVersionCalculators/TrackReleaseBranchesVersionStrategy.cs#L37) uses it and the version detection in the branch name could probably be parameterized.

## Expected Behavior
The strategy should consider the version in the current branch, e.g. `hotfix/17.1.0`

## Actual Behavior
The version in the hotfix branch name has no chance to make it to the [list](https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/VersionCalculation/BaseVersionCalculator.cs#L37) of possible candidates.

## Possible Fix
Make strategy branch agnostic.

## Context
1.We created a hotfix branch (`hotfix/1.0.1`) from the tag `v1.0.0`
2. However, the calculated version number was still `1.0.0-something` because we had set `prevent-increment-of-merged-branch-version` to `true`.
3. In this situation, we would have expected the version number from the hotfix branch name to make it to the list of candidates as well.

## Workaround
Set the  `prevent-increment-of-merged-branch-version` to `false`

## Environment
OS: Windows
GitVersion version: Latest

HHobeck:
Fix [Bug] VersionInBranchNameVersionStrategy only considers the release branch #2693

Is related to https://github.com/GitTools/GitVersion/issues/2336

Close #2693 

## Checklist:


HHobeck:
Okay I'm done please review and merge to main.

docs/input/docs/reference/configuration.md
```diff
@@ -258,9 +259,12 @@ and [tracks-release-branches](#tracks-release-branches).
 
 ### label-prefix
 
-A regex which is used to trim Git tags before processing (e.g., v1.0.0). Default
-is `[vV]`, although this is just for illustrative purposes as we do a IgnoreCase
-match and could be `v`.
+A regular expression which is used to trim Git tags before processing (e.g., v1.0.0). The default value is `[vV]`.
```
asbjornu, line 262: Please wrap all Markdown content at 80 characters.

```suggestion
A regular expression which is used to trim Git tags before processing (e.g.,
v1.0.0). The default value is `[vV]`.
```

docs/input/docs/reference/configuration.md
```diff
@@ -258,9 +259,12 @@ and [tracks-release-branches](#tracks-release-branches).
...
+
+### version-in-branch-pattern
+
+A regular expression which is used to determine the version number in the branch name or commit message (e.g., v1.0.0-LTS). 
+The default value is `(?<version>[vV]?\d+(\.\d+)?(\.\d+)?).*`.
```
asbjornu, line 267:
```suggestion
A regular expression which is used to determine the version number in the branch
name or commit message (e.g., v1.0.0-LTS). The default value is
`(?<version>[vV]?\d+(\.\d+)?(\.\d+)?).*`.
```

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
```diff
@@ -167,7 +167,7 @@ public void FeatureOnHotfixFeatureBranchDeleted()
         fixture.Checkout(hotfix451);
         fixture.MergeNoFF(featureBranch); // commit 2
         fixture.Repository.Branches.Remove(featureBranch);
-        fixture.AssertFullSemver("4.5.1-beta.2", configuration);
+        fixture.AssertFullSemver("4.5.1-beta.3", configuration);
```
asbjornu, line 170: Why did this go from `4.5.1-beta.2` to `4.5.1-beta.3`?

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
```diff
@@ -60,7 +60,7 @@ public void CanTakeVersionFromHotfixesBranch()
 
         // create hotfix branch
         Commands.Checkout(fixture.Repository, fixture.Repository.CreateBranch("hotfixes/1.1.1"));
-        fixture.AssertFullSemver("1.1.0"); // We are still on a tagged commit
+        fixture.AssertFullSemver("1.1.1+0");
```
asbjornu, line 63: Are we not still on a tagged commit here?

asbjornu (changes requested):
Nice! Does `version-in-branch-pattern` require `is-release-branch` to be `true`? If so, we should say something about that in the documentation.

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
HHobeck: Hmm very good quesiton. I think you have foun a bug. ;) The same question applies to the following scanrio:
```csharp
    [Test]
    public void __Just_A_Test__()
    {

        fixture.BranchTo("release/1.1.0");

        fixture.AssertFullSemver("1.1.0+0");
    }
```

Why it yields to 1.1.0 and not to 1.0.0?

docs/input/docs/reference/configuration.md
HHobeck: Done

docs/input/docs/reference/configuration.md
HHobeck: Done

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
HHobeck: Hmm very good question. I think it has something to do with the logic in GitVersionVariables.cs which you know will be changed in PR #2347. I have checked the source of the resulting base version and it looks okay for me:

image

image

image

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
asbjornu: I think this is one of those edge-cases that no matter which solution we pick, we are going to make some people unhappy. I'm in the "a tag should always win" camp, but I can definitely see how some people would expect `release/1.1.0` to yield `1.1.0` even though the same commit is tagged `1.0.0`. I don't think there's a right answer here, we just need to pick one way or the other, stick to it and document it well.

asbjornu:
> Nice! Does `version-in-branch-pattern` require `is-release-branch` to be `true`? If so, we should say something about that in the documentation.

Can you please answer this, @HHobeck? ☝🏼

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
asbjornu: But the `.3` in `4.5.1-beta.3` indicates that there's 3 commits since the version source, but it is only 2 indicated by the previous `.2` as in `4.5.1-beta.2`. So this feels like a bug to me? Why does it count to 3 and not 2?

HHobeck:
> > Nice! Does `version-in-branch-pattern` require `is-release-branch` to be `true`? If so, we should say something about that in the documentation.
> 
> Can you please answer this, @HHobeck? ☝🏼

Yes you are right the property `version-in-branch-pattern` effects only branches where the `is-release-branch` option is set to `true`. I have changed the documentation :)

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
HHobeck: I think this discussion is related to this issue here: [Same version computed on different branches](https://github.com/GitTools/GitVersion/issues/3453) #3453 

We need to introduce a new branch related property (like `take-incremetned-version`) or make the behavior dependent on the used `version-mode` option and make it configurable. The point is if you are on the hotfix branch you don't want to have properly the tagged version you would like to have the next version `1.1.0+0`. Other way around if you are on the main branch you would like to have the tagged version `1.0.0`.

image

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
HHobeck: If you take a look to the following figure:

image

then you see that GitVersion uses the highest version (4.5.1). If there are ambiguous results (like in this case) the oldest commit as a source will be taken. Here it is the commit with the identifier bf7a8f7f. Thus if you have this in mind `4.5.1-beta.3` is totally correct. See yor comment here https://github.com/GitTools/GitVersion/issues/2394#issuecomment-1062170487

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
asbjornu: If we were to base this on `mode`, how would you suggest we do it?

src/GitVersion.Core.Tests/VersionCalculation/Strategies/VersionInBranchNameBaseVersionStrategyTests.cs
```diff
@@ -80,6 +82,7 @@ public void CanTakeVersionFromNameOfConfiguredReleaseBranch(string branchName, s
         var repository = fixture.Repository.ToGitRepository();
 
         var configuration = GitFlowConfigurationBuilder.New
+            .WithLabelPrefix("([vV]|lts-)?")
```
asbjornu, line 85: Do we need to document this change being required, perhaps?

src/GitVersion.Core.Tests/VersionCalculation/Strategies/VersionInBranchNameBaseVersionStrategyTests.cs
```diff
@@ -68,8 +66,12 @@ public void ShouldNotTakeVersionFromNameOfNonReleaseBranch(string branchName)
 
     [TestCase("release-2.0.0", "2.0.0")]
     [TestCase("release/3.0.0", "3.0.0")]
-    [TestCase("support/lts-2.0.0", "2.0.0")]
```
asbjornu, line 71: With the new logic, `support/lts-2.0.0` is not parsed as before?

src/GitVersion.Core/Configuration/GitFlowConfigurationBuilder.cs
```diff
@@ -133,6 +134,7 @@ private GitFlowConfigurationBuilder()
                 HotfixBranch.Name
             },
             Label = "beta",
+            IsReleaseBranch = true,
```
asbjornu, line 137: So `hotfix` is considered a release branch now? We should probably document that.

src/GitVersion.Core/VersionCalculation/SemanticVersioning/ReferenceNameExtensions.cs
```diff
@@ -0,0 +1,45 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Diagnostics.Contracts;
+using System.Text.RegularExpressions;
+using GitVersion.Extensions;
+
+namespace GitVersion;
+
+public static class ReferenceNameExtensions
+{
+    public static bool TryGetSemanticVersion(this ReferenceName source,
```
asbjornu, line 10: Is there a reason for not having the method `TryGetSemanticVersion` on the `ReferenceName` class directly?

asbjornu (changes requested):

src/GitVersion.Core/VersionCalculation/SemanticVersioning/ReferenceNameExtensions.cs
HHobeck: Actually I have added it befor and got an compile error in the solution new-cli/GitVersion.sln. The ReferenceName.cs file has been added just as a reference... The SemanticVersion type is not known in this solution. @arturcic What do you suggest?

src/GitVersion.Core/Configuration/GitFlowConfigurationBuilder.cs
HHobeck: Yep I agree. I can add a point to the breaking change documenatation no problem.

src/GitVersion.Core.Tests/VersionCalculation/Strategies/VersionInBranchNameBaseVersionStrategyTests.cs
HHobeck: Like you know I have introduced the option `version-in-branch-pattern` and it defaults to the following value:

``` csharp
public const string DefaultVersionInBranchPattern = @"(?<version>[vV]?\d+(\.\d+)?(\.\d+)?).*";
```

That means the behavior is stricter then before right and allows only the name after the version number. If the user don't want this he or she can change it in the configuration very easily. But we want to not allow `release/downgrade-some-lib-to-3.2.1`. What is you suggestion?

src/GitVersion.Core.Tests/VersionCalculation/Strategies/VersionInBranchNameBaseVersionStrategyTests.cs
HHobeck: What is requirement behind it to support lts before the number will be present? For sure if you define support/2.0.0-lts it works per default if you define support/lts-2.0.0 it's not working per default and a change in the configuration needs to be done (either changing `label-prefix` or `version-in-branch-name`). I think this line is not necessary anymore because I have changed the test case.

src/GitVersion.Core.Tests/IntegrationTests/HotfixBranchScenarios.cs
HHobeck: Hmm I think it would be not a good idea to put it on the `version-mode` option because it is very confusing and would lead to issues complaining about that the tag is not recognice correctly. What is your opinion about that?

Also if I think about the behavior of real live deplyoment scenarios (trunk-based, continues-deployment, continues-delivery and manually-deployment) this question is totally independent of that. Because it depends on the fact: Do I create a artifact after I have tagged the commit or before. Or do I take a pre-release version and just redeclare it to production version or not.

What is your opinion about that?

HHobeck:
Okay I'm done with the integration of review comments. Please check and merge to main.

src/GitVersion.Core.Tests/VersionCalculation/Strategies/VersionInBranchNameBaseVersionStrategyTests.cs
asbjornu: Ok, I agree with the stricter behaviour. We just need to document it, like you have.

src/GitVersion.Core/Git/ReferenceName.cs
```diff
@@ -76,6 +78,38 @@ public static ReferenceName FromBranchName(string branchName)
     public override int GetHashCode() => equalityHelper.GetHashCode(this);
     public override string ToString() => Friendly;
 
+    public bool TryGetSemanticVersion([NotNullWhen(true)] out (SemanticVersion Value, string? Name) result,
+        Regex versionPatternRegex, string? labelPrefix, SemanticVersionFormat format)
```
asbjornu, line 82: When a method signature is so long it needs to be broken up, I prefer that every parameter is given its own line.

```suggestion
    public bool TryGetSemanticVersion([NotNullWhen(true)] out (SemanticVersion Value, string? Name) result,
                                      Regex versionPatternRegex,
                                      string? labelPrefix,
                                      SemanticVersionFormat format)
```

src/GitVersion.Core/Git/SemanticVersion.cs
```diff
@@ -225,39 +217,29 @@ private static bool TryParseLoose(string version, [NotNullWhen(true)] out Semant
...
     {
         if (value == null)
-        {
             return 1;
-        }
-        if (this.Major != value.Major)
+        if (Major != value.Major)
         {
-            if (this.Major > value.Major)
-            {
```
asbjornu, line 234: I don't agree with the removal of these curly braces. Please add them back and adjust the formatting settings accordingly.

src/GitVersion.Core/Git/SemanticVersionBuildMetaData.cs
```diff
@@ -87,15 +87,15 @@ public string ToString(string? format, IFormatProvider? formatProvider)
         if (formatProvider?.GetFormat(GetType()) is ICustomFormatter formatter)
             return formatter.Format(format, this, formatProvider);
 
-        if (format.IsNullOrEmpty())
+        if (string.IsNullOrEmpty(format))
```
asbjornu, line 90: I think `format.IsNullOrEmpty()` reads better than `string.IsNullOrEmpty(format)`. Why is it changed?

src/GitVersion.Core/Git/SemanticVersionBuildMetaData.cs
```diff
@@ -87,15 +87,15 @@ public string ToString(string? format, IFormatProvider? formatProvider)
...
             format = "b";
 
         format = format.ToLower();
         return format.ToLower() switch
         {
             "b" => $"{this.CommitsSinceTag}",
-            "s" => $"{this.CommitsSinceTag}{(this.Sha.IsNullOrEmpty() ? null : ".Sha." + this.Sha)}".TrimStart('.'),
-            "f" => $"{this.CommitsSinceTag}{(this.Branch.IsNullOrEmpty() ? null : ".Branch." + FormatMetaDataPart(this.Branch))}{(this.Sha.IsNullOrEmpty() ? null : ".Sha." + this.Sha)}{(this.OtherMetaData.IsNullOrEmpty() ? null : "." + FormatMetaDataPart(this.OtherMetaData))}".TrimStart('.'),
+            "s" => $"{this.CommitsSinceTag}{(string.IsNullOrEmpty(this.Sha) ? null : ".Sha." + this.Sha)}".TrimStart('.'),
```
asbjornu, line 97: Why all these changes to `string.IsNullOrEmpty(x)` instead of `x.IsNullOrEmpty()`?

asbjornu (changes requested):
There's suddenly a lot of formatting changes here. Please do formatting and refactoring in separate pull requests from bug fixes and feature implementations.

Also, I would love to see a test case added for `hotfix/downgrade-library-to-3.1.2` not yielding the version number `3.1.2` as that seems to be one of the primary reasons behind this PR.

src/GitVersion.Core/Git/SemanticVersion.cs
HHobeck: Really strange I didn't change it explicit by myselfe. It might be because of some tooling on my side. sorry for that I change it back

HHobeck:
> There's suddenly a lot of formatting changes here. Please do formatting and refactoring in separate pull requests from bug fixes and feature implementations.
> 
> Also, I would love to see a test case added for `hotfix/downgrade-library-to-3.1.2` not yielding the version number `3.1.2` as that seems to be one of the primary reasons behind this PR.

You mean such tests!? There are already existing. Nevermind let me know if you are missing test scenarios.

```csharp
[Test]
public void DoesNotTakeVersionFromNameOfNonReleaseBranch()
{
    using var fixture = new BaseGitFlowRepositoryFixture("1.0.0");
    fixture.CreateAndMergeBranchIntoDevelop("pull-request/improved-by-upgrading-some-lib-to-4.5.6");
    fixture.CreateAndMergeBranchIntoDevelop("hotfix/downgrade-some-lib-to-3.2.1-to-avoid-breaking-changes");

    fixture.AssertFullSemver("1.1.0-alpha.5");
}

[TestCase("release-2.0.0", "2.0.0")]
[TestCase("release/3.0.0", "3.0.0")]
[TestCase("support/2.0.0-lts", "2.0.0")]
[TestCase("support-3.0.0-lts", "3.0.0")]
[TestCase("hotfix/2.0.0", "2.0.0")]
[TestCase("hotfix-3.0.0", "3.0.0")]
[TestCase("hotfix/2.0.0-lts", "2.0.0")]
[TestCase("hotfix-3.0.0-lts", "3.0.0")]
public void CanTakeVersionFromNameOfConfiguredReleaseBranch(string branchName, string expectedBaseVersion)
{

    fixture.MakeACommit();
    fixture.CreateBranch(branchName);

    var repository = fixture.Repository.ToGitRepository();

    var configuration = GitFlowConfigurationBuilder.New
        .WithBranch("support", builder => builder.WithIsReleaseBranch(true))
   ...
}
```

src/GitVersion.Core/Git/SemanticVersionBuildMetaData.cs
HHobeck: I have changed it back

src/GitVersion.Core/Git/SemanticVersionBuildMetaData.cs
HHobeck: I have changed it back

src/GitVersion.Core/Git/SemanticVersion.cs
HHobeck: I have changed it back

src/GitVersion.Core.Tests/VersionCalculation/Strategies/VersionInBranchNameBaseVersionStrategyTests.cs
HHobeck: I have added a new test DoesNotTakeVersionFromNameWhenItHasBeenAccidentalSpecifiedInBranch

src/GitVersion.Core/Git/ReferenceName.cs
HHobeck: done

.github/ISSUE_TEMPLATE/documentation.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: Documentation
-about: Describe this issue template's purpose here.
+about: Describe the missing/wrong information in the documentation.
```
asbjornu, line 3: Seems like you've rebased some commits from `main` into your branch.

src/GitVersion.Core.Tests/IntegrationTests/VersionInMergedBranchNameScenarios.cs
```diff
@@ -26,6 +26,21 @@ public void DoesNotTakeVersionFromNameOfNonReleaseBranch()
         fixture.AssertFullSemver("1.1.0-alpha.5");
     }
 
+    [TestCase("release")]
+    [TestCase("hotfix")]
+    public void DoesNotTakeVersionFromNameWhenItHasBeenAccidentalSpecifiedInBranch(string branch)
```
asbjornu, line 31: This is a perfect test to showcase why this PR is needed. 👍🏼  I just think the method name can be simplified a bit.

```suggestion
    public void DoesNotTakeVersionFromBranchWithAccidentalVersion(string branch)
```

asbjornu (changes requested):

.github/ISSUE_TEMPLATE/documentation.md
HHobeck: I have used the tooling in github to rebase pending commits. I'm actually confused that this commits are shown as different. I don't know how to workaround this.

arturcic: seems like it was a UI issue in Github, I made a minor change in the latest commit message (added a space) and force-pushed. now it seems to display correctly

asbjornu (approved):
Fabulous! 🚀

PR #3585, opened 2023-06-09
Improve performance on repositories with a lot of tags

### LINKED ISSUE #2881: [Bug] GitVersion is extremely slow (25 minutes)
**Describe the bug**
Running `gitversion` to generate the JSON for a project is very slow; about 25 minutes.

## Expected Behavior
It should not take more than a few seconds.

## Steps to Reproduce
This is a private project I cannot share, unfortunately, but I can run additional tests if necessary.

Attached is a log with `/diag`: [gitversion-log.txt](https://github.com/GitTools/GitVersion/files/7351835/gitversion-log.txt)

## Your Environment

- Version Used: 5.7.0+Branch.main.Sha.8d177c6d666c8eeb1c6a6a2c71fd4b78741137d2
- Operating System and version (Windows 10, Ubuntu 18.04): Windows 10
- git version 2.33.1.windows.1
- The repository has about 12,000 commits in total and 11,000 tags.

### LINKED ISSUE #2885: Improve GitVersion performance
GitVersion unnecessarily slows down the build process on solutions with many projects.

## Detailed Description
GitVersion.exe has a performance problem when used in large solutions with several hundred projects.
Even though it takes less than a second for each project, this sums up to a couple of avoidable minutes for each (re-)build.
This is not so much of a problem on a build agent, but it is annoying during local development when the Git history hasn't changed but larger parts of the application need to be built due to my code changes.

Significant overhead is caused by the YamlDotNet library which is used for gitversion_cache and config file parsing.
The deserialization code is pretty slow on the first call ("cold start") for each deserialized type (here: Dictionary<string,string>, see VersionVariables.FromFileInternal and Config, see ConfigSerializer.Read). (Obviously it needs some internal "warmup" and caches the generated deserializers.)

Here's a performance trace (taken with JetBrains dotTrace) that shows how slow e. g. the gitversion_cache file parsing currently is:

image

And here's how fast it could be with some very simple/naive, hand-crafted parsing logic (that currently works only for the simply structured gitversion_cache file):

image

A similar performance hit (> 100 ms) can be seen when parsing the GitVersion Config file.

An additional performance gain seems to be achievable by streamlining the startup procedure (CreateHostBuilder and friends).

## Context
Performance improvements are important because I'm slowed by down GitVersion by additional build times during development.
gitversion.exe is executed thousands of times per day and should be strongly optimized performance.

## Possible Implementation
Replace YamlDotNet with a faster YAML parser, or find a way to somehow pre-compile the internal setup of its deserializers (optimized for the required data types).

### LINKED ISSUE #3212: [Bug]Regression perfomance since 5.2.4
**Describe the bug**
the code in TaggedCommitVersionStrategy has been change in 5.2.0 with this code 

            var tagsOnBranch = currentBranch
                .Commits
                .SelectMany(commit => { return allTags.Where(t => IsValidTag(t.Item1, commit)); })
                .Select(t =>
                {
                    if (t.Item1.PeeledTarget() is Commit)
                        return new VersionTaggedCommit(t.Item1.PeeledTarget() as Commit, t.Item2, t.Item1.FriendlyName);

                    return null;
                })
                .Where(a => a != null)
                .Take(5)
                .ToList();
in Mainline mode, it avoid to list all tags 

In 5.10.0, the code change and it list all tag, it cause gitversion is very slow in our buildchain (almost 40 minutes)

## Expected Behavior
gitversion eecute quickly

## Actual Behavior
gitversion very slow

## Possible Fix
same as 5.2.4

## Context
In our buildchain, we use repo with many version in githubflow each build is tagged so we have more than 3000 tags.
Gitversion start to be very slow since 5.0.0, so we try to upgrade to 5.10.0. after many research, we found something similar here 
https://github.com/GitTools/GitVersion/pull/2149
it appear that it fixes our problem

## Your Environment

- Version Used:
windows 10 64 bit

AntonMTolmachev:
## Description
Added a tag SHA cache into IncrementStrategyFinder to prevent read of all tags from repository each time when FindCommitMessageIncrement is called.

## Related Issue
Fixes #3212 Fixes #2881 Fixes #2885

## Motivation and Context
GitVersion runs slowly on repositories with a lot of tags when there is no gitversion_cache available.
Each tag produces a base version which goes through IncrementStrategyFinder, which in turn reads all tags from repository. This causes N^2 tags to be read from repository in total.

## How Has This Been Tested?
Manually checked performance improvement via profiler on a repository with 200 commits & tags. See screenshot attached below.
A test that recreates a repository with a similar structure is provided in pull request.

image
Performance before (right) and after (left).

## Checklist:

### DIFF
--- a/src/GitVersion.Core.Tests/IntegrationTests/PerformanceScenarios.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/PerformanceScenarios.cs
@@ -1,0 +1,29 @@
+using GitVersion.Configuration;
+using GitVersion.Core.Tests.Helpers;
+
+namespace GitVersion.Core.Tests.IntegrationTests;
+
+public class PerformanceScenarios : TestBase
+{
+    [Test]
+    public void RepositoryWithALotOfTags()
+    {
+        var configuration = GitFlowConfigurationBuilder.New.Build();
+
+        using var fixture = new EmptyRepositoryFixture();
+
+        const int maxCommits = 500;
+        for (int i = 0; i < maxCommits; i++)
+        {
+            fixture.MakeATaggedCommit($"1.0.{i}");
+        }
+
+        fixture.BranchTo("feature");
+        fixture.MakeACommit();
+
+        var sw = Stopwatch.StartNew();
+
+        fixture.AssertFullSemver($"1.0.{maxCommits}-feature.1+1", configuration);
+        sw.ElapsedMilliseconds.ShouldBeLessThan(5000);
+    }
+}
--- a/src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
+++ b/src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
@@ -16,6 +16,7 @@ internal class IncrementStrategyFinder : IIncrementStrategyFinder
     private readonly Dictionary<string, VersionField?> commitIncrementCache = new();
     private readonly Dictionary<string, Dictionary<string, int>> headCommitsMapCache = new();
     private readonly Dictionary<string, ICommit[]> headCommitsCache = new();
+    private readonly Lazy<IReadOnlySet<string?>> tagsShaCache;
 
     private static readonly Regex DefaultMajorPatternRegex = new(DefaultMajorPattern, RegexOptions.Compiled | RegexOptions.IgnoreCase);
     private static readonly Regex DefaultMinorPatternRegex = new(DefaultMinorPattern, RegexOptions.Compiled | RegexOptions.IgnoreCase);
@@ -24,7 +25,11 @@ internal class IncrementStrategyFinder : IIncrementStrategyFinder
 
     private readonly IGitRepository repository;
 
-    public IncrementStrategyFinder(IGitRepository repository) => this.repository = repository.NotNull();
+    public IncrementStrategyFinder(IGitRepository repository)
+    {
+        this.repository = repository.NotNull();
+        this.tagsShaCache = new Lazy<IReadOnlySet<string?>>(ReadRepositoryTagsSha);
+    }
 
     public VersionField DetermineIncrementedField(ICommit? currentCommit, BaseVersion baseVersion, EffectiveConfiguration configuration)
     {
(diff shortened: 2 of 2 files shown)


arturcic:
@AntonMTolmachev that is an impressive improvement, let's see the tests pass, then we can continue

arturcic (approved):

sreenivas-ps:
@AntonMTolmachev : I've the same problem with  one of my repo that has lots of tags, what exactly I should do to prevent reading 200+ tags? At the moment, I'm using this:

```
      - name: Install GitVersion
        uses: gittools/actions/gitversion/setup@v0.9.15
        with:
          versionSpec: '5.x'
      - name: Determine Version
        id: gitversion
        uses: gittools/actions/gitversion/execute@v0.9.15
        with:
          useConfigFile: true
          targetPath: ${{ inputs.targetPath }}
          configFilePath: ${{ inputs.configFilePath }}
```

arturcic:
@sreenivas-ps this should be fixed in 6.0.0-beta3 

````yaml
      - name: Install GitVersion
        uses: gittools/actions/gitversion/setup@v0.9.15
        with:
          versionSpec: '6.0.0-beta.3'
          includePrerelease: true
````

**Comment by sreenivas-ps** [comment]
@arturcic : thanks. I've tried using that version before but it is even taking long than 5.x. For each branch in the repo, it is cycling 236 tags ( to find merge base between)  the PR and  branch and then moves to next branch to find the same...

e.g..
```
End: Finding merge base between 'pull/1723/merge' and 'develop'. (Took: 109,096.17ms)
End: Finding merge base between 'pull/1723/merge' and 'feature/DATAENG-0000_Quick_fix_backward_compatibility'. (Took: 109,958.75ms)
End: Finding merge base between 'pull/1723/merge' and 'feature/DATAENG_4662_Experian_Files_2023_06_09_05_56_32'. (Took: 114,400.10ms)
```

I'm not sure if i'm doing anything wrong. Here is my Gitversion.yaml file and new to this gitversion.yaml

```
mode: Mainline
branches: {}
ignore:
  sha: []
merge-message-formats: {}
```

My repo has about 236 tags.
With 5.x, it was taking around 9 to 12 mins but with 6.0.0-beta3, it is running with 35mins and still going on.

Issue #3606, opened 2023-06-19
[Bug] CS0436 may be present with InternalsVisibleTo


Applesauce314:
**Describe the bug**
When `GitVersionTask `was replace with `GitVersion.MSBuild` in 5.6.0 the issue described in #448 has reemerged.
## Expected Behavior
Not sure, GitVersionTask -> GitVersion.MsBuild is obviously a breaking version change, but the inconsistent behavior does not appear to be documented. was this change in behavior intentional or an accidental regression between versions.
## Actual Behavior
CS0436 is present if you try to use the `GitVersionInformation` when it is also present in a referenced project which exposes its internals to the current assembly with `InternalsVisibleTo`

arturcic:
@Applesauce314 can you please create a test solution that reproduces the issue, and then we can decide if/when to fix it. If we decide to do it, then it will be implemented in v6 release cycle.

Applesauce314:
[repro repository](https://github.com/Applesauce314/GitVersion_3606)

Apparently the change made for #448 only existed in version 3.2.1, version 4.0.0 was what actually returned to the original behavior.

my thoughts on "fixing" it would be to add a configuration option, that way it would not be a breaking change.
```xml
<GenerateGitVersionInformationInUniqueNamespace>true</GenerateGitVersionInformationInUniqueNamespace>
```

I have a fork I can turn into a pull request with the above suggestion included

Applesauce314:
pull request is #3619 

I am not very happy with the way I handled the template for C♯, its kind of clunky,

Issue #448, opened 2015-05-28
Generated classes create conflicts with InternalsVisibleTo


onovotny:
Using the 3.0 b2 msbuild tasks, I see this:
```
C:\Users\Oren\Source\Git\MetroLog\MetroLog.WP8\obj\Release\GitVersionTaskAssemblyInfo.g.cs(8,12,8,23): warning CS0436: The type 'ReleaseDateAttribute' in 'C:\Users\Oren\Source\Git\MetroLog\MetroLog.WP8\obj\Release\GitVersionTaskAssemblyInfo.g.cs' conflicts with the imported type 'ReleaseDateAttribute' in 'MetroLog, Version=0.8.8.0, Culture=neutral, PublicKeyToken=ba4ace74c3b410f3'. Using the type defined in 'C:\Users\Oren\Source\Git\MetroLog\MetroLog.WP8\obj\Release\GitVersionTaskAssemblyInfo.g.cs'.
```
Problem is that the ReleaseDateAttribute and GitVersionInformation classes are created in the root namespace, so they'll conflict with friend access.

Proposal:
Put the attribute in a namespace underneath the assembly name.

I.e., for an assemlby Foobar.Frob:

namespace Foobar.Frob {
...
}

JakeGinnivan:
@SimonCropp, thoughts?

SimonCropp:
Dont object. But how bigger "greaking" change is this?

onovotny:
As these classes are generated at compile-time and included, how many people have references to those types?

I'd think it'd be more of a breaking change if it was generated on included somehow in VS so developers could use those types in their code directly?

SimonCropp:
well either way we r in v3beta so i say lets do it. @onovotny want to do a PR?

onovotny:
Wish I could, I just don't have the bandwidth right now :/

distantcam:
Should we also add a CompilerGeneratedAttribute too?

PR #3619, opened 2023-07-05
Unique namespace generation for GitVersionInformation class

Applesauce314:
## Description
adds a configuration option 
```xml
<GenerateGitVersionInformationInUniqueNamespace>true</GenerateGitVersionInformationInUniqueNamespace>
```
to allow the creation of the `GitVersionInformation` class in a namespace related to the project it is in.
this allows different projects to have the class in different namespaces so if internalsvisibleto attribute is used there is not class name collision (CS0436), eliminating the need to suppress the issue.

## Related Issue
issue #3606 
Issue #448 

## Motivation and Context
Elimination of CS0436

## How Has This Been Tested?
wrote unit tests to cover added code

## Checklist:


docs/input/docs/usage/msbuild.md
```diff
@@ -258,6 +258,18 @@ For SDK-style projects, `UpdateVersionProperties` controls setting the default
 variables: `Version`, `VersionPrefix`, `VersionSuffix`, `PackageVersion`,
 `InformationalVersion`, `AssemblyVersion` and `FileVersion`.
 
+### Namespace generation
+
+You can configure GitVersion to generate the `GitVersionInformation` class in a namespace that matches the current assembly. By default this class is created in the global namespace. If `GenerateGitVersionInformationInUniqueNamespace` is set to true, the `GitVersionInfomation` class will instead be generated in a namespace matching the current project. If the property `<RootNamespace>` is set that value will be used, otherwise the name of the project file is used.
```
asbjornu, line 263: Since the namespace matches the project, I think `GenerateGitVersionInformatoinInProjectNamespace` seems more accurate. But that is also a bit long and unwieldy, so I think we can just call it `UseProjectNamespace`, no?

```suggestion
You can configure GitVersion to generate the `GitVersionInformation` class in a namespace that matches the current assembly. By default this class is created in the global namespace. If `UseProjectNamespace` is set to true, the `GitVersionInfomation` class will instead be generated in a namespace matching the current project. If the property `<RootNamespace>` is set that value will be used, otherwise the name of the project file is used. 
```

docs/input/docs/usage/msbuild.md
```diff
@@ -258,6 +258,18 @@ For SDK-style projects, `UpdateVersionProperties` controls setting the default
...
 `InformationalVersion`, `AssemblyVersion` and `FileVersion`.
 
+### Namespace generation
+
+You can configure GitVersion to generate the `GitVersionInformation` class in a namespace that matches the current assembly. By default this class is created in the global namespace. If `GenerateGitVersionInformationInUniqueNamespace` is set to true, the `GitVersionInfomation` class will instead be generated in a namespace matching the current project. If the property `<RootNamespace>` is set that value will be used, otherwise the name of the project file is used. 
+
+```xml
+<PropertyGroup>
+  ...
+  <GenerateGitVersionInformationInUniqueNamespace>true</GenerateGitVersionInformationInUniqueNamespace>
```
asbjornu, line 268:
```suggestion
  <UseProjectNamespace>true</UseProjectNamespace>
```

asbjornu (changes requested):
I think this is a nice addition that may resolve a few bugs reported over the years, but I think the proposed name is a bit too long, unwieldy, and imprecise. How do you feel about `UseProjectNamespace`?

docs/input/docs/usage/msbuild.md
Applesauce314: I would be worried about such a short property name colliding with other build tasks. and a developer would not know what it was associated with just looking at a project file.
We could probably make it a tiny bit shorter by renaming it `UseProjectNamespaceForGetVersionInfomation`.
I'm not sure length is a huge issue though since it lives on a line in the project file by itself.

asbjornu: > I would be worried about such a short property name colliding with other build tasks. 

Good point.


> We could probably make it a tiny bit shorter and more precise by renaming it `UseProjectNamespaceForGitVersionInfomation`. 

👍🏼

src/GitVersion.Output/GitVersionInfo/Templates/GitVersionInformation.cs
```diff
@@ -23,10 +23,10 @@ namespace System.Diagnostics.CodeAnalysis
...
-static class GitVersionInformation
-{{
+{1}{2}
+{4}[global::System.Runtime.CompilerServices.CompilerGenerated]
+{4}[global::System.Diagnostics.CodeAnalysis.ExcludeFromCodeCoverage]
+{4}static class GitVersionInformation
+{4}{{
 {0}
-}}
+{4}}}{3}
```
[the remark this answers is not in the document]
asbjornu: Yeah, this is not very elegant. Pull requests to improve it are welcome.

asbjornu (approved):

PR #3628, opened 2023-07-18
docs: fix Conventional Commits PATCH rule

### LINKED ISSUE #3627: [Docs] Fix "Conventional Commit" commit message format
https://github.com/GitTools/GitVersion/blob/63f7d275955cc9596bb67715e7b2ec35acbdab60/docs/input/docs/reference/version-increments.md?plain=1#L91-L104
The current Conventional Commit configuration breaks the standard's rules. Where `PATCH` is supposed to be incremented by only `fix` or `perf`, the configuration increments `PATCH` when the commit Type is *any [Angular commit Type](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#type)* except `feat` and no breaking change is present.

Additionally, the following should be noted for consumers:
- In Conventional Commit 1.0.0, [`revert`-type commits do not explicitly increment the version](https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits) as their significance differs on a case-by-case basis. The convention gives the following format recommendation:
  ```
  revert: let us never again speak of the noodle incident
  
  Refs: 676104e, a215868
  ```
  - WARN: this recommendation prevents commit message analyzers from determining the reverted commit's significance. Analyzers which also parse commit hashes in the message *may* be able to determine the reverted commit's significance. In contrast, the `angular` convention [places the reverted commit's header in the header of the revert commit](https://github.com/angular/angular/blob/main/CONTRIBUTING.md#revert-commits).
- Some software used in tandem with GitVersion may use version-increment rules not present in a given commit convention preset (e.g. `angular`, `conventionalcommits`). `@semantic-release/commit-analyzer` v10.0.1, for example, will increment `PATCH` for any `revert` commits regardless of the (in)significance of the reverted changes. This semantic-release behavior can be counteracted by adding custom rules to `@semantic-release/commit-analyzer`'s configuration. Rather than RegEx, commit-analyzer uses [Micromatch's glob implementation](https://github.com/micromatch/micromatch#matching-features).

BinToss:
This PR changes the "Conventional Commit" RegEx to be more in line with Convention Commits 1.0.0 and commitlint specifications.

This PR also adds notes for configuring GitVersion and semantic-release to generate the same version.

I'm considering adding a separate section for the `angular` commit format since that is what semantic-release defaults to. It's...nearly identical to Conventional Commits, but is more concrete in its specifications e.g. a reverted commit's header (`type(scope): header message`) is to be the revert commit's header message.

## Related Issue

resolves #3627

## Motivation and Context

I use GitVersion for stamping .NET assemblies and NuGet packages with version information, but I use semantic-release for git tags, release commits, change logs, release notes, GitHub releases, and more.

Because they both have commit analysis and version generation capabilities, I've found that ensuring both generate the same version can be quite finicky. 

## How Has This Been Tested?

https://regexr.com/7h6lv

## Checklist:

### DIFF
--- a/docs/input/docs/reference/version-increments.md
+++ b/docs/input/docs/reference/version-increments.md
@@ -95,9 +95,9 @@ you can leverage this feature as follows:
 
 ```yaml
 mode: MainLine # Only add this if you want every version to be created automatically on your main branch.
-major-version-bump-message: "^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test)(\\([\\w\\s-]*\\))?(!:|:.*\\n\\n((.+\\n)+\\n)?BREAKING CHANGE:\\s.+)"
-minor-version-bump-message: "^(feat)(\\([\\w\\s-]*\\))?:"
-patch-version-bump-message: "^(fix|perf)(\\([\\w\\s-]*\\))?:"
+major-version-bump-message: "^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test)(\\([\\w\\s-,/\\\\]*\\))?(!:|:.*\\n\\n((.+\\n)+\\n)?BREAKING CHANGE:\\s.+)"
+minor-version-bump-message: "^(feat)(\\([\\w\\s-,/\\\\]*\\))?:"
+patch-version-bump-message: "^(fix|perf)(\\([\\w\\s-,/\\\\]*\\))?:"
 ```
 
 This will ensure that your version gets bumped according to the commits you've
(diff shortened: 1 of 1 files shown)


asbjornu (approved):
I don't use conventional commits myself, but this looks reasonable to me.

asbjornu:
I think we should just set `check_external_hash` to `false` in HTMLProofer to turn it green.

BinToss:
I didn't thoroughly test the theory, but I believe it's failing due to many of GitHub's "anchored" links not matching to an element's id attribute. That is, GitHub links "#L8" to line 8 of a given file via JavaScript and classes, but the HTML proofer is probably looking for something like `<a id="L8" />`.

PR #3631, opened 2023-07-23
Htmlproofer fix

Applesauce314:
htmlproofer seems to be ignoring the --disable-external option seems to be ignored during the CI run of htmlProofer

## Description
updated the --ignore-urls to have " around them as documented to see if that fixes the issue, perhaps it is not parsing the rest of the arguments properly due to that?

## Checklist:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -110,7 +110,7 @@ jobs:
       uses: chabad360/htmlproofer@master
       with:
         directory: ./artifacts/docs/preview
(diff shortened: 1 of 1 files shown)


Applesauce314:
it appears the version 5ba7f6e6685d04729c18f2453457881e9951382138abd7d374c441ff65a0e474 of the htmlproofer package introcuded this bug, i'm closing this as it appears to not be an issue with gitversions config, but insteat htmlproofer.

Issue #3083, opened 2022-04-14
[Bug] Azure DevOps: Empty git branch on PR triggered build agent when Gitversion Setup is run


shaneocall:
**Describe the bug**
With GitVersion 5.10.0, ion Azure DevOps, when GitVersion is installed on a build agent which was triggered by an automated PR Build Validation. The git branch is returned as an empty string and `Gitversion setup` fails.

## Expected Behavior
In Gitversion 5.9.0, for PR a numbered 1751, the below can be seen in the output:
![image](https://user-images.githubusercontent.com/16311180/163429081-37d97221-8fe4-44ca-a064-67ed674befeb.png)

## Actual Behavior
In Gitversion 5.10.0, for PR a numbered 1758, the below can be seen in the output:
![image](https://user-images.githubusercontent.com/16311180/163429368-91513dc2-5d18-4ff0-9157-54a7fe7c8afe.png)

## Steps to Reproduce
- Using AzureDev Ops repo
- Add a Build Validation in the Repo settings: 
![image](https://user-images.githubusercontent.com/16311180/163429754-562102da-6a64-40e4-95f6-4070d66fdeec.png)
- In the `azure-pipelines.yml` build file, install GitVersion and run `gitversion Setup`
![image](https://user-images.githubusercontent.com/16311180/163429991-d305b8d6-fad7-4d17-9922-368e398060eb.png)
- Create a PR in DevOps and Queue the build validation: 
![image](https://user-images.githubusercontent.com/16311180/163430293-ce7a7009-f562-47a3-aded-131b93631ca0.png)

## Context
Unable to use pre-merge Build Validation in Azure Devops with GitVersion 5.10.0

## Further information
Error further down the logs:

``
refs/remotes/pull/1758/merge (29cb882c4287dfc0c9ce867d43149a74154600cf)
  INFO [04/14/22 15:31:25:28] No local branch pointing at the commit '29cb882c4287dfc0c9ce867d43149a74154600cf'. Fake branch needs to be created.
  INFO [04/14/22 15:31:25:29] Fetching remote refs to see if there is a pull request ref
  INFO [04/14/22 15:31:25:33] End: Normalizing git directory for branch '' (Took: 182.17ms)
  ERROR [04/14/22 15:31:25:35] An unexpected error occurred:
LibGit2Sharp.LibGit2SharpException: this remote has never connected
[...]

Issue #3082, opened 2022-04-14
[Bug] LibGit2Sharp.LibGit2SharpException: this remote has never connected after updating to 5.10.0


IBlackI:
**Describe the bug**
Gitversion execute throws a LibGit2Sharp.LibGit2SharpException after updating to version 5.10.0

## Expected Behavior
GitVersion should be able to determine the correct version as when it was running 5.9.0
Logs have been stripped of sensitive information, and shortened.
Logs with 5.9.0:
```
2022-04-14T09:21:18.0677229Z ##[section]Starting: gitversionexecute
2022-04-14T09:21:18.0965352Z ==============================================================================
2022-04-14T09:21:18.0965732Z Task         : Execute GitVersion Task
2022-04-14T09:21:18.0966103Z Description  : Easy Semantic Versioning (http://semver.org) for projects using Git
2022-04-14T09:21:18.0966404Z Version      : 0.9.13
2022-04-14T09:21:18.0966648Z Author       : GitTools Contributors
2022-04-14T09:21:18.0966978Z Help         : See the [documentation](https://gitversion.net/docs/) for help
2022-04-14T09:21:18.0967577Z ==============================================================================
2022-04-14T09:21:18.5580529Z Command: dotnet-gitversion D:/a/1/s /output json /output buildserver /config D:\a\1\s\build\GitVersion.yml
2022-04-14T09:21:18.5623256Z [command]C:\hostedtoolcache\windows\GitVersion.Tool\5.9.0\x64\dotnet-gitversion.exe D:/a/1/s /output json /output buildserver /config D:\a\1\s\build\GitVersion.yml
2022-04-14T09:21:28.1605223Z INFO [04/14/22 9:21:19:05] Working directory: D:/a/1/s
2022-04-14T09:21:28.1606662Z INFO [04/14/22 9:21:19:16] Branch from build environment: refs/pull/12445/merge
2022-04-14T09:21:28.1607933Z INFO [04/14/22 9:21:19:17] Project root is: D:\a\1\s\
2022-04-14T09:21:28.1609543Z INFO [04/14/22 9:21:19:17] DotGit directory is: D:\a\1\s\.git
2022-04-14T09:21:28.1610670Z INFO [04/14/22 9:21:19:17] Begin: Normalizing git directory for branch 'refs/pull/12445/merge'
2022-04-14T09:21:28.1611984Z   INFO [04/14/22 9:21:19:24] One remote found (origin -> '[...]').
[...]

Issue #3081, opened 2022-04-14
[Bug] Version 5.10 is causing "SyntaxError: Unexpected end of JSON input" error


ebbz:
**Describe the bug**
Version 5.10 is causing "SyntaxError: Unexpected end of JSON input" error
![afbeelding](https://user-images.githubusercontent.com/6773985/163355572-aedeabb5-e858-4426-80e9-ac5d38bf508c.png)

Failing build:
![afbeelding](https://user-images.githubusercontent.com/6773985/163355111-0034f03f-c121-4a4e-a6da-2959abcc42e1.png)
Normal working build:
![afbeelding](https://user-images.githubusercontent.com/6773985/163355333-21e2dc15-a630-4a4e-be30-4a829c37f8a1.png)

## Expected Behavior
No errors

## Actual Behavior
I believe this is because since version 5.10 the variable 'Branch from build environment: ' is not being set.

## Possible Fix
Make sure the variable is set again

## Context
We always use the newest version fi gitvesion, since the update a few hours ago, our builds stopped working.

## Your Environment
Azure DevOps (private environment)

- Version Used:
Agent name: 'Azure Pipelines 2'
Agent machine name: 'WIN-MGDD5L23JQP'
Current agent version: '2.202.0'
Operating System
Virtual Environment
Virtual Environment Provisioner
Current image version: '20220330.1'
Agent running as: 'VssAdministrator'
Prepare build directory.
Set build variables.
Download all required tasks.
Downloading task: gitversion/setup (0.9.13)
Downloading task: gitversion/execute (0.9.13)
Downloading task: NodeTool (0.200.0)
Downloading task: Npm (1.198.0)
Downloading task: DotNetCoreCLI (2.198.0)
Downloading task: reportgenerator (5.1.4)
Downloading task: PublishCodeCoverageResults (1.200.0)
Downloading task: Gitleaks (2.2.16)
Downloading task: PublishBuildArtifacts (1.200.0)

cheesi:
We are having the same issue on Azure DevOps (using https://github.com/GitTools/actions/issues).

```
2022-04-14T08:54:08.7396902Z ##[section]Starting: Generate version
2022-04-14T08:54:08.7406213Z ==============================================================================
2022-04-14T08:54:08.7406533Z Task         : Execute GitVersion Task
2022-04-14T08:54:08.7406846Z Description  : Easy Semantic Versioning (http://semver.org) for projects using Git
[...]

PR #3033, opened 2022-03-06
Use GITHUB_REF only for CI builds on branches

ThomasPiskol:
Improve support for GitHub Actions and fix #2838 

## Description
In case GitVersion is running during GitHub Actions CI build, the environment variable `GITHUB_REF` is used by GitVersion to get the current branch.

But GitHub Actions only provide branch name here in case of a "normal" CI build. If the GitHub Action was triggered because of a tag event or pull request event, the `GITHUB_REF` environment variable does not contain an existing branch name [as described by GitHub](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables):

> The branch or tag ref that triggered the workflow run. For branches this is the format `refs/heads/<branch_name>`, for tags it is `refs/tags/<tag_name>`, and for pull requests it is `refs/pull/<pr_number>/merge`. This variable is only set if a branch or tag is available for the event type. For example, `refs/heads/feature-branch-1`.

## Related Issue
Fixes #2838.
Fixes #2852.
Fixes #2301.
Fixes #2928.
Fixes #2869.
Related to #2074.

## Motivation and Context
Bug fix and align with GitHub Actions concept.

## How Has This Been Tested?
Three GitHub Actions scenarios must be tested:
1. "normal" GitHub Actions CI build
2. Tag Event
3. Pull Request Event

https://github.com/ThomasPiskol/gitversion-repro/actions

There's no change to the "normal" the Pull Request events, they create the same version string as before.
So this change should fix the issue without causing some issue for the other workflows.

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [X] I have added tests to cover my changes.
- [X] All new and existing tests passed.

PR #3768, opened 2023-11-13
Re-fixed github actions tag handling

david-driscoll:
So similar changes to this were provided in PR #3033 (fix for #2838) however the changes Azure Pipelines caused it to quickly be reverted (#3081, #3082, #3083).  While I believe the Azure Pipelines issues were genuine, I don't believe the changes for Github Actions had any issues (that I'm aware of) however it currently causes mass headache when trying to push preview releases.

This change reintroduces only the github action changes.  I suspect the changes to Gitlab and others might still apply, however I would rather they get added as separate PRs in the event that one of them triggers another revert.

## Description
Updates the `GitHubActions` build agent to properly handle both tag builds and pull request builds based on the github actions docs. 

See: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

## Related Issue
#2838, #3033, https://github.com/GitTools/GitVersion/discussions/3653

## Motivation and Context
Today github actions cannot properly create a tagged preview build because the version comes out and some form of 1.5.0-tags-v1-5-0-preview-6.1 where the real base branch is not properly detected

## How Has This Been Tested?
Changes are based off the original PR with some changes based on additional environment variables that exist in a Github Action Runner environment (event name, ref type, head ref, etc)

## Checklist:

### DIFF
--- a/src/GitVersion.BuildAgents.Tests/Agents/GitHubActionsTests.cs
+++ b/src/GitVersion.BuildAgents.Tests/Agents/GitHubActionsTests.cs
@@ -18,6 +18,7 @@ public class GitHubActionsTests : TestBase
         this.environment = sp.GetRequiredService<IEnvironment>();
         this.buildServer = sp.GetRequiredService<GitHubActions>();
         this.environment.SetEnvironmentVariable(GitHubActions.EnvironmentVariableName, "true");
+        this.environment.SetEnvironmentVariable("GITHUB_REF_TYPE", "branch");
 
         this.githubSetEnvironmentTempFilePath = Path.GetTempFileName();
         this.environment.SetEnvironmentVariable(GitHubActions.GitHubSetEnvTempFileEnvironmentVariableName, this.githubSetEnvironmentTempFilePath);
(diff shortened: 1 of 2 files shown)


src/GitVersion.BuildAgents.Tests/Agents/GitHubActionsTests.cs
```diff
@@ -75,26 +76,30 @@ public void GetCurrentBranchShouldHandleBranches()
...
-        result.ShouldBe("refs/tags/1.0.0");
+        result.ShouldBeNull();
     }
 
     [Test]
     public void GetCurrentBranchShouldHandlePullRequests()
     {
         // Arrange
+        this.environment.SetEnvironmentVariable("GITHUB_EVENT_NAME", "pull_request");
+        this.environment.SetEnvironmentVariable("GITHUB_HEAD_REF", "some-branch");
```
david-driscoll, line 94: head ref already has the `refs/heads/` removed, this mirrors that behavior. Same for base ref as well.

asbjornu (approved):
LGTM!

PR #3838, opened 2023-12-14
Replace the version mode Mainline in 6.x

HHobeck:
## Description
Create TrunkBasedVersionStrategy implementation and replace the version mode Mainline in 6.x.

For more detail please see the following ADR:
- #3601

## Checklist:


src/GitVersion.Core/Git/VersionFieldExtensions.cs
arturcic, line 1: Please move `VersionFieldExtensions` to Extensions folder

src/GitVersion.Core/Git/ICommit.cs
```diff
@@ -2,6 +2,8 @@ namespace GitVersion;
 
 public interface ICommit : IEquatable<ICommit?>, IComparable<ICommit>, IGitObject
 {
+    public bool IsMergeCommit => Parents.Count() >= 2;
+
```
arturcic, line 6: I think public is redundant here. Please also remove the empty line to be consistent with other interfaces in the Git Folder

src/GitVersion.Core/Git/ITag.cs
```diff
@@ -2,7 +2,7 @@ namespace GitVersion;
 
 public interface ITag : IEquatable<ITag?>, IComparable<ITag>, INamedReference
 {
-    string? TargetSha { get; }
+    string TargetSha { get; }
```
arturcic, line 6: Please remove the empty line to be consistent with other interfaces

src/GitVersion.Core/MergeMessage.cs
```diff
@@ -93,4 +94,20 @@ private ReferenceName GetMergedBranchName(string mergedBranch)
         }
         return ReferenceName.FromBranchName(mergedBranch);
     }
+
+    public static bool TryParse(
+        [NotNullWhen(true)] out MergeMessage? mergeMessage, ICommit mergeCommit, IGitVersionConfiguration configuration)
```
arturcic, line 99: I think the `out` parameter should go at the end

src/GitVersion.Core/SemVer/SemanticVersion.cs
```diff
@@ -24,14 +24,16 @@ public class SemanticVersion : IFormattable, IComparable<SemanticVersion>, IEqua
...
+
     public SemanticVersionPreReleaseTag PreReleaseTag { get; init; }
 
     public SemanticVersionBuildMetaData BuildMetaData { get; init; }
 
     public bool IsLabeledWith(string value) => PreReleaseTag.HasTag() && PreReleaseTag.Name.IsEquivalentTo(value);
 
     public bool IsMatchForBranchSpecificLabel(string? value)
-        => PreReleaseTag.Name.Length == 0 || value is null || IsLabeledWith(value);
+        => (PreReleaseTag.Name == string.Empty && PreReleaseTag.Number is null) || value is null || IsLabeledWith(value);
```
arturcic, line 36: https://josefpihrt.github.io/docs/roslynator/analyzers/RCS1156/

src/GitVersion.Core/Git/ICommit.cs
HHobeck: Yes you are right. Unfortunately the design in .net is not consistent compared to a definition in classes because there it defaults to internal.

src/GitVersion.Core/Git/ITag.cs
HHobeck: Okay no problem I change it. Just want to point out that it is hard to ready and maybe it is worth to think about to change this rule to increase the accessability to the code.

src/GitVersion.Core/SemVer/SemanticVersion.cs
HHobeck: Actually I'm not constructing any new string. Thus it would be the same. If you not mind I would like to leave it like it is (see https://stackoverflow.com/a/10250).

arturcic: it's not a big thing, just wanted to have a consistent approach, we will have to agree on an approach, put it in the .editorconfig and then the analyzer will know how to handle it

src/GitVersion.Core/Git/ITag.cs
arturcic: Let's check this after we're done with the PR

src/GitVersion.Core/SemVer/SemanticVersion.cs
HHobeck: Done

HHobeck:
Actual version has been rebased from main branch without conflicts. Is now up to date

arturcic:
Hey @HHobeck I have a request, is it possible to split this PR in 2 - first PR that prepares the existing code for the second PR.

What I mean by that is to include in the first PR the existing classes with the changes required for the second PR, but that can still be part of the code without the second PR.

The reasoning is there are a huge number of files to review, and you keep adding/ changing the PR, and it's a bit hard to review.

If you can separate the changes into 2 PRs it will be easier to follow the changes. Thoughts?

HHobeck:
> Hey @HHobeck I have a request, is it possible to split this PR in 2 - first PR that prepares the existing code for the second PR.
> 
> What I mean by that is to include in the first PR the existing classes with the changes required for the second PR, but that can still be part of the code without the second PR.
> 
> The reasoning is there are a huge number of files to review, and you keep adding/ changing the PR, and it's a bit hard to review.
> 
> If you can separate the changes into 2 PRs it will be easier to follow the changes. Thoughts?

If it makes your live easier. I can split it no problem. But please keep in mind that all changes unitil 94fdd4f are just unit test. Or if you ignore the files in `.\src\GitVersion.Core.Tests\TrunkBased\` then you see it is not so much. ;)

HHobeck:
Maybe good to know: The TrunkBasedVersionStrategy class is the entry point for calculating the version based on the new trunk based strategy which will be replacing the MainlineVersionCaluclator. All classes under .\src\GitVersion.Core\VersionCalculation\TrunkBased\ are related to TrunkBasedVersionStrategy. If you ignore this as well then you can do a review with a clear focus on:
- Changes related to pre-peration
- Changes related to TrunkBasedVersionStrategy implementation
- Unit Tests

arturcic:
> Maybe good to know: The TrunkBasedVersionStrategy class is the entry point for calculating the version based on the new trunk based strategy which will be replacing the MainlineVersionCaluclator. All classes under .\src\GitVersion.Core\VersionCalculation\TrunkBased\ are related to TrunkBasedVersionStrategy. If you ignore this as well then you can do a review with a clear focus on:
> 
> * Changes related to preperation
> * Changes related to TrunkBasedVersionStrategy implementation
> * Unit Tests

Do you think you can create 3 PRs that are based on these 3 points?

HHobeck:
> As for the unit tests I noticed there are tests that have quite a lot of [TestCase] that are similar, maybe it makes sense for some of them to use the [TestCaseSource](https://docs.nunit.org/articles/nunit/writing-tests/attributes/testcasesource.html) instead, and have the cases created based on the values?

Hmm. I'm not sure if it is a good idea. Can we discuss this on a concrete example?

arturcic:
For example https://github.com/GitTools/GitVersion/blob/d276be90688c3bcf5809b32043fb16601f31189c/src/GitVersion.Core.Tests/TrunkBased/TrunkBasedScenariosWithAGitHubFlow%2BGivenAFeatureBranchWithOneCommitBranchedFromMainWhenCommitBHasBumpMessageMajor.cs#L139

arturcic:
As far as I understand it's a combinatory issue, so you can have a 3x3 for(each) loop that yields the parameters

HHobeck:
> As far as I understand it's a combinatory issue, so you can have a 3x3 for(each) loop that yields the parameters

But you see that it is not always the same. It depends also on the scenario. I'm not sure if the focus will be lost. Here it is very clear one test case one result

arturcic:
> > As far as I understand it's a combinatory issue, so you can have a 3x3 for(each) loop that yields the parameters
> 
> But you see that it is not always the same. It depends also on the scenario. I'm not sure if the focus will be lost. Here it is very clear one test case one result.
> 
> Maybe give an example how it would look like.

something like

```csharp
private static object[] Cases()
{
	var incrementOnMainStrategies = new[] { IncrementStrategy.None, IncrementStrategy.Patch, IncrementStrategy.Minor, IncrementStrategy.Major };
	var incrementStrategies = new[] { IncrementStrategy.None, IncrementStrategy.Patch, IncrementStrategy.Minor, IncrementStrategy.Major, IncrementStrategy.Inherit };
	var label = new[] { null, "", "foo", "bar", "{BranchName}" };

	var cases = (
		from incrementOnMainStrategy in incrementOnMainStrategies
		from incrementStrategy in incrementStrategies
		from labelStrategy in label
		let expectedResult = "" // here is the computation of the expected result
		select new object[] { incrementOnMainStrategy, incrementStrategy, labelStrategy, expectedResult }).Cast<object>().ToArray();
	return cases;
}
```

and usage:
```csharp
[TestCaseSource(nameof(Cases))]
public string GetVersionWithNoLabelOnMain(IncrementStrategy incrementOnMain, IncrementStrategy increment, string? label)
```

arturcic:
I'd say give it a try and see if that can be done, and if that makes it more "readable"

arturcic:
@HHobeck  Now that the other PR got merged yo can rebase this one on top and we can continue the review

HHobeck:
> @HHobeck I wonder what should we do with this PR? Should we rebase on the latest changes on `main` and then we can work on getting it merged, or better to create a new PR based on the remaining changes we have here, and close this PR?

Yes I think we can close the PR and create new one

PR #3858, opened 2023-12-29
Docker documentation (v6)

asbjornu:
## Description
Add documentation for Docker usage.

## Motivation and Context
It's currently not documented how to use GitVersion through Docker.

## How Has This Been Tested?
I've executed the docs build locally.

Docker usage documentation

## Checklist:

### DIFF
--- a/docs/input/docs/usage/docker.md
+++ b/docs/input/docs/usage/docker.md
@@ -1,0 +1,27 @@
+---
+Order: 50
+Title: Docker
+Description: |
+    Use GitVersion through one of its many published Docker containers.
+---
+
+GitVersion can be used through one of its many published Docker
+containers. The list of available containers can be found on
+[Docker Hub][docker-hub]. Once you've found the image you want to use,
+you can run it like this:
+
+```shell
+docker run --rm --volume "$(pwd):/repo" gittools/gitversion:6.0.0-fedora.36-6.0 /repo
+```
+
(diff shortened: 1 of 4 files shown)


docs/input/docs/usage/docker.md
```diff
@@ -0,0 +1,27 @@
...
+    Use GitVersion through one of its many published Docker containers.
+---
+
+GitVersion can be used through one of its many published Docker
+containers. The list of available containers can be found on
+[Docker Hub][docker-hub]. Once you've found the image you want to use,
+you can run it like this:
+
+```shell
+docker run --rm --volume "$(pwd):/repo" gittools/gitversion:6.0.0-beta.3-fedora.36-6.0-arm64 /repo
```
arturcic, line 14: I would say in this example you can better use the manifest version, without specifying the arm64. It will pick the correct version depending on the platform

arturcic: And maybe we do not specify beta as by the time we get to V6 to be released I don't think we'll remember to change here. As far as I remember we have the tag to 6.0.0 already currently pointing to the beta4, but eventually it will point to the GA version

asbjornu: Good point. I changed it to ´gittools/gitversion:6.0.0-fedora.36-6.0`.

arturcic (approved):

Issue #2670, opened 2021-04-26
[Bug] the presence of an outdated `next-version` configuration breaks the version resolution


franknarf8:
**Describe the bug**

See discussion here : https://github.com/GitTools/GitVersion/discussions/2660

With this dummy git repo :

```
*   0bac05d 3 minutes ago  (HEAD -> master)
|\
| * 2922409 3 minutes ago  (feature/first-feature)
| * 2246a3e 7 minutes ago
|/
* bc00c9f 7 minutes ago  (tag: 1.2.3)
```

The following config resolves (incorrectly) to `1.2.3` :

```yaml
mode: Mainline
next-version: 1.0.0
branches:
  release:
    tag: 'rc'
  feature:
    regex: ^features?[/-]|^(pull|pull\-requests|pr)[/-]
    tag: 'dev'
  develop:
    tag: 'a'
  hotfix:
    tag: 'post'
ignore:
  sha: []
merge-message-formats: {}
```

whereas this config resolves (correctly) to `1.2.4` :

```yaml
mode: Mainline
branches:
  release:
    tag: 'rc'
  feature:
    regex: ^features?[/-]|^(pull|pull\-requests|pr)[/-]
    tag: 'dev'
  develop:
    tag: 'a'
  hotfix:
    tag: 'post'
```

## Expected Behavior

GitVersion should resolve to  `1.2.4` with both configurations

## Actual Behavior

The first config resolves (incorrectly) to `1.2.3`

## Possible Fix (Workaround)

Removing the `next-version` configuration to achieve the expected behavior.

## Context

See discussion : https://github.com/GitTools/GitVersion/discussions/2660

## Your Environment
GitVersion 5.6.8
Windows 10

mchippingtonderrick-afl:
Aha! I've been trying to track down a tricky versioning issue for months; and it seems like this may well be the culprit.

asbjornu:
@DavidDeSloovere, are you sure you're not experiencing #2441? If you are, please remove `next-version` and add a `git tag` with the same version number, on the appropriate commit.

DavidDeSloovere:
Yes. Looks very much like it. Thanks for figuring that out.

What if the way forward to set the "next version" then? Using commit messages?
I automatically tags the commits in master/main when build has succeeded. So I don't want to manually add a tag before the build, or I need to skip the auto tagging in the build if there is already a tag.

asbjornu:
[...]

Issue #2661, opened 2021-04-15
[Bug]


td0wn:
**Describe the bug**
Setting next-version prevents Mainline mode from counting direct & merge commits when computing the version, since the changes in commit c701080bfb1eb5ca611fed209d6b6f3f0caef2bb, I believe.

## Expected Behavior
The next-version setting should affect only the baseline version and the commit traversal in Mainline mode should still be executed, as was the previous behavior.

## Actual Behavior
No commits are traversed/recognized as bumping the version so the next-version is just used as the computed version.

Prior to this change the next-version setting was used just for determining the baseline version, but now that setting is being consulted for commit traversal in Mainline mode and resulting in different behavior.  I don't _think_ this was intentional? 

## Possible Fix
Don't check for next-version being null/empty when deciding whether to walk the commit graph in Mainline mode.  Basically I don't think this conditional check should be there: https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/VersionCalculation/MainlineVersionCalculator.cs#L55

## Context
Computed version when in Mainline mode and setting next-version results in different values when upgrading to versions released after c701080bfb1eb5ca611fed209d6b6f3f0caef2bb merged.

## Your Environment
- Version Used: 5.6.4
- Operating System and version (Windows 10, Ubuntu 18.04): Windows 10 & Ubuntu 20.04

asbjornu:
Duplicate of #2441. Please remove `next-version` and add a `git tag` with the same version number, on the appropriate tag.

Issue #2441, opened 2020-10-29
[Bug] GitVersion no longer calculates version correctly for mainline mode


techgeek03:
**Describe the bug**
Since version 5.5 when using Mainline mode GitVersion is not calculating the version at all and it gives ether the value in `next-version` or the version from the last tag.

## Expected Behavior
To have correct version

## Actual Behavior
Version is stacked at `next-version` or the version from the last tag

## Steps to Reproduce
Here is a simple GitVersion.yaml file I have been using
```
mode: Mainline
next-version: 1.0.0
branches:
  feature:
    tag: beta.{BranchName}
ignore:
  sha: []
merge-message-formats: {}
```

After many commits in `master` branch this is the output of GitVersion

```
INFO [10/29/20 23:06:38:48] Dumping commit graph:
INFO [10/29/20 23:06:38:64] *   3cdb26c 23 minutes ago  (HEAD -> master)
|\
| * 41f9c14 24 minutes ago  (feature/y2)
|/
*   9367673 73 minutes ago
|\
| * eb5c77a 76 minutes ago  (feature/test)
|/
* f957b18 78 minutes ago  (tag: 1.1.0)
* 318e307 79 minutes ago

INFO [10/29/20 23:06:38:98] Using latest commit on specified branch
INFO [10/29/20 23:06:39:02] Running against branch: master (3cdb26cf21767b34a3ecc65662fa00717857100a)
INFO [10/29/20 23:06:39:02] Begin: Calculating base versions
  INFO [10/29/20 23:06:39:07] NextVersion in GitVersion configuration file: 1.0.0 with commit count source External Source
  INFO [10/29/20 23:06:39:16] Git tag '1.1.0': 1.1.0 with commit count source f957b18b8557ba7d40a4e7a1a72f160c250e386f
  INFO [10/29/20 23:06:39:18] Found multiple base versions which will produce the same SemVer (1.1.1), taking oldest source for commit counting (Git tag '1.1.0')
  INFO [10/29/20 23:06:39:18] Base version used: Git tag '1.1.0': 1.1.0 with commit count source f957b18b8557ba7d40a4e7a1a72f160c250e386f
  INFO [10/29/20 23:06:39:18] End: Calculating base versions (Took: 158.30ms)
  INFO [10/29/20 23:06:39:18] Begin: Using mainline development mode to calculate current version
    INFO [10/29/20 23:06:39:20] Found possible mainline branches: master
[...]

Issue #1996, opened 2019-12-17
Infer next semantic version on any branch


nfantone:
I was drawn into `GitVersion` after reading the tag line in your `README.md`:

> GitVersion looks at your git history and works out the semantic version of the commit being built.

...but I couldn't quite make that work.

I've been trying to setup `GitVersion` to infer next version numbers on a project following GitFlow. However, I noticed that the only property that seems to matter when computing the next SemVer number is `increment`. The actual git log and commits messages are completely ignored.

Here's what I mean. Given a project with a latest tag of `0.4.0` and three more `fix:` commits added afterwards + the `GitVersion.yml` shown below:

```sh
$ git log

commit ee6cc4d1317588b62307f15e9ea55e0d0de24863 (HEAD -> develop, origin/develop, refs/pipelines/1336)
Date:   Mon Dec 16 20:51:02 2019 -0300

    fix(session): extend session timeout to six months

commit a38af3f6ea94573da234bf92635b6cfa844889a7
Date:   Mon Dec 16 20:50:00 2019 -0300

    fix(sfs): prevent common js files from being downloaded from sfs

commit da8c2d2f80dc608d77f4b9de1c6504116847000a
Date:   Mon Dec 16 08:54:51 2019 -0800

    fix: include in the attachment the file used to create encounter (#217699)

commit 1aeff2869e840dbe5029205cd869c68020b7a253
Merge: af74a2f79 98f9a8e23
Date:   Fri Dec 13 12:50:18 2019 -0800

    Merge tag '0.4.0' into develop

    foo-project 0.4.0

commit 98f9a8e2304b363a686f5aa78597ad903308d661 (tag: 0.4.0, origin/master, master)
Merge: b19e0cbe4 eb317b818
Date:   Fri Dec 13 12:50:18 2019 -0800

    Merge branch 'release/0.4.0'
```

```yml
# GitVersion.yml
major-version-bump-message: "BREAKING CHANGES?:"
minor-version-bump-message: "^(feature|feat)"
patch-version-bump-message: "^fix"
```

I would expect `GitVersion` to output `0.4.1` as the next version to be released when on `develop` branch (matching log above). Instead, it **always bumps the minor version**, producing releases such as `0.5.0`, `0.6.0`, `0.7.0`, etc. forced by the `increment` setting value.

[...]

PR #3883, opened 2024-01-25
Replace the version mode Mainline in 6.x (Part IV)

### LINKED ISSUE #1839: Configurable next-version strategies and mode
Running GitVersion, most of the time running, which can be quite long on large-ish codebases, is spent on looping through all version candidates on all versioning strategies on the codebase.

This includes traversing backwards using:
* FallbackBaseVersionStrategy
* ConfigNextVersionBaseVersionStrategy,
* TaggedCommitVersionStrategy
* MergeMessageBaseVersionStrategy
* VersionInBranchNameBaseVersionStrategy
* TrackReleaseBranchesVersionStrategy

Some of these are rather expensive. On our codebase at work, a lot of time is spent in MergeMessageBaseVersionStrategy. I don't think we need this, as we use tag-based versioning. We would benefit a lot being able to enable just the versioning strategy desired, alternatively a list of strategies. Not all strategies are applicable for everyone.

So, 1st suggestion is:

**Making the strategies used configurable**

2nd, I'd like to question if it is really necessary to loop through all previous commits in parents to find the "largest version" candidate. I am very open to being wrong here, but wouldn't the newest version found be the correct one to use for calculating the next version in 99% of all cases? At least, there could be a configurable setting to make it use the first/newest ancestor version on each strategy. 

2nd suggestion is, then:

**Make a config switch available to always use the newest version when calculating next**

I'd be happy to take on this task, but I might need a little guidance on where I should put the new config values. Does it make sense to put them on the branch? Or globally?

I have attached a run of GitVersion of a codebase I am working on. In all strategies, it finds the correct version to use for calculating "next version" (2019.21.0) in the first try. First is after 3 secs. But it still loops through all commits in all strategies, and takes and additional 60secs (30 after my other #1838). I'd like it to stop after finding the first candidate for a strategy.

[gitversion-diag.txt](https://github.com/GitTools/GitVersion/files/3683460/gitversion-diag.txt)

### LINKED ISSUE #3308: [Improvement] Replacing the mainline version mode with the continuous deployment mode
In the discussion of issue https://github.com/GitTools/GitVersion/issues/2388 ([Feature] Root configuration should propagate to branches) I came up with the idea of replacing the mainline version mode with the continuous deployment mode.

(see https://github.com/GitTools/GitVersion/pull/3235#discussion_r1030158052 and https://github.com/GitTools/GitVersion/issues/2388#issuecomment-1273310849)

## Detailed Description
I would like to not having a separate main line mode and reuse the ContinuousDeployment mode for this scenario as well. If we think about it: What is the different of ContinuousDeployment and Mainline when it has been defined on the main branch?

Definition in AWS https://aws.amazon.com/devops/continuous-delivery/:
> Continuous Delivery vs. Continuous Deployment
> With continuous delivery, every code change is built, tested, and then pushed to a non-production testing or staging environment. There can be multiple, parallel test stages before a production deployment. The difference between continuous delivery and continuous deployment is the presence of a manual approval to update to production. With continuous deployment, production happens automatically without explicit approval.

In my opinion there is no different from the business process point of view.

## Context
I think we should first do the breaking change on the versioning mode ContinuousDeployment and ensure that it behaves like the mainline mode. In a second step we can remove the mainline mode and replace it with ContinuousDeployment.

## Possible Implementation
The business rules are the following to increment the next version number (`ShouldIncrement` is true) in MergeMessageVersionStrategy:
- the PreventIncrementOfMergedBranchVersion of the effective branch configuration is false or
- CurrentCommit is not equals to MergeCommit and VersioningMode of the effective branch configuration is not equals to ContinuousDelivery

```csharp
    [Test]
    public void __Just_A_Test__()
    {
        var mainlineConfiguration = TestConfigurationBuilder.New
            .WithoutVersioningMode().WithTag("release", "beta").WithTag("main", "ci")
            .WithVersioningMode("release", VersioningMode.Mainline)
            .WithVersioningMode("main", VersioningMode.Mainline).Build();

        var continuousDeploymentConfiguration = TestConfigurationBuilder.New
            .WithoutVersioningMode().WithTag("release", "beta").WithTag("main", "ci")
            .WithVersioningMode("release", VersioningMode.ContinuousDeployment)
            .WithVersioningMode("main", VersioningMode.ContinuousDeployment).Build();

        var continuousDeliveryConfiguration = TestConfigurationBuilder.New
            .WithoutVersioningMode().WithTag("release", "beta").WithTag("main", "ci")
            .WithVersioningMode("release", VersioningMode.ContinuousDelivery)
            .WithVersioningMode("main", VersioningMode.ContinuousDelivery).Build();

        using var fixture = new EmptyRepositoryFixture("main");

        fixture.MakeACommit("one");
        fixture.BranchTo("release/1.0.0");
        fixture.MakeACommit("two");
        fixture.MakeACommit("three");

        fixture.Checkout("main");
        fixture.MergeNoFF("release/1.0.0");

        fixture.AssertFullSemver("1.0.0-ci", mainlineConfiguration);
        fixture.AssertFullSemver("1.0.0-ci.4", continuousDeploymentConfiguration);
        fixture.AssertFullSemver("1.0.0-ci.1+4", continuousDeliveryConfiguration);

        fixture.Repository.Branches.Remove("release/1.0.0");

        fixture.AssertFullSemver("1.0.0-ci", mainlineConfiguration); // <<---- the same sem version like ContinuousDeployment
        fixture.AssertFullSemver("1.0.0-ci.4", continuousDeploymentConfiguration);
        fixture.AssertFullSemver("1.0.0-ci.1+4", continuousDeliveryConfiguration);

        fixture.MakeACommit("four");

        fixture.AssertFullSemver("1.0.1-ci", mainlineConfiguration); // <<---- the same sem version like ContinuousDeployment
        fixture.AssertFullSemver("1.0.1-ci.1", continuousDeploymentConfiguration); // <<---- this is important
        fixture.AssertFullSemver("1.0.0-ci.1+5", continuousDeliveryConfiguration); // <<---- this is important
    }
```

Edit:
Final design can be found here:
- #3601

### LINKED ISSUE #3877: Add configuration option to disable branch name version calculation
We are starting to create releases based off actual release branches.  The desire is to name them after the sprint they are finished in.  24.01 for example.  Using that in the branch name causes GitVersion to use it as a version number as documented here:

https://gitversion.net/docs/reference/version-increments

I'd simply like a way to disable that feature.  

## Context
Adding some alpha prefix works great, unless you or someone else on your team forgets.  It would just be nice to have that feature as optional.

## Possible Implementation
Add an option to the config file:

merge-message-formats: {}
branches:
  release:
    regex: ^(releases?|releases?[/-])
    mode: ContinuousDelivery
    tag: ''
    increment: Patch
    prevent-increment-of-version-based-on-name: true

HHobeck:
## Description
Create TrunkBasedVersionStrategy implementation and replace the version mode Mainline in 6.x.

Close #3877 
Close #1839 
Close #3308

## Motivation and Context
For more detail please see the following ADR:
- #3601

## Next steps are:
- Create TrunkBasedVersionStrategy implementation (done)
- Create integration tests for TrunkBasedVersionStrategy (I skip this because of the limit time of reviewer)
- Remove Mainline mode (done)
- Replace VersioningMode with DeploymentMode (done)
- **Add new property with name strategies (type VersionStrategies[]) to the configuration root level** <<-- Part of this PR
- **Create documentation for TrunkBased workflow and add breaking changes notes** <<-- Part of this PR

## Checklist:


BREAKING_CHANGES.md
```diff
@@ -32,12 +32,21 @@
...
     *   Default `RegularExpression` for unknown branches is changed from `.*` to `(?<BranchName>.*)` to support using `{BranchName}` out-of-the-box
+*   The `Mainline` mode and the related implementation has been removed completely. The new `TrunkBased` version strategy should be used instead.
 *   The branch related property `is-mainline` in the configuration system has been renamed to `is-main-branch`
 *   The versioning mode has been renamed to deployment mode and consists of following values:
     *   ManualDeployment (previously ContinuousDelivery)
     *   ContinuousDelivery (previously ContinuousDeployment)
     *   ContinuousDeployment (new)
-	
+*   On the configuration root level a new property with name `version-strategy` has been introduced with following values:
+    *   ConfigNext
```
asbjornu, line 42: What's the rationale behind the name `ConfigNext`?

BREAKING_CHANGES.md
```diff
@@ -32,12 +32,21 @@
...
+    *   MergeMessage
+    *   TaggedCommit
+    *   TrackReleaseBranches
```
asbjornu, line 45: I'm a bit unsure about the name `TrackReleaseBranches`. We should be consistent with whether we use a verb (`Track`) in the enum values  and whether we use singular (`Commit`) or plural (`Branches`) form.

BREAKING_CHANGES.md
```diff
@@ -32,12 +32,21 @@
...
+    *   VersionInBranchName
+    *   NonTrunkBased = ConfigNext | MergeMessage | TaggedCommit | TrackReleaseBranches | VersionInBranchName,
```
asbjornu, line 47: I'm not a fan of `NonTrunkBased`. Can we find another word that has the opposite meaning and thus doesn't require the `Non` prefix?

docs/input/docs/reference/configuration.md
```diff
@@ -53,6 +53,7 @@ commit-date-format: yyyy-MM-dd
 merge-message-formats: {}
 update-build-number: true
 semantic-version-format: Strict
+version-strategy: NonTrunkBased
```
asbjornu, line 56: Can we just call the property `strategy`?

```suggestion
strategy: NonTrunkBased
```

src/GitVersion.Configuration/SupportedWorkflows/GitFlow/v1.yml
```diff
@@ -10,6 +10,7 @@ commit-date-format: yyyy-MM-dd
 merge-message-formats: {}
 update-build-number: true
 semantic-version-format: Strict
+versioning-strategy: 'ConfigNext,MergeMessage,TaggedCommit,TrackReleaseBranches,VersionInBranchName'
```
asbjornu, line 13: YAML supports arrays, so we should use those instead.

```suggestion
versioning-strategy: [ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName]
```

src/GitVersion.Configuration/SupportedWorkflows/TrunkBased/v1.yml
asbjornu, line 1: Why is the folder called `SupportedWorkflows` when it's associated with a configuration property called `version-strategy`? Either we should rename the YAML property to `workflow` or we should rename the folder to `Strategies`.

src/GitVersion.Core/VersionCalculation/VersionStrategies.cs
```diff
@@ -0,0 +1,14 @@
...
+public enum VersionStrategies
+{
+    None = 0,
+    ConfigNext = 1,
+    MergeMessage = 2,
+    TaggedCommit = 4,
+    TrackReleaseBranches = 8,
+    VersionInBranchName = 16,
+    NonTrunkBased = ConfigNext | MergeMessage | TaggedCommit | TrackReleaseBranches | VersionInBranchName,
+    TrunkBased = 32,
```
asbjornu, line 13: I think combined flag values should be the last value in the enum.

```suggestion
    TrunkBased = 32,
    NonTrunkBased = ConfigNext | MergeMessage | TaggedCommit | TrackReleaseBranches | VersionInBranchName,
```

asbjornu (changes requested):

BREAKING_CHANGES.md
HHobeck: Good question: I have taken the name of the stragies and removed the trailing VersionStrategy postfix.
- ConfigNextVersionVersionStrategy
- MergeMessageVersionStrategy
- TaggedCommitVersionStrategy
- TrackReleaseBranchesVersionStrategy
- TrunkBasedVersionStrategy
- VersionInBranchNameVersionStrategy

asbjornu: But what is `ConfigNext`? The name should give a clear indication of what it does. To me, `ConfigNext` means nothing. 🤷🏼

BREAKING_CHANGES.md
HHobeck: Of course we can change it. Actually it's just a abbreviation for `ConfigNext,MergeMessage,TaggedCommit,TrackReleaseBranches,VersionInBranchName `for better usability. Do you have something in mind?

BREAKING_CHANGES.md
HHobeck: I'm open to any renaming as long we are renming the underlying class who implements the strety as well.

src/GitVersion.Configuration/SupportedWorkflows/GitFlow/v1.yml
HHobeck: But this is not an array it is a flags enum. Should I change the type from `VersionStrategies` to `VersionStrategies[]`? It feels a little bit repeated unnecessarily.

BREAKING_CHANGES.md
asbjornu: What about just calling it `Default`?

src/GitVersion.Configuration/SupportedWorkflows/TrunkBased/v1.yml
HHobeck: Hmm correct we have a clash with the term TrunkBased here... What is you suggestion? In the folder `SupportedWorkflows` we are talking from workflow where in the strategy we are talking from strategy which is something different.

We could define e.g. on the workflow template  `SupportedWorkflows/TrunkBased/v1.yml` to use the strategy `TrunkBased,NextVersion` to support the next-version configuration as well if it is applicable. Thus there is no one to one relation from workspace to strategy. It's up to you how you would define it.

BREAKING_CHANGES.md
HHobeck: That means you are saing in other words: In git version the strategy classes have not a clear name to get an indecation of for what it is good for!? Oka. It's the core domain and the most important code if you ask me. Thus it makes sense to refactor it.

I can tell you what ConfigNextVersionVersionStrategy is doing. I takes the next-version configuration property if existent and returns it is as next-version.

How would you name it?

BREAKING_CHANGES.md
HHobeck: In my opinion default is everything and nothing. I would say to delete it completely.

docs/input/docs/reference/configuration.md
HHobeck: I name it strategies intead because it is an array data type.

src/GitVersion.Core/VersionCalculation/VersionStrategies.cs
HHobeck: done

src/GitVersion.Configuration/SupportedWorkflows/GitFlow/v1.yml
HHobeck: done

docs/input/docs/reference/configuration.md
HHobeck: done

BREAKING_CHANGES.md
HHobeck: done

BREAKING_CHANGES.md
asbjornu: Aha, I see. I'm not sure that `next-version` should have its own strategy because it's a configuration parameter and not the Git repository that affects the version number. It would be a bit strange to have `next-version` configured, but not have the `ConfigNext` strategy configured, leading to `next-version` being ignored. I think the existence of `next-version` should imply that it should be considered for version calculation.

BREAKING_CHANGES.md
asbjornu: Is `TrackReleaseBranches` really a versioning strategy, though? Isn't it more like a branch configuration?

docs/input/docs/reference/configuration.md
```diff
@@ -53,6 +53,8 @@ commit-date-format: yyyy-MM-dd
 merge-message-formats: {}
 update-build-number: true
 semantic-version-format: Strict
+strategies:
+- ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName
```
asbjornu, line 57: Arrays in YAML are either wrapped in `[]`:

```suggestion
strategies: [ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName]
```

…or on separate lines starting with `-`:

```suggestion
strategies:
  - ConfigNext
  - MergeMessage
  - TaggedCommit
  - TrackReleaseBranches
  - VersionInBranchName
```

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
```diff
@@ -11,6 +11,8 @@ commit-date-format: yyyy-MM-dd
 merge-message-formats: {}
 update-build-number: true
 semantic-version-format: Strict
+strategies:
+- ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName
```
asbjornu, line 15:
```suggestion
strategies: [ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName]
```

src/GitVersion.Configuration.Tests/Configuration/Init/InitScenarios.CanSetNextVersion.approved.txt
```diff
@@ -2,6 +2,8 @@ next-version: 2.0.0
 merge-message-formats: {}
 update-build-number: true
 semantic-version-format: Strict
+strategies:
+- None
```
asbjornu, line 6:
```suggestion
strategies: [None]
```

src/GitVersion.Configuration/ConfigurationBuilderBase.cs
```diff
@@ -24,6 +24,7 @@ internal abstract class ConfigurationBuilderBase<TConfigurationBuilder> : IConfi
     private string? commitDateFormat;
     private bool updateBuildNumber;
     private SemanticVersionFormat semanticVersionFormat;
+    private VersionStrategies[] versionStrategies;
```
asbjornu, line 27: I don't think we should use an array for the `enum`. We should only use an array in the YAML file. There's a bit of relevant discussion in https://github.com/aaubry/YamlDotNet/issues/679. We may have to rely on custom deserialization logic for this specific property. The goal is to make YAML as developer-friendly as possible, without negatively affecting the `GitVersionConfiguration` domain model in C#.

```suggestion
    private VersionStrategies versionStrategies;
```

src/GitVersion.Core/VersionCalculation/VersionStrategies.cs
```diff
@@ -0,0 +1,13 @@
...
+[Flags]
+public enum VersionStrategies
+{
+    None = 0,
+    ConfigNext = 1,
+    MergeMessage = 2,
+    TaggedCommit = 4,
+    TrackReleaseBranches = 8,
+    VersionInBranchName = 16,
+    TrunkBased = 32,
```
asbjornu, line 12: Where did `Default` go?
```suggestion
    TrunkBased = 32,
    Default = ConfigNext | MergeMessage | TaggedCommit | TrackReleaseBranches | VersionInBranchName,
```

asbjornu (changes requested):

BREAKING_CHANGES.md
asbjornu: `Default` would be what GitVersion would use if `strategies` is missing in `GitVersion.yml` and would be the same as `NonTrunkBased`. I think that makes sense, no?

BREAKING_CHANGES.md
HHobeck: You mean I should delete the ConfigNextVersionVersionStrategy class and place the logic somewhere else?

BREAKING_CHANGES.md
HHobeck: At the moment it is a version strategy yes. What is your suggestion!? To delete it? I thought this is the concept of our core domain to have provider based (strategy based) approach to determine the next version. And you can control whether or not it is called using the strategies property on the configuration root level.

BREAKING_CHANGES.md
HHobeck: No that is not true default is empty. Because we have implemented it in the way it is abwarts compatible by using the GitFlowConfigurationBuilder if no workspace is defined. I future all configuration should have a workspace definition.

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
HHobeck: Actually the deserialize gives me this format. I have this not in control

src/GitVersion.Configuration.Tests/Configuration/Init/InitScenarios.CanSetNextVersion.approved.txt
HHobeck: I have changed it to empty

src/GitVersion.Core/VersionCalculation/VersionStrategies.cs
HHobeck: We have no default

HHobeck: Please forget the idea of trying to sum this up. This is the wrong place. We need to configure this semantically in the workspace template.

BREAKING_CHANGES.md
asbjornu: I think so. There's already a lot of confusion about how `next-version` works, because it doesn't interoperate with the other strategies the way people expect. I think we need to integrate `next-version` differently somehow, and by doing that, fix the many issues people have with it. See #303 for how it was first implemented, then the following issues caused by confusion about how it works: #937, #1474, #1996, #2441, #2661, and #2670.

If we are unable to fix any of those issues in v6, I think we should just remove the property altogether (but not in this PR).

asbjornu:
Relevant to this PR: #1839.

BREAKING_CHANGES.md
asbjornu: Yes, I like the configurable strategy approach. But like `next-version`, `tracks-release-branches` is a configuration property. It's strange to have `tracks-release-branches` configured for a branch, but ignored because `TrackReleaseBranches` isn't configured.

src/GitVersion.Configuration/ConfigurationBuilderBase.cs
HHobeck: done

BREAKING_CHANGES.md
asbjornu: You're right about `Default` being wrong here. I still think we should have an enum value for it, I just don't like the `NonTrunkBased` name. How about `BranchBased`?

With `workspace`, I suppose you're meaning `workflow`? If so, I agree.

src/GitVersion.Configuration/SupportedWorkflows/TrunkBased/v1.yml
asbjornu: Yeah, the overlap between `strategy` and `workflow` is a bit confusing. Would it be possible to remove `strategy` completely from configuration and only allow `workflow` to be configured? Perhaps not in v6, but in v7?

docs/input/docs/reference/configuration.md
HHobeck: done

src/GitVersion.Core/VersionCalculation/VersionStrategies.cs
asbjornu: Yeah, I agree. Moving this to `workflow` is a much better direction.

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
asbjornu: I know. Which is why we want to separate our domain classes from DTO objects used for de/serialization. We can just have a private class that we use for serialization and deserialization, close to where it is used in the perimeter of the codebase, and then map from that class to the core domain object. In our mapping code, we will do the translation between array and `[Flags] enum`.

BREAKING_CHANGES.md
HHobeck: Also here I would say like I already did previously: You need to think of the property like a characteristic of an branch or a characteristic of the repository which is generally isolated of the actual used strategies. If someone decide e.g. not to use the NextVersionVersionStrategy but define next-version then it would be fine to ignore it.

And you second example is even clear to argument. The track-release-brachnes property will be used in the TrackReleaseBranchesVersionStrategy and TrunkBaseVersionStrategy. If you decide to use TrunkBaseVersionStrategy you should not execute the TrackReleaseBranchesVersionStrategy because it gives you maybe a wrong behavior for the TrunkBased (Mainline) workflow.

Does it makes sense?

BREAKING_CHANGES.md
HHobeck: In my opinion the concept is quite clear and works very well now also with the TrunkBased workflow. Previously we have trouble in the Mainline mode but this problem doesn't remain. Thus I don't see any reason why we should remove the feature. Just my view: The next-version only pumps to a higher version if the next calculated version is lower.

But I agree it's not part of this PR. ;)

BREAKING_CHANGES.md
HHobeck: Yes I mean workflow. You are right at the moment we have no workspace term in our domain. Okay that is fine with me to name it BranchBased. What about the confusion with TrunkBased on workflow and strategy level? I thought maybe we should name the workflow TrunkBased and the strategy Mainline.

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
HHobeck: I think you are mistaken here. This is equivalent in yaml and the property is of type VersionStrategies[] already:

```yaml
strategies:
- ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName

strategies:
- ConfigNext
- MergeMessage
- TaggedCommit
- TrackReleaseBranches
- VersionInBranchName
```

BREAKING_CHANGES.md
asbjornu: I think we should use either `TrunkBased` or `Mainline`, not both. Would using `workflow: TrunkBased` be different from `strategies: [TrunkBased]`? If so, in what way?

src/GitVersion.Configuration/SupportedWorkflows/TrunkBased/v1.yml
asbjornu: Good point wrt. allowing configurability on individual strategies. So we probably want `workflow` and `strategies` to be mutually exclusive, no? If not, how would `workflow: TrunkBased` with `strategies: [TaggedCommit]` work?

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
asbjornu: Hm, I've never seen arrays like this before:

```yaml
strategies:
- ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName
```

I don't see it [in this documentation of YAML arrays](https://www.w3schools.io/file/yaml-arrays/) either. And if you try to [this online converter](https://onlineyamltools.com/convert-yaml-to-json), it turns this:

yaml

into this:

json

As you can see from the converted JSON:

```json
{
  "strategies": [
    "ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName"
  ]
}
```

`strategies` here has a single item which is the comma separated string `ConfigNext, MergeMessage, TaggedCommit, TrackReleaseBranches, VersionInBranchName`. That's not what we want, is it?

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
```diff
@@ -11,6 +11,12 @@ commit-date-format: yyyy-MM-dd
 merge-message-formats: {}
 update-build-number: true
 semantic-version-format: Strict
+strategies:
+- ConfigNextVersion
```
asbjornu, line 15: How about `ConfiguredNextVersion`?

```suggestion
- ConfiguredNextVersion
```

src/GitVersion.Configuration/GitVersionConfiguration.cs
```diff
@@ -124,6 +124,14 @@ private string GetVersionInBranchPattern()
     [JsonPropertyDefault(DefaultSemanticVersionFormat)]
     public SemanticVersionFormat SemanticVersionFormat { get; internal set; }
 
+    [JsonIgnore]
+    VersionStrategies IGitVersionConfiguration.VersionStrategy => VersionStrategies.Length == 0
+        ? VersionCalculation.VersionStrategies.None : VersionStrategies.Aggregate((one, another) => one | another);
+
+    [JsonPropertyName("strategies")]
+    [JsonPropertyDescription($"Specifies which version strategies (one or more) will be used to determine the next version. Following values are available: 'ConfigNextVersion', 'MergeMessage', 'TaggedCommit', 'TrackReleaseBranches', 'VersionInBranchName' and 'TrunkBased'.")]
+    public VersionStrategies[] VersionStrategies { get; internal set; } = [];
```
asbjornu, line 133: I don't understand how it works that `VersionStrategies` is an array where but not in the `IGitVersionConfiguration` interface.

asbjornu (commented):
I think we are close now. The YAML array vs. flags enum thing can probably be fixed in a separate PR where we:

1. Remove all `[Json…]` attributes from `GitVersionConfiguration`.
2. Add a new private class inside `GitVersion.Configuration.ConfigurationSerializer` called `GitVersionConfigurationDto` that is decorated with deserialization attributes and tailored towards simple de/serialization from/to YAML.
3. Change the implementation of `GitVersion.Configuration.ConfigurationSerializer.Read()` to deserialize to `GitVersionConfigurationDto` and then map from the DTO to the `GitVersionConfiguration` with manual handling of the array to `Flags enum` conversion.
4. Change the implementation of  `GitVersion.Configuration.ConfigurationSerializer.Write()` to first map from the received `IGitVersionConfiguration` object to `GitVersionConfigurationDto` and then serialize the DTO to disk.

BREAKING_CHANGES.md
HHobeck: > I think we should use either `TrunkBased` or `Mainline`, not both. Would using `workflow: TrunkBased` be different from `strategies: [TrunkBased]`? If so, in what way?

Actually I'm quite at the beginning to understand what a trunk-based workflow is and how it should be ideally behave. That means this is just a beginng of an jorney for me and I think the workflow will evolve when the feedback is comming back from the community. At this moment I think a TrunkBased workflow should use the TrunkBasedVersionStrategy and the ConfiguredNextVersionVersionStrategy. But this can be changed anytime in future. Please notice that I have changed the workflow TrunkBased/v1.yaml already.

src/GitVersion.Configuration/SupportedWorkflows/TrunkBased/v1.yml
HHobeck: > Good point wrt. allowing configurability on individual strategies. So we probably want `workflow` and `strategies` to be mutually exclusive, no? If not, how would `workflow: TrunkBased` with `strategies: [TaggedCommit]` work?

The VersionStrategies pattern is the main mechanism to control which strategy will be used to determine the next version. This can be of course changed by the user. Like every time if the user changes the configuration and customizes their own workflow he needs to ensure the correct behavior as well. Because it is hard work to build a workflow we as contributors are providing a supported workflow templates (at this moment GitHubFlow, GitFlow and TrunkBased and maybe more in future) to help the user get there build pipe up and running. And for this workflows we give a support.

Comming to your question: You only can answer this question if it is working or not if you know what workflow the user wants to implement. Because we don't know what the workflow might be we cannot do any assumptions about it. In best case it has just a performance impact in worst case the user gets a version number he would not expect.

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
HHobeck: Okay I think I need to take a look. That is indeed very strange

src/GitVersion.Configuration/GitVersionConfiguration.cs
HHobeck: image

From IGitVersionConfiguration.VersionStrategy you call VersionStrategies property.

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
HHobeck: At the moment it looks like the following:

image

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
HHobeck: done

BREAKING_CHANGES.md
asbjornu: I didn't think of the possibility of `tracks-release-branches` being used by more than one strategy. That's not the case for `next-version`, is it?

src/GitVersion.Configuration/SupportedWorkflows/TrunkBased/v1.yml
asbjornu: Right, I just want us to be on the same page here and offer the best possible developer experience when interacting with these new configuration properties. That's why I thought `strategies` and `workflow` could be made mutually exclusive; you either use one or the other. GitVersion would simply return an error if both properties are present in `GitVersion.yml`. 

I think that's going to lead to fewer problems (i.e. fewer filed issues on GitHub) than allowing developers to mix `workflow` and `strategies`. From my years of experience maintaining API surfaces like this, it's wise to start out restrictive and increase flexibility over time. If we start out flexible, allowing `strategies` and `workflow` to be mixed, it's going to be much harder adding restrictions on their usage in the future.

If we are to allow `strategies` and `workflow` simultaneously, we need to figure out whether `strategies` are additive or reductive to the ones implicitly added by `workflow`, for instance. I say we just avoid having to answer that question by not allowing it.

src/GitVersion.Configuration/GitVersionConfiguration.cs
asbjornu: Ah, I didn't notice that. Okay, that's fine for now. But I would like to see the split between the core domain model and DTO used for de/serialization in a future PR. I might contribute that, if time allows.

src/GitVersion.Core.Tests/VersionCalculation/Strategies/ConfigNextVersionBaseVersionStrategyTests.cs
```diff
@@ -7,7 +7,7 @@
 namespace GitVersion.Core.Tests.VersionCalculation.Strategies;
 
 [TestFixture]
-public class ConfigNextVersionBaseVersionStrategyTests : TestBase
+public class ConfiguredNextVersionBaseVersionStrategyTests : TestBase
```
asbjornu, line 10: It seems like the class name here contains `BaseVersion` for no particular reason?
```suggestion
public class ConfiguredNextVersionStrategyTests : TestBase
```

asbjornu (approved):
Great work, @HHobeck!

src/GitVersion.Configuration.Tests/Configuration/ConfigurationProviderTests.CanWriteOutEffectiveConfiguration.approved.txt
asbjornu: Yep, that looks good.

BREAKING_CHANGES.md
HHobeck: Sorry I was mistaken. I have mixed `tracks-release-branches` with `track-merge-message` in my mind. You are right `tracks-release-branches` will be used in `TrackReleaseBranchesVersionStrategy` and `next-version` in `ConfiguredNextVersionVersionStrategy` only.

src/GitVersion.Configuration/SupportedWorkflows/TrunkBased/v1.yml
HHobeck: I'm actually not a big fan of being a guardian for the user and protecting them for themselves. I would say to document it and make a attention stign that only advanced user who knows what they are doing should change this field.

Anyway if we do it like you mentioned how would you solve the following issued?
- https://github.com/GitTools/GitVersion/issues/3877
- https://github.com/GitTools/GitVersion/issues/1839

src/GitVersion.Core.Tests/VersionCalculation/Strategies/ConfigNextVersionBaseVersionStrategyTests.cs
HHobeck: done

HHobeck:
Actually what I like to discuss with you @asbjornu: With this PR the pump messages `+semver: none` is not usable anymore. What ware we doing with it?

PR #3898, opened 2024-02-06
[Feature] Same version computed on different branches

### LINKED ISSUE #3453: [Feature] Same version computed on different branches

HHobeck:
[Feature] Same version computed on different branches:
- #3453

Close #3453

## Checklist:


HHobeck:
Hi @asbjornu. I'm not sure if you are agree with the new branch related property with name `take-incremented-version`. Please take a look and give me feedback.

Thank's

docs/input/docs/reference/configuration.md
```diff
@@ -52,16 +53,22 @@ commit-date-format: yyyy-MM-dd
...
+strategies:
+- ConfiguredNextVersion
+- MergeMessage
+- TaggedCommit
+- TrackReleaseBranches
+- VersionInBranchName
 branches:
   develop:
-    mode: ContinuousDeployment
+    take-incremented-version: TakeAlwaysIncrementedVersion
```
asbjornu, line 64: Hm, I think `take-incremented-version: TakeAlwaysIncrementedVersion` have room for improvements. Give me some time to think through the use case and see if I can come up with better names.

asbjornu (changes requested):

asbjornu:
Do I understand this issue correctly when it's about whether we want to consider the `git tag` of the current commit or not? If so, perhaps `ignore-tag: False` would be more intuitive?

HHobeck:
> Do I understand this issue correctly when it's about whether we want to consider the `git tag` of the current commit or not? If so, perhaps `ignore-tag: False` would be more intuitive?

At the end the criteria you are mentioned is correct. But you need to think about the semantic meaning. We have one or more strategies who are determining the existing base versions with a indication should the base increment yes or no. We are calculating based on this the next version and take the one which is highest. The result is we have exactly one base version and one incremented version (which might be equal). Now this feature comes in place where we want for some branches not to use the incremented version but the base version. 

I can think of the following use cases:
- Take always the base version and not the incremented version,
- Take the base version when the current commit was released otherwise the incremented version,
- Take always the incremented version and not the base version

Your proposed property with name `ignore-tag` does not reflect this semantic meaning if you ask me. It's also not about ignoring tags because the current tag will be considered and used to calculate the incremented version. If someone uses the `ignore-tag` property he could expect that all tags are ignored on the branch and not considered for the version calculation (like you would specify to ignore commits).

Does it make sense?

HHobeck:
> I can think of the following use cases:
> 
> - Take always the base version and not the incremented version,
> - Take the base version when the current commit was released otherwise the incremented version,
> - Take always the incremented version and not the base version

While writing this I came up with the problem that we needs to extend the criteria e.g. for trunk based workflows. Because on trunk based workflows the released version on main will be not always marked with a tag (it's optional). So the criteria might:

- The current commit was tagged or the current branch IsMain (at least when trunk-based stratregy was used)

asbjornu:
Right. But wouldn't then `increment: None` work, since for the specific branch we don't want to increment either `Major`, `Minor` or `Patch`?

HHobeck:
> Right. But wouldn't then `increment: None` work, since for the specific branch we don't want to increment either `Major`, `Minor` or `Patch`?

But we want to increment the version on the branch but only if the commit was not released (marked with a tag).

HHobeck:
Maybe the following enum values are better:

```csharp
internal enum TakeIncrementedVersion
{
    Never,
    IfNotReleased/Tagged,
    Always
}
```

docs/input/docs/reference/configuration.md
asbjornu: Let's do as discussed on Slack and group this and `prevent-increment-of-merged-branch-version` into a new top-level property as such:

```yaml
prevent-increment:
    when-versioned-branch-merged: true
    when-tagged: true
```

HHobeck:
> ```yaml
> prevent-increment:
>     when-versioned-branch-merged: true
>     when-tagged: true
> ```

Actually I would like to do this in a separate PR if you not mind. Please review the changes and merge it to main. Thank you.

HHobeck:
I think this PR can be merged right!?

src/GitVersion.Core.Tests/IntegrationTests/DevelopScenarios.cs
```diff
@@ -46,7 +46,7 @@ public void WhenDevelopBranchedFromTaggedCommitOnMainVersionDoesNotChange()
         using var fixture = new EmptyRepositoryFixture();
         fixture.MakeATaggedCommit("1.0.0");
         fixture.BranchTo("develop");
-        fixture.AssertFullSemver("1.0.0");
+        fixture.AssertFullSemver("1.1.0-alpha.0");
```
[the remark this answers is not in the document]
asbjornu: Indeed! :)

asbjornu (approved):

Issue #3229, opened 2022-10-14
[Improvement] Revamp init wizard for v6


asbjornu:
If we want to keep the init wizard in v6, it needs a major revamp. This issue exists to track that either a full revamp of the init wizard, or its complete removal from v6, is done before the release of `6.0.0`.

If no one invests time to revamp the init wizard completely until the release of `6.0.0`, it will be gone.

_Originally posted by @asbjornu in https://github.com/GitTools/GitVersion/pull/3226#discussion_r993566889_

PR #3899, opened 2024-02-06
[Improvement] Revamp init wizard for v6

HHobeck:
[Improvement] Revamp init wizard for v6
- #3229

## Checklist:


HHobeck:
> I still think would be nice to be able from the `gitversion init` to be able to configure the workflows and all the options (now that we can Builders)

I agree it would be nice. If you like I can help you with the design. Would you like to implement this?

HHobeck:
I understand. The same on my side. I'm going to create a feature request and put it into 7.x so we have a reminder.

PR #3965, opened 2024-03-17
Extend MergeMessageVersionStrategy to support squash commits

DSpirit:
Extended the `TryParse` operation to include a regex check whether the commit message contains one of the specified CommitMessageFormats in order to support increment of merges from release into main branch when a squash merge has been applied.

## Motivation and Context

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/GitVersion.Core/MergeMessage.cs
+++ b/src/GitVersion.Core/MergeMessage.cs
@@ -98,7 +98,11 @@ public class MergeMessage
 
         mergeMessage = null;
 
-        if (mergeCommit.IsMergeCommit)
+        var mergeMessageFormats = DefaultFormats.Union(
+                configuration.MergeMessageFormats
+                    .Select(n => new MergeMessageFormat(n.Key, n.Value)));
+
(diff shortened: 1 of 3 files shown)


src/GitVersion.Core/MergeMessage.cs
```diff
@@ -98,7 +98,11 @@ private ReferenceName GetMergedBranchName(string mergedBranch)
 
         mergeMessage = null;
 
-        if (mergeCommit.IsMergeCommit)
+        var mergeMessageFormats = DefaultFormats.Union(
+                configuration.MergeMessageFormats
+                    .Select(n => new MergeMessageFormat(n.Key, n.Value)));
+
+        if (mergeCommit.IsMergeCommit || mergeMessageFormats.Any(format => format.Pattern.IsMatch(mergeCommit.Message)))
```
asbjornu, line 105: Can we expose this as `Commit.HasMergeMessage(…)` instead?

```suggestion
        if (mergeCommit.IsMergeCommit || mergeCommit.HasMergeMessage(mergeMessageFormats))
``````

asbjornu (changes requested):
This was less code than I expected. If the tests pass, I give this a 👍🏼.

PR #3966, opened 2024-03-17
Feature/merge message releases

DSpirit:
Extended the `TryParse` operation to check whether the commit source branch is a release branch in order to support increment of merges from release into main branch when a squash merge has been applied.

## Motivation and Context

## How Has This Been Tested?

❗ I had to comment one test, as the `hotfix-0.1.5` branch was somehow configured as a release branch and thus, the version was taken into respect. Would appreciate some feedback here!

## Checklist:

### DIFF
--- a/src/GitVersion.Core.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
+++ b/src/GitVersion.Core.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
@@ -52,6 +52,7 @@ public class MergeMessageBaseVersionStrategyTests : TestBase
     [TestCase("Merge branch 'Release-v0.2.0'", true, "0.2.0")]
     [TestCase("Merge remote-tracking branch 'origin/release/0.8.0' into develop/" + MainBranch, true, "0.8.0")]
     [TestCase("Merge remote-tracking branch 'refs/remotes/origin/release/2.0.0'", true, "2.0.0")]
+    [TestCase("Merge branch 'Releases/0.2.0'", false, "0.2.0")] // Support Squash Commits
     public void TakesVersionFromMergeOfReleaseBranch(string message, bool isMergeCommit, string expectedVersion)
     {
         var parents = GetParents(isMergeCommit);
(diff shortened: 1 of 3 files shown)


src/GitVersion.Core/MergeMessage.cs
```diff
@@ -96,13 +96,10 @@ private ReferenceName GetMergedBranchName(string mergedBranch)
         mergeCommit.NotNull();
         configuration.NotNull();
 
-        mergeMessage = null;
+        mergeMessage = new(mergeCommit.Message, configuration);
```
asbjornu, line 99: `mergeMessage` should not be set if the return value equals `false`.

src/GitVersion.Core.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
```diff
@@ -65,7 +66,7 @@ public void TakesVersionFromMergeOfReleaseBranch(string message, bool isMergeCom
         AssertMergeMessage(message + "\n ", expectedVersion, parents);
     }
 
-    [TestCase("Merge branch 'hotfix-0.1.5'", false)]
+    // [TestCase("Merge branch 'hotfix-0.1.5'", false)] this is somehow configured as a release branch
```
asbjornu, line 69: Can you please elaborate on how `hotfix-0.1.5` is configured as a release branch? Is `is-release-branch` set to `true`?

asbjornu (changes requested):
I think this looks mostly fine!

src/GitVersion.Core.Tests/VersionCalculation/Strategies/MergeMessageBaseVersionStrategyTests.cs
DSpirit: Yes, the `is-release-branch` is set to `true`.
Here's an excerpt of the generated configuration:

``` yaml
hotfix:
      mode: ManualDeployment
      label: beta
      increment: Inherit
      prevent-increment:
        when-current-commit-tagged: false
      regex: ^hotfix(es)?[/-](?<BranchName>.+)
      source-branches:
      - release
      - main
      - support
      - hotfix
      is-source-branch-for: []
      is-release-branch: true
      pre-release-weight: 30000
```

Unfortunately I can't tell whether this is on purpose or not.

src/GitVersion.Core/MergeMessage.cs
DSpirit: Understood, I've changed it.

DSpirit:
Closing this PR in favour to https://github.com/GitTools/GitVersion/pull/3984 with all current changes from master.

PR #3984, opened 2024-03-29
Add support for Squash Commit releases

DSpirit:
Extended the `TryParse` operation to check whether the commit source branch is a release branch in order to support increment of merges from release into main branch when a squash merge has been applied.

## Motivation and Context

## How Has This Been Tested?

## Checklist:

### DIFF
--- a/src/GitVersion.Core/MergeMessage.cs
+++ b/src/GitVersion.Core/MergeMessage.cs
@@ -97,14 +97,15 @@ public class MergeMessage
         mergeCommit.NotNull();
         configuration.NotNull();
 
-        mergeMessage = new(mergeCommit.Message, configuration);
+        mergeMessage = null;
 
-        var isReleaseBranch = mergeMessage.MergedBranch != null && configuration.IsReleaseBranch(mergeMessage.MergedBranch);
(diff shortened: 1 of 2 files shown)


DSpirit:
@asbjornu I've created a clean PR due to merge conflicts. 
Here I've fixed the failing hotfix tests and applied the changes you've requested. Would be glad if you could take another look!

src/GitVersion.Core/MergeMessage.cs
```diff
@@ -97,13 +97,16 @@ private ReferenceName GetMergedBranchName(string mergedBranch)
         mergeCommit.NotNull();
         configuration.NotNull();
 
-        mergeMessage = null;
+        mergeMessage = new(mergeCommit.Message, configuration);
```
asbjornu, line 100: Can you reverse this so it initializes to `null` by default and only to a `MergeMessage` instance in the case `isValidMergeCommit` is `true`?

asbjornu (changes requested):

src/GitVersion.Core/MergeMessage.cs
DSpirit: @asbjornu sure, I've added an additional variable to check the mergedBranch.

asbjornu (approved):

Issue #4057, opened 2024-06-08
[Feature]: Support of multi mainline branches in the Mainline version strategy


HHobeck:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/GitTools/GitVersion/issues) to ensure it has not already been reported

### GitVersion package

GitVersion.Tool

### GitVersion version

6.0.0-rc.1

### Operating system

N/A

### What are you seeing?

The support of multi mainline branches (where the `is-mainline` property is set to true) needs to be implemented. 

GIVEN a `Mainline` version strategy with the `GitFlow/v1` workflow
WHEN a main -> support/1.x branch is used
THEN a `NotImplementedException` in `MainlineVersionStrategy.cs` is thrown

### What is expected?

A semantic version needs to be calculated

### Steps to Reproduce

Create a commit on main
branch to support/1.x
create a commit on support/1.x
run GitVersion

PR #4076, opened 2024-06-24
Update mainline.md

zammitjohn:
As of v6.0.0 of the schema, the Mainline mode and the related implementation has been removed completely. The new Mainline version strategy should be used instead.

## Description

As per new version on the schema, we must use the new strategies array.

## Motivation and Context

Fixes to documentation.

## Checklist:

### DIFF
--- a/docs/input/docs/reference/modes/mainline.md
+++ b/docs/input/docs/reference/modes/mainline.md
@@ -71,7 +71,7 @@ Deployment][continuous-deployment] by default). To change the [versioning
 mode][modes] to Mainline Development, just change the [configuration][configuration] as such:
 
 ```yaml
-mode: Mainline
+strategies: [Mainline]
 ```
 
 [configuration]: /docs/reference/configuration


HHobeck:
I would recommend the following configuration when using the Mainline strategy:

1. From scratch:
```yaml
workflow: ''
strategies:
- Mainline
- ConfiguredNextVersion
...
```

2. Using the pre-defined `GitFlow` workflow as a base template:
```yaml
workflow: GitFlow/v1
strategies:
- Mainline
- ConfiguredNextVersion
```

3. Using the pre-defined `GitHubFlow` workflow as a base template:
```yaml
workflow: GitHubFlow/v1
strategies:
- Mainline
- ConfiguredNextVersion
```

4. Using the pre-defined `TrunkBased` workflow as a base template:
```yaml
workflow: TrunkBased/preview1
strategies:
- Mainline
- ConfiguredNextVersion
```

We need to make clear that it is a new implementation which needs to be used in experimental context only. In advance multi mainline branches are not supported in version 6.x.
- #4057

arturcic:
@HHobeck can we merge this one? The build will not continue as there are only documentation changes

HHobeck:
> @HHobeck can we merge this one? The build will not continue as there are only documentation changes

Actually I would like to see the points I mentioned in the documentation. The current change makes no sense because it lean on the fact that the default value of workflow is `GitFlow/v1` (This is because of backwards compatibility reason). In future releases we are going to no support this.

arturcic:
> > @HHobeck can we merge this one? The build will not continue as there are only documentation changes
> 
> Actually I would like to see the points I mentioned in the documentation. The current change makes no sense because it lean on the fact that the default value of workflow is `GitFlow/v1` (This is because of backwards compatibility reason). In future releases we are going to not support this (probably).

Can we update the docs based on your suggestion and then we can push the rc2?

HHobeck:
> Can we update the docs based on your suggestion and then we can push the rc2?

Yes I agree

arturcic:
@HHobeck please then close this PR

Issue #4184, opened 2024-08-29
[ISSUE]: Bumping major/breaking via commit message not working as described during 0.x


Sharparam:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/GitTools/GitVersion/issues) to ensure it has not already been reported

### GitVersion package

GitVersion.Tool

### GitVersion version

6.0.2+Branch.main.Sha.30211316bc16e481dc440baae39ff904c4fa4966

### Operating system

Linux, Windows

### What are you seeing?

When bumping the major version (via `+semver:breaking` in the commit message) during "initial development" phase (0.x.y), GitVersion bumps the normal major component (which is 0 during initial development) to 1 instead of the minor component.

### What is expected?

GitVersion should bump the minor version instead during initial development in 0.x versions.

[As described in the documentation:](https://gitversion.net/docs/reference/version-increments#commit-messages)

> One thing to be aware of: If the current version is an alpha-version (i.e. 0.x.y.), attempting to bump the major version **will merely bump the minor (eg from 0.2.0 to 0.3.0 instead of 1.0.0)**. Once the current version is greater than 1.0.0, bumping the major version works as expected.

(Emphasis mine)

### Steps to Reproduce

Commands to show the issue:

```sh
mkdir gv-test && cd gv-test && git init
dotnet new tool-manifest
dotnet tool install gitversion.tool
git commit --allow-empty -m "Initial commit"
dotnet gitversion /showvariable SemVer # Outputs 0.0.1-1 as expected
git tag v0.1.0 -a -m ""
dotnet gitversion /showvariable SemVer # Outputs 0.1.0 as expected
git commit --allow-empty -m "Breaking change!\n\n+semver:breaking"
dotnet gitversion /showvariable SemVer # Outputs 1.0.0-1 but should be 0.2.0-1!
```

### RepositoryFixture Test

New test can be found in PR #4185.

### Output log or link to your CI build (if appropriate).

```shell
N/A
```

Sharparam:
After checking the code, it seems this bug might stem from commit c89c94c0e6b70b953a5a3d05a8c913277e49a7a6 where the `if` statement to "cap" the severity was removed by @HHobeck.

If I downgrade GitVersion to the latest release prior to that commit (6.0.0-beta.1), the bumping works as expected.

HHobeck:
I think we need to change the documentation.

Sharparam:
[...]

PR #4185, opened 2024-08-29
Add test for commit bump during 0.x

Sharparam:
This test verifies issue #4184.

## Description
As [described in the documentation](https://gitversion.net/docs/reference/version-increments#commit-messages), bumping the major version during the 0.x phase should bump the minor version instead.

However, this does not currently work. This test sets up that scenario.

## Related Issue
#4184 

## Motivation and Context
This test will make sure the bump behaviour stays as intended.

## How Has This Been Tested?
I have run the test locally on my machine and it behaves as expected (currently failing of course).

N/A

## Checklist:

- [X] My code follows the code style of this project.
- [X] I have added tests to cover my changes.
- [ ] All new and existing tests passed. (The new test fails by design)

### DIFF
--- a/src/GitVersion.Core.Tests/IntegrationTests/VersionBumpingScenarios.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/VersionBumpingScenarios.cs
@@ -115,4 +115,17 @@ public class VersionBumpingScenarios : TestBase
 
         fixture.AssertFullSemver("2.0.1-1");
     }
+
+    [Theory]
+    [TestCase("Breaking change\n\n+semver:breaking", "0.2.0-1")]
+    [TestCase("Minor change\n\n+semver:minor", "0.1.1-1")]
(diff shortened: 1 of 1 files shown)


HHobeck:
I think that is not correct:
```csharp
[TestCase("Minor change\n\n+semver:minor", "0.1.1-1")]
```

Sharparam:
Since the middle/minor part becomes the major essentially during 0.x, I moved the minor down to the patch level to follow suit.

However, if we consider [the discussion](https://github.com/GitTools/GitVersion/issues/4184#issuecomment-2320533792) in the linked issue:

> Anyway it feels like +semver: breaking has a special semantic and needs to be (if we want to support this behavior) separated from the major-version-bump-message configuration.

Then it might make more sense to adjust this new test to only consider `+breaking`, `+feature` and '+fix', if they're going to be treated separately?

arturcic:
Closing in favor of https://github.com/GitTools/GitVersion/pull/4308

PR #4188, opened 2024-08-30
add link to docs version 5.12.0

arturcic:
Adds the link to the older version of the docs at https://gitversion.net/5.12.0/docs/


asbjornu:
Nice solution you found there, @arturcic! 👍🏼

arturcic:
I think we can keep this option for when we have a new major version, and we want to expose the older version docs. For 5.12.0 I had to copy the older version and adjust the relative paths, but hopefully we can improve this.

asbjornu:
It's not perfect, but it's much better than nothing. Good job!

Issue #4192, opened 2024-09-02
[ISSUE]: Building with target framework net8.0-windows requires a `<GitVersionTargetFramework>`


aristotelos:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/GitTools/GitVersion/issues) to ensure it has not already been reported

### GitVersion package

GitVersion.MsBuild

### GitVersion version

6.0.2

### Operating system

Windows

### What are you seeing?

Build fails when building a project that has `<TargetFramework>net8.0-windows</TargetFramework>` with a dependency on GitVersion.MsBuild 6.0.2. Build output:

```
Run dotnet build --no-restore --configuration $env:Configuration
  Possible reasons for this include:
    * You misspelled a built-in dotnet command.
    * You intended to execute a .NET program, but dotnet---roll-forward does not exist.
  Could not execute because the specified command or file was not found.
    * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
C:\Users\runneradmin\.nuget\packages\gitversion.msbuild\6.0.2\tools\GitVersion.MsBuild.targets(22,9): error MSB30[7](https://github.com/FlaUI/FlaUI.WebDriver/actions/runs/10612608289/job/29414614648?pr=68#step:5:8)3: The command "dotnet --roll-forward Major "C:\Users\runneradmin\.nuget\packages\gitversion.msbuild\6.0.2\tools\net[8](https://github.com/FlaUI/FlaUI.WebDriver/actions/runs/10612608289/job/29414614648?pr=68#step:5:9).0-windows\gitversion.dll" "D:\a\FlaUI.WebDriver\FlaUI.WebDriver\src\FlaUI.WebDriver"  -output file -outputfile "obj\gitversion.json"" exited with code 1. [D:\a\FlaUI.WebDriver\FlaUI.WebDriver\src\FlaUI.WebDriver\FlaUI.WebDriver.csproj]
  WpfApplication -> D:\a\FlaUI.WebDriver\FlaUI.WebDriver\src\TestApplications\WpfApplication\bin\Release\WpfApplication.dll

Build FAILED.

[...]

PR #4206, opened 2024-09-11
Normalization of GitVersionTargetFramework and fallback scenario

samtrion:
This is about making sure that compatible target frameworks are treated equally. The previous solution tried to use the TargetFramework, which led to an error when we used `net8.0-windows`.

## Description

The solution here normalizes `net8.0-windows` and all other platform-specific TargetFrameworks to  
`net8.0`. It also accounts for `net9.0` or higher. If the TargetFramework isn't compatible with `net8.0`, it checks for compatibility with `net6.0`. If that doesn't work, it falls back to `net8.0`.

## Related Issue

#4192

## Checklist:

In this case not necessary, i think.


arturcic:
@mergify rebase

arturcic (approved):

Issue #4159, opened 2024-08-13
[ISSUE]: GitVersion tool is slow / stalls


christophwille:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/GitTools/GitVersion/issues) to ensure it has not already been reported

### GitVersion package

GitVersion.Tool

### GitVersion version

6.0.2

### Operating system

Linux

### What are you seeing?

This is for 

```
      git status
      dotnet gitversion /version
      dotnet gitversion $(Build.SourcesDirectory) /output buildserver /nofetch
```

on Azure DevOps pipelines (git status only added for testing and because some old issue mentioned that the index might be the culprit).

I am seeing a "pause" in the log - and it always happens for the current PRs branch (pull/2271/merge) End and the Begin of the actual branch I created comparing to develop (GitFlow). Below are two runs, as you can see it seems to wait for 3 minutes for no reason at all.

```
2024-08-13T07:24:15.4623771Z     INFO [24-08-13 7:24:15:46] -< End: Finding merge base between 'pull/2271/merge' and 'origin/release/4.77.0'. (Took: 0.43ms) >-
2024-08-13T07:24:15.4667782Z   INFO [24-08-13 7:24:15:46] -< End: Finding branches source of 'pull/2271/merge' (Took: 368.42ms) >-
2024-08-13T07:27:26.9638056Z   INFO [24-08-13 7:27:26:96] -< Begin: Finding branches source of 'feature/gitversionperftest' >-
2024-08-13T07:27:26.9666509Z     INFO [24-08-13 7:27:26:96] -< Begin: Finding merge base between 'feature/gitversionperftest' and 'develop'. >-

2024-08-13T07:36:02.1232132Z     INFO [24-08-13 7:36:02:12] -< End: Finding merge base between 'pull/2271/merge' and 'origin/release/4.77.0'. (Took: 0.40ms) >-
2024-08-13T07:36:02.1277016Z   INFO [24-08-13 7:36:02:12] -< End: Finding branches source of 'pull/2271/merge' (Took: 376.28ms) >-
2024-08-13T07:39:15.4620032Z   INFO [24-08-13 7:39:15:46] -< Begin: Finding branches source of 'feature/gitversionperftest' >-
2024-08-13T07:39:15.4648616Z     INFO [24-08-13 7:39:15:46] -< Begin: Finding merge base between 'feature/gitversionperftest' and 'develop'. >-
```

[...]

PR #4208, opened 2024-09-13
attempt to fix performance issues with a cache

david-driscoll:
## Description
This is an attempt to fix some of the performance problems (see #4159).  As best I can tell each time that the commits are requested from a given `IBranch` the git library appears to enumerate again and again.

## Related Issue
https://github.com/GitTools/GitVersion/issues/4159

## Motivation and Context
1.5 minutes to compute a version is a lot! 😄 

## How Has This Been Tested?
Best way I was able to test, was use the current version (6.0.2) and then compare to the results after caching.  The difference in performance was 1 and a half minutes on my machine to 10 seconds.

## Checklist:

### DIFF
--- a/src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs
@@ -6,11 +6,16 @@ namespace GitVersion.Git;
 internal sealed class CommitCollection : ICommitCollection
 {
     private readonly ICommitLog innerCollection;
+    private readonly Lazy<IReadOnlyCollection<ICommit>> commits;
 
-    internal CommitCollection(ICommitLog collection) => this.innerCollection = collection.NotNull();
+    internal CommitCollection(ICommitLog collection)
+    {
+        this.innerCollection = collection.NotNull();
+        this.commits = new Lazy<IReadOnlyCollection<ICommit>>(() => this.innerCollection.Select(commit => new Commit(commit)).ToArray());
+    }
 
     public IEnumerator<ICommit> GetEnumerator()
-        => this.innerCollection.Select(commit => new Commit(commit)).GetEnumerator();
+        => this.commits.Value.GetEnumerator();
 
     IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
 
@@ -21,7 +26,13 @@ internal sealed class CommitCollection : ICommitCollection
     {
         var includeReachableFrom = GetReacheableFrom(commitFilter.IncludeReachableFrom);
         var excludeReachableFrom = GetReacheableFrom(commitFilter.ExcludeReachableFrom);
-        var filter = new LibGit2Sharp.CommitFilter { IncludeReachableFrom = includeReachableFrom, ExcludeReachableFrom = excludeReachableFrom, FirstParentOnly = commitFilter.FirstParentOnly, SortBy = (LibGit2Sharp.CommitSortStrategies)commitFilter.SortBy };
+        var filter = new LibGit2Sharp.CommitFilter
+        {
+            IncludeReachableFrom = includeReachableFrom,
+            ExcludeReachableFrom = excludeReachableFrom,
+            FirstParentOnly = commitFilter.FirstParentOnly,
+            SortBy = (LibGit2Sharp.CommitSortStrategies)commitFilter.SortBy
+        };
         var commitLog = ((IQueryableCommitLog)this.innerCollection).QueryBy(filter);
         return new CommitCollection(commitLog);
 
--- a/src/GitVersion.Core/Configuration/IgnoreConfigurationExtensions.cs
+++ b/src/GitVersion.Core/Configuration/IgnoreConfigurationExtensions.cs
@@ -5,28 +5,20 @@ namespace GitVersion.Configuration;
 
 internal static class IgnoreConfigurationExtensions
 {
-    public static IEnumerable<ITag> Filter(this IIgnoreConfiguration ignore, IEnumerable<ITag> source)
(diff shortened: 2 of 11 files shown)


arturcic:
@david-driscoll can you provide the logs before and after the change, with the timestamps? You can remove the sensitive information. Now if the fix works as you say, maybe it makes sense to use same caching approach in the other `collections` in that folder?

david-driscoll:
So logging just halts, while it's calculating source.  I found 2 gaps where it's sitting around, so it's concieveable that it's happening else where, just not as prevelent.

When I was debugging I did have 74 source branches, which isn't a lot of numbers, but when you're iterating at least 2 times, things add up.  I will do a quick test, cause I still have a 10 second wait or so while it's working.

```
      DEBUG [24-09-13 9:24:47:66] Breaking
      INFO [24-09-13 9:24:47:66] Merge base of 'gitversion-test' and 'origin/***' is ''0a556d1' - Merge branch 'main' of REPO'
    INFO [24-09-13 9:24:47:66] -< End: Finding merge base between 'gitversion-test' and 'origin/***'. (Took: 6.46ms) >-
  INFO [24-09-13 9:24:47:66] -< End: Finding branches source of 'gitversion-test' (Took: 484.98ms) >-
  INFO [24-09-13 9:25:45:30] -< Begin: Finding branches source of 'feature/***' >-
    INFO [24-09-13 9:25:45:30] -< Begin: Finding merge base between 'feature/***' and 'origin/***'. >-
      INFO [24-09-13 9:25:45:33] Found merge base of ''b1f00f5' - Automatically linting code'
...
      DEBUG [24-09-13 9:25:46:87] Breaking
      INFO [24-09-13 9:25:46:87] Merge base of 'origin/***' and 'origin/feature/***' is ''0a556d1' - Merge branch 'main' of REPO'
    INFO [24-09-13 9:25:46:87] -< End: Finding merge base between 'origin/***' and 'origin/feature/***'. (Took: 17.02ms) >-
  INFO [24-09-13 9:25:46:87] -< End: Finding branches source of 'origin/***' (Took: 1,449.28ms) >-
  INFO [24-09-13 9:26:07:41] -< Begin: Finding branches source of 'feature/***' >-
  INFO [24-09-13 9:26:07:41] -< End: Finding branches source of 'feature/***' (Took: 6.37ms) >-
```

Just a complete halt in logging for

david-driscoll:
I gave it a shot, and went from 14s this morning to 12s.  Honestly that's within margin of error, so I can't say it really did much.

If you want I'll be glad to push the code, it doesn't break anything that I'm aware of (unit tests are running now), however I don't have any evidence it will have any real effect.

david-driscoll:
I have discovered similar behavior with this repo as well: https://github.com/RocketSurgeonsGuild/Nuke which is public, so clone away!  As long as you're on a branch other than `master` the calculation took 1m5s with 6.0.2 and then it was 7s with these changes.

david-driscoll:
@arturcic anything else you need from me?

arturcic:
@david-driscoll first of all can you fist squash the commits you currently have, the result will have only the change that we need. Second, can you also use the Lazy collections for `BranchCollection`, `RefSpecCollection`, `ReferenceCollection`, `RemoteCollection` and `TagCollection`, that way we have improvements for all of them

david-driscoll:
@arturcic I've added in the other collection changes.  And looks like you squashed, so I did the same.

arturcic:
> @arturcic I've added in the other collection changes. And looks like you squashed, so I did the same.

@david-driscoll thank you, I will have a look tomorrow

src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs
```diff
@@ -6,11 +6,16 @@ namespace GitVersion.Git;
 internal sealed class CommitCollection : ICommitCollection
 {
     private readonly ICommitLog innerCollection;
+    private readonly Lazy<IReadOnlyCollection<ICommit>> commits;
 
-    internal CommitCollection(ICommitLog collection) => this.innerCollection = collection.NotNull();
+    internal CommitCollection(ICommitLog collection)
+    {
+        this.innerCollection = collection.NotNull();
```
asbjornu, line 13: Do we need `this.innerCollection` at all, or is it enough as a captured reference inside the Lazy lambda? If it can be removed, it would improve readability and avoid accidental references to `this.innerCollection`.

david-driscoll: `QueryBy` still uses the inner collection, so we need both currenttly.

arturcic: it's used here https://github.com/GitTools/GitVersion/blob/825f62ca5ad06532d671178c8a683aa200ffc659/src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs#L25-L40, and that's the challenging part, If we can filter on the cached collection the same way git does filtering then we're good

arturcic:
@christophwille it will be included in the next release, whenever this PR gets merged

src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs
asbjornu: Right, so this is an impedance mismatch between the lazy collection and LibGit2's `QueryBy` method? Also, trying to abstract it might put us well into [Hyrum's law](https://www.hyrumslaw.com/) territory. Although we could probably simplify the abstraction if we manage to reduce all of GitVersion's usage to a least common denominator. That's out of scope for this PR, though.

asbjornu (approved):

arturcic (approved):

PR #4403, opened 2025-01-31
Fixes to dev container

Bi0T1N:
## Description

One change avoids compiling git from source since the version provided by the OS is fine and can be installed quickly (compared to compiling git). The other one addresses the problem that the `dotnet:dev-8.0` doesn't work out-of-the-box because the `global.json` defines SDK 9.0 (not sure if this is the problem but my changes fixed the issue and got the devcontainer running) so that you need `dotnet:dev-9.0`.

## Motivation and Context
Better out-of-the-box experience.

## How Has This Been Tested?
Setting up the project in VS Code and running the testsuite with `dotnet test --framework net9.0 src/`

## Checklist:

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/dotnet:dev-8.0
+FROM mcr.microsoft.com/devcontainers/dotnet:dev-9.0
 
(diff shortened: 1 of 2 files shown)


arturcic:
@Bi0T1N thank you for the updates, nowadays I use less devcontainer for GitVersion but glad there is still interest in this

Issue #4329, opened 2024-11-22
[ISSUE]: Overrideconfig for strategies returns 'unknown key' error


ohlrogge:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched [issues](https://github.com/GitTools/GitVersion/issues) to ensure it has not already been reported

### GitVersion package

GitVersion.Tool

### GitVersion version

6.0.5+Branch.main.Sha.c78e7a3f34e0afe6ca0a3b4a03218f1b1280d434

### Operating system

Windows

### What are you seeing?

Hi there,
when I want to override the config for strategies I receive the following error:

```
Unhandled exception. GitVersion.WarningException: Could not parse /overrideconfig option: strategies=mainline. Unsupported 'key'.
   at GitVersion.ArgumentParser.ParseOverrideConfig(Arguments arguments, IReadOnlyCollection`1 values) in /_/src/GitVersion.App/ArgumentParser.cs:line 463
```

### What is expected?

I expect to be able to pass in the strategies as array. Later I want to override my config using the azure pipelines task.

### Steps to Reproduce

I run the following command on my main branch:
```
gitversion . /output json /output buildserver /overrideconfig mode=ContinuousDelivery /overrideconfig major-version-bump-message="^(build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test)(\([\w\s,\/\\-]*\))?(!:|:.*\n\n((.+\n)+\n)?BREAKING CHANGE:\s.+)" /overrideconfig minor-version-bump-message="^(feat)(\([\w\s,\/\\-]*\))?:" /overrideconfig patch-version-bump-message="^(build|chore|ci|docs|fix|perf|refactor|revert|style|test)(\([\w\s,\/\\-]*\))?:" /overrideconfig strategies=mainline
```

### Output log or link to your CI build (if appropriate).

```shell

```

HHobeck:
Very good. You have found a bug. If you not mind please fix this and create a pull-request.

arturcic:
Closed by #4405

PR #4405, opened 2025-02-03
Fix [ISSUE]: Overrideconfig for strategies returns 'unknown key' error #4329

schickst:
This pull request tries to fix [ISSUE]: Overrideconfig for strategies returns 'unknown key' error #4329.

## Description

There are multiple issues when using /overrideconfig with the 'strategies' property.

First: 
The overrideconfig uses a different format that the normal config. And they are then fed differently into the deserializer.

````
For the override config the following format seems to be used:
strategies=["Mainline","None"]

For the normal config the following format is used:
strategies:
- Mainline
- None
````

The second issue is that a cast from string[] to VersionStrategies[] has to happen. 

My approach was to implement a custom type converter and registered it with the deserializer. The type converter then detects the config format and parses it accordingly. 

In my tests it worked for both configuration sources. I am unsure if the location where I put the converter is the right place. I would prefer to avoid YamlDotNet as a direct dependency.

Also I hope that there is a simpler solution and avoid the converter all-together.  Maybe this splitting behavior of the value beforehand can be improved.

## Related Issue

#4329 

## How Has This Been Tested?

A test case has been added for the related overrideconfig strategies property. There are also many other test cases that might break, if the serialization/deserialization is wrong.

## Checklist:

* \[x] My code follows the code style of this project. (I hope so)

### CODE REVIEW

File: src/GitVersion.Core/VersionCalculation/VersionStrategiesConverter.cs [hunk 1]
```diff
@@ -0,0 +1,53 @@
+using YamlDotNet.Core;
+using YamlDotNet.Core.Events;
+using YamlDotNet.Serialization;
+using YamlDotNet.Serialization.NamingConventions;
+
+namespace GitVersion.VersionCalculation;
+
+public class VersionStrategiesConverter : IYamlTypeConverter
```
arturcic, line 8: Please move this to the Configuration module

src/GitVersion.Core/GitVersion.Core.csproj
```diff
@@ -15,6 +15,7 @@
         <PackageReference Include="Polly" />
         <PackageReference Include="Microsoft.Extensions.DependencyInjection.Abstractions" />
         <PackageReference Include="Microsoft.Extensions.Options" />
+        <PackageReference Include="YamlDotNet" />
```
arturcic, line 18: We should no include `YamlDotNet`, we just removed a couple of releases ago, so if you need anything configuration related do it in Configuration module

arturcic (changes requested):

arturcic (approved):
LGTM, thanks for fixing this

PR #4420, opened 2025-02-14
Feature: Support ignoring (filtering) paths in git tree

### LINKED ISSUE #2436: filter on tags on prefix or regex, to make it possible to use multiple git versions in one repository
### Context
We are trying to merge a solution that build nuget packages from one repo to another.
This repo already has a tagging structure for other nuget packages. we do not want to mix up these tags.
The most obvious thing for us, seems to be able to have a filter, in which we can state a prefix or regex, so that a certain git version task, will only analyze the ones having this prefix or regex.

Other issue that we see: the analysis of the already existing tags takes about 10 minutes, because we have a lot of them. having the filter possibility, just as command git tag -l "prefix*" has, would make the analysis a lot quicker. Or is there any other way?

### Solution that we think would help:
specify a prefix, that is used both for tagging and for filtering during the analysis to specify which git tag will have to be added.

Or am I missing a feature that already does this?

Roei-Levi:
## Description

This PR adds a new option to the git version config, allowing users to filter specific paths in the directory tree using the "ignore/paths" config.

## Related Issue

Resolves #2436 

## Motivation and Context
This feature enables better versioning capability when using GitVersion in a *monorepo*, where different components need to be versioned independently. 
It is based on work done previously and mentioned in the related issue #2436. See discussion for more info.

## How Has This Been Tested?
All existing tests pass. 
New sanity tests were added and passed.
Additional tests will be included before closing this PR.

## Checklist:

* \[X] My code follows the code style of this project.
* \[X] My change requires a change to the documentation.

### DIFF
--- a/src/GitVersion.Core/Configuration/EffectiveConfiguration.cs
+++ b/src/GitVersion.Core/Configuration/EffectiveConfiguration.cs
@@ -68,7 +68,6 @@ public record EffectiveConfiguration
         PatchVersionBumpMessage = configuration.PatchVersionBumpMessage;
         NoBumpMessage = configuration.NoBumpMessage;
         CommitMessageIncrementing = branchConfiguration.CommitMessageIncrementing.Value;
-        VersionFilters = configuration.Ignore.ToFilters();
         Ignore = configuration.Ignore;
         TracksReleaseBranches = branchConfiguration.TracksReleaseBranches ?? false;
         IsReleaseBranch = branchConfiguration.IsReleaseBranch ?? false;
@@ -123,8 +122,6 @@ public record EffectiveConfiguration
     public string? NoBumpMessage { get; }
 
     public CommitMessageIncrementMode CommitMessageIncrementing { get; }
-
-    public IEnumerable<IVersionFilter> VersionFilters { get; }
 
     public IIgnoreConfiguration Ignore { get; }
 
--- a/src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
+++ b/src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
@@ -8,5 +8,5 @@ public interface IIgnoreConfiguration
 
     IReadOnlyCollection<string> Paths { get; }
 
-    bool IsEmpty => Before == null && Shas.Count == 0 && Paths.Count == 0;
+    bool IsEmpty { get; }
 }
--- a/src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
+++ b/src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
@@ -6,5 +6,7 @@ public interface IIgnoreConfiguration
 
     IReadOnlySet<string> Shas { get; }
 
-    bool IsEmpty { get; }
+    IReadOnlyCollection<string> Paths { get; }
+
+    bool IsEmpty => Before == null && Shas.Count == 0 && Paths.Count == 0;
 }
--- a/src/GitVersion.Core/Git/ICommit.cs
+++ b/src/GitVersion.Core/Git/ICommit.cs
@@ -7,5 +7,6 @@ public interface ICommit : IEquatable<ICommit?>, IComparable<ICommit>, IGitObjec
     DateTimeOffset When { get; }
 
     string Message { get; }
+
     IReadOnlyList<string> DiffPaths { get; }
 }
--- a/src/GitVersion.Core/Git/ICommit.cs
+++ b/src/GitVersion.Core/Git/ICommit.cs
@@ -7,5 +7,5 @@ public interface ICommit : IEquatable<ICommit?>, IComparable<ICommit>, IGitObjec
     DateTimeOffset When { get; }
 
     string Message { get; }
-    IEnumerable<string> DiffPaths { get; }
+    IReadOnlyList<string> DiffPaths { get; }
 }
--- a/src/GitVersion.Core/Git/ICommit.cs
+++ b/src/GitVersion.Core/Git/ICommit.cs
@@ -7,4 +7,5 @@ public interface ICommit : IEquatable<ICommit?>, IComparable<ICommit>, IGitObjec
     DateTimeOffset When { get; }
 
     string Message { get; }
+    IEnumerable<string> DiffPaths { get; }
 }
--- a/src/GitVersion.Core/VersionCalculation/Abstractions/IVersionFilter.cs
+++ b/src/GitVersion.Core/VersionCalculation/Abstractions/IVersionFilter.cs
@@ -1,6 +1,9 @@
+using GitVersion.Git;
+
 namespace GitVersion.VersionCalculation;
 
 public interface IVersionFilter
 {
     bool Exclude(IBaseVersion baseVersion, out string? reason);
+    bool Exclude(ICommit commit, out string? reason);
 }
--- a/src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
+++ b/src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
@@ -7,7 +7,9 @@ using GitVersion.Git;
 
 namespace GitVersion.VersionCalculation;
 
-internal class IncrementStrategyFinder(IRepositoryStore repositoryStore, ITaggedSemanticVersionRepository taggedSemanticVersionRepository)
+internal class IncrementStrategyFinder(
+    IRepositoryStore repositoryStore,
+    ITaggedSemanticVersionRepository taggedSemanticVersionRepository)
     : IIncrementStrategyFinder
 {
     private readonly Dictionary<string, VersionField?> commitIncrementCache = [];
--- a/src/GitVersion.Core/VersionCalculation/PathFilter.cs
+++ b/src/GitVersion.Core/VersionCalculation/PathFilter.cs
@@ -35,21 +35,22 @@ internal class PathFilter(IReadOnlyList<string> paths, PathFilterMode mode = Pat
     public bool Exclude(ICommit? commit, [NotNullWhen(true)] out string? reason)
     {
         reason = null;
-
-        if (commit != null)
+        if (commit == null)
         {
-            switch (mode)
-            {
-                case PathFilterMode.Inclusive:
+            return false;
+        }
+
+        switch (mode)
+        {
+            case PathFilterMode.Inclusive:
+                {
+                    if (commit.DiffPaths.All(this.IsMatch))
                     {
-                        if (commit.DiffPaths.All(this.IsMatch))
-                        {
-                            reason = "Source was ignored due to all commit paths matching ignore regex";
-                            return true;
-                        }
-                        break;
+                        reason = "Source was ignored due to all commit paths matching ignore regex";
+                        return true;
                     }
-            }
+                    break;
+                }
         }
 
         return false;
--- a/src/GitVersion.Core/VersionCalculation/PathFilter.cs
+++ b/src/GitVersion.Core/VersionCalculation/PathFilter.cs
@@ -1,13 +1,14 @@
+using System.Collections.Concurrent;
 using System.Diagnostics.CodeAnalysis;
 using System.Text.RegularExpressions;
 using GitVersion.Git;
 
 namespace GitVersion.VersionCalculation;
 
-internal class PathFilter( IEnumerable<string> paths) : IVersionFilter
+internal class PathFilter(IReadOnlyList<string> paths) : IVersionFilter
 {
-    private readonly List<Regex> pathsRegexes = paths.Select(path => new Regex(path, RegexOptions.IgnoreCase | RegexOptions.Compiled)).ToList();
-    private readonly Dictionary<string, bool> pathMatchCache = [];
+    private readonly List<Regex> pathsRegexes = [.. paths.Select(path => new Regex(path, RegexOptions.IgnoreCase | RegexOptions.Compiled))];
+    private readonly ConcurrentDictionary<string, bool> pathMatchCache = [];
 
     public bool Exclude(IBaseVersion baseVersion, [NotNullWhen(true)] out string? reason)
     {
@@ -27,21 +28,18 @@ internal class PathFilter( IEnumerable<string> paths) : IVersionFilter
         {
             var patchPaths = commit.DiffPaths;
 
-            if (patchPaths != null)
+            foreach (var path in patchPaths)
             {
-                foreach (var path in patchPaths)
+                if (!pathMatchCache.TryGetValue(path, out var isMatch))
                 {
-                    if (!pathMatchCache.TryGetValue(path, out var isMatch))
-                    {
-                        isMatch = this.pathsRegexes.Any(regex => regex.IsMatch(path));
-                        pathMatchCache[path] = isMatch;
-                    }
+                    isMatch = this.pathsRegexes.Any(regex => regex.IsMatch(path));
+                    pathMatchCache[path] = isMatch;
+                }
 
-                    if (isMatch)
-                    {
-                        reason = "Source was ignored due to commit path matching ignore regex";
-                        return true;
-                    }
+                if (isMatch)
+                {
+                    reason = "Source was ignored due to commit path matching ignore regex";
+                    return true;
                 }
             }
         }
--- a/src/GitVersion.Core/VersionCalculation/PathFilter.cs
+++ b/src/GitVersion.Core/VersionCalculation/PathFilter.cs
@@ -7,16 +7,12 @@ namespace GitVersion.VersionCalculation;
 
 internal class PathFilter(IReadOnlyList<string> paths) : IVersionFilter
 {
-    private readonly List<Regex> pathsRegexes = [.. paths.Select(path => new Regex(path, RegexOptions.IgnoreCase | RegexOptions.Compiled))];
+    private readonly List<Regex> pathsRegexes = [.. paths.Select(path => new Regex(path, RegexOptions.Compiled))];
     private readonly ConcurrentDictionary<string, bool> pathMatchCache = [];
 
     public bool Exclude(IBaseVersion baseVersion, [NotNullWhen(true)] out string? reason)
     {
         ArgumentNullException.ThrowIfNull(baseVersion);
-
-        reason = null;
-        if (baseVersion.Source.StartsWith("Fallback") || baseVersion.Source.StartsWith("Git tag") || baseVersion.Source.StartsWith("NextVersion")) return false;
-
         return Exclude(baseVersion.BaseVersionSource, out reason);
     }
 
@@ -26,9 +22,7 @@ internal class PathFilter(IReadOnlyList<string> paths) : IVersionFilter
 
         if (commit != null)
         {
-            var patchPaths = commit.DiffPaths;
-
-            foreach (var path in patchPaths)
+            foreach (var path in commit.DiffPaths)
             {
                 if (!pathMatchCache.TryGetValue(path, out var isMatch))
                 {
--- a/src/GitVersion.Core/VersionCalculation/PathFilter.cs
+++ b/src/GitVersion.Core/VersionCalculation/PathFilter.cs
@@ -1,0 +1,51 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Text.RegularExpressions;
+using GitVersion.Git;
+
+namespace GitVersion.VersionCalculation;
+
+internal class PathFilter( IEnumerable<string> paths) : IVersionFilter
+{
+    private readonly List<Regex> pathsRegexes = paths.Select(path => new Regex(path, RegexOptions.IgnoreCase | RegexOptions.Compiled)).ToList();
+    private readonly Dictionary<string, bool> pathMatchCache = [];
+
+    public bool Exclude(IBaseVersion baseVersion, [NotNullWhen(true)] out string? reason)
+    {
+        ArgumentNullException.ThrowIfNull(baseVersion);
+
+        reason = null;
+        if (baseVersion.Source.StartsWith("Fallback") || baseVersion.Source.StartsWith("Git tag") || baseVersion.Source.StartsWith("NextVersion")) return false;
+
+        return Exclude(baseVersion.BaseVersionSource, out reason);
+    }
+
+    public bool Exclude(ICommit? commit, [NotNullWhen(true)] out string? reason)
+    {
+        reason = null;
+
+        if (commit != null)
+        {
+            var patchPaths = commit.DiffPaths;
+
+            if (patchPaths != null)
+            {
+                foreach (var path in patchPaths)
+                {
+                    if (!pathMatchCache.TryGetValue(path, out var isMatch))
+                    {
+                        isMatch = this.pathsRegexes.Any(regex => regex.IsMatch(path));
+                        pathMatchCache[path] = isMatch;
+                    }
+
+                    if (isMatch)
+                    {
+                        reason = "Source was ignored due to commit path matching ignore regex";
+                        return true;
+                    }
+                }
+            }
+        }
+
+        return false;
+    }
+}
--- a/src/GitVersion.Core/VersionCalculation/PathFilter.cs
+++ b/src/GitVersion.Core/VersionCalculation/PathFilter.cs
@@ -5,9 +5,15 @@ using GitVersion.Git;
 
 namespace GitVersion.VersionCalculation;
 
-internal class PathFilter(IReadOnlyList<string> paths) : IVersionFilter
+internal enum PathFilterMode
 {
-    private readonly List<Regex> pathsRegexes = [.. paths.Select(path => new Regex(path, RegexOptions.Compiled))];
+    Inclusive,  // All commit paths must match for commit to be excluded
+    //Exclusive // Any commit path must match for commit to be excluded
+}
+
+internal class PathFilter(IReadOnlyList<string> paths, PathFilterMode mode = PathFilterMode.Inclusive) : IVersionFilter
+{
+    private readonly IReadOnlyList<Regex> pathsRegexes = [.. paths.Select(path => new Regex(path, RegexOptions.Compiled))];
     private readonly ConcurrentDictionary<string, bool> pathMatchCache = [];
 
     public bool Exclude(IBaseVersion baseVersion, [NotNullWhen(true)] out string? reason)
@@ -16,25 +22,33 @@ internal class PathFilter(IReadOnlyList<string> paths) : IVersionFilter
         return Exclude(baseVersion.BaseVersionSource, out reason);
     }
 
+    private bool IsMatch(string path)
+    {
+        if (!pathMatchCache.TryGetValue(path, out var isMatch))
+        {
+            isMatch = this.pathsRegexes.Any(regex => regex.IsMatch(path));
+            pathMatchCache[path] = isMatch;
+        }
+        return isMatch;
+    }
+
     public bool Exclude(ICommit? commit, [NotNullWhen(true)] out string? reason)
     {
         reason = null;
 
         if (commit != null)
         {
-            foreach (var path in commit.DiffPaths)
+            switch (mode)
             {
-                if (!pathMatchCache.TryGetValue(path, out var isMatch))
-                {
-                    isMatch = this.pathsRegexes.Any(regex => regex.IsMatch(path));
-                    pathMatchCache[path] = isMatch;
-                }
-
-                if (isMatch)
-                {
-                    reason = "Source was ignored due to commit path matching ignore regex";
-                    return true;
-                }
+                case PathFilterMode.Inclusive:
+                    {
+                        if (commit.DiffPaths.All(this.IsMatch))
+                        {
+                            reason = "Source was ignored due to all commit paths matching ignore regex";
+                            return true;
+                        }
+                        break;
+                    }
             }
         }
 
--- a/src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/ConfiguredNextVersionVersionStrategy.cs
+++ b/src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/ConfiguredNextVersionVersionStrategy.cs
@@ -18,7 +18,7 @@ internal sealed class ConfiguredNextVersionVersionStrategy(Lazy<GitVersionContex
     {
         configuration.NotNull();
 
-        if (!this.Context.Configuration.VersionStrategy.HasFlag(VersionStrategies.ConfiguredNextVersion))
+        if (!Context.Configuration.VersionStrategy.HasFlag(VersionStrategies.ConfiguredNextVersion))
             yield break;
 
         var nextVersion = Context.Configuration.NextVersion;
--- a/src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/ConfiguredNextVersionVersionStrategy.cs
+++ b/src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/ConfiguredNextVersionVersionStrategy.cs
@@ -18,7 +18,7 @@ internal sealed class ConfiguredNextVersionVersionStrategy(Lazy<GitVersionContex
     {
         configuration.NotNull();
 
-        if (!Context.Configuration.VersionStrategy.HasFlag(VersionStrategies.ConfiguredNextVersion))
+        if (!this.Context.Configuration.VersionStrategy.HasFlag(VersionStrategies.ConfiguredNextVersion))
             yield break;
 
         var nextVersion = Context.Configuration.NextVersion;
--- a/src/GitVersion.LibGit2Sharp/Git/Commit.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/Commit.cs
@@ -27,7 +27,6 @@ internal sealed class Commit : GitObject, ICommit
     public IReadOnlyList<ICommit> Parents => this.parentsLazy.Value;
     public DateTimeOffset When { get; }
     public string Message => this.innerCommit.Message;
-    // TODO implement tag prefix filtering before returning the paths.
     public IReadOnlyList<string> DiffPaths
     {
         get
--- a/src/GitVersion.LibGit2Sharp/Git/Commit.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/Commit.cs
@@ -44,5 +44,5 @@ internal sealed class Commit : GitObject, ICommit
     public override int GetHashCode() => equalityHelper.GetHashCode(this);
     public override string ToString() => $"'{Id.ToString(7)}' - {this.innerCommit.MessageShort}";
     public static implicit operator LibGit2Sharp.Commit(Commit d) => d.innerCommit;
-    private TreeChanges CommitChanges => new TreeChanges(this.repoDiff.Compare<LibGit2Sharp.TreeChanges>(this.innerCommit.Tree, this.innerCommit.Parents.FirstOrDefault()?.Tree));
+    private TreeChanges CommitChanges => new(this.repoDiff.Compare<LibGit2Sharp.TreeChanges>(this.innerCommit.Tree, this.innerCommit.Parents.FirstOrDefault()?.Tree));
 }
--- a/src/GitVersion.LibGit2Sharp/Git/Commit.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/Commit.cs
@@ -28,19 +28,21 @@ internal sealed class Commit : GitObject, ICommit
     public DateTimeOffset When { get; }
     public string Message => this.innerCommit.Message;
     // TODO implement tag prefix filtering before returning the paths.
-    public IEnumerable<string> DiffPaths
+    public IReadOnlyList<string> DiffPaths
     {
         get
         {
-            if (pathsCache.TryGetValue(this.Sha, out var paths))
-                return paths;
-            else
-                return this.CommitChanges?.Paths ?? [];
+            if (!pathsCache.TryGetValue(this.Sha, out var paths))
+            {
+                paths = this.CommitChanges?.Paths ?? [];
+                pathsCache[this.Sha] = paths;
+            }
+            return paths;
         }
     }
     public override bool Equals(object? obj) => Equals(obj as ICommit);
     public override int GetHashCode() => equalityHelper.GetHashCode(this);
     public override string ToString() => $"'{Id.ToString(7)}' - {this.innerCommit.MessageShort}";
     public static implicit operator LibGit2Sharp.Commit(Commit d) => d.innerCommit;
-    private ITreeChanges CommitChanges => new TreeChanges(this.repoDiff.Compare<LibGit2Sharp.TreeChanges>(this.innerCommit.Tree, this.innerCommit.Parents.FirstOrDefault()?.Tree));
+    private TreeChanges CommitChanges => new TreeChanges(this.repoDiff.Compare<LibGit2Sharp.TreeChanges>(this.innerCommit.Tree, this.innerCommit.Parents.FirstOrDefault()?.Tree));
 }
--- a/src/GitVersion.LibGit2Sharp/Git/Commit.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/Commit.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using GitVersion.Extensions;
 using GitVersion.Helpers;
 
@@ -5,17 +6,20 @@ namespace GitVersion.Git;
 
 internal sealed class Commit : GitObject, ICommit
 {
+    private static readonly ConcurrentDictionary<string, IReadOnlyList<string>> pathsCache = new();
     private static readonly LambdaEqualityHelper<ICommit> equalityHelper = new(x => x.Id);
     private static readonly LambdaKeyComparer<ICommit, string> comparerHelper = new(x => x.Sha);
     private readonly Lazy<IReadOnlyList<ICommit>> parentsLazy;
 
     private readonly LibGit2Sharp.Commit innerCommit;
+    private readonly LibGit2Sharp.Diff repoDiff;
 
-    internal Commit(LibGit2Sharp.Commit innerCommit) : base(innerCommit)
+    internal Commit(LibGit2Sharp.Commit innerCommit, LibGit2Sharp.Diff repoDiff) : base(innerCommit)
     {
         this.innerCommit = innerCommit.NotNull();
-        this.parentsLazy = new(() => innerCommit.Parents.Select(parent => new Commit(parent)).ToList());
+        this.parentsLazy = new(() => innerCommit.Parents.Select(parent => new Commit(parent, repoDiff)).ToList());
         When = innerCommit.Committer.When;
+        this.repoDiff = repoDiff;
     }
 
     public int CompareTo(ICommit? other) => comparerHelper.Compare(this, other);
@@ -23,8 +27,20 @@ internal sealed class Commit : GitObject, ICommit
     public IReadOnlyList<ICommit> Parents => this.parentsLazy.Value;
     public DateTimeOffset When { get; }
     public string Message => this.innerCommit.Message;
+    // TODO implement tag prefix filtering before returning the paths.
+    public IEnumerable<string> DiffPaths
+    {
+        get
+        {
+            if (pathsCache.TryGetValue(this.Sha, out var paths))
+                return paths;
+            else
+                return this.CommitChanges?.Paths ?? [];
+        }
+    }
     public override bool Equals(object? obj) => Equals(obj as ICommit);
     public override int GetHashCode() => equalityHelper.GetHashCode(this);
     public override string ToString() => $"'{Id.ToString(7)}' - {this.innerCommit.MessageShort}";
     public static implicit operator LibGit2Sharp.Commit(Commit d) => d.innerCommit;
+    private ITreeChanges CommitChanges => new TreeChanges(this.repoDiff.Compare<LibGit2Sharp.TreeChanges>(this.innerCommit.Tree, this.innerCommit.Parents.FirstOrDefault()?.Tree));
 }
--- a/src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using GitVersion.Extensions;
 using GitVersion.Helpers;
 using LibGit2Sharp;
@@ -7,6 +8,7 @@ namespace GitVersion.Git;
 internal sealed partial class GitRepository
 {
     private Lazy<IRepository>? repositoryLazy;
+    private readonly ConcurrentDictionary<string, IReadOnlyList<string>?> patchPathsCache = new();
 
     private IRepository RepositoryInstance
     {
@@ -16,17 +18,16 @@ internal sealed partial class GitRepository
             return lazy.Value;
         }
     }
-
     public string Path => RepositoryInstance.Info.Path;
     public string WorkingDirectory => RepositoryInstance.Info.WorkingDirectory;
     public bool IsHeadDetached => RepositoryInstance.Info.IsHeadDetached;
     public bool IsShallow => RepositoryInstance.Info.IsShallow;
-    public IBranch Head => new Branch(RepositoryInstance.Head);
+    public IBranch Head => new Branch(RepositoryInstance.Head, RepositoryInstance.Diff);
 
-    public ITagCollection Tags => new TagCollection(RepositoryInstance.Tags);
+    public ITagCollection Tags => new TagCollection(RepositoryInstance.Tags, RepositoryInstance.Diff);
     public IReferenceCollection Refs => new ReferenceCollection(RepositoryInstance.Refs);
-    public IBranchCollection Branches => new BranchCollection(RepositoryInstance.Branches);
-    public ICommitCollection Commits => new CommitCollection(RepositoryInstance.Commits);
+    public IBranchCollection Branches => new BranchCollection(RepositoryInstance.Branches, RepositoryInstance.Diff);
+    public ICommitCollection Commits => new CommitCollection(RepositoryInstance.Commits, RepositoryInstance.Diff);
     public IRemoteCollection Remotes => new RemoteCollection(RepositoryInstance.Network.Remotes);
 
     public void DiscoverRepository(string? gitDirectory)
@@ -49,7 +50,7 @@ internal sealed partial class GitRepository
             var first = (Commit)commit;
             var second = (Commit)otherCommit;
             var mergeBase = RepositoryInstance.ObjectDatabase.FindMergeBase(first, second);
-            return mergeBase == null ? null : new Commit(mergeBase);
+            return mergeBase == null ? null : new Commit(mergeBase, RepositoryInstance.Diff);
         });
     }
 
@@ -88,7 +89,7 @@ internal sealed partial class GitRepository
         }
 
         // gets all changes of the last commit vs Staging area and WT
-        var changes = RepositoryInstance.Diff.Compare<TreeChanges>(RepositoryInstance.Head.Tip.Tree,
+        var changes = RepositoryInstance.Diff.Compare<LibGit2Sharp.TreeChanges>(RepositoryInstance.Head.Tip.Tree,
             DiffTargets.Index | DiffTargets.WorkingDirectory);
 
         return changes.Count;
--- a/src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
@@ -1,4 +1,3 @@
-using System.Collections.Concurrent;
 using GitVersion.Extensions;
 using GitVersion.Helpers;
 using LibGit2Sharp;
@@ -8,7 +7,6 @@ namespace GitVersion.Git;
 internal sealed partial class GitRepository
 {
     private Lazy<IRepository>? repositoryLazy;
-    private readonly ConcurrentDictionary<string, IReadOnlyList<string>?> patchPathsCache = new();
 
     private IRepository RepositoryInstance
     {
--- a/docs/input/docs/reference/configuration.md
+++ b/docs/input/docs/reference/configuration.md
@@ -191,6 +191,7 @@ branches:
     is-main-branch: false
 ignore:
   sha: []
+  paths: []
 mode: ContinuousDelivery
 label: '{BranchName}'
 increment: Inherit
@@ -208,7 +209,7 @@ tracks-release-branches: false
 is-release-branch: false
 is-main-branch: false
 ```
-<sup><a href='/docs/workflows/GitFlow/v1.yml#L1-L166' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/GitFlow/v1.yml' title='Start of snippet'>anchor</a></sup>
+<sup><a href='/docs/workflows/GitFlow/v1.yml#L1-L167' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/GitFlow/v1.yml' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
 The supported built-in configuration for the `GitHubFlow` workflow (`workflow: GitHubFlow/v1`) looks like:
@@ -315,6 +316,7 @@ branches:
     is-main-branch: false
 ignore:
   sha: []
+  paths: []
 mode: ContinuousDelivery
 label: '{BranchName}'
 increment: Inherit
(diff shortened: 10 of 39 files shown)


Roei-Levi:
@asbjornu Please have a look at this :)

src/GitVersion.Core/Configuration/EffectiveConfiguration.cs
```diff
@@ -122,8 +122,6 @@ public EffectiveConfiguration(
 
     public CommitMessageIncrementMode CommitMessageIncrementing { get; }
 
-    public IEnumerable<IVersionFilter> VersionFilters { get; }
```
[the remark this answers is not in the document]
Roei-Levi: Because it had 0 references across the solution. I can restore it...

HHobeck:
This changes are integral and I need time to review it.

HHobeck:
image

You are defining an enum with inclusive and exclusive mode. Can you answer the following question please:
1. Why is the exclusive mode not used?
1. In the ignore configuration section you are defining pathes which will be treated as inclusive. Is this correct?
1. Why are the enum values pre-defined with zero and one?

Roei-Levi:
Hi @HHobeck,
Thanks for the questions.

1. Because I missed it. Thank you for noticing I'll add it to the config.
2. Yes. But I'll update it to include both.
3. No reason really.

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+        commit.NotNull();
+        reason = null;
+
+        if (commit != null)
+        {
+            var patchPaths = repository.FindPatchPaths(commit, context.Configuration.TagPrefixPattern);
+
+            if (patchPaths != null)
+            {
+                switch (mode)
```
asbjornu, line 35: Please prefix local field access with `this.`.

```suggestion
                switch (this.mode)
```

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public bool Exclude(ICommit? commit, out string? reason)
+    {
+        commit.NotNull();
+        reason = null;
+
+        if (commit != null)
+        {
+            var patchPaths = repository.FindPatchPaths(commit, context.Configuration.TagPrefixPattern);
```
asbjornu, line 31:
```suggestion
            var patchPaths = repository.FindPatchPaths(commit, this.context.Configuration.TagPrefixPattern);
```

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            if (patchPaths != null)
+            {
+                switch (mode)
+                {
+                    case PathFilterMode.Inclusive:
+                        if (!paths.Any(path => patchPaths.Any(p => p.StartsWith(path, StringComparison.OrdinalIgnoreCase))))
```
asbjornu, line 38:
```suggestion
                        if (!this.paths.Any(path => patchPaths.Any(p => p.StartsWith(path, StringComparison.OrdinalIgnoreCase))))
```

asbjornu (changes requested):

HHobeck:
> 2\. Yes. But I'll update it to include both.

Thank you for your answers. Would it be an idea to just support regular expressions? This would skip the need of the inclusive and exclusive path filter mode enum and make it more generic.

Roei-Levi:
Hi,
I'm not sure.
I agree it would be more generic. However, in my opinion there are 2 things
to keep in mind:
a) Which option would be more intuitive for users?
b) What would be the performance implication?

A bit far-fetched, but maybe we should try reaching out to @sebastienwarin,
to get his opinion, since this is his original implementation.

And speaking of performance, I noticed that the current implementation
doesn't scale well. The next version calculation becomes prohibitively
long.. This is due to the long patch calculation step, which calculates the
diff between two trees. To overcome this, I think it requires slightly
modifying the next version calculation step, to defer filtering to later
time.

I can commit more changes, but waiting for your response on these topics
first :)


‫בתאריך יום ג׳, 18 בפבר׳ 2025 ב-0:03 מאת ‪Hardy Hobeck‬‏ <‪
***@***.***‬‏>:‬

HHobeck:
> However, in my opinion there are 2 things to keep in mind: a) Which option would be more intuitive for users? b) What would be the performance implication?

Okay I would say just start with the simple path evaluation and if we see the need of using regular expressions, we can change it later.

> And speaking of performance, I noticed that the current implementation doesn't scale well. The next version calculation becomes prohibitively long.. This is due to the long patch calculation step, which calculates the diff between two trees. To overcome this, I think it requires slightly modifying the next version calculation step, to defer filtering to later time.

Would be good to have the performance in mind. But I think it is not so easy to do the filtering later, because e.g. the Mainline version strategy calculates the version by traversing the commits (which are not ignored).

Roei-Levi:
@HHobeck I've included a new implementation that uses regex to match paths.
I also tried to defer the logic which filters out, in the IncrementStrategyFinder, MainlineStrategy and NextVersionCalculator.

arturcic:
@Roei-Levi please rebase onto main instead of merging main into your branch. You need to exclude the merge commits from your PR

HHobeck:
I would prefer more clean code in GitRepository.cs:

```csharp
    private readonly ConcurrentDictionary<string, IReadOnlyList<string>?> patchPathsCache = new();

    public IReadOnlyList<string>? FindPatchPaths(ICommit commit, string? tagPrefix)
    {
        ArgumentNullException.ThrowIfNull(commit, nameof(commit));

        return patchPathsCache.GetOrAdd(commit.Sha, commitSha =>
        {
            if (!IsCommitTagged(commitSha, tagPrefix))
            {
                var innerCommit = this.RepositoryInstance.Commits.First(c => c.Sha == commitSha);
                Tree commitTree = innerCommit.Tree; // Main Tree
                Tree? parentCommitTree = innerCommit.Parents.FirstOrDefault()?.Tree; // Secondary Tree
                Patch patch = this.RepositoryInstance.Diff.Compare<Patch>(parentCommitTree, commitTree); // Difference
                return patch.Select(p => p.Path).ToList();
            }
            return null;
        });
    }

    private bool IsCommitTagged(string commitSha, string? tagPrefix)
    {
        if (string.IsNullOrEmpty(tagPrefix))
        {
            return false;
        }

        foreach (var tag in this.RepositoryInstance.Tags.Where(element => element.Target.Sha == commitSha))
        {
            if (tag.FriendlyName.StartsWith(tagPrefix))
            {
                return true;
            }
        }
        return false;
    }
```

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/MainlineVersionStrategy.cs
```diff
@@ -355,15 +357,20 @@ internal static BaseVersion DetermineBaseVersionRecursive(MainlineIteration iter
...
     {
-        MainlineContext context = new(incrementStrategyFinder, configuration)
+        MainlineContext context = new(incrementStrategyFinder, configuration, repository, gitverContext)
         {
             TargetLabel = targetLabel
         };
 
         foreach (var commit in iteration.Commits)
         {
+            if (configuration.Ignore.ToFilters(repository, gitverContext).OfType<PathFilter>().Any(f => f.Exclude(commit.Value, out _)))
```
HHobeck, line 369: This looks strange to me. In my opinion the iteration.Commits list should only consist the relevant commits. That mean in the IterateOverCommitsRecursive method the filtering should be alread happened.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/MainlineVersionStrategy.cs
```diff
@@ -355,15 +357,20 @@ internal static BaseVersion DetermineBaseVersionRecursive(MainlineIteration iter
     }
 
     private static IEnumerable<IBaseVersionIncrement> GetIncrements(MainlineIteration iteration, string? targetLabel,
-        IIncrementStrategyFinder incrementStrategyFinder, IGitVersionConfiguration configuration)
+        IIncrementStrategyFinder incrementStrategyFinder, IGitVersionConfiguration configuration, IGitRepository repository, GitVersionContext gitverContext)
```
HHobeck, line 360: IGitRepository and GitVersionContext parameter is not necessary. Please see comment below

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,44 @@
+using System.Text.RegularExpressions;
+using GitVersion.Extensions;
+using GitVersion.Git;
+
+namespace GitVersion.VersionCalculation;
+
+internal class PathFilter(IGitRepository repository, GitVersionContext context, IEnumerable<string> paths) : IVersionFilter
+{
+    public enum PathFilterMode { Inclusive = 0, Exclusive = 1 }
```
HHobeck, line 9: Is this enum still required?

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,44 @@
...
+    {
+        reason = null;
+
+        if (commit != null)
+        {
+            var patchPaths = repository.FindPatchPaths(commit, this.context.Configuration.TagPrefixPattern);
+
+            if (patchPaths != null)
+            {
+                if (this.paths.Any(path => patchPaths.All(p => Regex.IsMatch(p, path, RegexOptions.IgnoreCase))))
```
HHobeck, line 34: Regex.IsMatch parses the regular expression any time from new. Please cache it somehow and use the compile option.

src/GitVersion.Core/VersionCalculation/Mainline/MainlineContext.cs
```diff
@@ -4,11 +4,13 @@
 
 namespace GitVersion.VersionCalculation.Mainline;
 
-internal record MainlineContext(IIncrementStrategyFinder IncrementStrategyFinder, IGitVersionConfiguration Configuration)
+internal record MainlineContext(IIncrementStrategyFinder IncrementStrategyFinder, IGitVersionConfiguration Configuration, IGitRepository Repository, GitVersionContext GitverContext)
 {
     public IIncrementStrategyFinder IncrementStrategyFinder { get; } = IncrementStrategyFinder.NotNull();
 
     public IGitVersionConfiguration Configuration { get; } = Configuration.NotNull();
+    public IGitRepository Repository { get; } = Repository.NotNull();
```
HHobeck, line 12: Not necessary in my opinion to extend the Context with IGitRepository and GitVersionContext.

HHobeck (changes requested):
Thank you very much that you are investing time on this. I appriciate it very much. Please see me code review remarks. I think we need more then one iteration to finalize this PR. Happy coding!

src/GitVersion.Core/Configuration/IIgnoreConfiguration.cs
```diff
@@ -6,5 +6,7 @@ public interface IIgnoreConfiguration
 
     IReadOnlySet<string> Shas { get; }
 
-    bool IsEmpty => Before == null && Shas.Count == 0;
+    IReadOnlyCollection<string> Paths { get; }
+
+    public bool IsEmpty => Before == null && Shas.Count == 0 && Paths.Count == 0;
```
arturcic, line 11:
```suggestion
    bool IsEmpty => Before == null && Shas.Count == 0 && Paths.Count == 0;
```

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
```diff
@@ -53,6 +55,30 @@ public void DiscoverRepository(string? gitDirectory)
         });
     }
 
+    public IEnumerable<string>? FindPatchPaths(ICommit commit, string? tagPrefix)
```
arturcic, line 58: this piece should be moved to `RepositoryStore` instead, we keep IGitRepository and its implementation only as an abstraction on top of the git repository

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
```diff
@@ -53,6 +55,30 @@ public void DiscoverRepository(string? gitDirectory)
...
+    {
+        Patch? patch = null;
+        var innerCommit = this.RepositoryInstance.Commits.First(c => c.Sha == commit.Sha);
+        var match = new Regex($"^({tagPrefix ?? ""}).*$", RegexOptions.Compiled);
```
arturcic, line 62: please do not create Regex in line, consider moving it to https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/Core/RegexPatterns.cs

arturcic (changes requested):

HHobeck:
Hi @Roei-Levi,

Hope you are doing well. May I kindly ask you to rebase your branch with the latest version of main? Please let me know if you need help from the conceptual point of view.

Regards
Hardy

Roei-Levi:
Hey Hardy, sure I'll do it as soon as I find the time, hopefully in the
next couple of days.

בתאריך יום ג׳, 11 במרץ 2025, 10:28, מאת Hardy Hobeck ‏<
***@***.***>:

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
Roei-Levi: I'm not sure how exactly. Currently the GitRepository abstraction doesn't contain the neccessary interface to externally calculate the patch. It doesn't seem like a straightforward implementation, I'd prefer leaving it to you to decide how it should be done.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/MainlineVersionStrategy.cs
Roei-Levi: @HHobeck I've reverted the changes in the Mainline strategy. Now, if I understand correctly, the filtering would only occur when using NextVersionCalculator.

HHobeck:
@Roei-Levi thank's for your work. I have taken a look into the code and think we need some more code changes. At the moment the feature is not implemented completely:

1. The `EffectiveConfiguration` record class should not have additional dependencies. Fortunately the `VersionFilters` property has no usage. Please refactor the code and remove the `VersionFilters` property and replace it with the `Ignore` property of type `IIgnoreConfiguration`.
2. I think the usage of the `Ignore` property in the interface `IGitVersionConfiguration` is a good starting point to see the location where you need to do your changes. In my opinion the business logic to ignore the commits should be executed somewhere at the following locations:
image

Please notice the following class as well:
image

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
```diff
@@ -264,7 +269,8 @@ IEnumerable<NextVersion> GetNextVersionsInternal()
                             foreach (var baseVersion in versionStrategy.GetBaseVersions(effectiveBranchConfiguration))
                             {
                                 log.Info(baseVersion.ToString());
-                                if (IncludeVersion(baseVersion, configuration.Ignore))
+                                if (versionStrategy is not FallbackVersionStrategy || IncludeVersion(baseVersion, configuration.Ignore))
```
HHobeck, line 272: please revert the logic to exclude the FallbackVersionStrategy.The FallackVersionStrategy class is like every other version strategy class and should not handled explicit.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/BaseVersionOperand.cs
```diff
@@ -12,6 +12,8 @@ public BaseVersionOperand() : this(string.Empty, SemanticVersion.Empty)
 
     public string Source { get; init; } = Source.NotNull();
 
+    public VersionIncrementSourceType SourceType { get; init; } = SourceType;
```
HHobeck, line 15: Can you please explain me for what the enum `VersionIncrementSourceType` is good for? In my opinion the hole SourceType property can be removed.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/BaseVersion.cs
```diff
@@ -10,13 +10,13 @@ public BaseVersion() : this(new BaseVersionOperand())
...
-    public BaseVersion(string source, SemanticVersion semanticVersion, ICommit? baseVersionSource = null)
-        : this(new BaseVersionOperand(source, semanticVersion, baseVersionSource))
-    {
-    }
+    public BaseVersion(string source, SemanticVersion semanticVersion, ICommit? baseVersionSource = null, VersionIncrementSourceType sourceType = VersionIncrementSourceType.Tree)
+        : this(new BaseVersionOperand(source, semanticVersion, baseVersionSource, sourceType)) => this.SourceType = sourceType;
 
     public string Source => (Operator?.Source).IsNullOrEmpty() ? Operand.Source : Operator.Source;
 
+    public VersionIncrementSourceType SourceType { get; init; } = VersionIncrementSourceType.Tree;
```
HHobeck, line 18: Can you please explain me for what the enum `VersionIncrementSourceType` is good for? In my opinion the hole SourceType property can be removed.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/BaseVersionOperator.cs
```diff
@@ -6,6 +6,8 @@ public sealed record BaseVersionOperator : IBaseVersionIncrement
 {
     public string Source { get; init; } = string.Empty;
 
+    public VersionIncrementSourceType SourceType { get; } = VersionIncrementSourceType.Tree;
```
HHobeck, line 9: Can you please explain me for what the enum `VersionIncrementSourceType` is good for? In my opinion the hole SourceType property can be removed.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/IBaseVersionIncrement.cs
```diff
@@ -6,5 +6,7 @@ public interface IBaseVersionIncrement
 {
     string Source { get; }
 
+    VersionIncrementSourceType SourceType { get; }
```
HHobeck, line 9: Can you please explain me for what the enum `VersionIncrementSourceType` is good for? In my opinion the hole SourceType property can be removed.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/VersionIncrementSourceType.cs
```diff
@@ -0,0 +1,9 @@
+namespace GitVersion.VersionCalculation;
+
+public enum VersionIncrementSourceType
```
HHobeck, line 3: Can you please explain me for what the enum `VersionIncrementSourceType` is good for? In my opinion the hole SourceType property can be removed.

HHobeck (changes requested):

HHobeck:
I have found the following explanation:
The Windows and macOS file systems are case-insensitive (but case-preserving) by default. Most Linux filesystems are case-sensitive. Git was built originally to be the Linux kernel's version control system, so unsurprisingly, it's case-sensitive.

I think the regular expression in PathFilter.cs needs to be case sensitive. What do you think?

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
arturcic: I need to think about this. The reason we want to keep `IGitRepository` as lightweight as possible is to be able later to have a managed implementation of it. Now if we add to many Libgit2Sharp primitives to it, it will be much harder, especially if the contributor of these changes will not be able to help in supporting the changes

arturcic:
@HHobeck @asbjornu do you think we need to postpone v6.2 to include this PR or we can release v6.2 and move this PR to v6.*?

HHobeck:
> @HHobeck @asbjornu do you think we need to postpone v6.2 to include this PR or we can release v6.2 and move this PR to v6.*?

I think we should not postpone v6.2 because of this feature.

asbjornu:
> > @HHobeck @asbjornu do you think we need to postpone v6.2 to include this PR or we can release v6.2 and move this PR to v6.*?
> 
> I think we should not postpone v6.2 because of this feature.

I agree.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -51,15 +55,14 @@ public VersionField DetermineIncrementedField(
         var minorRegex = TryGetRegexOrDefault(configuration.MinorVersionBumpMessage, RegexPatterns.VersionCalculation.DefaultMinorRegex);
         var patchRegex = TryGetRegexOrDefault(configuration.PatchVersionBumpMessage, RegexPatterns.VersionCalculation.DefaultPatchRegex);
         var noBumpRegex = TryGetRegexOrDefault(configuration.NoBumpMessage, RegexPatterns.VersionCalculation.DefaultNoBumpRegex);
+        var pathFilters = configuration.Ignore.ToFilters(repository, versionContext.Value).OfType<PathFilter>();
```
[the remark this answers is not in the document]
Roei-Levi: I see.
Let me just clarify my concern, that if filtering based on paths is done eagerly, this will cause performance issues, especially for large commit history, due to the costly patch calculation (from what I saw on my machine, every patch calc takes as long as ~100ms). 
I believe that current implementation defers the filtering well till it is required. 
This issue applies generally whenever filtering needs to be done.

With that in mind, my suggestion is not to filter earlier, but rather later.
Would that make sense?

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
```diff
@@ -155,10 +156,10 @@ private SemanticVersion CalculateSemanticVersion(
 
     private NextVersion CalculateNextVersion(IBranch branch, IGitVersionConfiguration configuration)
     {
-        var nextVersions = GetNextVersions(branch, configuration);
+        var nextVersions = GetNextVersions(branch, configuration).OrderByDescending(v => v.IncrementedVersion);
```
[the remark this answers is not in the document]
Roei-Levi: The idea here is similar to the one before - defer the filtering of commits. Without these changes, the filtering would have happened on every single commit which is considered as a base version for the calculation. 
On the other hand, since it handles the base versions only, maybe many of commits won't be flowing through this filtering, therefore performance is less of an issue here IDK.

Regarding your comment about it being a safe guard - my understanding is that this part is responsible for filtering the base versions, which are only partially filtered by the commit SHA alone in `versionStrategy.GetBaseVersions` implementations.

Let me know what you think we should do here, and I'll adjust the logic accordingly.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/BaseVersionOperand.cs
Roei-Levi: It is used to distinguish between different strategies when processing a commit. 
Hoenstly, I'm not entirely sure what's the motivation here, let's just say that, if we decide to exclude commits regardless of the strategy, we can remove it entirely.

Roei-Levi:
> I have found the following explanation:
> 
> > The Windows and macOS file systems are case-insensitive (but case-preserving) by default. Most Linux filesystems are case-sensitive. Git was built originally to be the Linux kernel's version control system, so unsurprisingly, it's case-sensitive.
> 
> I think the regular expression in PathFilter.cs needs to be not case insensitive. What do you think?
> 
> Edit: Maybe we can use the (?i) mode inside the Regular expression: https://www.regular-expressions.info/modifiers.html

I agree, we can omit the `RegexOptions.IgnoreCase` in code.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/BaseVersionOperand.cs
HHobeck: > It is used to distinguish between different strategies when processing a commit.
> Hoenstly, I'm not entirely sure what's the motivation here, let's just say that, if we decide to exclude commits regardless of the strategy, we can remove it entirely.

Yes, I agree to remove it completely. I don't see any reason why we should treat the BaseVersion differently just because it was created through a specific version strategy.

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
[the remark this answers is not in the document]
HHobeck: It makes the code absolutely unmaintainable and error prone if we have a ignore configuration with different filtering methods, which are executed on different code level. I see your motivation... But we need to find a way where both aspects (performance/eager loading and business logic filtering) are considered. Does this make sense?

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
HHobeck: You see, this is not a good design (and makes it hard to understand) if you apply a implicit knowledge about the fact that the base version fo `FallbackVersionStrategy` is already evaluated. You are skipping it because of performance reason.. We need to find a design on an abstracion level e.g. ICommit interface where we ensure this aspect. We need to come to a design where we have transparent calls.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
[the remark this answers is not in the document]
HHobeck: I have tought about it and it is indeed very bad from the performance perspective, if in a potential scenario the `CommitsSinceTag` and `CommitsSinceVersionSource` variable will be calculated but the user don't use it.

This applies for `ManualDeployment` and `ContinuousDeployment` mode if you are only interested in e.g. the `SemVer` variable.

https://github.com/GitTools/GitVersion/blob/dd9434b0132a8030dca9e7d8f2633b1c8f9d67bc/src/GitVersion.Core/VersionCalculation/VersionCalculators/ManualDeploymentVersionCalculator.cs#L23-L26

and

https://github.com/GitTools/GitVersion/blob/dd9434b0132a8030dca9e7d8f2633b1c8f9d67bc/src/GitVersion.Core/VersionCalculation/VersionCalculators/ContinuousDeploymentVersionCalculator.cs#L23-L31

Roei-Levi:
Hey @HHobeck, a few updates:

1. After looking into the patch calculation logic a bit more, it turns out that the time-consuming part is about finding the correct commit to calculate the diff on, as in: `var innerCommit = this.RepositoryInstance.Commits.First(c => c.Sha == commitSha);
`. It is not the actual diff/patch calculation. It can be easily solved by a cache on commits, which makes things a lot easier.
2. My understanding is that calculating a diff does not require the actual repo to calculate it. It requires only the two commits  being compared, and a Diff functionality. Therefore, the `FindPatchPaths` can be moved out of the `GitRepository` class.
3. Following your suggestions, please see below a rough extended `Commit` implementation, which adds a PatchPaths property, describing the commit's paths to previous commit. Note that the diff functionality is still being done using the `Diff` in `Repository` class, but I think this can be put elsewhere, we can decide about it. This implementation greatly simplifies filtering later on, and indeed we should be able to use the same flow as we have today for filtering commits. wdyt?
4. If you recall, the `PathFilter` class has two modes of filtering - `Exclusive` and `Inclusive`. Now, I'm thinking maybe we should have something similar, but maybe different. For example, consider two filtering use cases - one where we want to match ANY path in a commit, and another where we want ALL paths to match. wdyt?

Here's the `Commit` class

```diff
@@ -1,5 +1,7 @@
+using System.Collections.Concurrent;
 using GitVersion.Extensions;
 using GitVersion.Helpers;
+using LibGit2Sharp;

 namespace GitVersion.Git;

@@ -8,6 +10,9 @@ internal sealed class Commit : GitObject, ICommit
     private static readonly LambdaEqualityHelper<ICommit> equalityHelper = new(x => x.Id);
     private static readonly LambdaKeyComparer<ICommit, string> comparerHelper = new(x => x.Sha);
     private readonly Lazy<IReadOnlyList<ICommit>> parentsLazy;
+    private static Lazy<IRepository>? repositoryLazy;
+    private static readonly ConcurrentDictionary<string, IReadOnlyList<string>?> patchPathsCache = new();
+    private static readonly ConcurrentDictionary<string, LibGit2Sharp.Commit> commitsShaCache = new();

     private readonly LibGit2Sharp.Commit innerCommit;

@@ -16,8 +21,51 @@ internal Commit(LibGit2Sharp.Commit innerCommit) : base(innerCommit)
         this.innerCommit = innerCommit.NotNull();
         this.parentsLazy = new(() => innerCommit.Parents.Select(parent => new Commit(parent)).ToList());
         When = innerCommit.Committer.When;
+        DiscoverRepository(".");
     }

+    // The following property can be ommitted if the Repository.Diff functionality is implemented elsewhere
+    private static IRepository RepositoryInstance
+    {
+        get
+        {
+            var lazy = repositoryLazy ?? throw new NullReferenceException("Repository not initialized. Call DiscoverRepository() first.");
+            return lazy.Value;
+        }
+    }
+
+    // The following method can be ommitted if the Repository.Diff functionality is implemented elsewhere
+    public static void DiscoverRepository(string? gitDirectory)
+    {
+        if (repositoryLazy != null)
+        {
+            return;
+        }
+        if (gitDirectory?.EndsWith(".git") == false)
+        {
+            gitDirectory = Repository.Discover(gitDirectory);
+        }
+        repositoryLazy = new(() => new Repository(gitDirectory));
+
+        foreach (var commit in RepositoryInstance.Commits)
+        {
+            commitsShaCache[commit.Sha] = commit;
+        }
+    }
+
+    public IReadOnlyList<string>? PatchPaths =>
+        patchPathsCache.GetOrAdd(this.Sha, commitSha =>
+        {
+            if (!commitsShaCache.TryGetValue(this.Sha, out var innerCommit))
+            {
+                return null;
+            }
+            Tree commitTree = innerCommit.Tree; // Main Tree
+            Tree? parentCommitTree = innerCommit.Parents.FirstOrDefault()?.Tree; // Secondary Tree
+            TreeChanges patch = RepositoryInstance.Diff.Compare<TreeChanges>(parentCommitTree, commitTree); // Difference
+            return patch.Select(element => element.Path).ToList();
+        });
+
     public int CompareTo(ICommit? other) => comparerHelper.Compare(this, other);
     public bool Equals(ICommit? other) => equalityHelper.Equals(this, other);
     public IReadOnlyList<ICommit> Parents => this.parentsLazy.Value;
```

Also, the IgnoreConfigurationExtensions.cs might be like:
```diff
@@ -1,10 +1,13 @@
 using GitVersion.Extensions;
 using GitVersion.Git;
+using GitVersion.VersionCalculation;

 namespace GitVersion.Configuration;

 internal static class IgnoreConfigurationExtensions
 {
+    private static PathFilter? pathFilter;
+
     public static IEnumerable<ITag> Filter(this IIgnoreConfiguration ignore, ITag[] source)
     {
         ignore.NotNull();
@@ -22,5 +25,20 @@ public static IEnumerable<ICommit> Filter(this IIgnoreConfiguration ignore, ICom
     }

     private static bool ShouldBeIgnored(ICommit commit, IIgnoreConfiguration ignore)
-        => !(commit.When <= ignore.Before) && !ignore.Shas.Contains(commit.Sha);
+    {
+        if (ignore.IsEmpty)
+        {
+            return false;
+        }
+        pathFilter ??= ignore.ToFilters().OfType<PathFilter>().FirstOrDefault();
+
+        if (ignore.Before.HasValue && commit.When <= ignore.Before ||
+            ignore.Shas.Contains(commit.Sha) ||
+            pathFilter?.Exclude(commit, out var _) == true)
+        {
+            return true;
+        }
+
+        return false;
+    }
 }
```

HHobeck:
Thank you for your input. We need to refactore it step by step.

Can you please start with:

1.

1.
```

**Comment by HHobeck** [comment]
Thank you for your input. I think we need to do a refactoring of the code step by step.

Please start with the following points:

1. Remove VersionFilters property in EffectiveConfiguration.cs and replace it with the Ignore property.

1. Change the interface and add (for now) an new method:
```csharp
public interface IVersionFilter
{
    bool Exclude(IBaseVersion baseVersion, out string? reason);

    bool Exclude(ICommit commit, out string? reason);
}
```

1. Change the ToFilter extension method in ConfigurationExtensions.cs as following:
```csharp
public static IEnumerable<IVersionFilter> ToFilters(this IGitVersionConfiguration configuration)
{
    configuration.NotNull();

    var source = configuration.Ignore;

    if (source.Shas.Count != 0) yield return new ShaVersionFilter(source.Shas);
    if (source.Before.HasValue) yield return new MinDateVersionFilter(source.Before.Value);
    if (source.Paths.Count != 0) yield return new PathFilter(configuration.TagPrefixPattern, source.Paths);
}
```

1. Move the FindPatchPaths logic in GitRepository.cs to the PathFilter.cs class (this is necessary to encapsulate the logic and remove the dependency to the Repository class)

1. Instead of  `this.RepositoryInstance.Commits.First(c => c.Sha == commitSha);` use ((Commit)commit).InnerCommit. Therefore you need to change the Commit.cs class an make the innerCommit public.

Does that make sense to you? Please let me know what do you think.


> If you recall, the PathFilter class has two modes of filtering - Exclusive and Inclusive. Now, I'm thinking maybe we should have something similar, but maybe different. For example, consider two filtering use cases - one where we want to match ANY path in a commit, and another where we want ALL paths to match. wdyt?

Can you please give some use cases where I would use the one to the other?

HHobeck:
> 2\. Not a prob, but may I ask why would you want to do that?

In the `IgnoreConfigurationExtensions::Filter` you are filtering the tags based on the commit. The business logic here is violating the separation of concern aspect because all filtering logic needs to be go to the same code path. We need to have one place for filtering and this is in my opinion the implementation of IVersionFilter. Of course maybe we need to rename it to ICommitFilter. Anyway, to use the native BL we need to extend the interface `IVersionFilter` and make it usable for ICommit as well. Please let me know what do you think.

> 3\. No problem. Just as a remark, if the reason for this refactor is merely for accessing TagPrefixPattern, then IMHO the `TagPrefixPattern` shouldn't be a part of `PathFilter` - it is used to skip the filtering of commits that are tagged with this prefix – an equivalent logic can be implemented at the caller to `PathFilter.Exclude()`.

Yes exactly that was the reason. Yes agree we can place this outside of the PathFilter filtering and say it is BL to filter the tags.

> 4\. This is the most problematic part of the implementation. Perhaps I'm missing your entire design. You suggest moving `FindPatchPaths` out to `PathFilter`, yet it conflicts with previous requests in this PR, to keep the `LibGit2Sharp` dependencies only in the namespace `GitVersion.Git`. The diff/patch calculation is the issue here, which requires two things basically: `LibGit2Sharp.Commit` instances, and the actual diff logic. The actual diff logic is currently implemented in `LibGit2Sharp.Repository.Diff` – but as I mentioned in the previous comment, it is not mandatory, and I believe it can be implemented outside of it, if needed. In short, this is the reason I suggested extending the `GitVersion.Git.Commit` class previously.

I agree and follow your argumentation. Important is that we ensure that the IGitRepository is available in PathFilter.

> 5\. I think this is discouraged, since it: a) breaks the encapsulation; b) introduces a dependency in `LibGit2Sharp`, where it might not be needed.

I agree it is not nice. But it solves the performance problem (hopefully). You need to keep in mind that RepositoryInstance.Commit is of type IQuerable (means it will be evaluated each time?):

image

> I understand. What if the suggested logic is implemented elsewhere as an extension to this class? If the `GitVersion.Git.Commit` is just a wrapper to the `LibGit2Sharp.Commit`, then I think it won't make sense to make `innerCommit` publicly available.

Okay, can you do a prototype how the extension will look like? In my opinion following needs to be considered:
- the call needs to go to one instance of IGitRepository (no extra initialization of GitLib2Sharp.IRepository)
- FindPatchPaths needs to be cached because of performance reason

> Well, I guess the "typical" use case would be calculating the sem ver for a sub project in a repository, thus a user would want GitVersion to consider commits where ANY changed paths match the sub project path.
> But another use case might be, where a user wants GitVersion to account for those commits where ONLY the relevant (i.e. non-ignored) paths where changed. If, for example, this user would want to avoid GitVersion processing a commit where multiple sub projects where changed at once. In such cases, the commit message might not correctly reflect the change that was made to all sub projects, leading to an incorrect version bump.

Thank you for explanation. Can we postpone this discussion for a moment and concentrate first on the exclusion filter? If we have this in place it is easy to introduce a new include filter.

Roei-Levi:
Let me try to refine the points that I think we would like to agree upon, before creating a prototype implementation... 

1. **Where should reside the logic of finding a Commit's patch (diff from parent)?** I see the following options: 
    a. Implement the logic in `Commit` class itself. To do so, add `PatchPaths` property in `ICommit`, `Commit`, as well as a new dependency in `Commit` constructor: `GitLib2Sharp.IRepository`, or `GitLib2Sharp.Diff`. The repo/diff instance must be used to implement `PatchPaths`.
    b. Create a `FilterableCommit`, `IFilterableCommit`, which inherit from `Commit`, `ICommit`. Implement the logic as in option (a) but in subclass. This requires making the `Commit` class not `sealed`, and `innerCommit` `protected` instead of `private`.
    c. Implement the logic in a seperate static class, say we call it `CommitExtensions` (as in `ConfigurationExtensions` etc.). Make `innerCommit` `internal`/`public`, instead of `private` (break encapsulation), to make it accessible in the extension class. 
    d. Implement the logic in a seperate static class, as in option (c), but without breaking encapsulation. To do so, extend `GitVersion.Git` with types of `Tree`, `Diff`, and `TreeChanges`, which will be very simple wrappers around the corresponding `GitLib2Sharp` types. Then, expose `commitTree` in `GitVersion.Git.Commit`, and expose `Diff` in `GitVersion.Git.GitRepository` - both are required to calculate `PatchPaths`. Then (unless I'm missing smthing 😄) we should be able to implement `FindPatches` as an extension, which accepts (existing) instance of `IGitRepository` as argument, to calculate the patches (diff).

Unless you're ok with extending logic of `Commit` class, we can try option (d) above.

2. **How to solve the performance issue when finding a commit**, i.e. `this.RepositoryInstance.Commits.First(c => c.Sha == commit.Sha);`?
    I believe this is solved in all options above, since this kind of lookup is no longer needed, the appplicable commit properties are accesible. Patch calculation itself is quite fast.
 
I think we already agree on the other points, so no need to repeat.

Please let me know that you think :)

arturcic:
> Let me try to refine the points that I think we would like to agree upon, before creating a prototype implementation...
> 
> 1. **Where should reside the logic of finding a Commit's patch (diff from parent)?** I see the following options:
>    a. Implement the logic in `Commit` class itself. To do so, add `PatchPaths` property in `ICommit`, `Commit`, as well as a new dependency in `Commit` constructor: `GitLib2Sharp.IRepository`, or `GitLib2Sharp.Diff`. The repo/diff instance must be used to implement `PatchPaths`.
>    b. Create a `FilterableCommit`, `IFilterableCommit`, which inherit from `Commit`, `ICommit`. Implement the logic as in option (a) but in subclass. This requires making the `Commit` class not `sealed`, and `innerCommit` `protected` instead of `private`.
>    c. Implement the logic in a seperate static class, say we call it `CommitExtensions` (as in `ConfigurationExtensions` etc.). Make `innerCommit` `internal`/`public`, instead of `private` (break encapsulation), to make it accessible in the extension class.
>    d. Implement the logic in a seperate static class, as in option (c), but without breaking encapsulation. To do so, extend `GitVersion.Git` with types of `Tree`, `Diff`, and `TreeChanges`, which will be very simple wrappers around the corresponding `GitLib2Sharp` types. Then, expose `commitTree` in `GitVersion.Git.Commit`, and expose `Diff` in `GitVersion.Git.GitRepository` - both are required to calculate `PatchPaths`. Then (unless I'm missing smthing 😄) we should be able to implement `FindPatches` as an extension, which accepts (existing) instance of `IGitRepository` as argument, to calculate the patches (diff).
> 
> Unless you're ok with extending logic of `Commit` class, we can try option (d) above.
> 
> 2. **How to solve the performance issue when finding a commit**, i.e. `this.RepositoryInstance.Commits.First(c => c.Sha == commit.Sha);`?
>    I believe this is solved in all options above, since this kind of lookup is no longer needed, the appplicable commit properties are accesible. Patch calculation itself is quite fast.
> 
> I think we already agree on the other points, so no need to repeat.
> 
> Please let me know that you think :)

My expectation is in any way we decouple from LibGit2Sharp and not couple more, as the main goal long term is to have another implementation of access to git which might use a managed code instead of LibGit2Sharp. So in that case we need to carefully think of "extending" any lower level classes/interfaces

HHobeck:
> Unless you're ok with extending logic of Commit class, we can try option (d) above.

Encapsulating is fine with me.

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,51 @@
+using System.Diagnostics.CodeAnalysis;
+using System.Text.RegularExpressions;
+using GitVersion.Git;
+
+namespace GitVersion.VersionCalculation;
+
+internal class PathFilter( IEnumerable<string> paths) : IVersionFilter
+{
+    private readonly List<Regex> pathsRegexes = paths.Select(path => new Regex(path, RegexOptions.IgnoreCase | RegexOptions.Compiled)).ToList();
+    private readonly Dictionary<string, bool> pathMatchCache = [];
```
HHobeck, line 10: Maybe ConcurrentDictionary?

src/GitVersion.LibGit2Sharp/Git/Commit.cs
```diff
@@ -1,30 +1,46 @@
...
     public bool Equals(ICommit? other) => equalityHelper.Equals(this, other);
     public IReadOnlyList<ICommit> Parents => this.parentsLazy.Value;
     public DateTimeOffset When { get; }
     public string Message => this.innerCommit.Message;
+    // TODO implement tag prefix filtering before returning the paths.
+    public IEnumerable<string> DiffPaths
+    {
+        get
+        {
+            if (pathsCache.TryGetValue(this.Sha, out var paths))
```
HHobeck, line 35: You are only reading from the pathsCache member.

schemas/6.1/GitVersion.configuration.json
```diff
@@ -189,6 +189,13 @@
             "null"
           ]
         },
+        "paths": {
```
arturcic, line 192: please revert this file, please rebase on latest main and if needed you can regenerate this file

arturcic (changes requested):

HHobeck (changes requested):
Your code changes are looking good. Great work. Can you write some additional integration tests to ensue this functionality please?

Roei-Levi:
Hey @HHobeck,
I added tests. 
Also, I have a few points in mind that might require attention:
1. Regarding existing `MinDateVersionFilter` - it has slightly changed. It now filters only commits whose commit date is strictly less than the requested 'before' date. (Originally, there was an incosistency in the implementation, where in one place the filtering was like `commit.When < before` in one place, and `commit.When <= before` in another place, so I chose the second option).
2. The current PathFilter doesn't take into consideration the fact that filtered commits might be tagged. So, in real use cases, a tag which is assigned to filtered commits, will not be considered as a base version. IMHO, this could be ok in most scenarios. What do you think?

src/GitVersion.Core/Git/ICommit.cs
```diff
@@ -7,4 +7,5 @@ public interface ICommit : IEquatable<ICommit?>, IComparable<ICommit>, IGitObjec
     DateTimeOffset When { get; }
 
     string Message { get; }
+    IReadOnlyList<string> DiffPaths { get; }
```
HHobeck, line 11: Please move the property to the next line (formatting issue)

src/GitVersion.Core/VersionCalculation/Abstractions/IVersionFilter.cs
```diff
@@ -1,6 +1,9 @@
+using GitVersion.Git;
+
 namespace GitVersion.VersionCalculation;
 
 public interface IVersionFilter
 {
     bool Exclude(IBaseVersion baseVersion, out string? reason);
```
HHobeck, line 7: Do you think we need the IBaseVersion variation? In my opinion all base version who are returned by IVersionStrategy should be already valid.

src/GitVersion.LibGit2Sharp/Git/Commit.cs
```diff
@@ -1,30 +1,48 @@
...
         When = innerCommit.Committer.When;
+        this.repoDiff = repoDiff;
     }
 
     public int CompareTo(ICommit? other) => comparerHelper.Compare(this, other);
     public bool Equals(ICommit? other) => equalityHelper.Equals(this, other);
     public IReadOnlyList<ICommit> Parents => this.parentsLazy.Value;
     public DateTimeOffset When { get; }
     public string Message => this.innerCommit.Message;
+    // TODO implement tag prefix filtering before returning the paths.
```
HHobeck: What about this todo?

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,43 @@
+using System.Collections.Concurrent;
+using System.Diagnostics.CodeAnalysis;
+using System.Text.RegularExpressions;
+using GitVersion.Git;
+
+namespace GitVersion.VersionCalculation;
+
+internal class PathFilter(IReadOnlyList<string> paths) : IVersionFilter
+{
+    private readonly List<Regex> pathsRegexes = [.. paths.Select(path => new Regex(path, RegexOptions.Compiled))];
```
HHobeck: Maybe use a IReadOnlyList data type here

HHobeck:
In my opinion the following steps are:
- implement include pattern (??)
- reabase with main
- document this feature (e.g. the need of defining the regular expression to case insensitive on windows)

HHobeck:
> 1. Regarding existing `MinDateVersionFilter` - it has slightly changed. It now filters only commits whose commit date is strictly less than the requested 'before' date. (Originally, there was an incosistency in the implementation, where in one place the filtering was like `commit.When < before` in one place, and `commit.When <= before` in another place, so I chose the second option).

That is fine. Probably you have the date time of the first commit you want to filter. Your change makes sense to me.

> 2. The current PathFilter doesn't take into consideration the fact that filtered commits might be tagged. So, in real use cases, a tag which is assigned to filtered commits, will not be considered as a base version. IMHO, this could be ok in most scenarios. What do you think?

This would be also my expectation. If you have a tag on a filtered commit this tag will be ignored.

src/GitVersion.Core/VersionCalculation/Abstractions/IVersionFilter.cs
Roei-Levi: I suppose not, since as you mention filtering has already happened by the time we exclude base version. 
I can remove these.

src/GitVersion.LibGit2Sharp/Git/Commit.cs
Roei-Levi: no longer relevant.

Roei-Levi:
> In my opinion the following steps are:
> 
> * implement include pattern (??)
> * reabase with main
> * document this feature (e.g. the need of defining the regular expression to by case insensitive on windows)

Great. 
regrading include pattern - what do you mean by that? Is this the exclusive/inclusive modes of filtering?
regrading the documentation - how/where would you like it to be written?

HHobeck:
Good morning @Roei-Levi,

> regrading include pattern - what do you mean by that? Is this the exclusive/inclusive modes of filtering?

Yes correct. This is what you also mentioned last time and we have posponed the discussion, right?

> regrading the documentation - how/where would you like it to be written?

The documentation is public available [here](https://gitversion.net/docs/reference/configuration). You can edit the pages at following location:
- https://github.com/GitTools/GitVersion/tree/main/docs/input/docs

The configuration options are documented here:
- https://github.com/GitTools/GitVersion/blob/main/docs/input/docs/reference/configuration.md

The breaking changes are documented here (not sure if this PR inroduces breaking changes):
- https://github.com/GitTools/GitVersion/blob/main/BREAKING_CHANGES.md

(Please see also at https://github.com/GitTools/GitVersion/blob/main/CONTRIBUTING.md#contribution-guidelines for additional information)

src/GitVersion.Core/VersionCalculation/Abstractions/IVersionFilter.cs
HHobeck: I have thought about it and think this is something which is out of scope of this PR. Because you are having a cache in place the additional check of IBaseVersion might be not a performance problem, right? If so I would say to skpip this refactoring and do it next time. What do you think?

asbjornu:
Do we need an `include` pattern? How would it interact with an `exclude` pattern? Would having both be an error? I think a possible way forward is to ship `exclude` and then add `include` if we see a need for it.

Regarding breaking changes, I think the addition of a configuration property as this should be safe as long as existing tests pass.

Roei-Levi:
In my view it will be good to have both filtering modes, since they could serve different use cases. The two modes (exclusive and inclusive) aren't about including vs excluding a path, but rather about how to handle multiple paths in a commit - i.e. consider any of the paths, or all paths.
Anyway, I think we can continue with having only one mode (inclusive, which is, I belive, the most useful). I thought it's a good idea to add the foundation for implementation of the modes, in case it will be required in the future... 

Added docs, that currently describe the paths filter as of now (inclusive mode only).

HHobeck:
> In my view it will be good to have both filtering modes, since they could serve different use cases.

How would the logic look like for Exclusive?

```csharp
  public bool Exclude(ICommit? commit, [NotNullWhen(true)] out string? reason)
  {
      reason = null;

      if (commit != null)
      {
          switch (mode)
          {
              case PathFilterMode.Inclusive:
                  {
                      if (commit.DiffPaths.All(this.IsMatch))
                      {
                          reason = "Source was ignored due to all commit paths matching ignore regex";
                          return true;
                      }
                      break;
                  }
          }
      }
      return false;
  }
```

I'm asking because I'm not certain if we have the same use cases in mind.

HHobeck:
> Added docs, that currently describe the paths filter as of now (inclusive mode only).

Or let me describe the scenario that I think is most common. In a mono repository you have multiple products which are separated from each other. The products have there own release and bugfix cycles with it's own version increment. Let's say you have a product A in path `/ProductA/` and a product B in path `/ProductB/` which are sharing a common library C in path `/LibraryC/`.

Suppose you are in the role of a release manager for product A and you want a version increase when something changes in the path `/ProductA/*` or `/LibraryC/*`.

The commit should be not ignored when changes existing in path:
- `/ProductA/*`
- `/ProductA/*` and `/ProductB/*`
- `/ProductA/*` and `/ProductB/*` and `/LibraryC/*`
- `/ProductA/*` and  `/LibraryC/*`
- `/ProductB/*` and `/LibraryC/*`

The commit should be ignored when changes existing in path:
- `/ProductB/*`

How would the configuration look like if you want to use the regular expression for path `/ProductA/*` and `/LibraryC/*`?

Roei-Levi:
> How would the logic look like for Exclusive?

similar, with different exclusion criteria:
```
public bool Exclude(ICommit? commit, [NotNullWhen(true)] out string? reason)
{
    reason = null;

    if (commit != null)
    {
        switch (mode)
        {
            case PathFilterMode.Exclusive:
                {
                    if (commit.DiffPaths.Any(this.IsMatch))
                    {
                        reason = "Source was ignored due to one or more commit paths matching ignore regex";
                        return true;
                    }
                    break;
                }
            case PathFilterMode.Inclusive:
                {
                    if (commit.DiffPaths.All(this.IsMatch))
                    {
                        reason = "Source was ignored due to all commit paths matching ignore regex";
                        return true;
                    }
                    break;
                }
        }
    }

    return false;
}
```

> How would the configuration look like if you want to use the regular expression for path `/ProductA/*` and `/LibraryC/*`?

I guess it can be either a regex containing ProductB, or more generically a negative lookahead on everything which is not ProductA or LibraryC. Meaning either of:
- `^\/ProductB\/.*`
- `^(?!\/ProductA\/|\/LibraryC\/).*`


> The commit should be not ignored when changes existing in path:
> 
> * `/ProductA/*`
> * `/ProductA/*` and `/ProductB/*`
> * `/ProductA/*` and `/ProductB/*` and `/LibraryC/*`
> * `/ProductA/*` and  `/LibraryC/*`
> * `/ProductB/*` and `/LibraryC/*`

In this case, the `Inclusive` mode, as I called it, taking into consideration ALL paths in a commit, would handle this case well. 
does that make sense?

HHobeck:
> In this case, the `Inclusive` mode, as I called it, taking into consideration ALL paths in a commit, would handle this case well.
> does that make sense?

Yes this makes sense, aboslutely. Thank's for clarification. If you not mind please use this or a similar scenario with example in the documentation section. This feature is awesome. Thank you for your work.

HHobeck (approved):

HHobeck:
The configuaration.md doc has been reverted. Why? @arturcic

HHobeck:
This branch is out-of-date with the base branch. Are you planning to do a rebase? Please recover your documentation change as well.

Roei-Levi:
I can do it. Is the issue with the docs being reverted fixed?

HHobeck:
> I can do it. Is the issue with the docs being reverted fixed?

What do you mean? @arturcic reverted the documentation. Don't know why. Is it generated from somewhere else?

Roei-Levi:
I think it is automatic, it is not manually committed. I didn't check why I
thought you know why..

בתאריך יום ב׳, 23 ביוני 2025, 19:08, מאת Hardy Hobeck ‏<
***@***.***>:

arturcic:
@HHobeck @Roei-Levi there is this workflow
https://github.com/GitTools/GitVersion/blob/main/.github/workflows/mkdocs.yml
It is compiling the output `configuration.md` from the source https://github.com/GitTools/GitVersion/blob/main/docs/input/docs/reference/mdsource/configuration.source.md, so if you need to make changes to the `configuration.md` you need to actually change the `configuration.source.md` instead

Roei-Levi:
@HHobeck it is now up to date with main.

HHobeck:
> @HHobeck it is now up to date with main.

One last question: Is it possible to document the fact that the ignore path is case sensitive and that this might be problematic on windows based systems? You can metigate this by defining regular expressions with case insensitive option like: `(?i)abc`

Roei-Levi:
> > @HHobeck it is now up to date with main.
> 
> One last question: Is it possible to document the fact that the ignore path is case sensitive and that this might be problematic on windows based systems? You can metigate this by defining regular expressions with case insenstive option like: `(?i)abc`

How about
```
Note: The `ignore.paths` configuration is case-sensitive. This can lead to unexpected behavior on case-insensitive file systems, such as Windows. 
To ensure consistent matching regardless of case, you can prefix your regular expressions with the case-insensitive flag `(?i)`. 
For example, `(?i)^docs\/` will match both `docs/` and `Docs/`. 
```

HHobeck:
Approved. @arturcic @asbjornu

arturcic:
> Approved. @arturcic @asbjornu 

I will hopefully have a look this week

src/GitVersion.Core/VersionCalculation/PathFilter.cs
```diff
@@ -0,0 +1,57 @@
...
+        }
+        return isMatch;
+    }
+
+    public bool Exclude(ICommit? commit, [NotNullWhen(true)] out string? reason)
+    {
+        reason = null;
+
+        if (commit != null)
+        {
```
asbjornu: Please avoid nesting by returning early.

```suggestion
        if (commit == null)
        {
            return false;
        }
```

asbjornu (changes requested):
This is looking really good now! Just one little nit before it can be merged, imho.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/ConfiguredNextVersionVersionStrategy.cs
```diff
@@ -18,7 +18,7 @@ public IEnumerable<BaseVersion> GetBaseVersions(EffectiveBranchConfiguration con
     {
         configuration.NotNull();
 
-        if (!Context.Configuration.VersionStrategy.HasFlag(VersionStrategies.ConfiguredNextVersion))
+        if (!this.Context.Configuration.VersionStrategy.HasFlag(VersionStrategies.ConfiguredNextVersion))
```
arturcic: please revert this part, the `Context` field is used without `this.` in this file, let's keep it consistent

arturcic:
> This is looking really good now! Just one little nit before it can be merged, imho.

Agree, after applying @asbjornu's and mine suggestions, @Roei-Levi do you mind to work a bit on the commit history? Can you check what should be squashed together to make it a bit more descriptive. After that you have approval from me as well. Please rebase your branch onto `main` as well

arturcic (changes requested):

arturcic (approved):

arturcic:
Thank you @Roei-Levi for taking the time and the patience to create this PR.

HHobeck (approved):

arturcic:
@mergifyio refresh

CharliePoole:
Will a new release containing this be out soon? Or is there a feed where the build is available?

asbjornu:
@CharliePoole, things are going a bit slow now during summer vacationing, but I expect us to cut a 6.4.0 release in a few weeks time.

arturcic:
> @CharliePoole, things are going a bit slow now during summer vacationing, but I expect us to cut a 6.4.0 release in a few weeks time.

as @asbjornu we will have a release soon, but you can also consume the package from GitHub packages https://github.com/GitTools/GitVersion/pkgs/nuget/GitVersion.Tool

CharliePoole:
Thanks to both of you... I'll do some experiments with the pre-release.

PR #4432, opened 2025-02-22
Support for dotted versions of the GitVersion config files by default

### LINKED ISSUE #4431: [Feature]: Support for dotted versions of the GitVersion config files by default
### Prerequisites

### GitVersion package

GitVersion.Tool

### GitVersion version

current main

### Operating system

N/A

### What are you seeing?

Currently only `GitVersion.yml` and `GitVersion.yaml` are searched if no config filename is specified on the command-line. Many other tools follow a convention of having a leading dot for their configuration files, e.g. `.gitignore`, `.editorconfig`, `.gitattributes`, etc.

### What is expected?

Look for the following configuration files by default: `GitVersion.yml`, `GitVersion.yaml`, `.GitVersion.yml` and `.GitVersion.yaml`  

### Steps to Reproduce

Run `gitversion` and have your configuration only in a dotted config file (`.GitVersion.yml` or `.GitVersion.yaml`).

Bi0T1N:
## Description
Follow the convention of many other tools which use a leading dot for their configuration files, e.g. `.gitignore`, `.editorconfig`, `.gitattributes`, etc.

## Related Issue
Closes #4431 

## Motivation and Context
With this change GitVersion will look for the following configuration files by default:  
`GitVersion.yml`, `GitVersion.yaml`, `.GitVersion.yml` and `.GitVersion.yaml` (in this order)

## How Has This Been Tested?
By extending the unit tests.

## Checklist:

### DIFF
--- a/src/GitVersion.Configuration/ConfigurationFileLocator.cs
+++ b/src/GitVersion.Configuration/ConfigurationFileLocator.cs
@@ -74,7 +74,7 @@ internal class ConfigurationFileLocator(
 
         if (!hasConfigInProjectRootDirectory && !hasConfigInWorkingDirectory)
         {
-            if (!SupportedConfigFileNames.Any(entry => entry.Equals(this.ConfigurationFile)))
+            if (!SupportedConfigFileNames.Any(entry => entry.Equals(this.ConfigurationFile, StringComparison.OrdinalIgnoreCase)))
             {
                 workingConfigFile = PathHelper.Combine(workingDirectory, this.ConfigurationFile);
                 projectRootConfigFile = PathHelper.Combine(projectRootDirectory, this.ConfigurationFile);
--- a/src/GitVersion.Configuration/ConfigurationFileLocator.cs
+++ b/src/GitVersion.Configuration/ConfigurationFileLocator.cs
@@ -15,7 +15,7 @@ internal class ConfigurationFileLocator(
     public const string DefaultAlternativeFileName = "GitVersion.yaml";
     public const string DefaultFileNameDotted = $".{DefaultFileName}";
     public const string DefaultAlternativeFileNameDotted = $".{DefaultAlternativeFileName}";
-    public List<string> PossibleConfigFileNames = [DefaultFileName, DefaultAlternativeFileName, DefaultFileNameDotted, DefaultAlternativeFileNameDotted];
+    public List<string> SupportedConfigFileNames = [DefaultFileName, DefaultAlternativeFileName, DefaultFileNameDotted, DefaultAlternativeFileNameDotted];
 
     private readonly IFileSystem fileSystem = fileSystem.NotNull();
     private readonly ILog log = log.NotNull();
@@ -33,7 +33,7 @@ internal class ConfigurationFileLocator(
     public string? GetConfigurationFile(string? directory)
     {
         if (directory is null) return null;
-        var candidateList = new List<string>(PossibleConfigFileNames);
+        var candidateList = new List<string>(SupportedConfigFileNames);
         if (!this.ConfigurationFile.IsNullOrEmpty())
         {
             // give configuration value the highest priority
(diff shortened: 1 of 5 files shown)


src/GitVersion.Configuration/ConfigurationFileLocator.cs
```diff
@@ -13,6 +13,9 @@ internal class ConfigurationFileLocator(
 {
     public const string DefaultFileName = "GitVersion.yml";
     public const string DefaultAlternativeFileName = "GitVersion.yaml";
+    public const string DefaultFileNameDotted = $".{DefaultFileName}";
+    public const string DefaultAlternativeFileNameDotted = $".{DefaultAlternativeFileName}";
+    public List<string> PossibleConfigFileNames = [DefaultFileName, DefaultAlternativeFileName, DefaultFileNameDotted, DefaultAlternativeFileNameDotted];
```
asbjornu, line 18: Perhaps `Supported` is a better word to use here? 

```suggestion
    public List<string> SupportedConfigFileNames = [DefaultFileName, DefaultAlternativeFileName, DefaultFileNameDotted, DefaultAlternativeFileNameDotted];
```

src/GitVersion.Configuration/ConfigurationFileLocator.cs
```diff
@@ -65,7 +74,7 @@ private void WarnAboutAmbiguousConfigFileSelection(string? workingDirectory, str
 
         if (!hasConfigInProjectRootDirectory && !hasConfigInWorkingDirectory)
         {
-            if (this.ConfigurationFile is not (DefaultFileName or DefaultAlternativeFileName))
+            if (!PossibleConfigFileNames.Any(entry => entry.Equals(this.ConfigurationFile)))
```
asbjornu, line 77: Should we make the file comparison here case insensitive as well? I suppose that could be considered a breaking change, as someone might be crazy enough to have both `gitversion.yml` and `GitVersion.yml` on an ext4 or similar case sensitive file system, but I think `.gitversion.yml` somehow looks more Unix idiomatic than `.GitVersion.yml`. Thoughts?

asbjornu (commented):
Looks good to me!

src/GitVersion.Configuration/ConfigurationFileLocator.cs
arturcic: I agree, let's make it insensitive, I don't think we should treat it as breaking change as this does not affect most of the users.

src/GitVersion.Configuration/ConfigurationFileLocator.cs
```diff
@@ -30,11 +33,17 @@ public void Verify(string? workingDirectory, string? projectRootDirectory)
     public string? GetConfigurationFile(string? directory)
     {
         if (directory is null) return null;
-        string?[] candidates = [this.ConfigurationFile, DefaultFileName, DefaultAlternativeFileName];
-        var candidatePaths =
-            from candidate in candidates
-            where !candidate.IsNullOrWhiteSpace()
-            select PathHelper.Combine(directory, candidate);
+        var candidateList = new List<string>(PossibleConfigFileNames);
```
arturcic, line 36: I'd prefer to keep the old syntax

```suggestion
        string?[] candidates = [this.ConfigurationFile, ..SupportedConfigFileNames];
        var candidatePaths =
            from candidate in candidates
            where !candidate.IsNullOrWhiteSpace()
            select PathHelper.Combine(directory, candidate);
```

arturcic (changes requested):

src/GitVersion.Configuration/ConfigurationFileLocator.cs
arturcic: Agree with the suggestion

Bi0T1N:
All suggestions applied. :+1:

arturcic:
please run `dotnet format ./src/ --exclude **/AddFormats/` to fix formatting

Bi0T1N:
> please run `dotnet format ./src/ --exclude **/AddFormats/` to fix formatting

Its the line you suggested: `string?[] candidates = [this.ConfigurationFile, ..SupportedConfigFileNames];`
Changing it to `string?[] candidates = [this.ConfigurationFile, .. SupportedConfigFileNames];` fixes the warning but looks weird to me.

arturcic:
> > please run `dotnet format ./src/ --exclude **/AddFormats/` to fix formatting
> 
> Its the line you suggested: `string?[] candidates = [this.ConfigurationFile, ..SupportedConfigFileNames];` Changing it to `string?[] candidates = [this.ConfigurationFile, .. SupportedConfigFileNames];` fixes the warning but looks weird to me.

`string?[]` it's because `this.ConfigurationFile` can be null, that's why we have `where !candidate.IsNullOrWhiteSpace()`

Bi0T1N:
To make the linter happy you need to have a whitespace between the two dots `..` and the variable `SupportedConfigFileNames`. However, fixed and pipeline succeeded now.

PR #4483, opened 2025-04-02
Make sure custom configuration file path is returned when custom configuration is set

### LINKED ISSUE #4480: [ISSUE]: 6.2.0 seems to be adding unnecessary pre-release suffix to the tag
### Prerequisites

### GitVersion package

GitVersion.Tool

### GitVersion version

6.2.0

### Operating system

Windows, macOS, Linux

### What are you seeing?

Adds a pre-release suffix differently regardless of the operating system:

```
❯ dotnet-gitversion version
6.2.0+Branch.main.Sha.c2bc09f9ee048c6bdce6ec494fa48cb74420f608
❯ dotnet-gitversion . /config ../common-ci/GitVersion.yml /format '{SemVer}'
5.0.64-1
```

You can see our GitVersion.yml here: https://github.com/51Degrees/common-ci/blob/main/GitVersion.yml

### What is expected?

Previously it was not adding a pre-release suffix:
```
❯ dotnet-gitversion version
6.1.0+Branch.main.Sha.8856e3041dbb768118a55a31ad4e465ae70c6767
❯ dotnet-gitversion . /config ../common-ci/GitVersion.yml /format '{SemVer}'
5.0.64
```
We expect it to stay that way.

### Steps to Reproduce

Steps are specified above

### RepositoryFixture Test

_No response_

```shell
f.e. here: https://github.com/51Degrees/device-detection-nginx/actions/runs/14210425608/job/39816479668#step:5:38

it gets worse on non-main branches: 
https://github.com/51Degrees/device-detection-cxx/actions/runs/14209141831/job/39824275265#step:5:35
```

kucuk-furkan:
## Description

In this pr, `/config` argument usage controlled and issue when using it fixed.

- Introduce a new test method `ReturnConfigurationFilePathIfCustomConfigurationIsSet` in `ConfigurationFileLocatorTests.cs` to 
verify custom config file handling.
- Update `GitVersion.Configuration.Tests.csproj` to copy `Configuration/CustomConfig.yaml` to the output directory.
- Enhance `GetConfigurationFile` method in `ConfigurationFileLocator.cs` to return the custom config file path if set and exists.
- Add `CustomConfig.yaml` with various versioning settings.

## Related Issue

Resolves #4480 

## Motivation and Context

This bugfix resolves #4480 issue. In earlier changes there is a change to how should tool locating configuration files. It cause a problem if `/config` arguments is used. Now if its used and configuration file is presents, locator returns custom configuration file path directly.

This isn't a problem before [this change](https://github.com/GitTools/GitVersion/pull/4452/files#diff-698746acc424792fb49b9893dfe1d64685618dc599f54c8c8063c79af7262abb) merged. In dotnet Path.Combine method handles if path is rooted automatically. But with this pr, tool stops using Path.Combine and starts directly looking under givin directory path. For further information, you can check [this Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.path.combine?view=net-9.0#system-io-path-combine(system-string-system-string)). In this documentation; we are interested with combine method's examples section - path1 and path3 combine example -. 

## How Has This Been Tested?

All existing tests pass.
New tests added.

## Checklist:

### DIFF
--- a/src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
+++ b/src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
@@ -154,19 +154,39 @@ public static class ConfigurationFileLocatorTests
         }
 
         [Test]
-        public void DoNotThrowWhenFileNameAreSame_WithDifferentCasing()
+        public void ReturnConfigurationFilePathIfCustomConfigurationIsSet()
+        {
+            this.workingPath = this.repoPath;
+            string configurationFilePath = Path.Combine(this.workingPath, "Configuration", "CustomConfig.yaml");
+
+            this.gitVersionOptions = new() { ConfigurationInfo = { ConfigurationFile = configurationFilePath } };
+
+            var serviceProvider = GetServiceProvider(this.gitVersionOptions);
+            this.fileSystem = serviceProvider.GetRequiredService<IFileSystem>();
+
+            using var _ = this.fileSystem.SetupConfigFile(
+                path: Path.Combine(this.workingPath, "Configuration"), fileName: "CustomConfig.yaml"
+            );
+            this.configFileLocator = serviceProvider.GetRequiredService<IConfigurationFileLocator>();
+
+            var config = this.configFileLocator.GetConfigurationFile(this.workingPath);
+            config.ShouldBe(configurationFilePath);
+        }
+
+        [TestCase(null)]
+        [TestCase("")]
+        [TestCase(" ")]
+        [TestCase("Configuration/CustomConfig2.yaml")]
+        public void ReturnConfigurationFilePathIfCustomConfigurationIsSet_InvalidConfigurationFilePaths(string? configFile)
         {
             this.workingPath = this.repoPath;
 
-            this.gitVersionOptions = new() { ConfigurationInfo = { ConfigurationFile = "MyConfig.yaml" } };
+            this.gitVersionOptions = new() { ConfigurationInfo = { ConfigurationFile = configFile } };
             var sp = GetServiceProvider(this.gitVersionOptions);
             this.configFileLocator = sp.GetRequiredService<IConfigurationFileLocator>();
-            this.fileSystem = sp.GetRequiredService<IFileSystem>();
 
-            using var _ = this.fileSystem.SetupConfigFile(path: this.workingPath, fileName: ConfigFile.ToLower());
-
-            var config = Should.NotThrow(() => this.configFileLocator.GetConfigurationFile(this.workingPath));
-            config.ShouldNotBe(null);
+            var config = this.configFileLocator.GetConfigurationFile(this.workingPath);
+            config.ShouldBe(null);
         }
 
         [Test]
--- a/src/GitVersion.Configuration/ConfigurationFileLocator.cs
+++ b/src/GitVersion.Configuration/ConfigurationFileLocator.cs
@@ -38,32 +38,48 @@ internal class ConfigurationFileLocator(
         WarnAboutAmbiguousConfigFileSelection(workingDirectory, projectRootDirectory);
     }
 
-    public string? GetConfigurationFile(string? directory)
+    public string? GetConfigurationFile(string? directoryPath)
     {
-        if (directory is null) return null;
-
-        string[] candidates = !string.IsNullOrWhiteSpace(this.ConfigurationFile)
-            ? [this.ConfigurationFile, .. this.SupportedConfigFileNames]
-            : this.SupportedConfigFileNames;
-
-        foreach (var fileName in candidates)
+        string? customConfigurationFile = GetCustomConfigurationFilePathIfEligable(directoryPath);
+        if (!string.IsNullOrWhiteSpace(customConfigurationFile))
         {
-            this.log.Debug($"Trying to find configuration file {fileName} at '{directory}'");
-            if (directory != null && fileSystem.Directory.Exists(directory))
+            return customConfigurationFile;
+        }
+
+        if (string.IsNullOrWhiteSpace(directoryPath) || !fileSystem.Directory.Exists(directoryPath))
+        {
+            return null;
+        }
+
+        string[] files = fileSystem.Directory.GetFiles(directoryPath);
+        foreach (var fileName in this.SupportedConfigFileNames)
+        {
+            this.log.Debug($"Trying to find configuration file {fileName} at '{directoryPath}'");
+            string? matchingFile = files.FirstOrDefault(file => string.Equals(PathHelper.GetFileName(file), fileName, StringComparison.OrdinalIgnoreCase));
+            if (matchingFile != null)
             {
-                var files = fileSystem.Directory.GetFiles(directory);
+                this.log.Info($"Found configuration file at '{matchingFile}'");
+                return matchingFile;
+            }
+        }
 
-                var matchingFile = files.FirstOrDefault(file =>
-                    string.Equals(fileSystem.Path.GetFileName(file), fileName, StringComparison.OrdinalIgnoreCase));
+        return null;
+    }
 
-                if (matchingFile != null)
-                {
-                    this.log.Info($"Found configuration file at '{matchingFile}'");
-                    return matchingFile;
-                }
+    private string? GetCustomConfigurationFilePathIfEligable(string? directoryPath)
+    {
+        if (!string.IsNullOrWhiteSpace(this.ConfigurationFile))
+        {
+            string configurationFilePath = this.ConfigurationFile;
+            if (!string.IsNullOrWhiteSpace(directoryPath))
+            {
+                configurationFilePath = Path.Combine(directoryPath, this.ConfigurationFile);
             }
 
-            this.log.Debug($"Configuration file {fileName} not found at '{directory}'");
+            if (fileSystem.File.Exists(configurationFilePath))
+            {
+                return configurationFilePath;
+            }
         }
 
         return null;
--- a/src/GitVersion.Core/Configuration/IConfigurationFileLocator.cs
+++ b/src/GitVersion.Core/Configuration/IConfigurationFileLocator.cs
@@ -3,5 +3,5 @@ namespace GitVersion.Configuration;
 public interface IConfigurationFileLocator
 {
     void Verify(string? workingDirectory, string? projectRootDirectory);
-    string? GetConfigurationFile(string? directory);
+    string? GetConfigurationFile(string? directoryPath);
 }
--- a/src/GitVersion.Core/PublicAPI.Shipped.txt
+++ b/src/GitVersion.Core/PublicAPI.Shipped.txt
@@ -113,7 +113,7 @@ GitVersion.Configuration.IConfigurationBuilder
 GitVersion.Configuration.IConfigurationBuilder.AddOverride(System.Collections.Generic.IReadOnlyDictionary<object!, object?>! value) -> void
 GitVersion.Configuration.IConfigurationBuilder.Build() -> GitVersion.Configuration.IGitVersionConfiguration!
 GitVersion.Configuration.IConfigurationFileLocator
-GitVersion.Configuration.IConfigurationFileLocator.GetConfigurationFile(string? directory) -> string?
+GitVersion.Configuration.IConfigurationFileLocator.GetConfigurationFile(string? directoryPath) -> string?
 GitVersion.Configuration.IConfigurationFileLocator.Verify(string? workingDirectory, string? projectRootDirectory) -> void
 GitVersion.Configuration.IConfigurationProvider
 GitVersion.Configuration.IConfigurationProvider.Provide(System.Collections.Generic.IReadOnlyDictionary<object!, object?>? overrideConfiguration = null) -> GitVersion.Configuration.IGitVersionConfiguration!


src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
```diff
@@ -169,6 +169,35 @@ public void DoNotThrowWhenFileNameAreSame_WithDifferentCasing()
...
+        public void ReturnConfigurationFilePathIfCustomConfigurationIsSet_InvalidConfigurationFilePaths(string? configFile)
+        {
+            this.workingPath = this.repoPath;
+
+            this.gitVersionOptions = new() { ConfigurationInfo = { ConfigurationFile = configFile } };
+            var sp = GetServiceProvider(this.gitVersionOptions);
+            this.configFileLocator = sp.GetRequiredService<IConfigurationFileLocator>();
+
+            var config = this.configFileLocator.GetConfigurationFile(this.workingPath);
+            config.ShouldNotBe(null);
```
HHobeck: ShouldBe instead of ShouldNotBe?

src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
```diff
@@ -169,6 +169,35 @@ public void DoNotThrowWhenFileNameAreSame_WithDifferentCasing()
             config.ShouldNotBe(null);
         }
 
+        [Test]
+        public void ReturnConfigurationFilePathIfCustomConfigurationIsSet()
```
HHobeck, line 173: Change to:

```csharp
[Test]
public void ReturnConfigurationFilePathIfCustomConfigurationIsSet()
{
    this.workingPath = this.repoPath;

    this.gitVersionOptions = new() { ConfigurationInfo = { ConfigurationFile = "Configuration/CustomConfig.yaml" } };

    var serviceProvider = GetServiceProvider(this.gitVersionOptions);
    this.fileSystem = serviceProvider.GetRequiredService<IFileSystem>();

    using var _ = this.fileSystem.SetupConfigFile(
        path: Path.Combine(this.workingPath, "Configuration"), fileName: "CustomConfig.yaml"
    );
    this.configFileLocator = serviceProvider.GetRequiredService<IConfigurationFileLocator>();

    var config = this.configFileLocator.GetConfigurationFile(this.workingPath);
    Path.GetFileName(config).ShouldBe("CustomConfig.yaml");
}
```

src/GitVersion.Configuration/ConfigurationFileLocator.cs
```diff
@@ -40,6 +40,14 @@ public void Verify(string? workingDirectory, string? projectRootDirectory)
 
     public string? GetConfigurationFile(string? directory)
     {
+        // If configuration file overriden and its exists, return it
```
HHobeck, line 43: Please change to:

```csharp
public string? GetConfigurationFile(string? directoryPath)
{
    string[] configurationFilePaths = string.IsNullOrWhiteSpace(this.ConfigurationFile)
        ? this.SupportedConfigFileNames : [this.ConfigurationFile, .. this.SupportedConfigFileNames];

    foreach (var item in configurationFilePaths)
    {
        this.log.Debug($"Trying to find configuration file {item} at '{directoryPath}'");

        var configurationFilePath = item;
        if (!PathHelper.IsPathRooted(configurationFilePath))
        {
            if (string.IsNullOrEmpty(directoryPath))
            {
                throw new WarningException(
                    $"The configuration file '{configurationFilePath}' is relative and no directory path known."
                );
            }
            configurationFilePath = Path.Combine(directoryPath, configurationFilePath);
        }

        if (fileSystem.File.Exists(configurationFilePath))
        {
            this.log.Info($"Found configuration file at '{configurationFilePath}'");
            return configurationFilePath;
        }

        this.log.Debug($"Configuration file {configurationFilePath} not found at '{directoryPath}'");
    }

    return null;
}
```

HHobeck (changes requested):
I have reviewed your code changes.

src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
kucuk-furkan: Yes, i fixed it in newer commit

src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
kucuk-furkan: The reason of i'm checking fullpath not filename is; this issue occurs for using full path of given configuration file. I'm changing it.

src/GitVersion.Configuration/ConfigurationFileLocator.cs
kucuk-furkan: I'm controlling this.ConfigurationFile above foreach loop, because this senerio is only eligable for this.ConfigurationFile. Its not necessary to make this rooted control inside foreach, SupportedConfigFileNames does not contains rooted paths. 

Also if i apply your suggestion, tool stops working as case insensitive. Case insensitive support was added couple weeks ago. I'm not sure this suggestion should be applied or not

src/GitVersion.Configuration.Tests/Configuration/CustomConfig.yaml
```diff
@@ -0,0 +1,75 @@
+assembly-versioning-scheme: MajorMinorPatch
```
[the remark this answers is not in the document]
kucuk-furkan: I understand but I'm new in this project and need help for this one. How should i generate mock config in fileSystem?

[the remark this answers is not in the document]
kucuk-furkan: Sorry, i found out. You make it in another comment. Thank you!

src/GitVersion.Configuration/ConfigurationFileLocator.cs
HHobeck: I don't understand. We have the call `IsPathRooted` to check whether it is a relative or a absolute file path (it's IMO very clean and understandable code). It doesn't matter where the file path comes from.

Yes that is correct we need to be case sensitive on unix based system and be case insensitive on windows systems. In my opinion we ensure this be using the `fileSystem.File.Exists(configurationFilePath)` call which is OS specific.

Does this make sense to you?

src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
HHobeck: At the moment the test is a false positive test because you get null back.

If you like you can change it to:

```csharp
config.ShouldBe(Path.GetFullPath("Configuration/CustomConfig.yaml"));
```

src/GitVersion.Configuration/ConfigurationFileLocator.cs
kucuk-furkan: Yes i'm fully on your side, this way is more clean and understandable but less performant. But in this scenerio performance is not an issue. One if condition or five if condition doesnt change anythink literally. - I'm working in a company that we developing an low code workflow platform for companies. We have to deal with high concurrent loads times to times. For that i'm chasing every  milliseconds and nanoseconds. This is a side effect for me. Also sorry for long comment -

Again i'm fully on your side, if os handles case sensitive; we should handle case sensitive. If os handles case insensitive; we should handle case insensitive too. But like i told earlier, there was a spesific issue and development for running tool in case insensitive for every os. Thats why i mentioned it.

I'm applied your suggestions. Thank you for your time and help.

HHobeck: Do you think it is really a performance problem? Can you do a measuring how much it costs?

kucuk-furkan: For this project it wont be a problem at all. Because this method won't be called that much that made a difference. We can ignore it. For this method in this project clear code and readablity has more value than this small performance improvement.

- I don't think this is worth the effort, but I'm curious about the outcome. :) -
Yes, it will operate slower for most of the scenerios i think. I've created a second method called GetConfigurationFileTemp. This method contains older codes. And GetConfigurationFile method contains your suggested codes. Also i wrote a test method called "ReturnConfigurationFilePathIfCustomConfigurationIsSet_PerformanceComparison". In this test method, custom configuration file created, then called once with both methods and compared elapsedticks for both methods. Test method will and should fail if my theory is right. Because its controls as GetConfigurationFile method should be run shorter than GetConfigurationFileTemp method. If this assertion commented out, test method will call both methods 1M times than compare results. Again test method will and should fail if my theory is right.

For best case scenario - Config file is rooted, so foreach will break in the first loop or in Temp method it wont start a foreach loop and returns immediately -
In my test iterations when both methods called once;
image
image
image
Based on compared elapsed ticks GetConfigurationFileTemp method 5x faster than GetConfigurationFile method.

In my test iterations when both methods called 1M times;
image
image
image
Based on compared elapsed milliseconds GetConfigurationFileTemp method approximately between %5 and %10 faster than GetConfigurationFile method.

For the worst case scenari - Config files not found in directories - your suggestted method gives better performance because you directly control path is exists, not getting all files under directory.

I'm pushing the changes for you can try it out if you want. When we dicide what should we choose, i'll remove unnecessary codes.

Btw if we combined two methods; we can get better(IMO) overall result - performans, clean code, readablity and behaviour(case sensitive) - 
What about this?
``` csharp
public string? GetConfigurationFileCombined(string? directoryPath)
{
    if (!string.IsNullOrWhiteSpace(this.ConfigurationFile) &&
        PathHelper.IsPathRooted(this.ConfigurationFile) &&
        fileSystem.File.Exists(this.ConfigurationFile))
    {
        return this.ConfigurationFile;
    }

    if (directoryPath is null) return null;

    string[] configurationFilePaths = string.IsNullOrWhiteSpace(this.ConfigurationFile)
    ? this.SupportedConfigFileNames : [this.ConfigurationFile, .. this.SupportedConfigFileNames];

    foreach (var item in configurationFilePaths)
    {
        this.log.Debug($"Trying to find configuration file {item} at '{directoryPath}'");

        string configurationFilePath;
        if (!PathHelper.IsPathRooted(item))
        {
            configurationFilePath = Path.Combine(directoryPath, item);
        }
        else
        {
            configurationFilePath = item;
        }

        if (fileSystem.File.Exists(configurationFilePath))
        {
            this.log.Info($"Found configuration file at '{configurationFilePath}'");
            return configurationFilePath;
        }

        this.log.Debug($"Configuration file {configurationFilePath} not found at '{directoryPath}'");
    }

    return null;
}
```

Last thing; we shouldn't throw exception if directoryPath is null. It changes tool's behaviour.

HHobeck: In my opinion the directory path is always set and cannot be null. It's strange that it is defined here with as a nullable string. Anyway, what do you think about thie following code:

```csharp
public string? GetConfigurationFileCombined(string? directoryPath)
{
    string[] configurationFilePaths = string.IsNullOrWhiteSpace(this.ConfigurationFile)
    ? this.SupportedConfigFileNames : [this.ConfigurationFile, .. this.SupportedConfigFileNames];

    foreach (var item in configurationFilePaths)
    {
        this.log.Debug($"Trying to find configuration file {item} at '{directoryPath}'");

        string configurationFilePath;
        if (!PathHelper.IsPathRooted(item))
        {
            if (string.IsNullOrEmpty(directoryPath)) continue;
            configurationFilePath = Path.Combine(directoryPath, item);
        }
        else
        {
            configurationFilePath = item;
        }

        if (fileSystem.File.Exists(configurationFilePath))
        {
            this.log.Info($"Found configuration file at '{configurationFilePath}'");
            return configurationFilePath;
        }

        this.log.Debug($"Configuration file {configurationFilePath} not found at '{directoryPath}'");
    }

    return null;
}
```

kucuk-furkan: [This line](https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Configuration/ConfigurationProvider.cs#L30) is why directoryPath is nullable :D I dont know why but it is what it is. I felt same thing like youi than found this line but still doesnt make any sense. I think there is a logical error.

Method is delicious, i'm changing then push again. Also removing other performance test methods too.

HHobeck (approved):

arturcic:
@kucuk-furkan  please check why the the tests fail on ubuntu
image
then I can review the code

kucuk-furkan:
@arturcic i didn't fix it on purpose to ask this. I shouldn't be the decider i think. I'm just a user of this tool and try to help fix my issue.

Tests fails because of [this](https://github.com/GitTools/GitVersion/pull/4483#discussion_r2026501904) `Also if i apply your suggestion, tool stops working as case insensitive. Case insensitive support was added couple weeks ago. I'm not sure this suggestion should be applied or not`

Either i can remove those tests too or put back case insensivity support back. What should i do? Someone who responsible for this project tell me what to do. I think leaving managing case sensivity to OS is better choice. But someone opened this as an issue then fixed it.

Which direction should i go?

arturcic:
@kucuk-furkan please make sure you do rebase of your branch on top of `main` branch, instead or mergin `main` into your branch. We prefer to have clean history and it's clear where the PR branch started in the history

kucuk-furkan:
@arturcic i think its ready now. Tests looks fine. I rebased my fork and for cleaner look, i squashed my commits.

asbjornu (approved):

src/GitVersion.Configuration.Tests/Configuration/ConfigurationFileLocatorTests.cs
```diff
@@ -154,19 +154,38 @@ public void DoNotThrowWhenWorkingAndRepoPathsAreSame_WithDifferentCasing()
         }
 
         [Test]
-        public void DoNotThrowWhenFileNameAreSame_WithDifferentCasing()
+        public void ReturnConfigurationFilePathIfCustomConfigurationIsSet()
         {
             this.workingPath = this.repoPath;
 
-            this.gitVersionOptions = new() { ConfigurationInfo = { ConfigurationFile = "MyConfig.yaml" } };
+            this.gitVersionOptions = new() { ConfigurationInfo = { ConfigurationFile = Path.Combine(this.workingPath, "Configuration", "CustomConfig.yaml") } };
```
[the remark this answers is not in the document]
kucuk-furkan: i pushed requested change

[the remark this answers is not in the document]
kucuk-furkan: squash completed too

arturcic (approved):
LGTM!

PR #4553, opened 2025-05-20
Reuse created commits for all branches as commits are shared.

ganju001:
Observed huge memory usage (40GB+) seems to be related on wrapping same commits over and over again.

## Description

Try to cache the commit objects and reuse them.


arturcic:
This issue was closed because it has been stalled for 30 days with no activity. Thank you for your contributions

PR #4561, opened 2025-05-30
Feature: Support shallow cloning of git repositories

joaopbnogueira:
## Description

This PR adds support for shallow cloning of git repositories. While having the full git history available is the only way to truly be certain of the next version, this is not desirable on larger repositories with many thousands of commits as a full clone can take several minutes and is usually neither desirable nor needed as there is sufficient information (e.g. git tags) available on the N most recent commits to be able to reliably determine the next version without requiring a full clone.

Since this is a special scenario it put behind an explicit flag `allowshallow` to ensure both backwards compatibility and awareness of the user of the tool.

## Motivation and Context

Observation that the git checkout step on CI pipelines is taking 5-20 minutes on large internal repositories due to having to fetch the whole history as required by GitVersion. This step sometimes represents more time than the rest of the steps, leading to costs and undesirable long iteration cycles.

## How Has This Been Tested?

- Unit tests

## Checklist:

### DIFF
--- a/docs/input/docs/reference/requirements.md
+++ b/docs/input/docs/reference/requirements.md
@@ -14,8 +14,9 @@ build server, needs to adhere to the below requirements.
 
 ### Unshallow
 
-The repository needs to be an [unshallow][git-unshallow] clone. This means
-that the `fetch-depth` in GitHub Actions needs to be set to `0`, for instance.
+The repository should be an [unshallow][git-unshallow] clone. This means
+that the `fetch-depth` in GitHub Actions should set to `0`, unless
+the `allowshallow` flag is used.
 Check with your [build server][build-servers] to see how it can be configured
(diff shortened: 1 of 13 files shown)


joaopbnogueira:
Thanks @arturcic will do, I'll be off for 2 weeks but will pick this up as soon as I'm back.

Edit: found a few minutes and just did it, ready to take it for a spin @arturcic .

joaopbnogueira:
@arturcic thank you for the release. I'm using gitversion on Azure DevOps. Am I correct that I also need to create a PR here: https://github.com/GitTools/actions/blob/main/src/tools/gitversion/tool.ts to make sure that when `disableShallowCloneCheck` is true that the `/allowshallow` argument needs to also be passed (same for other actions of course)?

Thank you

PR #4585, opened 2025-06-21
Extend git flow complex example to post merge and rebase behaviour

9swampy:
## Description

Extend the GitFlowComplexExample to document unexpected post merge and rebase behaviour.
Fix the integration test fixtures so they can accept a "master" main branch instead of hardcoding "main".

## Motivation and Context

I have been really struggling to get GitVersion to work post the v6 changes so I checked out the repo to run through sample scenarios to try and adapt but I either need more documentation to understand what the new behaviour is or the new behaviour is possibly unintended.

Rather than just rely on the Integration tests with injected configuration, I wanted to be able to revisit the repository that is created by the integration test and I observe 
1. Inconsistent behaviour between the TestFixture and when running dotnet-gitversion against GitVersion.yml output by the test fixture (problem number 1 documented in the penultimate commit in the amended GitFlowComplexExample - I set this assertion to pass but behaviour in the test does not match behaviour observed with dotnet-gitversion 6.3.0 and a persisted GitVersion.yml)
2. Consistent but unanticipated behaviour on a rebase (problem number 2 which is the current failure in the amended GitFlowComplexExample)

## How Has This Been Tested?

GitFlowComplexExample 

image

## Checklist:

This is a WIP proposal to document a problem & ask for help; if I get direction I'm happy to update the PR as offered/advised.

### DIFF
--- a/src/GitVersion.Core.Tests/Helpers/TestBase.cs
+++ b/src/GitVersion.Core.Tests/Helpers/TestBase.cs
@@ -7,7 +7,7 @@ namespace GitVersion.Core.Tests.Helpers;
 
 public class TestBase
 {
-    public const string MainBranch = RepositoryFixtureBase.MainBranch;
+    public const string MainBranch = "main";
 
     protected static IServiceProvider ConfigureServices(Action<IServiceCollection>? overrideServices = null)
     {
--- a/src/GitVersion.Core.Tests/Helpers/TestBase.cs
+++ b/src/GitVersion.Core.Tests/Helpers/TestBase.cs
@@ -7,7 +7,7 @@ namespace GitVersion.Core.Tests.Helpers;
 
 public class TestBase
 {
-    public const string MainBranch = "main";
+    public const string MainBranch = RepositoryFixtureBase.MainBranch;
 
     protected static IServiceProvider ConfigureServices(Action<IServiceCollection>? overrideServices = null)
     {
--- a/src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
+++ b/src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
@@ -1,13 +1,23 @@
 using GitVersion.Configuration;
 using GitVersion.Core.Tests.Helpers;
 using GitVersion.Helpers;
+using GitVersion.Logging;
 using LibGit2Sharp;
+using Microsoft.Extensions.DependencyInjection;
 
 namespace GitVersion.Core.Tests.IntegrationTests;
 
 [TestFixture]
 public class GitflowScenarios : TestBase
 {
+    private readonly ILog log;
+
+    public GitflowScenarios()
+    {
+        var sp = ConfigureServices();
+        this.log = sp.GetRequiredService<ILog>();
+    }
+
     [Test]
     public void GitflowComplexExample()
     {
@@ -21,9 +31,10 @@ public class GitflowScenarios : TestBase
 
         var configuration = GitFlowConfigurationBuilder.New.Build();
 
-        using var fixture = new BaseGitFlowRepositoryFixture(initialMainAction, deleteOnDispose: false);
+        using var fixture = new BaseGitFlowRepositoryFixture(InitialMainAction, deleteOnDispose: false);
         var fullSemver = "1.1.0-alpha.1";
         fixture.AssertFullSemver(fullSemver, configuration);
+        fixture.AssertCommitsSinceVersionSource(1, configuration);
 
         // Feature 1
         fixture.BranchTo(feature1Branch);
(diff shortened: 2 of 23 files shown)


asbjornu:
@HHobeck, I think these changes look reasonable. What are your thoughts on this?

src/GitVersion.Core.Tests/Helpers/TestBase.cs
```diff
@@ -7,7 +7,7 @@ namespace GitVersion.Core.Tests.Helpers;
 
 public class TestBase
 {
-    public const string MainBranch = "main";
+    public const string MainBranch = RepositoryFixtureBase.MainBranch;
```
9swampy: https://github.com/GitTools/GitVersion/discussions/4589 Magic strings >> BranchConfigurationKey.

src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
```diff
@@ -8,75 +11,140 @@ public class GitflowScenarios : TestBase
...
+        "and your config inherits from develop." + System.Environment.NewLine + System.Environment.NewLine +
+        "GitVersion uses the merge base between the feature and develop to determine the version." + System.Environment.NewLine + System.Environment.NewLine +
+        "As develop progresses (e.g., by releasing 1.2.0), rebuilding old feature branches can" + System.Environment.NewLine +
+        "produce different versions.");
+
+        fullSemver = "1.3.0-f2.1+1";
+        fixture.MakeACommit(
+            "feature 2 additional commit after original feature has been merged to develop " + System.Environment.NewLine +
+            $"and release/1.2.0 has already happened >> {fullSemver}" +
+            "Problem #1: 1.3.0-f2.1+0 is what I observe when I run dotnet-gitversion 6.3.0 but in the repo the assertion is 1.3.0-f2.1+1" +
```
9swampy: Problem 1: 
- 1.3.0-f2.1+1 is what the test asserts
- 1.3.0-f2.1+0 is what I observe when I run dotnet-gitversion 6.3.0

src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
```diff
@@ -8,75 +11,140 @@ public class GitflowScenarios : TestBase
...
+            "After rebase 1.3.0-f2.1+3 is both what the test asserts and what I observe when I run dotnet-gitversion 6.3.0." +
+            "Problem #2: I expected to get the same before and after the rebase." +
```
9swampy: Problem 2: 
- Before rebase 1.3.0-f2.1+1 is what the test asserts (but not what dotnet-gitversion produces - see Problem 1).
- After rebase 1.3.0-f2.1+3 is what both tests and dotnet-gitversion 6.3.0 produce.

I'd have anticipated1.3.0-f2.1+1 both before and after rebase, in this specific scenario.

src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
```diff
@@ -8,75 +11,140 @@ public class GitflowScenarios : TestBase
...
+        fixture.AssertFullSemver(fullSemver, configuration, customMessage: "I expected to get the same before and after the rebase.");
+
+        void initialMainAction(IRepository r)
+        {
+            if (configuration is GitVersionConfiguration concreteConfig)
+            {
+                var yaml = new ConfigurationSerializer().Serialize(concreteConfig);
+                const string fileName = "GitVersion.yml";
+                var filePath = FileSystemHelper.Path.Combine(r.Info.Path, "..", fileName);
+                File.WriteAllText(filePath, yaml);
```
9swampy, line 186: Intentionally writing out the GitVersion.yml to demonstrate inconsistency between UnitTest version calculation and running dotnet-gitversion from the commandline. Refer to 2 problems in commentary.

src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
HHobeck: I have executed gitversion (6.3.0) after the test step you mentioned and got `1.3.0-f2.1+1`:

image

src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
HHobeck: The status of the repository is not the same and the information Gitversion uses to determine the next version is different:

Before rebase:
Repo_Befor_Rebase_1 3 0-f2 1+1

After rebase:
Repo_After_Rebase_1 3 0-f2 1+3

What is your point here?

src/GitVersion.Core.Tests/Helpers/TestBase.cs
HHobeck: May I ask you: Why do you care about how the branch is named in the unit tests? Does it make any different if it is named master or main? From the configuration point of view both are included and treated the same. At the end everyone can use their own name and their own configuration in the unit tests. There is no need for unification IMO.

src/GitVersion.Core.Tests/IntegrationTests/GitflowScenarios.cs
9swampy: Odd. If I run the test now and flip out to the command line the config that's being persisted by the test is no longer valid for running with 6.3.0 but it works for you? I'll look again when it's a more sensible o'clock.

In the meantime, as I said on my opening post, if my expectations aren't correct then I would like to understand what would be correct.
1 - The merge bumps the root version on the same f2 commit from 1.2 to 1.3; that's an expected change due to changed state, and the +0 resets since there have now been no commits since version source. All fine.
2 - I'd expected the first commit on the branch since the last version bump to be f2.1+1 as the test asserts before rebase. I concur with the test. Dotnet-gitversion 6.3.0 tbc.
3 - If I rebase that commit, yes the whole tree is different but that first commit is no longer considered the first commit on the branch since the version source? I'd not expected that.

How does +1 jump to +3? If there's documentation, point me at it as I do appreciate you've been generous with your time already.

To be fair working out how to configure the majority minor patch bumps work now is what I'd expected to be playing with in the tests, not the CommitsSinceVersionSource idiosyncrasies...

HHobeck: > Odd. If I run the test now and flip out to the command line the config that's being persisted by the test is no longer valid for running with 6.3.0 but it works for you? I'll look again when it's a more sensible o'clock.

I have just opend my cmd-shell and executed gitversion in version 6.3.0

> How does +1 jump to +3? If there's documentation, point me at it as I do appreciate you've been generous with your time already.

First of all the algorithm calcualtes the next versions by executing the differnt version strategy implementations. After that the highest increment version will be selected. If there are more then one increment with the same value, the newest base version source will be used.

In both cases (before and after the rebase) the version increment of tag 1.2.1 on main wins which will be returned by the TaggedCommitVersionStrategy class. So far so good. We can assure that the base version source is in both cases the same. Please notice that the base version source (commit) is not part of the feature branch but main branch.

The answer why a different commit count is calulcated is based in the following logic (calculating the CommitsSinceVersionSource):

https://github.com/GitTools/GitVersion/blob/d2829920c6d8888033348433022aac701821d180/src/GitVersion.Core/VersionCalculation/VersionCalculators/VersionCalculatorBase.cs#L17-L26

Before the reabase you have only one commit which differs from the main branch. After the rebase you have three commits who are not part of main (the branch where the base version source was selected).

Hope that answers your question.

9swampy:
`RepositoryStoreTests.FindsCorrectMergeBaseForForwardMergeMovesOn` from

image

to

image

to educate myself. 

Only draft code but would you consider that an improvement in documentation capability? Could make the test that I've applied it to brittle, but there would be appropriate tests to apply it to?

HHobeck:
> Only draft code but would you consider that an improvement in documentation capability? Could make the test that I've applied it to brittle, but there would be appropriate tests to apply it to?

https://gitversion.net/docs/learn/branching-strategies/gitflow/examples

```
Source
See DocumentationSamplesForGitFlow.cs. To update, modify then run test.
```

and 

https://gitversion.net/docs/learn/branching-strategies/githubflow/examples

```
Source
See DocumentationSamplesForGitHubFlow.cs. To update, modify then run test.
```

If you like you can extend the tests with comments (PlantUML) to make it more clear. For me this is the reference which describes the existing branching and merging use cases for the GitHub and GitHubFlow workflow. The TrunkBased workflow needs to be added here IMO.

9swampy:
There's lots of scope for more docs that would be helpful I think; case in hand myself. As you've concurred that what I've been putting in for my own education could be helpful additions I'll likely split off to something more mergeable in the meantime. Had noted the issue asking for some trunk-based docs too.

arturcic:
This issue was closed because it has been stalled for 30 days with no activity. Thank you for your contributions

PR #4588, opened 2025-06-22
Sanitize Participant

9swampy:
## Description

The SequenceDiagrams generated by the ComplexGitflowExample were invalid because the participants were incompatible with PlantUml. I've sanitized them so the SequenceDiagrams output by tests are cut & paste viewable.

## Related Issue

#4585

## Motivation and Context

So the SequenceDiagrams output by tests are cut & paste viewable.

## How Has This Been Tested?

Test coverage and output of ComplexGitflowExample is now cut-paste viewable.

For example feature/f1 is invalid - participant can't have '/' in it.

image

## Checklist:

### DIFF
--- a/src/GitVersion.Testing/Fixtures/SequenceDiagram.cs
+++ b/src/GitVersion.Testing/Fixtures/SequenceDiagram.cs
@@ -1,3 +1,4 @@
+using GitVersion.Testing.Helpers;
 using GitVersion.Testing.Internal;
 
 namespace GitVersion.Testing;
@@ -39,11 +40,12 @@ public class SequenceDiagram
     /// </summary>
     public void Participant(string participant, string? @as = null)
     {
-        this.participants.Add(participant, @as ?? participant);
-        if (@as == null)
+        var cleanParticipant = ParticipantSanitizer.SanitizeParticipant(@as ?? participant);
+        this.participants.Add(participant, cleanParticipant);
+        if (participant == cleanParticipant)
             this.diagramBuilder.AppendLineFormat("participant {0}", participant);
         else
-            this.diagramBuilder.AppendLineFormat("participant \"{0}\" as {1}", participant, @as);
+            this.diagramBuilder.AppendLineFormat("participant \"{0}\" as {1}", participant, cleanParticipant);
     }
 
     /// <summary>
--- a/src/GitVersion.Testing/GitVersion.Testing.csproj
+++ b/src/GitVersion.Testing/GitVersion.Testing.csproj
@@ -9,9 +9,6 @@
         <PackageReference Include="System.Reflection.Metadata" />
     </ItemGroup>
     <ItemGroup>
-        <Compile Include="..\GitVersion.Core\Helpers\FileSystemHelper.cs" Link="Helpers\FileSystemHelper.cs" />
-    </ItemGroup>
-    <ItemGroup>
         <InternalsVisibleTo Include="GitVersion.BuildAgents.Tests" />
         <InternalsVisibleTo Include="GitVersion.Configuration.Tests" />
         <InternalsVisibleTo Include="GitVersion.Core.Tests" />
--- a/src/GitVersion.Testing/GitVersion.Testing.csproj
+++ b/src/GitVersion.Testing/GitVersion.Testing.csproj
@@ -19,4 +19,7 @@
         <InternalsVisibleTo Include="GitVersion.App.Tests" />
         <InternalsVisibleTo Include="GitVersion.MsBuild.Tests" />
     </ItemGroup>
+    <ItemGroup>
+      <ProjectReference Include="..\GitVersion.Core\GitVersion.Core.csproj" />
+    </ItemGroup>
 </Project>
--- a/src/GitVersion.Core.Tests/Helpers/ParticipantSanitizerTests.cs
+++ b/src/GitVersion.Core.Tests/Helpers/ParticipantSanitizerTests.cs
@@ -1,0 +1,67 @@
+using GitVersion.Testing.Helpers;
+
+namespace GitVersion.Core.Tests.Helpers;
+
+[TestFixture]
+public class ParticipantSanitizerTests
+{
+    [TestCase("feature/1234-is-id-with-something-kebab", "feature_1234_is_id_with_something_kebab")]
+    [TestCase("feature/1234-IsSomethingPascalCase", "feature_1234_IsSomethingPascalCase")]
+    [TestCase("feature/Caps-lower-something-kebab", "feature_Caps_lower_something_kebab")]
+    [TestCase("feature/Caps-lower-is-kebab", "feature_Caps_lower_is_kebab")]
+    [TestCase("kebab-folder/1234-is-id-with-something-kebab", "kebab_folder_1234_is_id_with_something_kebab")]
+    [TestCase("kebab-folder/1234-IsSomethingPascalCase", "kebab_folder_1234_IsSomethingPascalCase")]
+    [TestCase("kebab-folder/Caps-lower-something-kebab", "kebab_folder_Caps_lower_something_kebab")]
+    [TestCase("kebab-folder/Caps-lower-is-kebab", "kebab_folder_Caps_lower_is_kebab")]
+    [TestCase("PascalCaseFolder/1234-is-id-with-something-kebab", "PascalCaseFolder_1234_is_id_with_something_kebab")]
+    [TestCase("PascalCaseFolder/1234-IsSomethingPascalCase", "PascalCaseFolder_1234_IsSomethingPascalCase")]
+    [TestCase("PascalCaseFolder/Caps-lower-something-kebab", "PascalCaseFolder_Caps_lower_something_kebab")]
+    [TestCase("PascalCaseFolder/Caps-lower-is-kebab", "PascalCaseFolder_Caps_lower_is_kebab")]
+    [TestCase("1234-is-id-with-something-kebab", "1234_is_id_with_something_kebab")]
+    [TestCase("1234-IsSomethingPascalCase", "1234_IsSomethingPascalCase")]
+    [TestCase("Caps-lower-something-kebab", "Caps_lower_something_kebab")]
+    [TestCase("Caps-lower-is-kebab", "Caps_lower_is_kebab")]
+    [TestCase("feature/all-lower-is-kebab", "feature_all_lower_is_kebab")]
+    [TestCase("feature/24321-Upperjustoneword", "feature_24321_Upperjustoneword")]
+    [TestCase("feature/justoneword", "feature_justoneword")]
+    [TestCase("feature/PascalCase", "feature_PascalCase")]
+    [TestCase("feature/PascalCase-with-kebab", "feature_PascalCase_with_kebab")]
+    [TestCase("feature/12414", "feature_12414")]
+    [TestCase("feature/12414/12342-FeatureStoryTaskWithShortDescription", "feature_12414_12342_FeatureStoryTaskWithShortDescription")]
+    [TestCase("feature/12414/12342-Short-description", "feature_12414_12342_Short_description")]
+    [TestCase("feature/12414/12342-short-description", "feature_12414_12342_short_description")]
+    [TestCase("feature/12414/12342-Short-Description", "feature_12414_12342_Short_Description")]
+    [TestCase("release/1.0.0", "release_1_0_0")]
+    [TestCase("releases", "releases")]
+    [TestCase("feature", "feature")]
+    [TestCase("feature/tfs1-Short-description", "feature_tfs1_Short_description")]
+    [TestCase("feature/f2-Short-description", "feature_f2_Short_description")]
+    [TestCase("feature/bug1", "feature_bug1")]
+    [TestCase("f2", "f2")]
+    [TestCase("feature/f2", "feature_f2")]
+    [TestCase("feature/story2", "feature_story2")]
+    [TestCase("master", "master")]
+    [TestCase("develop", "develop")]
+    [TestCase("main", "main")]
+    public void SanitizeValidParticipant_ShouldReturnExpectedResult(string input, string expected)
+    {
+        var actual = ParticipantSanitizer.SanitizeParticipant(input);
+        actual.ShouldBe(expected);
+    }
+
+    [TestCase("")]
+    [TestCase(" ")]
+    public void SanitizeEmptyOrWhitespaceParticipant_ShouldThrow(string value)
+    {
+        var exception = Should.Throw<ArgumentException>(() => ParticipantSanitizer.SanitizeParticipant(value));
+        exception.Message.ShouldBe("The value cannot be an empty string or composed entirely of whitespace. (Parameter 'participant')");
+    }
+
+    [TestCase("feature/")]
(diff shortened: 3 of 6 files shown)


asbjornu (commented):
Thanks for noticing this and providing a fix! By the test cases, it seems like a simple `.RegexReplace("[^a-zA-Z0-9]", "_")` would do the trick – am I missing something? :)

https://github.com/GitTools/GitVersion/blob/f8a95ab18574ff4ccf625d67bd9502161cc0661a/src/GitVersion.Core/Extensions/StringExtensions.cs#L14-L18

9swampy:
Don't let perfection be enemy of good enough; fair. I'm a little OCD when it comes to clean consistency so my initial proposal's a bit opionated; but also cleaner IMHO but the goals to be able to copy paste in to PlantUml and get an output. There are likely other curiousities i'd not encountered so the simplicity of the RegEx you propose wouldn't go amiss. If you're happy to take one or other; or both options; let me know how to tidy up the PR to meet acceptable standards to get a merge. Thanks for consideration.

| Input                                                  | PascalCase Output                             | Underscore Output                              |
|--------------------------------------------------------|-----------------------------------------------|------------------------------------------------|
| feature/1234-is-id-with-something-kebab                | feature_1234_IsIdWithSomethingKebab           | feature_1234_is_id_with_something_kebab        |
| feature/1234-IsSomethingPascalCase                     | feature_1234_IsSomethingPascalCase            | feature_1234_IsSomethingPascalCase             |
| feature/Caps-lower-something-kebab                     | feature_CapsLowerSomethingKebab               | feature_Caps_lower_something_kebab             |
| feature/Caps-lower-is-kebab                            | feature_CapsLowerIsKebab                      | feature_Caps_lower_is_kebab                    |
| kebab-folder/1234-is-id-with-something-kebab           | KebabFolder_1234_IsIdWithSomethingKebab       | kebab_folder_1234_is_id_with_something_kebab   |
| kebab-folder/1234-IsSomethingPascalCase                | KebabFolder_1234_IsSomethingPascalCase        | kebab_folder_1234_IsSomethingPascalCase        |
| kebab-folder/Caps-lower-something-kebab                | KebabFolder_CapsLowerSomethingKebab           | kebab_folder_Caps_lower_something_kebab        |
| kebab-folder/Caps-lower-is-kebab                       | KebabFolder_CapsLowerIsKebab                  | kebab_folder_Caps_lower_is_kebab               |
| PascalCaseFolder/1234-is-id-with-something-kebab       | PascalCaseFolder_1234_IsIdWithSomethingKebab  | PascalCaseFolder_1234_is_id_with_something_kebab |
| PascalCaseFolder/1234-IsSomethingPascalCase            | PascalCaseFolder_1234_IsSomethingPascalCase   | PascalCaseFolder_1234_IsSomethingPascalCase    |
| PascalCaseFolder/Caps-lower-something-kebab            | PascalCaseFolder_CapsLowerSomethingKebab      | PascalCaseFolder_Caps_lower_something_kebab    |
| PascalCaseFolder/Caps-lower-is-kebab                   | PascalCaseFolder_CapsLowerIsKebab             | PascalCaseFolder_Caps_lower_is_kebab           |
| 1234-is-id-with-something-kebab                        | 1234_IsIdWithSomethingKebab                   | 1234_is_id_with_something_kebab                |
| 1234-IsSomethingPascalCase                             | 1234_IsSomethingPascalCase                    | 1234_IsSomethingPascalCase                     |
| Caps-lower-something-kebab                             | CapsLowerSomethingKebab                       | Caps_lower_something_kebab                     |
| Caps-lower-is-kebab                                    | CapsLowerIsKebab                              | Caps_lower_is_kebab                            |
| feature/all-lower-is-kebab                             | feature_AllLowerIsKebab                       | feature_all_lower_is_kebab                     |
| feature/24321-Upperjustoneword                         | feature_24321_Upperjustoneword                | feature_24321_Upperjustoneword                 |
| feature/justoneword                                    | feature_Justoneword                           | feature_justoneword                            |
| feature/PascalCase                                     | feature_PascalCase                            | feature_PascalCase                             |
| feature/PascalCase-with-kebab                          | feature_PascalCaseWithKebab                   | feature_PascalCase_with_kebab                  |
| feature/12414                                          | feature_12414                                 | feature_12414                                  |
| feature/12414/12342-FeatureStoryTaskWithShortDescription | feature_12414_12342_FeatureStoryTaskWithShortDescription | feature_12414_12342_FeatureStoryTaskWithShortDescription |
| feature/12414/12342-Short-description                  | feature_12414_12342_ShortDescription          | feature_12414_12342_Short_description          |
| feature/12414/12342-short-description                  | feature_12414_12342_ShortDescription          | feature_12414_12342_short_description          |
| feature/12414/12342-Short-Description                  | feature_12414_12342_ShortDescription          | feature_12414_12342_Short_Description          |
| release/1.0.0                                          | release_1.0.0                                 | release_1_0_0                                  |
| releases                                               | releases                                      | releases                                       |
| feature                                                | feature                                       | feature                                        |
| feature/tfs1-Short-description                         | feature_tfs1_ShortDescription                 | feature_tfs1_Short_description                 |
| feature/f2-Short-description                           | feature_f2_ShortDescription                   | feature_f2_Short_description                   |
| feature/bug1                                           | feature_bug1                                  | feature_bug1                                   |
| f2                                                     | f2                                            | f2                                             |
| feature/f2                                             | feature_f2                                    | feature_f2                                     |
| feature/story2                                         | feature_story2                                | feature_story2                                 |
| master                                                 | master                                        | master                                         |
| develop                                                | develop                                       | develop                                        |
| main                                                   | main                                          | main                                           |

9swampy:
So having circled back to use the output in earnest I'm seeing that the simple Regex really would be sufficient due to the @as mechanism that had already been implemented. The sanitized participant doesn't appear in the ui anyhow, the `@as` takes care of PlantUml while the original still gets displayed. 

feature/f1 maps to feature_f1 so it doesn't break PlantUml but the original feature/f1 is still displayed.

FWIW I see what I was missing now 😁

image

src/GitVersion.Testing/Helpers/RegexReplacer.cs
```diff
@@ -0,0 +1,9 @@
+using System.Text.RegularExpressions;
+
+namespace GitVersion.Testing.Helpers;
+
+public static partial class RegexReplacer
```
asbjornu: Do we need the `partial class` or could the `NonAlphanumericRegex` method just be jammed into the (already existing) `StringExtensions` class?

src/GitVersion.Testing/Fixtures/SequenceDiagram.cs
```diff
@@ -39,11 +40,12 @@ public SequenceDiagram()
     /// </summary>
     public void Participant(string participant, string? @as = null)
     {
-        this.participants.Add(participant, @as ?? participant);
-        if (@as == null)
+        var cleanParticipant = ParticipantSanitizer.SanitizeParticipant(@as ?? participant);
```
asbjornu, line 43: Could we not use the `RegexReplace()` method here directly?

```suggestion
        var cleanParticipant = (@as ?? participant).RegexReplace(…);
```

9swampy:
## 🧵 Comparing `Regex` Usage Styles in C#

There are two common approaches to working with regular expressions in modern C#:

### 1. **Traditional Runtime Regex (with Optional Caching)**

Example:
```csharp
public static string RegexReplace(this string input, string pattern, string replace) 
{ 
    var regex = RegexPatterns.Cache.GetOrAdd(pattern); 
    return regex.Replace(input, replace); 
}
```

Used with:
```csharp
private const string TriviaRegexPattern = "...";
public static Regex TriviaRegex { get; } = new(TriviaRegexPattern, RegexOptions.Singleline | ...);
```

#### ✅ Pros
- Supports **dynamic** and **runtime-defined** patterns.
- Allows **complex, multi-line** regex using `RegexOptions.IgnorePatternWhitespace`.
- Can use caching to **reduce recompilation overhead**.

#### ❌ Cons
- Patterns are **compiled at runtime**, which is slower.
- Errors in patterns aren't caught until runtime.
- Refactoring tools don't understand regex inside strings.

---

### 2. **[GeneratedRegex] Attribute (Compile-Time Regex)**

Example:
```csharp
[GeneratedRegex("[^a-zA-Z0-9]")]
public static partial Regex NonAlphanumericRegex();
```

#### ✅ Pros
- Patterns are **compiled at build time** — no runtime parsing.
- Catches **syntax errors at compile time**.
- Fully compatible with **AOT** and **IL trimming**.
- **Better performance** than runtime regex.

#### ❌ Cons
- Requires **.NET 7 or later**.
- Only supports **compile-time constant patterns**.
- Verbose/complex patterns can be harder to maintain in this format.

---

## ✅ When to Use Which?

| **Use Case**                    | **Recommended Approach**                     |
|----------------------------------|----------------------------------------------|
| Known, constant patterns         | `GeneratedRegex`                             |
| Dynamic or user-defined patterns | Runtime + caching (`RegexPatterns.Cache`)    |
| Performance-critical code        | `GeneratedRegex`                             |
| Multi-line, complex patterns     | Traditional regex + `IgnorePatternWhitespace`|
| Needs AOT compatibility          | `GeneratedRegex` (strongly preferred)        |

If you prefer I can just follow the RegexPatterns convention for consistency, but especially for the preexisting as you've dropped net6 I'd strongly suggest converting as they're heavily used in the calculation so the compile time optimization would be worthwhile?

Let me know which route to go and I'll rip out the initial one, leave only regex.

asbjornu:
> If you prefer I can just follow the RegexPatterns convention for consistency, but especially for the preexisting as you've dropped net6 I'd strongly suggest converting as they're heavily used in the calculation so the compile time optimization would be worthwhile?

As long as all tests pass, I'm all for converting to the new, precompiled variant! 🙂👍🏼

arturcic:
> ## 🧵 Comparing `Regex` Usage Styles in C#
> There are two common approaches to working with regular expressions in modern C#:
> 
> ### 1. **Traditional Runtime Regex (with Optional Caching)**
> Example:
> 
> ```cs
> public static string RegexReplace(this string input, string pattern, string replace) 
> { 
>     var regex = RegexPatterns.Cache.GetOrAdd(pattern); 
>     return regex.Replace(input, replace); 
> }
> ```
> 
> Used with:
> 
> ```cs
> private const string TriviaRegexPattern = "...";
> public static Regex TriviaRegex { get; } = new(TriviaRegexPattern, RegexOptions.Singleline | ...);
> ```
> 
> #### ✅ Pros
> * Supports **dynamic** and **runtime-defined** patterns.
> * Allows **complex, multi-line** regex using `RegexOptions.IgnorePatternWhitespace`.
> * Can use caching to **reduce recompilation overhead**.
> 
> #### ❌ Cons
> * Patterns are **compiled at runtime**, which is slower.
> * Errors in patterns aren't caught until runtime.
> * Refactoring tools don't understand regex inside strings.
> 
> ### 2. **[GeneratedRegex] Attribute (Compile-Time Regex)**
> Example:
> 
> ```cs
> [GeneratedRegex("[^a-zA-Z0-9]")]
> public static partial Regex NonAlphanumericRegex();
> ```
> 
> #### ✅ Pros
> * Patterns are **compiled at build time** — no runtime parsing.
> * Catches **syntax errors at compile time**.
> * Fully compatible with **AOT** and **IL trimming**.
> * **Better performance** than runtime regex.
> 
> #### ❌ Cons
> * Requires **.NET 7 or later**.
> * Only supports **compile-time constant patterns**.
> * Verbose/complex patterns can be harder to maintain in this format.
> 
> ## ✅ When to Use Which?
> **Use Case**	**Recommended Approach**
> Known, constant patterns	`GeneratedRegex`
> Dynamic or user-defined patterns	Runtime + caching (`RegexPatterns.Cache`)
> Performance-critical code	`GeneratedRegex`
> Multi-line, complex patterns	Traditional regex + `IgnorePatternWhitespace`
> Needs AOT compatibility	`GeneratedRegex` (strongly preferred)
> If you prefer I can just follow the RegexPatterns convention for consistency, but especially for the preexisting as you've dropped net6 I'd strongly suggest converting as they're heavily used in the calculation so the compile time optimization would be worthwhile?
> 
> Let me know which route to go and I'll rip out the initial one, leave only regex.

There is this class https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core/Core/RegexPatterns.cs where we have all our regex, we will migrate this to newer GeneratedRegex in one go later on

9swampy:
Took reference to GitVersion.Core from  GitVersion.Testing but with InternalsVisibleTo to minimize impact and cleared out the overdone one. Tests probably overegging it but do no harm?

src/GitVersion.Testing/Fixtures/SequenceDiagram.cs
9swampy: Doen

src/GitVersion.Testing/Helpers/RegexReplacer.cs
9swampy: Done.

src/GitVersion.Testing/GitVersion.Testing.csproj
```diff
@@ -19,4 +19,7 @@
         <InternalsVisibleTo Include="GitVersion.App.Tests" />
         <InternalsVisibleTo Include="GitVersion.MsBuild.Tests" />
     </ItemGroup>
+    <ItemGroup>
+      <ProjectReference Include="..\GitVersion.Core\GitVersion.Core.csproj" />
```
asbjornu, line 20: The reference to `GitVersion.Core` now seems to provoke the following warning:

> The type 'FileSystemHelper' in '/Users/runner/work/GitVersion/GitVersion/src/GitVersion.Testing/../GitVersion.Core/Helpers/FileSystemHelper.cs' conflicts with the imported type 'FileSystemHelper' in 'GitVersion.Core, Version=6.3.1.0, Culture=neutral, PublicKeyToken=null'. Using the type defined in '/Users/runner/work/GitVersion/GitVersion/src/GitVersion.Testing/../GitVersion.Core/Helpers/FileSystemHelper.cs'.

Perhaps the `FileSystemHelper` can be fully qualified to remove this warning?

asbjornu (changes requested):
Looking very close to mergeable to me! 👍🏼

src/GitVersion.Testing/GitVersion.Testing.csproj
9swampy: FileSystemHelper is a linked file in GitVersion.Testing, but now it's also coming in as a proper reference in GitVersion.Core. Dropping the linked file sorts the warning:

9swampy: Done

asbjornu (approved):
LGTM! Please rebase and resolve the conflicts, and this should be good to merge. 👍🏼

global.json
```diff
@@ -5,6 +5,6 @@
     "src"
   ],
   "sdk": {
-    "version": "9.0.301"
+    "version": "9.0.300"
```
arturcic: please revert this file

arturcic (changes requested):

global.json
9swampy: whoops, sorry. Missed that on the conflict dialog. Reverted.

arturcic (approved):

PR #4681, opened 2025-09-30
Reduce memory usage by caching commits on their sha hash

### LINKED ISSUE #4409: [ISSUE]: gitversion 6.10.0 (and 6.5.0 as well) hangs for a while when calculating version
### Prerequisites

### GitVersion package

GitVersion.Tool, gittools/actions, AzureDevops task

### GitVersion version

6.10.0

### Operating system

Linux, Windows

### What are you seeing?

We are using git flow and mark specific commits on the release branch with a tag (looks like vX.X.X, v1.2.3 for example) and then run a Azure Devops build pipeline selecting that tag. Here is the GitVersion.yml
```mode: ContinuousDelivery
assembly-informational-format: "{FullSemVer}"
branches:
  develop:
    label: develop
  release:
    label: release
    is-release-branch: true
  hotfix:
    label: hotfix
ignore:
  sha: []
merge-message-formats: {}
```
Relevant release pipeline:
```
  - checkout: self
    clean: true
    fetchDepth: 0
  - task: gitversion/setup@3
    displayName: gitversion/setup
    inputs:
      versionSpec: 6.0.x
  - task: gitversion/execute@3
    displayName: gitversion/execute
```
We just updated from gitversion 5.12.0 to 6.0.5 and the gitversion/execute task takes now 1.5 hours for a first run and 3.5 hours a week later.
v5.12.0 finishes within a minute.
I can reproduce that behavior locally, by checking out our repository on a specific tag.

Gitversion starts as `/opt/hostedtoolcache/GitVersion.Tool/6.0.5/dotnet-gitversion --roll-forward Major /home/vsts/work/1/s /output json /l console` on the azure hosted build agent.
I can't provide the full output, but some findings:

Azure build agent reports memory and CPU warnings `Agent environment resources - Disk: / Available 18306.55 MB out of 74244.74 MB, Memory: Used 6846.00 MB out of 6921.00 MB, CPU: Usage 94.92%`

Gitversion runs fast in the beginning on tags with the output like `Begin: Finding merge base between 'tags/v2.17.0' and 'origin/release/2.9.0'.` etc
Then it stops for like 50 minutes and start to work on branches `Begin: Finding merge base between 'hotfix/2.15.1' and 'master'.`

Our repo has around a 100 active tags and somewhat about 15 active branches.

### What is expected?

Gitversion finishes within few minutes

### Steps to Reproduce

Run on a repo when a tag is checked out?

### RepositoryFixture Test

_No response_

```shell

```

### LINKED ISSUE #4680: [ISSUE]: High memory usage
### Prerequisites

### GitVersion package

GitVersion.Tool

### GitVersion version

6.4

### Operating system

Windows, Linux

### What are you seeing?

Running the `dotnet-gitversion` tool with default configuration results in high memory usage of around 10GB, this is on a repository with 25k commits.

This amount is a bit too much for our build agents resulting in builds to fail... and is also a bit unexpectedly high.

PR with investigation is on it's way...

### What is expected?

A more reasonable memory footprint of a few 100MBs would be reasonable I think in this case.

### Steps to Reproduce

Looking into anonymizing the private repository.

### RepositoryFixture Test

_No response_

```shell

```

pvanbuijtene:
## Description
Reduce the amount of allocations of `Commit` instances.

## Related Issue
Fix #4680, Fix #4409

## Motivation and Context

After some investigation it turned out that a lot of commits are being re-created during the process of calculating the version.
By implementing a cache for the commits the overall memory consumption went from ~10GB to ~400MB.

## How Has This Been Tested?

The solution was tested by running the project again with the changes against the same repository resulting in less memory consumption as shown below.

The problem of the memory usage was internally reported on Linux environments and reproduced on Windows.
My machine has 64GB & 22 logical processors.

The CPU and Memory consumption before the change, peaking at ~10GB.
before
The CPU and Memory consumption after the change, peaking at ~400MB.
after

## Checklist:

* \[ ] My code follows the code style of this project.

### DIFF
--- a/src/GitVersion.LibGit2Sharp/Git/BranchCollection.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/BranchCollection.cs
@@ -8,12 +8,14 @@ internal sealed class BranchCollection : IBranchCollection
     private readonly LibGit2Sharp.BranchCollection innerCollection;
     private readonly Lazy<IReadOnlyCollection<IBranch>> branches;
     private readonly Diff diff;
+    private readonly GitRepository repo;
 
-    internal BranchCollection(LibGit2Sharp.BranchCollection collection, Diff diff)
+    internal BranchCollection(LibGit2Sharp.BranchCollection collection, Diff diff, GitRepository repo)
     {
         this.innerCollection = collection.NotNull();
-        this.branches = new Lazy<IReadOnlyCollection<IBranch>>(() => [.. this.innerCollection.Select(branch => new Branch(branch, diff))]);
+        this.branches = new Lazy<IReadOnlyCollection<IBranch>>(() => [.. this.innerCollection.Select(branch => repo.GetOrCreate(branch, diff))]);
         this.diff = diff.NotNull();
+        this.repo = repo.NotNull();
     }
 
     public IEnumerator<IBranch> GetEnumerator()
@@ -27,7 +29,7 @@ internal sealed class BranchCollection : IBranchCollection
         {
             name = name.NotNull();
             var branch = this.innerCollection[name];
-            return branch is null ? null : new Branch(branch, this.diff);
+            return branch is null ? null : this.repo.GetOrCreate(branch, this.diff);
         }
     }
 
--- a/src/GitVersion.LibGit2Sharp/Git/Commit.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/Commit.cs
@@ -14,10 +14,12 @@ internal sealed class Commit : GitObject, ICommit
     private readonly LibGit2Sharp.Commit innerCommit;
     private readonly LibGit2Sharp.Diff repoDiff;
 
-    internal Commit(LibGit2Sharp.Commit innerCommit, LibGit2Sharp.Diff repoDiff) : base(innerCommit)
+    internal Commit(LibGit2Sharp.Commit innerCommit, LibGit2Sharp.Diff repoDiff, GitRepository repo) : base(innerCommit)
     {
+        repoDiff.NotNull();
+        repo.NotNull();
         this.innerCommit = innerCommit.NotNull();
-        this.parentsLazy = new(() => innerCommit.Parents.Select(parent => new Commit(parent, repoDiff)).ToList());
+        this.parentsLazy = new(() => innerCommit.Parents.Select(parent => repo.GetOrCreate(parent, repoDiff)).ToList());
         When = innerCommit.Committer.When;
         this.repoDiff = repoDiff;
     }
--- a/src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs
@@ -8,12 +8,14 @@ internal sealed class CommitCollection : ICommitCollection
     private readonly ICommitLog innerCollection;
     private readonly Lazy<IReadOnlyCollection<ICommit>> commits;
     private readonly Diff diff;
+    private readonly GitRepository repo;
 
-    internal CommitCollection(ICommitLog collection, Diff diff)
+    internal CommitCollection(ICommitLog collection, Diff diff, GitRepository repo)
     {
         this.innerCollection = collection.NotNull();
-        this.commits = new Lazy<IReadOnlyCollection<ICommit>>(() => [.. this.innerCollection.Select(commit => new Commit(commit, diff))]);
+        this.commits = new Lazy<IReadOnlyCollection<ICommit>>(() => [.. this.innerCollection.Select(commit => repo.GetOrCreate(commit, diff))]);
         this.diff = diff.NotNull();
+        this.repo = repo.NotNull();
     }
 
     public IEnumerator<ICommit> GetEnumerator()
@@ -36,7 +38,7 @@ internal sealed class CommitCollection : ICommitCollection
             SortBy = (LibGit2Sharp.CommitSortStrategies)commitFilter.SortBy
         };
         var commitLog = ((IQueryableCommitLog)this.innerCollection).QueryBy(filter);
-        return new CommitCollection(commitLog, this.diff);
+        return new CommitCollection(commitLog, this.diff, this.repo);
 
         static object? GetReacheableFrom(object? item) =>
             item switch
--- a/src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
@@ -1,3 +1,4 @@
+using System.Collections.Concurrent;
 using GitVersion.Extensions;
 using GitVersion.Helpers;
 using LibGit2Sharp;
@@ -7,6 +8,10 @@ namespace GitVersion.Git;
 internal sealed partial class GitRepository
 {
     private Lazy<IRepository>? repositoryLazy;
+
+    private readonly ConcurrentDictionary<string, Branch> cachedBranches = new();
+    private readonly ConcurrentDictionary<string, Commit> cachedCommits = new();
+    private readonly ConcurrentDictionary<string, Tag> cachedTags = new();
 
     private IRepository RepositoryInstance
     {
@@ -20,12 +25,12 @@ internal sealed partial class GitRepository
     public string WorkingDirectory => RepositoryInstance.Info.WorkingDirectory;
     public bool IsHeadDetached => RepositoryInstance.Info.IsHeadDetached;
     public bool IsShallow => RepositoryInstance.Info.IsShallow;
-    public IBranch Head => new Branch(RepositoryInstance.Head, RepositoryInstance.Diff);
+    public IBranch Head => GetOrCreate(RepositoryInstance.Head, RepositoryInstance.Diff);
 
-    public ITagCollection Tags => new TagCollection(RepositoryInstance.Tags, RepositoryInstance.Diff);
+    public ITagCollection Tags => new TagCollection(RepositoryInstance.Tags, RepositoryInstance.Diff, this);
     public IReferenceCollection Refs => new ReferenceCollection(RepositoryInstance.Refs);
-    public IBranchCollection Branches => new BranchCollection(RepositoryInstance.Branches, RepositoryInstance.Diff);
-    public ICommitCollection Commits => new CommitCollection(RepositoryInstance.Commits, RepositoryInstance.Diff);
+    public IBranchCollection Branches => new BranchCollection(RepositoryInstance.Branches, RepositoryInstance.Diff, this);
+    public ICommitCollection Commits => new CommitCollection(RepositoryInstance.Commits, RepositoryInstance.Diff, this);
     public IRemoteCollection Remotes => new RemoteCollection(RepositoryInstance.Network.Remotes);
 
     public void DiscoverRepository(string? gitDirectory)
@@ -48,7 +53,7 @@ internal sealed partial class GitRepository
             var first = (Commit)commit;
             var second = (Commit)otherCommit;
             var mergeBase = RepositoryInstance.ObjectDatabase.FindMergeBase(first, second);
-            return mergeBase == null ? null : new Commit(mergeBase, RepositoryInstance.Diff);
+            return mergeBase == null ? null : GetOrCreate(mergeBase, RepositoryInstance.Diff);
         });
     }
 
@@ -56,6 +61,26 @@ internal sealed partial class GitRepository
     {
         var retryAction = new RetryAction<LibGit2Sharp.LockedFileException, int>();
         return retryAction.Execute(GetUncommittedChangesCountInternal);
+    }
+
+    public Branch GetOrCreate(LibGit2Sharp.Branch innerBranch, Diff repoDiff)
+    {
+        if (innerBranch.Tip is null)
+        {
+            return new Branch(innerBranch, repoDiff, this);
+        }
+
+        var cacheKey = $"{innerBranch.CanonicalName}|{innerBranch.Tip.Sha}|{innerBranch.RemoteName}";
+        return cachedBranches.GetOrAdd(cacheKey, new Branch(innerBranch, repoDiff, this));
+    }
+
+    public Commit GetOrCreate(LibGit2Sharp.Commit innerCommit, Diff repoDiff) =>
+        cachedCommits.GetOrAdd(innerCommit.Sha, new Commit(innerCommit, repoDiff, this));
+
+    public Tag GetOrCreate(LibGit2Sharp.Tag innerTag, Diff repoDiff)
+    {
+        var cacheKey = $"{innerTag.CanonicalName}|{innerTag.Target.Sha}";
+        return cachedTags.GetOrAdd(cacheKey, new Tag(innerTag, repoDiff, this));
     }
 
     public void Dispose()
--- a/src/GitVersion.LibGit2Sharp/Git/Branch.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/Branch.cs
@@ -10,16 +10,18 @@ internal sealed class Branch : IBranch
 
     private readonly LibGit2Sharp.Branch innerBranch;
 
-    internal Branch(LibGit2Sharp.Branch branch, LibGit2Sharp.Diff diff)
+    internal Branch(LibGit2Sharp.Branch branch, LibGit2Sharp.Diff diff, GitRepository repo)
     {
+        diff.NotNull();
+        repo.NotNull();
         this.innerBranch = branch.NotNull();
         Name = new(branch.CanonicalName);
 
         var commit = this.innerBranch.Tip;
-        Tip = commit is null ? null : new Commit(commit, diff);
+        Tip = commit is null ? null : repo.GetOrCreate(commit, diff);
 
         var commits = this.innerBranch.Commits;
-        Commits = new CommitCollection(commits, diff);
+        Commits = new CommitCollection(commits, diff, repo);
     }
 
     public ReferenceName Name { get; }
--- a/src/GitVersion.LibGit2Sharp/Git/Tag.cs
+++ b/src/GitVersion.LibGit2Sharp/Git/Tag.cs
@@ -11,12 +11,14 @@ internal sealed class Tag : ITag
     private readonly LibGit2Sharp.Tag innerTag;
     private readonly Diff diff;
     private readonly Lazy<ICommit?> commitLazy;
+    private readonly GitRepository repo;
 
-    internal Tag(LibGit2Sharp.Tag tag, Diff diff)
+    internal Tag(LibGit2Sharp.Tag tag, Diff diff, GitRepository repo)
     {
         this.innerTag = tag.NotNull();
         this.commitLazy = new(PeeledTargetCommit);
         this.diff = diff.NotNull();
+        this.repo = repo.NotNull();
         Name = new(this.innerTag.CanonicalName);
     }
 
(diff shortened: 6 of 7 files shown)


src/GitVersion.LibGit2Sharp/Git/CommitFactory.cs
```diff
@@ -0,0 +1,11 @@
+using System.Collections.Concurrent;
+
+namespace GitVersion.Git;
+
+internal sealed class CommitFactory
```
arturcic: I'm not sure I like the name, I'd rather name it `GitCache`, and I think we could extend this class to have cache for other git primitives like `branch`, `tag`, `reference` and `remote`

pvanbuijtene: Applied the changes.

new-cli/GitVersion.Core.Libgit2Sharp/GitVersion.Core.Libgit2Sharp.csproj
```diff
@@ -21,6 +21,9 @@
       <Compile Include="..\..\src\GitVersion.LibGit2Sharp\Git\CommitCollection.cs">
         <Link>Git\CommitCollection.cs</Link>
       </Compile>
+      <Compile Include="..\..\src\GitVersion.LibGit2Sharp\Git\GitCache.cs">
```
asbjornu: Not sure why these explicit `Compile` items are needed, as the default value of [`EnableDefaultCompileItems`](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props?WT.mc_id=DOP-MVP-5001511#enabledefaultcompileitems) should be `true`. Ideas @arturcic? 🤔

src/GitVersion.LibGit2Sharp/Git/BranchCollection.cs
```diff
@@ -8,12 +8,14 @@ internal sealed class BranchCollection : IBranchCollection
...
     private readonly Lazy<IReadOnlyCollection<IBranch>> branches;
     private readonly Diff diff;
+    private readonly GitCache gitCache;
 
-    internal BranchCollection(LibGit2Sharp.BranchCollection collection, Diff diff)
+    internal BranchCollection(LibGit2Sharp.BranchCollection collection, Diff diff, GitCache gitCache)
     {
         this.innerCollection = collection.NotNull();
-        this.branches = new Lazy<IReadOnlyCollection<IBranch>>(() => [.. this.innerCollection.Select(branch => new Branch(branch, diff))]);
+        this.branches = new Lazy<IReadOnlyCollection<IBranch>>(() => [.. this.innerCollection.Select(branch => gitCache.GetOrCreate(branch, diff))]);
```
asbjornu: `this.` should be used to reference private local fields in the GitVersion codebase.

```suggestion
        this.branches = new Lazy<IReadOnlyCollection<IBranch>>(() => [.. this.innerCollection.Select(branch => this.gitCache.GetOrCreate(branch, diff))]);
```

src/GitVersion.LibGit2Sharp/Git/BranchCollection.cs
```diff
@@ -27,7 +29,7 @@ public IBranch? this[string name]
         {
             name = name.NotNull();
             var branch = this.innerCollection[name];
-            return branch is null ? null : new Branch(branch, this.diff);
+            return branch is null ? null : gitCache.GetOrCreate(branch, this.diff);
```
asbjornu:
```suggestion
            return branch is null ? null : this.gitCache.GetOrCreate(branch, this.diff);
```

src/GitVersion.LibGit2Sharp/Git/Commit.cs
```diff
@@ -14,10 +14,10 @@ internal sealed class Commit : GitObject, ICommit
     private readonly LibGit2Sharp.Commit innerCommit;
     private readonly LibGit2Sharp.Diff repoDiff;
 
-    internal Commit(LibGit2Sharp.Commit innerCommit, LibGit2Sharp.Diff repoDiff) : base(innerCommit)
+    internal Commit(LibGit2Sharp.Commit innerCommit, LibGit2Sharp.Diff repoDiff, GitCache gitCache) : base(innerCommit)
     {
         this.innerCommit = innerCommit.NotNull();
-        this.parentsLazy = new(() => innerCommit.Parents.Select(parent => new Commit(parent, repoDiff)).ToList());
+        this.parentsLazy = new(() => innerCommit.Parents.Select(parent => gitCache.GetOrCreate(parent, repoDiff)).ToList());
```
asbjornu:
```suggestion
        this.parentsLazy = new(() => innerCommit.Parents.Select(parent => this.gitCache.GetOrCreate(parent, repoDiff)).ToList());
```

src/GitVersion.LibGit2Sharp/Git/CommitCollection.cs
```diff
@@ -8,12 +8,14 @@ internal sealed class CommitCollection : ICommitCollection
...
     private readonly Lazy<IReadOnlyCollection<ICommit>> commits;
     private readonly Diff diff;
+    private readonly GitCache gitCache;
 
-    internal CommitCollection(ICommitLog collection, Diff diff)
+    internal CommitCollection(ICommitLog collection, Diff diff, GitCache gitCache)
     {
         this.innerCollection = collection.NotNull();
-        this.commits = new Lazy<IReadOnlyCollection<ICommit>>(() => [.. this.innerCollection.Select(commit => new Commit(commit, diff))]);
+        this.commits = new Lazy<IReadOnlyCollection<ICommit>>(() => [.. this.innerCollection.Select(commit => gitCache.GetOrCreate(commit, diff))]);
```
asbjornu:
```suggestion
        this.commits = new Lazy<IReadOnlyCollection<ICommit>>(() => [.. this.innerCollection.Select(commit => this.gitCache.GetOrCreate(commit, diff))]);
```

src/GitVersion.LibGit2Sharp/Git/GitCache.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace GitVersion.Git;
+
+internal sealed class GitCache
+{
+    private static readonly ConcurrentDictionary<string, Branch> _cachedBranches = new();
+    private static readonly ConcurrentDictionary<string, Commit> _cachedCommits = new();
+    private static readonly ConcurrentDictionary<string, Reference> _cachedReferences = new();
+    private static readonly ConcurrentDictionary<string, Remote> _cachedRemotes = new();
+    private static readonly ConcurrentDictionary<string, Tag> _cachedTags = new();
```
asbjornu: We don't use `_` as a private field prefix in the GitVersion codebase. :)

```suggestion
    private static readonly ConcurrentDictionary<string, Branch> cachedBranches = new();
    private static readonly ConcurrentDictionary<string, Commit> cachedCommits = new();
    private static readonly ConcurrentDictionary<string, Reference> cachedReferences = new();
    private static readonly ConcurrentDictionary<string, Remote> cachedRemotes = new();
    private static readonly ConcurrentDictionary<string, Tag> cachedTags = new();
```

arturcic:
@pvanbuijtene I'm quite happy with the direction you're heading with these changes, I will have a look later as well. Thanks for shared the anonymous repo

src/GitVersion.LibGit2Sharp/Git/BranchCollection.cs
pvanbuijtene: This is a tricky one as the field and constructor parameter have the same name, is it ok in this case or do you prefer to use a different name and only use the field?

Will change all the other occasions.

asbjornu: Ah, yes. As this is a reference to the parameter, `this.` doesn't apply here. Please ignore. :)

pvanbuijtene:
As you might have noticed, there are some tests failing.

I think one of the reasons for this is that the cache's life span is longer than you would normally expect I think.

From a CLI perspective you would request the version using GitVersion and you're done.

Within the tests there are complete scenarios being checked and I'm not sure if and how the cache should behave.
Maybe it should be cleared with certain steps, or it should be disabled 🤔 

As an example this one, it fails with caching of branches and succeeds without branch caching: https://github.com/GitTools/GitVersion/blob/main/src/GitVersion.Core.Tests/VersionCalculation/NextVersionCalculatorTests.cs#L124

```cs
    [Test]
    public void MergeFeatureIntoMainline()
    {
        var configuration = TrunkBasedConfigurationBuilder.New.Build();

        fixture.MakeACommit();
        fixture.ApplyTag("1.0.0");
        fixture.AssertFullSemver("1.0.0", configuration);

        fixture.BranchTo("feature/foo");
        fixture.MakeACommit();
        fixture.AssertFullSemver("1.1.0-foo.1", configuration);
        fixture.ApplyTag("1.1.0-foo.1");

        fixture.Checkout(MainBranch);
        fixture.MergeNoFF("feature/foo");
        fixture.AssertFullSemver("1.1.0", configuration);
    }
```

src/GitVersion.LibGit2Sharp/Git/ReferenceCollection.cs
```diff
@@ -5,13 +5,18 @@ namespace GitVersion.Git;
...
 {
     private readonly LibGit2Sharp.ReferenceCollection innerCollection;
+    private readonly GitCache gitCache;
     private IReadOnlyCollection<IReference>? references;
 
-    internal ReferenceCollection(LibGit2Sharp.ReferenceCollection collection) => this.innerCollection = collection.NotNull();
+    internal ReferenceCollection(LibGit2Sharp.ReferenceCollection collection, GitCache gitCache)
+    {
+        this.innerCollection = collection.NotNull();
+        this.gitCache = gitCache;
```
asbjornu: Perhaps we should ensure that the `GitCache` isn't `null`?

```suggestion
        this.gitCache = gitCache.NotNull();
```

arturcic:
> > Within the tests there are complete scenarios being checked and I'm not sure if and how the cache should behave. Maybe it should be cleared with certain steps, or it should be disabled 🤔
> > As an example this one, it fails with caching of branches and succeeds without branch caching […]
> 
> If the cache breaks tests, I'm thinking it might be uncovering a problem with the cache.
> 
> Since this PR was originally about caching the `Commit` class, perhaps we should keep it at that and ensure that works with 100% green tests before we move on with a future followup PR where we attempt to cache even more Git objects?

I will try to investigate why those tests are failing as well

new-cli/GitVersion.Core.Libgit2Sharp/GitVersion.Core.Libgit2Sharp.csproj
arturcic: these files are linked files into the `new-cli` solution but part of the current solution

pvanbuijtene:
> > Within the tests there are complete scenarios being checked and I'm not sure if and how the cache should behave. Maybe it should be cleared with certain steps, or it should be disabled 🤔
> > As an example this one, it fails with caching of branches and succeeds without branch caching […]
> 
> If the cache breaks tests, I'm thinking it might be uncovering a problem with the cache.
> 
> Since this PR was originally about caching the `Commit` class, perhaps we should keep it at that and ensure that works with 100% green tests before we move on with a future followup PR where we attempt to cache even more Git objects?

Think I would agree with that, caching branches and tags is as you might expect not as straight forward as the commits.

arturcic:
@pvanbuijtene I see you managed to fix the CI build, please rebase onto `main` and perform the `dotnet format`

pvanbuijtene:
@arturcic it was more a test to see if I could get it work.

I couldn't get the tests to work with cached `References` and `Remotes` so I removed those.

Let me know what you think, and if you want some tests specifically for the `GitCache`.

arturcic:
> @arturcic it was more a test to see if I could get it work.
> 
> I couldn't get the tests to work with cached `References` and `Remotes` so I removed those.
> 
> Let me know what you think, and if you want some tests specifically for the `GitCache`.

I think for now we can skip those 2.

arturcic:
@pvanbuijtene have you run this version of GitVersion for that repo with the issues? If so can you provide some statistics/graphs with the comparison?

src/GitVersion.LibGit2Sharp/Git/GitCache.cs
```diff
@@ -0,0 +1,30 @@
+using System.Collections.Concurrent;
+
+namespace GitVersion.Git;
+
+internal sealed class GitCache
```
arturcic: @pvanbuijtene can we move the GitCache into the GitRepository itself? The dictionaries and the methods, and then pass the repository to the constructors instead of the GitCache instance?

arturcic (changes requested):

pvanbuijtene:
> @pvanbuijtene have you run this version of GitVersion for that repo with the issues? If so can you provide some statistics/graphs with the comparison?

I didn't, but here are the results.

Main @ 9939c56:
image

This PR:
image

src/GitVersion.LibGit2Sharp/Git/GitCache.cs
pvanbuijtene: Sure, it's done.

arturcic:
> > @pvanbuijtene have you run this version of GitVersion for that repo with the issues? If so can you provide some statistics/graphs with the comparison?
> 
> I didn't, but here are the results.
> 
> Main @ [9939c56](https://github.com/GitTools/GitVersion/commit/9939c56a30f0c1aa45ae21a0d9287ccf7b15667f): image
> 
> This PR: image

So there was quite a pressure on the GC

arturcic (approved):
Looks good to me, approved, thanks for the time you spent

arturcic:
@asbjornu time for a second review?

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
```diff
@@ -58,6 +63,26 @@ public int UncommittedChangesCount()
...
     }
 
+    public Branch GetOrCreate(LibGit2Sharp.Branch innerBranch, Diff repoDiff)
+    {
+        if (innerBranch.Tip is null)
+        {
+            return new Branch(innerBranch, repoDiff, this);
+        }
+
+        var cacheKey = $"{innerBranch.CanonicalName}|{innerBranch.Tip.Sha}|{innerBranch.RemoteName}";
```
asbjornu, line 73: Sorry for not discovering this earlier. It's not very important, but will improve debugging if the format of the key matches what one would expect to find used in Git and GitHub a bit more than this.

```suggestion
        var cacheKey = $"{innerBranch.RemoteName}/{innerBranch.CanonicalName}@{innerBranch.Tip.Sha}|";
```

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
```diff
@@ -58,6 +63,26 @@ public int UncommittedChangesCount()
...
+        return cachedBranches.GetOrAdd(cacheKey, new Branch(innerBranch, repoDiff, this));
+    }
+
+    public Commit GetOrCreate(LibGit2Sharp.Commit innerCommit, Diff repoDiff) =>
+        cachedCommits.GetOrAdd(innerCommit.Sha, new Commit(innerCommit, repoDiff, this));
+
+    public Tag GetOrCreate(LibGit2Sharp.Tag innerTag, Diff repoDiff)
+    {
+        var cacheKey = $"{innerTag.CanonicalName}|{innerTag.Target.Sha}";
```
asbjornu, line 82:
```suggestion
        var cacheKey = $"{innerTag.CanonicalName}@{innerTag.Target.Sha}";
```

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
```diff
@@ -58,6 +63,26 @@ public int UncommittedChangesCount()
...
+        return cachedTags.GetOrAdd(cacheKey, new Tag(innerTag, repoDiff, this));
```
asbjornu, line 83:
```suggestion
        return cachedTags.GetOrAdd(cacheKey, () => new Tag(innerTag, repoDiff, this));
```

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
```diff
@@ -58,6 +63,26 @@ public int UncommittedChangesCount()
...
+        {
+            return new Branch(innerBranch, repoDiff, this);
+        }
+
+        var cacheKey = $"{innerBranch.CanonicalName}|{innerBranch.Tip.Sha}|{innerBranch.RemoteName}";
+        return cachedBranches.GetOrAdd(cacheKey, new Branch(innerBranch, repoDiff, this));
+    }
+
+    public Commit GetOrCreate(LibGit2Sharp.Commit innerCommit, Diff repoDiff) =>
+        cachedCommits.GetOrAdd(innerCommit.Sha, new Commit(innerCommit, repoDiff, this));
```
asbjornu, line 78:
```suggestion
        cachedCommits.GetOrAdd(innerCommit.Sha, () => new Commit(innerCommit, repoDiff, this));
```

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
```diff
@@ -58,6 +63,26 @@ public int UncommittedChangesCount()
...
 
+    public Branch GetOrCreate(LibGit2Sharp.Branch innerBranch, Diff repoDiff)
+    {
+        if (innerBranch.Tip is null)
+        {
+            return new Branch(innerBranch, repoDiff, this);
+        }
+
+        var cacheKey = $"{innerBranch.CanonicalName}|{innerBranch.Tip.Sha}|{innerBranch.RemoteName}";
+        return cachedBranches.GetOrAdd(cacheKey, new Branch(innerBranch, repoDiff, this));
```
asbjornu, line 74: Wouldn't lazy initialization here be even more optimal, avoiding the construction of a `Branch` instance altogether if it already exists?

```suggestion
        return cachedBranches.GetOrAdd(cacheKey, () => new Branch(innerBranch, repoDiff, this));
```

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
asbjornu, line 1: Why the rename from `GitCache` to `GitRepository`? I don't feel like the provided functionality is quite that of a repository, which is often used as an abstraction over a database. 🤔

asbjornu (approved):
This is most excellent! Well done! 👏🏼 I have a few questions and comments that can be ignored if there's good reason to, which is why I'm approving. But I would appreciate if they were addressed. :)

asbjornu:
Oh, it was set to auto-merge. 😅 Well well. You can work through my comments and see if you think they are worthy a followup PR, @pvanbuijtene.

arturcic:
> Oh, it was set to auto-merge. 😅 Well well. You can work through my comments and see if you think they are worthy a followup PR, @pvanbuijtene.

oh sorry about that, I might create a PR with the suggestions

src/GitVersion.LibGit2Sharp/Git/GitRepository.cs
arturcic: Agree

HHobeck:
@pvanbuijtene: Thank you for your contribution and the work you have done.

I have following remarks/suggestions:

- The classes Commit, Branch and so on are wrapper classes around the LibGit2Sharp objects. I would say the wrapper classes should not have a reference to GitRepostory. Is it an idea to extract the functionality of caching in a separate class and use this instead?
- Because of the nature of a wrapper class I would expect that the LibGit2Sharp objects are reused as well and the hash code is the same (otherwise we would have still a memory problem). Probably you can use the hash code instead of using an arbitrary string  (needs to be tested).

arturcic:
@asbjornu @HHobeck @pvanbuijtene follow up here https://github.com/GitTools/GitVersion/pull/4685

PR #4782, opened 2025-12-12
[WIP] Provide version dropdown in docs

### LINKED ISSUE #4537: [ISSUE]: Documentation "5.12.0"/"latest" link is confusing
### Prerequisites

### GitVersion package

N/A

### GitVersion version

N/A

### Operating system

N/A

### What are you seeing?

The documentation page has a header with a link between "latest" and "5.12.0" docs, introduced in [PR 4188](https://github.com/GitTools/GitVersion/pull/4188). This is certainly nice to have, but as a new user, the fact that the "latest" documentation shows "5.12.0" in the header was initially quite confusing. This would probably have been fine if there were several different versions to choose from in the header, but with only one link displayed in the header, the obvious interpretation is that the header shows "this is the version you're on", not "click here to go to this version". 

An additional contributing factor in my case is that my browser (Arc) by default only shows the domain name. Looking at the full URL makes it quite clear which version you're looking at, but this should not be necessary.

### What is expected?

Some possible ways to improve the UI here:
* Add some phrasing to the link between versions (e.g. "Go to [version]"). Clunky solution that doesn't scale if we want to add more links.
* Add more version links, and optionally turn the version selection into a dropdown. This would more clearly communicate that you're making a selection that takes you elsewhere.
* Show which version of the docs you're looking at somewhere, e.g. in the footer. 

### Steps to Reproduce

N/A

### RepositoryFixture Test

_No response_

```shell

```

Bi0T1N:
## Description
This provides a dropdown in the docs on the right side to select the desired docs for a specific version.

## Related Issue
Resolves #4537

## Motivation and Context
A dropdown in the docs will improve the user experience and ease the understanding which version of the docs are currently viewed (see #4537).  
It also allows to switch between multiple versions but this needs more changes since the docs for all versions need to be published. Dropdown selection for versions is used very often in documentation (e.g. https://docs.python.org/3.14/) and the design is inspired by [ROS 2](https://docs.ros.org/en/kilted/index.html)

## How Has This Been Tested?
By running the commands in a devcontainer that are provided in the [docs subfolder](https://github.com/GitTools/GitVersion/tree/main/docs). Browsing the docs was tested with Firefox under Linux.

Header

Header_dropdown

header_documentation

header_api

## Checklist:

### DIFF
--- a/docs/input/_Navbar.cshtml
+++ b/docs/input/_Navbar.cshtml
@@ -3,11 +3,65 @@
     {
         Tuple.Create("Documentation", Context.GetLink("docs")),
         Tuple.Create("API", Context.GetLink("api")),
-        Tuple.Create("<i></i> 5.12.0", "/5.12.0/docs"),
     };
+
+    // Define a list of versions with their respective links and status (e.g., latest, EOL).
+    var VersionList = new List<(string Name, string Link, string Status, bool IsReleased)>
+    {
+        ("5.12", "/5.12.0/docs", "(EOL)", true),
+        ("6.0", "/6.0.0/docs", "", true),
+        ("6.1-6.3", "/", "(latest)", true),
+        ("main branch", "/dev/docs", "(unstable)", false),
+    };
+
+    // determine currently viewed version
+    var CurrentlyViewedVersion = VersionList.Last().Name;
+    @foreach (var version in VersionList)
+    {
+        if (Context.GetLink(Document).StartsWith(version.Link))
+        {
+            CurrentlyViewedVersion = version.Name;
+            break;
+        }
+    }
+
     foreach(Tuple<string, string> p in pages)
     {
         string active = Context.GetLink(Document).StartsWith(p.Item2) ? "active" : null;
         <li class="@active"><a href="@p.Item2">@Html.Raw(p.Item1)</a></li>
     }
+
+    <li class="dropdown">
+        <a href="javascript:void(0)" class="dropbtn">
+            Other Versions - v: @Html.Raw(@CurrentlyViewedVersion)
+            <span class="fa fa-caret-down"></span>
+        </a>
+        <div class="dropdown-content">
+            <dl>
+                <dt>Releases</dt>
+                @foreach (var version in VersionList)
+                {
+                    if (version.IsReleased)
+                    {
+                        <dd>
+                        <a href="@version.Link">@version.Name @version.Status</a>
+                        </dd>
+                    }
+                }
+            </dl>
+            <dl>
+                <dt>In Development</dt>
+                @foreach (var version in VersionList)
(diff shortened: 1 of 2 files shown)


Bi0T1N:
The remaining issue is that the docs for different versions need to be created and published. Currently there are only docs for [5.12](https://gitversion.net/5.12.0/docs/) and [latest](https://gitversion.net/docs/). But I couldn't find where this is done. Or is this a manual task? :shrug:

arturcic:
> The remaining issue is that the docs for different versions need to be created and published. Currently there are only docs for [5.12](https://gitversion.net/5.12.0/docs/) and [latest](https://gitversion.net/docs/). But I couldn't find where this is done. Or is this a manual task? 🤷

v5.12 I manually ported back into the `main` branch, but I think we can automate this at publishing minor/major versions, @gep13 could probably help us with some suggestions

Bi0T1N:
Currently the website uses the `gh-pages` branch in this repository for GitHub Pages and I think we could use an approach as described [here](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#creating-a-custom-github-actions-workflow-to-publish-your-site). This approach might also work by adapting the code in the `PublishDocs.cs` but I assume this might be a bit more complex than using the existing features offered by GitHub workflow.
- The workflow for the default branch would build the HTML files for each commit and place them into e.g. a `/dev` folder for the latest docs
- Once a release is created, the main branch gets forked and the workflow detects that it is a release/branch by matching `github.ref_name` against `^[0-9]+\.[0-9]+\.[0-9]+$` and thus publishes to e.g. the `/v6.0.0` folder now. This would also allow fixes to docs in the release branch afterwards since it would just update the files inside that folder - if this is desired at all because normally a release is frozen. Otherwise it is run on a tag (which seems to be the case for 99% of the releases of GitVersion) and forget about fixing docs for released versions.
- The drawback is that you cannot link to the `v6.0.1` release from the `v6.0.0` docs because that would require that you always update the Navbar file in all existing releases for each new release but it will provide links to older and the latest (`/dev`) release. From there you can reach all other releases. (If this is important I guess it could be solved by using a single Navbar file for all version folders)
- The [existing publish workflow](https://github.com/GitTools/GitVersion/blob/main/.github/workflows/docs.yml#L123) would probably only need to build the HTML docs, run [actions/upload-pages-artifact](https://github.com/actions/upload-pages-artifact) and finally [actions/deploy-pages](https://github.com/actions/deploy-pages).
- The `gh-pages` branch is no longer needed. Otherwise it should be possible to use an similar approach and push the folders to the branch and trigger a workflow to deploy the pages from there.

arturcic:
Closing in favor of #5185

PR #4804, opened 2026-01-27
Expose VersionSourceSemVer as Output Variable

jakublatkowski:
This PR adds `VersionSourceSemVer` to the standard output variables exposed by GitVersion.

## Description

This PR adds `VersionSourceSemVer` to the standard output variables exposed by GitVersion, making it available alongside existing version fields in:
- Environment variables (e.g., `GitVersion_VersionSourceSemVer`)
- JSON output
- Build agent-specific outputs (Azure Pipelines, GitHub Actions, etc.)

The `VersionSourceSemVer` represents the base semantic version calculated by GitVersion before any pre-release tags or metadata are applied. This is the foundational version derived from the repository's version tags and configuration.

## Related Issue

This PR addresses a feature request for exposing VersionSourceSemVer in output variables.
Resolves #4803

### The Problem

GitVersion determines the version by identifying a "Base Version" (from a tag, merge message, or branch config) and then incrementally applying rules to calculate the new version. While we currently expose the distance from this base (via CommitsSinceVersionSource), the BaseVersion itself is internal. Users verify this "anchor" point only by reading verbose logs, making it difficult to debug why a specific version was calculated or to perform logic based on the version calculation's starting point.

### Why VersionSourceSemVer Matters

- Traceability & Debugging: It allows consumers to see exactly which version GitVersion selected as the anchor. This confirms whether the version is derived from the expected tag 1.0.0 or unexpectedly from a merge commit or branch configuration.
Workflow Flexibility: In strict tag-based workflows (like GitFlow), VersionSourceSemVer often correlates to the "previous release," enabling scripts to generate diffs or changelogs from that point.
- Consistency: It provides the missing "start" value to the existing "end" value (FullSemVer) and "distance" value (CommitsSinceVersionSource), completing the version calculation picture for CI/CD integrations.

> Note: Since GitVersion supports multiple strategies (e.g., specific branch naming), VersionSourceSemVer represents the  calculation baseline found by the strategy, which is not strictly always the historically previous release, but it is always the  deterministic anchor for the current build. 

### Why This Change is Valuable

By exposing `VersionSourceSemVer` as a standard output variable, we:
- **Eliminate workarounds**: Teams no longer need custom parsing logic
- **Improve consistency**: All version components are available through the same interface
- **Enable new integrations**: Simplifies integration with tools that need base version information
- **Follow principle of least surprise**: If GitVersion calculates it, users should be able to access it

## How Has This Been Tested?

**Unit Tests**: Added/updated tests to verify:
   - VersionSourceSemVer is included in output variable collections
   - Value is correctly propagated to all output channels
   - Backward compatibility with existing outputs

### Verification Steps
- Confirmed VersionSourceSemVer appears in all output formats
- Verified value matches the internally calculated base version
- Checked that existing output variables remain unchanged
- Tested with different GitVersion configuration scenarios

Example JSON output showing the new `VersionSourceSemVer` field:
```json
{
  "Major": 5,
  "Minor": 12,
  "Patch": 0,
  "SemVer": "5.12.0-beta.1",
  "FullSemVer": "5.12.0-beta.1+42",
  ...
  "VersionSourceSemVer": "5.12.0",
  ...
}
```

## Additional Notes
This change is purely additive - it introduces a new output variable without modifying any existing behavior or outputs. The risk of regression is minimal, and it opens up new integration possibilities for teams using GitVersion in their CI/CD pipelines.

## Checklist:


src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersionFormatValues.cs
```diff
@@ -52,6 +52,8 @@ public class SemanticVersionFormatValues(SemanticVersion semver, IGitVersionConf
 
     public string InformationalVersion => semver.ToString("i");
 
+    public string? VersionSourceSemVer => semver.BuildMetaData.VersionSourceSemVer?.ToString() ?? $"{semver.Major}.{semver.Minor}.{semver.Patch}";
```
HHobeck: In my opinion you should leave it empty or null instead of returning the calculated semantic version. What do you think?

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
```diff
@@ -98,7 +98,13 @@ public virtual SemanticVersion FindVersion()
             };
         }
 
-        return semanticVersion;
+        return new SemanticVersion(semanticVersion)
```
HHobeck: Please set the VersionSourceSemVer in VersionCalculatorBase::CreateVersionBuildMetaData method. Therefor you need to change the parameter from ICommit to IBaseVersion

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
```diff
@@ -121,7 +127,8 @@ private bool TryGetSemanticVersion(
             commitShortSha: Context.CurrentCommit.Id.ToString(7),
             commitDate: Context.CurrentCommit.When,
             numberOfUnCommittedChanges: Context.NumberOfUncommittedChanges
-        );
+        )
+        { VersionSourceSemVer = currentCommitTaggedVersion.Value };
```
HHobeck: Add the VersionSourceSemVer as a new parameter in the non-default constructor, please.

src/GitVersion.Core/VersionCalculation/SemanticVersioning/SemanticVersionFormatValues.cs
jakublatkowski: You're right. It might be confusing to calculate that version in the previous way. Removed completely.

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
jakublatkowski: Done, thank you.

src/GitVersion.Core/VersionCalculation/VersionCalculators/NextVersionCalculator.cs
jakublatkowski: Done, thank you

arturcic:
@jakublatkowski please make sure you rebase your branch onto `main` and not merge `main` into your branch

src/GitVersion.Schema/Program.cs
```diff
@@ -29,7 +29,7 @@
 }
 
 var builder = new JsonSchemaBuilder();
-const string jsonSchemaUri = "https://json-schema.org/draft-07/schema#";
+const string jsonSchemaUri = "http://json-schema.org/draft-07/schema#";
```
arturcic: I think this needs to be reverted

jakublatkowski: I am not sure about it. I added that because it seems it has some issues with https in json schema when building documentation. But unfortunately it does not look like it helped.

jakublatkowski: JSON Schema, $schema is primarily an identifier for “which dialect rules to use”, not “a URL the library will download”.

The JsonSchema.Net library we’re using resolves that identifier by doing an exact lookup in its DialectRegistry. Draft-07 is registered under:

http://json-schema.org/draft-07/schema# (canonical ID used by many implementations)

We can see that schema in previous generated schemas as well.

arturcic: I think reverting it to http will fix the schema tool

jakublatkowski: Updating it here didn't help. Are you suggesting a separate PR for that?

arturcic: it should be http, just revert the change in your PR as this is not needed. Keep the code in the schema project as was before

jakublatkowski: I reverted the change. Would you be able to fix the https issue?

arturcic: will have a look why it didn't pick it up

jakublatkowski: Https is committed to main branch. I tried to change it back to http (now reverted back to https), but it seems like it is cached, or this CI is using prevoius, or main build to run this tool

jakublatkowski: Https is committed to main branch. I tried to change it back to http (now reverted back to https), but it seems like it is cached, or this CI is using prevoius, or main build to run this tool

arturcic: It seems there is a thing I missed that Copilot changed, I will fix on main, you should rebase on main

jakublatkowski: Thanks, I completed the rebase. Unfortunately, I'm not sure if that helped.

arturcic: after rebase there is a commit 99eb699c4f93971cd444e5f70de56ce37a90888c - revert GitVersion.Schema schema uri change - drop this commit

jakublatkowski: You're right, thanks. I missed that.

PR #4944, opened 2026-05-19
feat: add support for environment variables in branch labels

### LINKED ISSUE #4745: [ISSUE]: Provide the ability to use environment variables in labels
### Prerequisites

### GitVersion package

gittools/actions

### GitVersion version

6

### Operating system

macOS

### What are you seeing?

GitVersion supports "{group-name}" style placeholders in labels via regex named groups. This works great for my team EXCEPT on pull requests. The label is limited to regular expression matching of the merge branch name (eg. `refs/pull/68/head`)

My team would really like to have the pull request source name (eg `github.head_ref`) available to place in the version label. This makes our version more readable when the team has several PR's in the same repository. 

### What is expected?

**Plan**: Support environment variables in GitVersion Labels

**Current State**
- GitVersion supports {regex-group-name} placeholders in labels via regex named groups.
- {env:VARIABLE_NAME} is supported in format strings (e.g., AssemblyInformationalFormat) via FormatWith.
- Labels are processed in GetBranchSpecificLabel in ConfigurationExtensions.cs, which only handles regex placeholders.
- IEnvironment is available in VariableProvider but not in label processing.

**Solution Approach**
Extend `GetBranchSpecificLabel` to also process `{env:...} `placeholders (e.g., `{env:GITHUB_HEAD_REF}`) after regex processing, using the existing FormatWith logic.

### Steps to Reproduce

GitVersion only supports {regex-group-name} placeholders in labels via regex named groups.
Unable to add branch name workflows running on a pull request.

### RepositoryFixture Test

_No response_

```shell

```

robertcoltheart:
## Description

Supports environment variable substitution in the label configuration for a branch using the established format used elsewhere. Fallbacks are also supported if the environment variable is not found. The changes are backwards compatible and should not fail for existing configuration.

Documentation detailing the changes has been added to `configuration.md`.

## Related Issue

Resolves #4745 

## Motivation and Context

Allow labels to include environment variables as well as regex-captured variables. This is useful for CICD environments like GitHub where the pull request head_ref can be provided instead of the merge commit target.

## How Has This Been Tested?

Unit tests covering the new functionality have been added.

## Checklist:

### DIFF
--- a/docs/input/docs/learn/who.md
+++ b/docs/input/docs/learn/who.md
@@ -12,7 +12,6 @@ that we know about today.
 * [ChocolateyGUI](https://github.com/chocolatey/ChocolateyGUI)
 * [GitLink](https://github.com/GitTools/GitLink)
 * [OctopusDeploy](https://github.com/OctopusDeploy)
-* [NUKE](https://nuke.build)
 * [Orc.\* packages](https://github.com/wildgums?query=orc)
 * [Orchestra](https://github.com/wildgums/orchestra)
 * [Shouldly](https://github.com/shouldly/shouldly)
--- a/docs/input/docs/reference/configuration.md
+++ b/docs/input/docs/reference/configuration.md
@@ -903,6 +903,8 @@ of `alpha.foo` with `label: 'alpha.{BranchName}'` and `regex: '^features?[\/-](?
 
 Another example: branch `features/sc-12345/some-description` would become a pre-release label of `sc-12345` with `label: '{StoryNo}'` and `regex: '^features?[\/-](?<StoryNo>sc-\d+)[-/].+'`.
 
+You can also use environment variable placeholders with the `{env:VARIABLE_NAME}` syntax. Environment variable placeholders can also be combined with regex placeholders, for example `{BranchName}-{env:VARIABLE_NAME}`, and support fallback values using the `{env:VARIABLE_NAME ?? "fallback"}` syntax.
+
 **Note:** To clear a default use an empty string: `label: ''`
 
 ### increment
--- a/docs/input/docs/reference/configuration.md
+++ b/docs/input/docs/reference/configuration.md
@@ -903,8 +903,6 @@ of `alpha.foo` with `label: 'alpha.{BranchName}'` and `regex: '^features?[\/-](?
 
 Another example: branch `features/sc-12345/some-description` would become a pre-release label of `sc-12345` with `label: '{StoryNo}'` and `regex: '^features?[\/-](?<StoryNo>sc-\d+)[-/].+'`.
 
-You can also use environment variable placeholders with the `{env:VARIABLE_NAME}` syntax. Environment variable placeholders can also be combined with regex placeholders, for example `{BranchName}-{env:VARIABLE_NAME}`, and support fallback values using the `{env:VARIABLE_NAME ?? "fallback"}` syntax.
-
 **Note:** To clear a default use an empty string: `label: ''`
 
 ### increment
--- a/src/GitVersion.Configuration.Tests/Configuration/ConfigurationExtensionsTests.cs
+++ b/src/GitVersion.Configuration.Tests/Configuration/ConfigurationExtensionsTests.cs
@@ -6,6 +6,8 @@ namespace GitVersion.Configuration.Tests;
 [TestFixture]
 public class ConfigurationExtensionsTests : TestBase
 {
+    private const string BranchName = "pull-request";
+
     [TestCase("release/2.0.0",
         "refs/heads/release/2.0.0", "release/2.0.0", "release/2.0.0",
         true, false, false, false, true)]
@@ -63,13 +65,13 @@ public class ConfigurationExtensionsTests : TestBase
 
         var configuration = GitFlowConfigurationBuilder.New
             .WithoutBranches()
-            .WithBranch("pull-request", builder => builder
+            .WithBranch(BranchName, builder => builder
                 .WithLabel("pr-{env:GITHUB_HEAD_REF}")
                 .WithRegularExpression(@"^pull[/-]"))
             .Build();
 
-        var effectiveConfiguration = configuration.GetEffectiveConfiguration(ReferenceName.FromBranchName("pull-request"));
-        var actual = effectiveConfiguration.GetBranchSpecificLabel(ReferenceName.FromBranchName("pull-request"), null, environment);
+        var effectiveConfiguration = configuration.GetEffectiveConfiguration(ReferenceName.FromBranchName(BranchName));
+        var actual = effectiveConfiguration.GetBranchSpecificLabel(ReferenceName.FromBranchName(BranchName), null, environment);
         actual.ShouldBe("pr-feature-branch");
     }
 
@@ -81,13 +83,13 @@ public class ConfigurationExtensionsTests : TestBase
 
         var configuration = GitFlowConfigurationBuilder.New
             .WithoutBranches()
-            .WithBranch("pull-request", builder => builder
+            .WithBranch(BranchName, builder => builder
                 .WithLabel("pr-{env:GITHUB_HEAD_REF ?? \"unknown\"}")
                 .WithRegularExpression(@"^pull[/-]"))
             .Build();
 
-        var effectiveConfiguration = configuration.GetEffectiveConfiguration(ReferenceName.FromBranchName("pull-request"));
-        var actual = effectiveConfiguration.GetBranchSpecificLabel(ReferenceName.FromBranchName("pull-request"), null, environment);
+        var effectiveConfiguration = configuration.GetEffectiveConfiguration(ReferenceName.FromBranchName(BranchName));
+        var actual = effectiveConfiguration.GetBranchSpecificLabel(ReferenceName.FromBranchName(BranchName), null, environment);
         actual.ShouldBe("pr-unknown");
     }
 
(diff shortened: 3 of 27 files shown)


docs/input/docs/reference/configuration.md
arturcic, line 1: please update `configuration.source.md` instead, this file gets generated from source file

robertcoltheart: Done. Do I need to run a tool to generate the `configuration.md` file for this PR?

arturcic: no, should be handled by the CI

robertcoltheart:
1 check is failing because the Nuke website is down.

docs/input/docs/learn/who.md
```diff
@@ -12,7 +12,6 @@ that we know about today.
 * [ChocolateyGUI](https://github.com/chocolatey/ChocolateyGUI)
 * [GitLink](https://github.com/GitTools/GitLink)
 * [OctopusDeploy](https://github.com/OctopusDeploy)
-* [NUKE](https://nuke.build)
```
robertcoltheart, line 15: Nuke is abandoned now and the website is unlikely to come back online. This was failing the build.

src/GitVersion.Core/Extensions/ConfigurationExtensions.cs
```diff
@@ -97,38 +98,25 @@ private static bool ShouldBeIgnored(ICommit commit, IIgnoreConfiguration ignore)
...
-                {
-                    var escapedGroupValue = escapedGroupValueLazy.Value;
-                    label = label.Remove(index, placeholder.Length).Insert(index, escapedGroupValue);
-                    startIndex = index + escapedGroupValue.Length;
-                }
-            }
-            return label;
+            var labelPlaceholders = BuildLabelPlaceholders(configuration.RegularExpression, effectiveBranchName);
+
+            return label.FormatWith(labelPlaceholders, environment);
```
robertcoltheart, line 119: Throws if environment variable is not found and has no fallback. This is deliberate to stay consistent with other usages of environment variables, for example in assembly file version.

robertcoltheart:
@HHobeck @arturcic Fixed a couple of issues that I missed. This is ready for review and merge now, thanks!

src/GitVersion.Configuration.Tests/Configuration/ConfigurationExtensionsTests.cs
```diff
@@ -51,7 +51,115 @@ public void EnsureGetBranchSpecificLabelWorksAsExpected(string branchName, strin
...
+                .WithRegularExpression(@"^pull[/-]"))
+            .Build();
+
+        var effectiveConfiguration = configuration.GetEffectiveConfiguration(ReferenceName.FromBranchName("pull-request"));
+        Should.Throw<ArgumentException>(() =>
+            effectiveConfiguration.GetBranchSpecificLabel(ReferenceName.FromBranchName("pull-request"), null, environment));
+    }
+
+    [Test]
+    public void EnsureGetBranchSpecificLabelProcessesEnvironmentVariablesEvenWhenRegexDoesNotMatch()
```
HHobeck: I don't understand the purpose of this test. Even if I write it as following it succeeded:

```csharp
    [Test]
    public void EnsureGetBranchSpecificLabelProcessesEnvironmentVariablesEvenWhenRegexDoesNotMatch()
    {
        var environment = new TestEnvironment();

        var configuration = GitFlowConfigurationBuilder.New
            .WithoutBranches()
            .WithBranch("feature", builder => builder
                .WithLabel("{BranchName}")
                .WithRegularExpression(@"^features?[\/-](?<BranchName>.+)"))
            .Build();

        var branchName = "feature/other-branch";
        var effectiveConfiguration = configuration.GetEffectiveConfiguration(
            ReferenceName.FromBranchName(branchName)
        );
        var actual = effectiveConfiguration.GetBranchSpecificLabel(
            branchName, null, environment
        );
        actual.ShouldBe("other-branch");
    }
```

src/GitVersion.Core/Formatting/StringFormatWithExtension.cs
HHobeck, line 1: Here are a few suggestions regarding this implementation:

1. **Use `object` instead of generics:** Change the generic type `T` to `object`. It makes more sense here since a generic type parameter provides no extra benefit for this use case.

2. **Clarify supported types:** The method currently supports instances of type `object` and `IDictionary<string, object>`, but this intent is not clearly expressed in the method body. I suggest splitting this into explicit overloads for better readability, like this:

```csharp
internal static class StringFormatWithExtension
{
    extension(string template)
    {
        public string FormatWith(object? source, IEnvironment environment)
        {
            object? GetFromSource(string value)
            {
                // ...
            }

            object? GetFromEnvironment(string value) => environment.GetEnvironmentVariable(value);
            
            return template.FormatWith(GetFromSource, GetFromEnvironment);
        }

        public string FormatWith(IDictionary<string, object> source, IEnvironment environment)
        {
            object? GetFromSource(string value)
            {
                source.TryGetValue(value, out var result);
                return result;
            }

            object? GetFromEnvironment(string value) => environment.GetEnvironmentVariable(value);

            return template.FormatWith(GetFromSource, GetFromEnvironment);
        }

        private string FormatWith(
            Func<string, object?> getFromSource, Func<string, object?> getFromEnvironment)
        {
            // ...
        }
    }
}
```

3. **Allow flexible placeholder combinations:** It does not make sense to restrict how placeholders can occur. Why shouldn't all of the following combinations be valid?
* `{JustAProperty}`
* `{JustAProperty ?? "JustAFallback"}`
* `{JustAProperty} ?? {env:JustAVariable}`
* `{JustAProperty} ?? {env:JustAVariable} ?? "JustAFallback"`
* `{env:JustAVariable}`
* `{env:JustAVariable} ?? "JustAFallback"`
* `{env:JustAVariable} ?? {JustAProperty}`
* `{env:JustAVariable} ?? {JustAProperty} ?? "JustAFallback"`

HHobeck (changes requested):

src/GitVersion.Core/Formatting/StringFormatWithExtension.cs
robertcoltheart: 1 and 2, fair enough, will make these changes.

For 3, it was written this way (and I'm guessing the original implementor in #4746 did it this way for the same reason) to remain compatible/easy to understand as the existing configuration options for `label` and for `assembly-file-versioning-format`, where the `{env:}` syntax is derived from. The understanding I have is that `{Name}` syntax is reserved for regex captures, whereas `{env:}` is used for environment variables. I think it would be more confusing given the language that the configuration has cultivated to conflate these 2 concepts.

If the longer-term intention is to further enhance the regex/env var to support this conflation, I'd suggest doing this in a follow up change rather than here. Let me know your thoughts on this.

src/GitVersion.Configuration.Tests/Configuration/ConfigurationExtensionsTests.cs
robertcoltheart: Agree. I copied this from the original implementor, will delete.

src/GitVersion.Core/Formatting/StringFormatWithExtension.cs
HHobeck: > For 3, it was written this way (and I'm guessing the original implementor in #4746 did it this way for the same reason) to remain compatible/easy to understand as the existing configuration options for `label` and for `assembly-file-versioning-format`, where the `{env:}` syntax is derived from. The understanding I have is that `{Name}` syntax is reserved for regex captures, whereas `{env:}` is used for environment variables. I think it would be more confusing given the language that the configuration has cultivated to conflate these 2 concepts.

Good question. @asbjornu @arturcic: What do you think?

robertcoltheart: Any thoughts on this guys?

asbjornu: I'm not sure I understand the following statement:

> The understanding I have is that `{Name}` syntax is reserved for regex captures

From [the documentation](https://gitversion.net/docs/reference/configuration#assembly-file-versioning-format):

> Expressions in curly braces reference one of the [variables](https://gitversion.net/docs/reference/variables) or a process-scoped environment variable (when prefixed with env:).

So for `{Name}`, `Name` would represent a GitVersion variable called `Name`. I think @HHobeck's example (in the configuration context):

```yaml
assembly-file-versioning-format: '{env:JustAVariable} ?? {JustAProperty} ?? "JustAFallback"`'
```

Should be possible as follows:

```yaml
assembly-file-versioning-format: '{env:JustAVariable ?? JustAProperty ?? "JustAFallback"}'
```

…the difference being the placement of the braces, which IIRC needs to embrace the entire expression. Whether one, two or multiple fallback values are possible already is not something I have tested, but I think it's something that should be supported.

HHobeck: > Should be possible as follows:
> 
> ```yaml
> assembly-file-versioning-format: '{env:JustAVariable ?? JustAProperty ?? "JustAFallback"}'
> ```

I agree. The would be the correction:
* `{JustAProperty}`
* `{JustAProperty ?? "JustAFallback"}`
* `{JustAProperty ?? env:JustAVariable}`
* `{JustAProperty ?? env:JustAVariable ?? "JustAFallback"}`
* `{env:JustAVariable}`
* `{env:JustAVariable ?? "JustAFallback"}`
* `{env:JustAVariable ?? JustAProperty}`
* `{env:JustAVariable ?? JustAProperty ?? "JustAFallback"}`

robertcoltheart: Fair enough, if that's the consensus I'll have a go at implementing it in this PR.

asbjornu: Awesome! But feel free to create a separate PR if you'd like.

arturcic:
@robertcoltheart please rebase onto `main` and fix the sonarcloud

arturcic:
@robertcoltheart could you please rebase instead of merging main into your branch? Thanks in advance. See https://github.com/GitTools/GitVersion/blob/main/CONTRIBUTING.md#git-history--pull-request-updates

arturcic:
@robertcoltheart please fix the Sonar Cloud issue as well https://sonarcloud.io/project/issues?id=GitTools_GitVersion&pullRequest=4944&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

arturcic:
@robertcoltheart is there anything left to be fixed/implemented based on the feedback. I'm considering to prepare a new release soon and was wondering if we can include this in the new release

robertcoltheart:
No I'm happy with this for now. I can address the updated fallback logic in a separate pull request.

arturcic:
@asbjornu @HHobeck can you have a second review and approve it if you're fine with the current state

HHobeck (approved):

arturcic:
@mergifyio refresh

PR #5018, opened 2026-07-03
feat(ci): sync GitVersion schema to SchemaStore on new schema releases

### LINKED ISSUE #5017: Automate SchemaStore PR creation for new minor/major releases
## Summary

Automate updating the GitVersion JSON schema entry on [SchemaStore/schemastore](https://github.com/SchemaStore/schemastore) whenever a new GitVersion version with a new schema is released, following the pattern of [SchemaStore/schemastore#5485](https://github.com/SchemaStore/schemastore/pull/5485).

Currently this update is done manually after each release that ships a new schema version. It should be automated instead.

## What the SchemaStore PR does

For a new version `X.Y`, it:
1. Adds `"X.Y": "https://gitversion.net/schemas/X.Y/GitVersion.configuration.json"` to the `versions` map for the GitVersion entry in `src/api/json/catalog.json`.
2. Bumps the top-level `"url"` for that entry to point at the new `X.Y` schema.
3. Updates the `$ref` in `src/schemas/json/gitversion.json` to point at the new `X.Y` schema URL.

## Acceptance criteria

- [ ] When a new schema version is published, a PR is automatically opened on `SchemaStore/schemastore` updating `catalog.json` and `gitversion.json` for the new version, without manual intervention.
- [ ] Releases that don't introduce a new schema version don't trigger a SchemaStore PR.
- [ ] It's possible to trigger the update manually for a specific version, in case the automatic detection is missed or needs to be re-run.

arturcic:
## Summary

Adds `.github/workflows/schemastore.yml`, which automates opening a PR against [SchemaStore/schemastore](https://github.com/SchemaStore/schemastore) whenever a new GitVersion schema version is published, per #5017.

## How it works

- **Trigger**: pushes to `gh-pages` that touch `schemas/**` (i.e. right after `docs.yml` publishes docs and the generated schema for the release), or `workflow_dispatch` with a `version` input for manual re-runs.
- **Detection**: on the automatic trigger, diffs `before`/`after` on `gh-pages` for newly *added* `schemas/*/GitVersion.configuration.json` files and derives the version from the folder name. If nothing new was added (e.g. a patch release with no new schema), the job no-ops. On manual trigger, the provided `version` input is used directly instead.
- **Auth**: uses a single token from `gittools/cicd/github-creds` (same mechanism as `homebrew.yml` uses for opening PRs against `homebrew-core`) for both pushing to the [GitTools/schemastore](https://github.com/GitTools/schemastore) fork and opening the PR against the upstream repo. A GitHub App token was considered (as used in `gittools-actions.yml`), but App installation tokens are scoped to repos the app is installed on and can't open a PR against `SchemaStore/schemastore`, which GitTools doesn't control — so a single PAT-based identity (like homebrew.yml) is the correct fit here.
- **Update logic**:
  1. `gh repo sync` brings the `GitTools/schemastore` fork's `master` up to date with upstream.
  2. Clones the fork, branches as `gitversion-schema-<version>`.
  3. Uses `jq` to bump the GitVersion entry's `url`, add the new `versions` map entry in `src/api/json/catalog.json`, and update the `$ref` in `src/schemas/json/gitversion.json` — matching [SchemaStore/schemastore#5485](https://github.com/SchemaStore/schemastore/pull/5485).
  4. Commits, pushes, and opens the PR via `gh pr create --repo SchemaStore/schemastore`.

## Test plan

- [ ] `actionlint` passes on the new workflow (verified locally).
- [ ] Manually trigger via `workflow_dispatch` with a real historical version (e.g. `6.7`) against a test fork to confirm the `jq` edits and PR creation work end-to-end, since this can't be exercised from a local sandbox (needs `OP_SERVICE_ACCOUNT_TOKEN` and push access to `GitTools/schemastore`).
- [ ] Confirm the automatic trigger correctly detects a newly-added schema folder on the next real release that ships a new schema version.

### DIFF
--- a/.github/workflows/homebrew.yml
+++ b/.github/workflows/homebrew.yml
@@ -61,7 +61,6 @@ jobs:
           brew bump-formula-pr gitversion \
             --url "$url" \
             --sha256 "$sha256" \
-            --fork-org gittools
             --no-audit \
             --no-browse \
             --message "For additional details see https://github.com/GitTools/GitVersion/releases/tag/${version}"
--- a/.github/workflows/schemastore.yml
+++ b/.github/workflows/schemastore.yml
@@ -1,0 +1,88 @@
+name: Update SchemaStore
+
+on:
+  push:
+    branches:
+      - gh-pages
+    paths:
+      - schemas/**
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'GitVersion schema version to publish (e.g. 6.7)'
+        required: true
+
+permissions:
+  contents: read
+
+jobs:
+  update-schemastore:
+    name: Open PR to SchemaStore
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Harden the runner (Audit all outbound calls)
+        uses: step-security/harden-runner@9af89fc71515a100421586dfdb3dc9c984fbf411 # v2.19.4
+        with:
+          egress-policy: audit
+
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          ref: gh-pages
+          fetch-depth: 2
+
+      - name: Detect schema version
+        id: detect
+        shell: bash
+        run: |
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ]; then
+            version="${{ inputs.version }}"
+          else
+            version="$(git diff --name-only --diff-filter=A "${{ github.event.before }}" "${{ github.event.after }}" -- 'schemas/*/GitVersion.configuration.json' \
+              | sed -E 's#schemas/([^/]+)/.*#\1#' | sort -V | tail -1)"
+          fi
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Load GitHub release token
+        id: github-creds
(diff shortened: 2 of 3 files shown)


arturcic:
Good catches, thanks — both fixed in 91fa2c3e2:

- Trigger: switched from `push: branches: [gh-pages]` to `workflow_run` keyed off `docs.yml` ("Verify & Publish Docs") completing. Confirmed `gh-pages` has no `.github/workflows` directory at all, so the push trigger would indeed never have fired.
- Diff: since `workflow_run` doesn't give a push `before`/`after`, the version detection now diffs `HEAD~1..HEAD` on `gh-pages` instead (docs.yml pushes a single commit per run), which also sidesteps the fetch-depth/before-SHA-availability issue.

arturcic:
Update: reverted to a pure `push: branches: [gh-pages]` + `workflow_dispatch` trigger (per maintainer request — the workflow_run approach changed *what* triggers it, which wasn't the intent).

To make the push trigger actually work, `docs.yml`'s `PublishDocs` task now also copies `.github/workflows/schemastore.yml` into the gh-pages commit it publishes, so the workflow file exists on that branch (GitHub resolves push-triggered workflows using the file as it exists on the pushed ref). Also widened checkout to `fetch-depth: 0` so the before/after diff is always valid regardless of how many commits a gh-pages push contains.

Issue #5037, opened 2026-07-11
GitVersion.Git adapter: managed reader behind the backend switch (Phase B.5)


arturcic:
Thin adapter implementing `IGitRepository` over the managed reader (mutations reuse the Phase A CLI mutator), registered when `GITVERSION_GIT_BACKEND=managed`; `DualBackendParityTests` harness asserting deep equality of ref enumeration, walk sequences, merge-bases, diffs and status between backends; real-world corpus JSON diffing as the flip gate.

Part of #5031 — see `docs/design/managed-git-migration.md` for the full design.

arturcic:
Done via #5051 (merged into feature/managed-git).

arturcic:
Reopening: feature/managed-git was rolled back to pre-B.5 (e5b7c1a5f), undoing #5051's merge. Phase B.5 is now tracked by #5053, which carries the complete change set including the Sonar fixes and the dedup refactor.

arturcic:
Done via #5053 (merged into feature/managed-git); parity test harness follows in #5055.

PR #5055, opened 2026-07-12
test(git): dual-backend parity tests; align annotated-tag TargetSha with libgit2

arturcic:
Adds the **DualBackendParityTests** validation layer from the managed-git migration plan (Phase B): the same on-disk fixture is opened with both backends (libgit2 and managed) in a single test process, independent of the `GITVERSION_GIT_BACKEND` selection, and every read surface is deep-compared:

- reference / branch / tag / remote (+refspec) enumeration, including indexer lookups by friendly and canonical names and `FromGlob`
- order-sensitive `QueryBy` sequences across all sort-strategy combinations, first-parent-only and exclusion variants, over a criss-cross history with equal-timestamp merge commits
- `FindMergeBase` over **all commit pairs** of the criss-cross fixture
- `DiffPaths` for every commit
- `UncommittedChangesCount` (modified + staged + untracked), empty-repository and detached-HEAD edge cases, `Path`/`WorkingDirectory`

The suite immediately caught one real divergence: for an **annotated tag**, libgit2's `ITag.TargetSha` is the sha of the object the tag annotation points at (a single dereference — usually the commit), not the tag object's own sha. `ManagedTag` now matches (`ManagedRepositorySession.GetTagTargetSha`).

Verified: solution builds with 0 warnings; Core.Tests **36138/36138** under both `GITVERSION_GIT_BACKEND=libgit2` and `managed`; `dotnet format --verify-no-changes` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Part of #5031 (Phase B parity hardening, follow-up to #5037/#5053).

### DIFF
--- a/src/GitVersion.Core.Tests/Core/DualBackendParityTests.cs
+++ b/src/GitVersion.Core.Tests/Core/DualBackendParityTests.cs
@@ -1,0 +1,355 @@
+using GitVersion.Git;
+using GitVersion.Helpers;
+
+namespace GitVersion.Tests;
+
+/// <summary>
+/// Opens the same on-disk repository with both Git backends (libgit2 and managed) and
+/// deep-compares every read surface: reference/branch/tag enumeration, order-sensitive
+/// commit walks, merge bases over all commit pairs, diff paths and the uncommitted
+/// changes count. This is the direct parity oracle demanded by the managed-git
+/// migration plan, independent of which backend the test run selects globally.
+/// </summary>
+[TestFixture]
+public class DualBackendParityTests : TestBase
+{
+    private static readonly CommitSortStrategies[] SortCombinations =
+    [
+        CommitSortStrategies.None,
+        CommitSortStrategies.Time,
+        CommitSortStrategies.Topological,
+        CommitSortStrategies.Topological | CommitSortStrategies.Time,
+        CommitSortStrategies.Time | CommitSortStrategies.Reverse,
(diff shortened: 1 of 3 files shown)


src/GitVersion.Core.Tests/Core/DualBackendParityTests.cs
```diff
@@ -0,0 +1,355 @@
...
+
+        return commits;
+    }
+
+    private sealed record BackendPair(IGitRepository LibGit2, IGitRepository Managed) : IDisposable
+    {
+        public void Dispose()
+        {
+            LibGit2.Dispose();
+            Managed.Dispose();
```
asbjornu, line 352: If `LibGit2.Dispose()` throws, `Managed.Dispose()` won't be invoked, right? So perhaps wrap the first one in a `try/finally` and do `Managed.Dispose()` in the `finally`?
```suggestion
            try {
                LibGit2.Dispose();
            } finally {
                Managed.Dispose();
            }
```

asbjornu (commented):
I love the direction this is taking! Here's just a minor thought I had.

src/GitVersion.Core.Tests/Core/DualBackendParityTests.cs
arturcic: Hey @asbjornu I will continue the work and when I will be done with Claude, let's definitely have a proper review. for now I iterate with Claude to get to a point where I can say it's more or less implemented, then I will have once more a proper review of the `feature/managed-git` branch. v7 will be a big version.

PR #5062, opened 2026-07-14
test(msbuild): serialize process-wide env mutation across MsBuild fixtures

arturcic:
## Description

Fixes an intermittent `WriteVersionInfoTask` test failure that showed up far more often under the managed Git backend. Investigation (dual-backend) traced it to a **pre-existing process-global environment-variable race in the MsBuild test fixtures**, not the git backend.

Part of #5031.

### Root cause

`MsBuildTaskFixture` mutates the real process environment (set/reset `TF_BUILD` etc.) so the task can detect the build agent, and `MsBuildExeFixture` launches a child process that inherits it. Under `[assembly: Parallelizable(ParallelScope.Fixtures)]` these race:
- an Azure task sees `TF_BUILD` cleared by a concurrent fixture → falls back to `LocalBuild`, emits no `##vso` output (empty log); or
- a non-CI MSBuild child inherits `TF_BUILD=true` left set by a concurrent Azure fixture → wrongly emits `##vso`.

The git backend computes the **correct version in every case** — the managed backend only *widened the race window* (~14×) because its CLI-driven normalization makes each task execution much slower (~2.8s vs libgit2 ~0.2s). The race is latent on libgit2 too.

### Fix (test-infra only; product code untouched)

New `MsBuildProcessEnvironment` helper with a shared lock + `Reset()`/`Apply()`. Both fixtures hold the lock across the whole reset→set→execute→reset window (and the exe fixture resets before launching the child), making the environment-sensitive sections mutually exclusive without disabling fixture parallelism.

### Verification

- Before: full managed-mode suite failed within 2–6 parallel runs.
- After: 14/14 (agent) + 5/5 (independent) clean full-suite managed-mode runs.
- `GitVersion.Git.Managed.Tests` 141/141; `src` + `new-cli` build 0/0; `dotnet format --verify-no-changes` clean. No product or libgit2 code changed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/GitVersion.MsBuild.Tests/Helpers/MsBuildProcessEnvironment.cs
+++ b/src/GitVersion.MsBuild.Tests/Helpers/MsBuildProcessEnvironment.cs
@@ -1,0 +1,80 @@
+using GitVersion.Agents;
+
+namespace GitVersion.MsBuild.Tests.Helpers;
+
+/// <summary>
+/// An exclusive scope over the *process-wide* environment-variable manipulation that the MsBuild
+/// task and MsBuild.exe fixtures depend on. The MsBuild task detects the current build agent from
+/// the real process environment, and the MsBuild.exe fixture launches a child process that inherits
+/// it, so both are sensitive to the environment being mutated by another fixture.
+/// </summary>
+/// <remarks>
+/// Under <c>[assembly: Parallelizable(ParallelScope.Fixtures)]</c> several fixtures run at the same
+/// time. Without coordination, one fixture setting a CI variable such as <c>TF_BUILD</c> could leak
+/// into another fixture's task or child process (making it wrongly detect a build agent), and one
+/// fixture's reset could clear another's variables mid-execution. <see cref="Enter"/> takes an
+/// exclusive lock, resets the build-agent variables to a clean baseline and applies the requested
+/// variables; disposing the scope resets them again and releases the lock, so the whole window is
+/// mutually exclusive:
+/// <code>
+/// using var environment = MsBuildProcessEnvironment.Enter(variables);
+/// // ... run the task or build ...
+/// </code>
+/// </remarks>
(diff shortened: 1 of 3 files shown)


src/GitVersion.MsBuild.Tests/Helpers/MsBuildProcessEnvironment.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            return;
+        }
+
+        foreach (var (key, value) in environmentVariables)
+        {
+            SysEnv.SetEnvironmentVariable(key, value);
+        }
+    }
+}
```
asbjornu: Perhaps the `IDisposable` pattern would make usage of this more elegant? The lock object could be made private by obtaining the lock inside the constructor and releasing it in the `Dispose()` method. Just an idea.

arturcic: Good idea — done. `MsBuildProcessEnvironment` is now an `IDisposable` scope: the lock object is private, `Enter()` acquires it (and resets the build-agent variables to a clean baseline) in the constructor, and `Dispose()` resets again and releases it. The call sites read much better now:

```csharp
using var environment = MsBuildProcessEnvironment.Enter();
environment.Apply(this.environmentVariables);
return func();
```

and in the exe fixture:

```csharp
using (MsBuildProcessEnvironment.Enter())
{
    results = analyzer.Build(environmentOptions);
}
```

The public `Lock` and the separate `Reset()`/`Apply()` calls are gone; the reset-on-enter / reset-on-dispose sequencing is encapsulated in the scope's lifecycle. Verified the managed-backend MsBuild.Tests suite stays green in parallel and `dotnet format` is clean. Thanks!

PR #4659, opened 2025-08-16
Introduce `CustomVersion` variable

phatcher:
Introduces a new output variable `CustomVersion` and an associated formatting variable `custom-version-format`

## Description

Intent is to allow the user to define an output variable without changing the semantics of existing ones. This can then be used to resolve issues such as providing an appropriate version to tools such as NuGet which do not fully followed semantic versioning rules.

## Motivation and Context

See above

## How Has This Been Tested?

Additional tests in `ConfigurationProviderTest` for custom-version-format and updated test outputs elsewhere to align
## Checklist:

* \[x] My code follows the code style of this project.
* \[x] My change requires a change to the documentation.
* \[x] I have updated the documentation accordingly.
* \[x] I have added tests to cover my changes.
* \[ ] All new and existing tests passed.

PR #5098, opened 2026-07-24
Add CustomVersion output variable

### LINKED ISSUE #1674: Add configuration param to define PreReleaseTag separator
I'm trying to use SemVer as a single identifier for all build artifacts. GitVersion Babmoo plugin is a great solution for almost all cases, but unfortunately I have linux RPM+YUM-based distribution. 

[RPM+YUM versioning](https://fedoraproject.org/wiki/Package_Versioning_Examples#Complex_versioning_examples) technicaly maps to SemVer abstractions (semver pre-release + build metadata = rpm release), but doesn't allow dashes in rpm-release string. Standard GitVersion config for feature branch `feature/JIRA-123` generates invalid artifacts like `name-1.4.3-JIRA-123.rpm`. Moreover, branch `feature/JIRA.123` also generates invalid rpm version. They say similar problem appears in other package managers, but i have no details. 

Current behavior comes from regex replacement of `[^a-zA-Z0-9]` symbols with universal separator. Now this separator is hardcoded in [core](https://github.com/asbjornu/GitVersion/blob/feature/ignore-merge-messages/src/GitVersionCore/VersionCalculation/NextVersionCalculator.cs#L126) and has dash = `-` as default value. I just want to add config parameter to specify other separator, i.e. dot = `.`

### LINKED ISSUE #2065: Feature request : Generating PEP440 compliant versions
Hello,

We are using GitVersion to version packages written in many different languages and up to now; most package managers are supporting classic SemVer.

We now want to use GitVersion to version for our Python packages; thing is, the Python community decided to invent their own version standard which is not compatible with SemVer : PEP440.

e.g. a pre-release tag can only be equal to a predefined list (So something like `0.6.3-featurebranch.10` is forbidden) (See https://www.python.org/dev/peps/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions)

I know GitVersion is mainly .Net oriented, but would it be possible to also print out a PEP440 compliant version number along with the other versions numbers? And while leveraging the various development mode GitVersion offers?

Thanks again, we love your tool!

- Frank

### LINKED ISSUE #3340: [Feature]
Feature request to include some value that only increases so that it can be used as an int.

## Context
Some systems, like Google Play Store for Android apps, have something called `ApplicationVersion` which is an integer that should only increase.

Using GitVersion is great for SemVer, for example, but it doesn't seem to have anything that outputs a straightforward integer version that could only increase.

Using a trick like concatenating major, minor and patch version would be undesirable as 
0.0.123 would be concatenated to 123 but it would be smaller than 0.1.0 which would be concatenated to 10

## Possible Implementation
Include a total git commit or similar
```
git rev-list --count --all
```

arturcic:
## Description

Adds a configurable `CustomVersion` output variable controlled by `custom-version-format`. The format supports the existing version-variable and environment-variable formatters and defaults to `{SemVer}` without changing the semantics of existing outputs.

The value is available in JSON output, generated version files, WiX output, and the MSBuild `GetVersion` task. Configuration, CLI override, output approvals, documentation, and generated schemas are updated as well.

## Related Issue

Closes #1674
Closes #2065
Closes #3340
Closes #4659

Supersedes the implementation attempted in #4659 and follows the design described in [discussion #4272](https://github.com/GitTools/GitVersion/discussions/4272#discussioncomment-11335454).

## Motivation and Context

Consumers sometimes need a version format with semantics different from GitVersion's standard outputs. Examples covered by this change include:

- PEP 440-compatible pre-release versions such as `0.6.3beta10`.
- Fixed-width monotonically increasing integer versions for Android, such as `0001000`.
- CI-owned build numbers supplied through environment variables.
- RPM/YUM-safe versions without dashes in the release value.

For #1674, separators in a branch-derived label can be removed with the PascalCase formatter:

```yaml
custom-version-format: '{MajorMinorPatch}.{PreReleaseLabel:c}.{PreReleaseNumber}'
```

When the exact replacement must be controlled—for example, preserving `JIRA.123` instead of reducing `JIRA-123` to `Jira123`—replace `{PreReleaseLabel:c}` with an environment variable containing the desired RPM-safe label:

```yaml
custom-version-format: '{MajorMinorPatch}.{env:RPM_PRERELEASE_LABEL}.{PreReleaseNumber}'
```

## How Has This Been Tested?

- `dotnet build ./src/GitVersion.slnx --no-restore`
- `dotnet test --project ./src/GitVersion.Configuration.Tests/GitVersion.Configuration.Tests.csproj --no-restore` — 97 passed
- App parser override tests — 40 passed
- Relevant Core variable, JSON, and executor tests — 37 passed
- `dotnet test --project ./src/GitVersion.Output.Tests/GitVersion.Output.Tests.csproj --no-restore` — 123 passed
- `dotnet test --project ./src/GitVersion.MsBuild.Tests/GitVersion.MsBuild.Tests.csproj --no-restore` — 151 passed
- `dotnet format ./src/GitVersion.slnx --verify-no-changes --no-restore`
- Configuration and output schemas regenerated successfully

A broader Core test run reached an unrelated timing-sensitive failure in `PerformanceScenarios.RepositoryWithALotOfTags` (3518 ms versus the 2500 ms threshold); all CustomVersion-related tests pass.

## Checklist

* [x] My code follows the code style of this project.
* [x] My change requires a change to the documentation.
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [ ] All new and existing tests passed. See the performance-test note above.


HHobeck:
Nice catch. I haven't found any issues with the implementation. The only point I'd like to raise is that, from a conceptual point of view, I think this property should be configurable on a per-branch basis if we want to properly support e.g. PEP 440.

See:
- https://github.com/GitTools/GitVersion/discussions/4272#discussioncomment-11335635

arturcic:
@HHobeck move the property to BranchConfiguration instead, please review

HHobeck:
```csharp
internal record BranchConfiguration : IBranchConfiguration
{
    [JsonPropertyName("custom-version-format")]
    [JsonPropertyDescription($"Specifies the format of CustomVersion. Defaults to '{DefaultCustomVersionFormat}'.")]
    [JsonPropertyDefault($"'{DefaultCustomVersionFormat}'")]
    public string? CustomVersionFormat { get; set; }
    ...
}
```

I would prefer not to define fallback configuration values in code. We usually rely on predefined workflows to provide default values, which makes those defaults transparent to the end user.

Also, I'm not sure a default value of `SemVer` for `custom-version-format` is necessary. As the name suggests, it's intended for custom formatting. Why would we output the semantic version twice?

Edit:
That means for GitHubFlow, GitFow and TrunkBased workflow I would leave it empty.

arturcic:
> We usually rely on predefined workflows to provide default values, which makes those defaults transparent to the end user.

Good catch

HHobeck:
I was also thinking about removing DeploymentMode and using the new formatting scheme to generate the different version formats instead. That would require a more generic property name than custom-version-format. The DeploymentMode effects only how the semver will be formatted.

What do you think? @asbjornu

HHobeck:
Maybe I'm wrong and it is more then just formatting:

image

arturcic:
I will focus only on adding the possibility to specify a custom version format so it can be used to generate custom version and later we can decide on how to extend or make it more generci

src/GitVersion.Core.Tests/VersionCalculation/VariableProviderTests.cs
```diff
@@ -247,8 +247,8 @@ public void ProvidesVariablesInContinuousDeliveryModeForFeatureBranchWithCustomA
         var configuration = GitFlowConfigurationBuilder.New.WithTagPreReleaseWeight(0)
             .WithAssemblyInformationalFormat("{Major}.{Minor}.{Patch}+{VersionSourceDistance}.Branch.{BranchName}.Sha.{ShortSha}")
             .Build();
-        var preReleaseWeight = configuration.GetEffectiveConfiguration(ReferenceName.FromBranchName("develop")).PreReleaseWeight;
-        var variables = this.variableProvider.GetVariablesFor(semanticVersion, configuration, preReleaseWeight);
+        var effectiveConfiguration = configuration.GetEffectiveConfiguration(ReferenceName.FromBranchName("develop"));
```
HHobeck: A little bit strange here to search for an non-existing branch:

Maybe we should extend the extensions as following:

```csharp
internal static class ConfigurationExtensions
{
    extension(IGitVersionConfiguration configuration)
    {
        public EffectiveConfiguration GetEffectiveConfiguration()
        {
            var branchConfiguration = configuration.GetEmptyBranchConfiguration();
            return new EffectiveConfiguration(configuration, branchConfiguration, null);
        }
    ...
    }
}
```

or just create the EffectiveConfiguration in the unit test or just pass in just the preReleaseWeight and the customVersionFormat

HHobeck (changes requested):

HHobeck (approved):

PR #5123, opened 2026-08-12
feat: support version bump reset messages

### LINKED ISSUE #4144: [FEATURE]: Support of '=semver: none|minor|major|patch' bump message to suppress the version increment
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched [issues](https://github.com/GitTools/GitVersion/issues) to ensure it has not already been reported

### GitVersion package

GitVersion.Tool

### GitVersion version

6.0.1

### Operating system

Windows

### What are you seeing?

[FEATURE]: Support of '=semver: none|minor|major|patch' bump message to suppress the version increment

This feature was inspired from the following discussion:
- https://github.com/GitTools/GitVersion/discussions/4143

We need a new bump message format to suppress the version increment of a version calculation run. I think e.g. `=semver: none|minor|major|patch` bump message should be used as a default format for the following workflows:
- GitFlow/v1
- GitHubFlow/v1
- TrunkBased/preview1

### What is expected?

If you specify `=semver: none` the increment value of the e.g. branch configuration will be overruled.

### Steps to Reproduce

Using the configuration:
```yaml
workflow: TrunkBased/preview1
branches:
  main:
    mode: ContinuousDelivery
    increment: Patch
```

**USE CASE 1:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one =semver: none'
- 'message two =semver: none'
- 'message three =semver: none'

should yield to:
1.0.0-3

**USE CASE 2:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one +semver: minor'
- 'message two =semver: none'
- 'message three =semver: none'

should yield to:
1.1.0-3

**USE CASE 3:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one =semver: none'
- 'message two +semver: minor'
- 'message three =semver: none'

should yield to:
1.1.0-2

**USE CASE 4:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one =semver: none'
- 'message two =semver: none'
- 'message three +semver: minor'

should yield to:
1.1.0-1

Using the configuration:
```yaml
workflow: GitFlow/v1
branches:
  main:
    mode: ContinuousDelivery
    increment: Patch
```

**USE CASE 5:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one =semver: none'
- 'message two =semver: none'
- 'message three =semver: none'

should yield to:
1.0.0-3

**USE CASE 6:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one +semver: minor'
- 'message two =semver: none'
- 'message three =semver: none'

should yield to:
1.0.0-3

**USE CASE 7:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one =semver: none'
- 'message two +semver: minor'
- 'message three =semver: none'

should yield to:
1.0.0-3

**USE CASE 8:**
with the following commits on main:
- 'a commit with a tag' (tag: 1.0.0)
- 'message one =semver: none'
- 'message two =semver: none'
- 'message three +semver: minor'

should yield to:
1.1.0-3

### RepositoryFixture Test

_No response_

_No response_

fzlzjerry:
## Description

Add a configurable `version-bump-reset-message` pattern with the default
`=semver:` for the GitFlow, GitHubFlow, and TrunkBased workflows.

The default major, minor, patch, and no-bump patterns now accept either `+semver`
or `=semver`. Those existing patterns select the increment value; a matching
reset message discards increments accumulated before that commit and suppresses
the configured branch increment. The existing `+semver` minimum-increment
behavior is unchanged.

Reset detection is centralized in `IncrementStrategyFinder` and exposed through
`CommitMessageIncrement`. Range-based strategies scan from the newest commit,
consolidate newer bump messages, and stop at the first reset. Mainline clears its
accumulated increment at the reset commit and keeps the branch increment
suppressed until the current mainline segment ends.

This also updates effective configuration output, CLI overrides, the 7.0 schema,
workflow snapshots, and the configuration and version-increment documentation.

## Related Issue

Resolves #4144

## Motivation and Context

Regular `+semver` messages treat the configured branch increment as a minimum,
so they cannot lower a Patch branch to None for documentation-only changes. The
new `=semver` reset form can suppress that branch increment without changing the
behavior of existing `+semver` messages.

## How Has This Been Tested?

- `dotnet test ./src/GitVersion.Core.Tests/GitVersion.Core.Tests.csproj --no-restore`
  (36,211 passed)
- `dotnet test ./src/GitVersion.Configuration.Tests/GitVersion.Configuration.Tests.csproj --no-restore`
  (100 passed)
- `dotnet test ./src/GitVersion.App.Tests/GitVersion.App.Tests.csproj --no-restore`
  (356 passed)
- `dotnet format ./src/GitVersion.slnx --verify-no-changes --no-restore`
- `git diff --check`

The Core integration coverage includes all eight acceptance scenarios from the
issue, all four reset values across the three built-in workflows, a custom reset
pattern, disabled commit-message incrementing, the requested merged-branch
regression, newest-first traversal stopping at the most recent reset, and a
malformed reset marker without a valid increment (32 focused cases).

Not applicable.

## Checklist:

* [x] My code follows the code style of this project.
* [x] My change requires a change to the documentation.
* [x] I have updated the documentation accordingly.
* [x] I have added tests to cover my changes.
* [x] All new and existing tests passed.


arturcic:
@fzlzjerry please rebase your branch onto `main` (do not merge `main` into your branch)

arturcic:
hey @HHobeck mind to have a look?

src/GitVersion.Core/VersionCalculation/Mainline/MainlineContext.cs
```diff
@@ -20,6 +20,8 @@ internal record MainlineContext(IIncrementStrategyFinder IncrementStrategyFinder
 
     public VersionField Increment { get; set; }
 
+    public VersionField? IncrementOverride { get; set; }
```
HHobeck: In my opinion this information is not necessary. It is enough to store a boolean to supress the increment forced by the branch configuration. e.g. bool SuppressBranchIncrement { get; set; }

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/MainlineVersionStrategy.cs
```diff
@@ -450,7 +451,22 @@ private static IEnumerable<IBaseVersionIncrement> GetIncrements(MainlineIteratio
             {
                 foreach (var item in incrementer.GetIncrements(iteration, commit, context))
                 {
-                    yield return item;
+                    if (item is BaseVersionOperand)
```
HHobeck: Don't understand the code change. In my opinion it is not necessary to change the current logic.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/MainlineVersionStrategy.cs
```diff
@@ -450,7 +451,22 @@ private static IEnumerable<IBaseVersionIncrement> GetIncrements(MainlineIteratio
...
+                        hasBaseVersionOperator = false;
+                    }
+
+                    if (item is not BaseVersionOperator
+                        || context.IncrementOverride != VersionField.None
+                        || !hasBaseVersionOperator)
+                    {
+                        yield return item;
+                    }
+
```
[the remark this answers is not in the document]
fzlzjerry: `null` intentionally means there is no override, so existing operator processing must remain unchanged. Requiring `HasValue` causes two TrunkBased acceptance cases with a later `+semver: minor` bump to regress. I rewrote the condition to make that distinction explicit. All 29 `VersionBumpOverrideScenarios` and the full 36,215-test Core suite pass.

fzlzjerry:
@arturcic Rebased onto current `main` (`7505ddb6f`) without a merge commit and force-updated the branch. I also clarified why a null increment override must preserve normal operator processing. Local checks: 36,215 Core tests, 101 Configuration tests, 356 App tests, and `dotnet format --verify-no-changes` all pass.

src/GitVersion.Core/VersionCalculation/Mainline/EnrichIncrement.cs
```diff
@@ -15,7 +15,15 @@ public void Enrich(MainlineIteration iteration, MainlineCommit commit, MainlineC
             ? VersionField.None
             : GetIncrementForcedByCommit(context, commit.Value, effectiveConfiguration);
         commit.Increment = incrementForcedByCommit;
-        context.Increment = context.Increment.Consolidate(incrementForcedByBranch, incrementForcedByCommit);
+        context.IncrementOverride = commit.IsDummy
```
HHobeck: Don't understand the logic here. It looks quite complex. If you ask me I would expect the following code:

```csharp
(var incrementForcedByCommit, bool suppressBranchIncrement) = GetIncrementForcedByCommit(
    context, commit.Value, effectiveConfiguration
);
suppressBranchIncrement |= context.SuppressBranchIncrement;
context.SuppressBranchIncrement = suppressBranchIncrement;
context.Increment = context.Increment.Consolidate(incrementForcedByCommit);
if(!suppressBranchIncrement)
{
    context.Increment = context.Increment.Consolidate(incrementForcedByBranch);
}
```

docs/input/docs/workflows/GitFlow/v1.yml
```diff
@@ -22,6 +22,7 @@ major-version-bump-message: "\\+semver:\\s?(breaking|major)"
 minor-version-bump-message: "\\+semver:\\s?(feature|minor)"
 patch-version-bump-message: "\\+semver:\\s?(fix|patch)"
 no-bump-message: "\\+semver:\\s?(none|skip)"
+override-version-bump-message: "=semver:\\s?(?<increment>none|patch|minor|major)"
```
HHobeck: I would say to change the configuration as following:

```yaml
minor-version-bump-message: "[\\+=]semver:\\s?(feature|minor)"
patch-version-bump-message: "[\\+=]semver:\\s?(fix|patch)"
no-bump-message: "[\\+=]semver:\\s?(none|skip)"
version-bump-reset-message: "=semver:"
```

But I would not call the last one override-version-bump-message, because = isn't really an override in the sense of "this value wins forever". It's a reset/baseline operator. Thus, my recommendation would be version-bump-reset-message or version-bump-baseline-message (RegEx)

HHobeck (changes requested):
Thank you for your work on this! I’ve left a few comments for your consideration.

src/GitVersion.Core/VersionCalculation/Mainline/EnrichIncrement.cs
fzlzjerry: Implemented this shape in 97edd62ce. `GetIncrementForcedByCommit` now returns `(Increment, SuppressBranchIncrement)`, carries the flag in the context, consolidates the commit increment first, and only adds the branch increment when suppression is false. All 26 focused scenarios and the full 36,205-test Core suite pass.

src/GitVersion.Core/VersionCalculation/VersionSearchStrategies/MainlineVersionStrategy.cs
fzlzjerry: Agreed. I reverted the changes to `MainlineVersionStrategy`; its incrementer iteration is now identical to `main`. With branch suppression handled in `EnrichIncrement`, the extra operator filtering is no longer needed.

src/GitVersion.Core/VersionCalculation/Mainline/MainlineContext.cs
fzlzjerry: Agreed. I replaced `IncrementOverride` with `bool SuppressBranchIncrement` in `MainlineContext` and removed the nullable override/parser model. The flag is reset at the same main-branch boundary as `context.Increment`.

docs/input/docs/workflows/GitFlow/v1.yml
fzlzjerry: Agreed on the reset terminology. The option is now `version-bump-reset-message` with the default `=semver:`; the major, minor, patch, and no-bump defaults accept `[+=]`. I applied the same change to major because #4144 explicitly includes `=semver: major`. The named capture parser and validation are gone, and the schema, snapshots, CLI coverage, and docs now describe reset/baseline semantics.

fzlzjerry:
The new CI run is green except for `Test / macos-26 - net10.0 - libgit2` again. All three built-in retry attempts failed in different `GitVersion.MsBuild.Tests` cases with the same `ArgumentNullException (BuildEventContext)` from `ParallelConsoleLogger` / `MsBuildPipeLogger`; every other test module in those attempts passed. The previous head failed in the same job with the same exception, while the other platforms, format, docs, CodeQL, and Sonar checks pass. I do not have upstream Actions admin permission to rerun it. @arturcic, could you please rerun job 94801513682?

fzlzjerry:
@HHobeck I reworked this around the reset/baseline model in 97edd62ce and replied to all four review threads. `MainlineContext` now only stores `SuppressBranchIncrement`, the `MainlineVersionStrategy` iteration is unchanged from `main`, and the option is `version-bump-reset-message`. The focused scenarios and full Core suite pass. Would you mind taking another look?

arturcic:
> The new CI run is green except for `Test / macos-26 - net10.0 - libgit2` again. All three built-in retry attempts failed in different `GitVersion.MsBuild.Tests` cases with the same `ArgumentNullException (BuildEventContext)` from `ParallelConsoleLogger` / `MsBuildPipeLogger`; every other test module in those attempts passed. The previous head failed in the same job with the same exception, while the other platforms, format, docs, CodeQL, and Sonar checks pass. I do not have upstream Actions admin permission to rerun it. @arturcic, could you please rerun job 94801513682?

please rebase on main again - it should be fixed

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -244,4 +261,6 @@ public VersionField GetIncrementForcedByCommit(ICommit commit, IGitVersionConfig
 
         return GetIncrementFromCommit(commit, majorRegex, minorRegex, patchRegex, none) ?? VersionField.None;
     }
+
+    private readonly record struct CommitMessageIncrement(VersionField Increment, bool SuppressBranchIncrement);
```
HHobeck: I would name it to VersionBumpNeedsToBeReset or similar. Maybe move this record class and make it public

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
HHobeck: Return CommitMessageIncrement

HHobeck, line 203: return CommitMessageIncrement

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -35,34 +35,51 @@ public VersionField DetermineIncrementedField(
...
-        var increments = commits
-            .Select(c => GetIncrementFromCommit(c, majorRegex, minorRegex, patchRegex, noBumpRegex))
-            .Where(v => v != null)
-            .ToList();
+        CommitMessageIncrement? result = null;
+        foreach (var commit in commits.Reverse())
+        {
+            var increment = GetIncrementFromCommit(commit, majorRegex, minorRegex, patchRegex, noBumpRegex);
+            var isVersionBumpReset = versionBumpResetRegex.IsMatch(commit.Message);
+            if (isVersionBumpReset)
```
HHobeck: isVersionBumpReset should be part of the result of CommitMessageIncrement

src/GitVersion.Core/VersionCalculation/Mainline/EnrichIncrement.cs
```diff
@@ -41,14 +47,28 @@ private static VersionField GetIncrementForcedByCommit(
...
             )
         };
     }
+
+    private static bool IsVersionBumpReset(ICommit commit, string? pattern)
+    {
+        var regex = pattern is null
+            ? RegexPatterns.VersionCalculation.DefaultVersionBumpResetRegex
+            : RegexPatterns.Cache.GetOrAdd(pattern);
+        return regex.IsMatch(commit.Message);
```
HHobeck: This logic seems to me redundant. Please consider to move this into IncrementStrategyFinder

src/GitVersion.Core/VersionCalculation/Mainline/EnrichIncrement.cs
```diff
@@ -11,11 +11,17 @@ public void Enrich(MainlineIteration iteration, MainlineCommit commit, MainlineC
...
         var effectiveConfiguration = commit.GetEffectiveConfiguration(context.Configuration);
         var incrementForcedByBranch = effectiveConfiguration.Increment.ToVersionField();
-        var incrementForcedByCommit = commit.IsDummy
-            ? VersionField.None
+        var (incrementForcedByCommit, suppressBranchIncrement) = commit.IsDummy
+            ? (VersionField.None, false)
             : GetIncrementForcedByCommit(context, commit.Value, effectiveConfiguration);
         commit.Increment = incrementForcedByCommit;
-        context.Increment = context.Increment.Consolidate(incrementForcedByBranch, incrementForcedByCommit);
+        suppressBranchIncrement |= context.SuppressBranchIncrement;
```
HHobeck: I know I have provided you this algo. But the implementation is not right. Please see following test:

    [TestCase(
        Workflow.TrunkBased,
        "message one +semver: major",
        "message two =semver: none",
        "message three +semver: patch",
        "1.0.1-4"
    )]
    public void __Just_A_Test__(
        Workflow workflow, string firstMessage, string secondMessage, string thirdMessage, string expectedVersion)
    {
        var configuration = BuildConfiguration(workflow, IncrementStrategy.Minor);

        fixture.BranchTo("feature/foo");
        fixture.MakeACommit(firstMessage);
        fixture.MakeACommit(secondMessage);
        fixture.MakeACommit(thirdMessage);
        fixture.MergeTo("main");

        fixture.AssertFullSemver(expectedVersion, configuration);
    }

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -35,34 +35,51 @@ public VersionField DetermineIncrementedField(
...
         var noBumpRegex = TryGetRegexOrDefault(configuration.NoBumpMessage, RegexPatterns.VersionCalculation.DefaultNoBumpRegex);
+        var versionBumpResetRegex = TryGetRegexOrDefault(
+            configuration.VersionBumpResetMessage, RegexPatterns.VersionCalculation.DefaultVersionBumpResetRegex);
 
-        var increments = commits
-            .Select(c => GetIncrementFromCommit(c, majorRegex, minorRegex, patchRegex, noBumpRegex))
-            .Where(v => v != null)
-            .ToList();
+        CommitMessageIncrement? result = null;
+        foreach (var commit in commits.Reverse())
```
HHobeck: You are iterating backwards beginning with the newest commit, right? My expectation would be if you are facing a reset bump message commit to break the loop, because it doesn't matter if any other bump messages existing after. Can you provide a unit test to ensure this scenario, please?

HHobeck (changes requested):

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Done. `GetIncrementFromCommit` now returns a nullable `CommitMessageIncrement`, so the increment and reset match are calculated and cached together.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Done. The reset match is now part of `CommitMessageIncrement`, and range aggregation carries that state with the consolidated increment.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Done. Range traversal now starts with the newest commit and breaks at the first reset. `VersionBumpResetStopsCommitHistoryAtTheMostRecentReset` covers the older-major, reset-none, newer-patch sequence for GitFlow and GitHubFlow.

src/GitVersion.Core/VersionCalculation/Mainline/EnrichIncrement.cs
fzlzjerry: Done. I removed the duplicate reset regex logic from `EnrichIncrement`; `IncrementStrategyFinder` now owns both increment and reset detection.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Done in `8d87db300`. I moved `CommitMessageIncrement` to its own public type and renamed the flag to `VersionBumpNeedsToBeReset`; `IIncrementStrategyFinder` now returns that result.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Done. `GetIncrementForcedByCommit` now returns `CommitMessageIncrement`, including the reset state.

src/GitVersion.Core/VersionCalculation/Mainline/EnrichIncrement.cs
fzlzjerry: Fixed and covered by `VersionBumpResetDiscardsEarlierBranchIncrementBeforeMerge`. Mainline now clears the accumulated increment at the reset commit while preserving branch-increment suppression, producing the requested `1.0.1-4`.

fzlzjerry:
@arturcic Rebased onto current `main` (`c5aa98937`), including the MSBuild pipe logger fix from #5138, without a merge commit. The updated head is `8d87db300`. The focused reset scenarios and the full Core, Configuration, and App suites pass locally.

fzlzjerry:
@HHobeck I addressed all seven follow-up threads in `8d87db300`, including the exact `1.0.1-4` regression scenario and newest-first reset traversal coverage. All focused and full local suites pass. Would you mind taking another look?

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -35,34 +35,52 @@ public VersionField DetermineIncrementedField(
...
+        foreach (var commit in commits)
+        {
+            var commitMessageIncrement = GetIncrementFromCommit(
+                commit, majorRegex, minorRegex, patchRegex, noBumpRegex, versionBumpResetRegex);
+            if (!commitMessageIncrement.HasValue)
+            {
+                continue;
+            }
+
+            result = result.HasValue
```
HHobeck, line 68: May I ask you to move this logic into CommitMessageIncrement? It would make the code cleaner and readable. Name the method e.g. Consolidate

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -127,7 +145,7 @@ [.. this.taggedSemanticVersionRepository
             }
         }
 
-        return commitLog.Values;
+        return [.. intermediateCommits.Where(commit => commitLog.ContainsKey(commit.Sha))];
```
HHobeck: Don't understand why you have changed this. Is it not the same to return commitLog.Values?

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
```diff
@@ -171,9 +189,17 @@ private ICommit[] GetHeadCommits(ICommit? headCommit, IIgnoreConfiguration ignor
...
-    private VersionField? GetIncrementFromCommit(ICommit commit, Regex majorRegex, Regex minorRegex, Regex patchRegex, Regex noBumpRegex) =>
+    private CommitMessageIncrement? GetIncrementFromCommit(
+        ICommit commit, Regex majorRegex, Regex minorRegex, Regex patchRegex, Regex noBumpRegex, Regex versionBumpResetRegex) =>
         this.commitIncrementCache.GetOrAdd(commit.Sha, () =>
-            GetIncrementFromMessage(commit.Message, majorRegex, minorRegex, patchRegex, noBumpRegex));
+        {
+            var increment = GetIncrementFromMessage(commit.Message, majorRegex, minorRegex, patchRegex, noBumpRegex);
+            var versionBumpNeedsToBeReset = versionBumpResetRegex.IsMatch(commit.Message);
+
+            return increment.HasValue || versionBumpNeedsToBeReset
```
HHobeck: This seems to me not right. If you have a commit message like:
```
=semver: I FORGOT THE VERSION INCREMENT
```
Should we treated this as valid? IMO only parse versionBumpResetRegex  if incremetn is not null here. Please add an unit test for this scenario. Thank you.

HHobeck (changes requested):
Okay looks quite good. Good work. If you fix the last three open comments I will approve it.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Done in c9b131e45. `CommitMessageIncrement.Consolidate` now owns both the `VersionField` consolidation and reset-state combination, and `IncrementStrategyFinder` delegates to it.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Agreed — restored `commitLog.Values` and the original `Dictionary<string, ICommit>.ValueCollection` return type in c9b131e45. The extra projection was unnecessary.

src/GitVersion.Core/VersionCalculation/IncrementStrategyFinder.cs
fzlzjerry: Fixed in c9b131e45. Reset matching now only runs after a valid increment has been parsed, so a message such as `=semver: I FORGOT THE VERSION INCREMENT` is ignored and the configured increment remains in effect. Added regression coverage for all three workflows.

HHobeck (approved):

Issue #5130, opened 2026-08-12
[Feature] Support regex-based branch and tag exclusions in ignore configuration


arturcic:
## Summary

Extend the `ignore` configuration with regex-based `branches` and `tags` collections. Matching branches should be excluded from version-source discovery and build-server branch normalization, while matching tags should be excluded as version-source candidates.

The existing ignore configuration supports `commits-before`, `sha`, and `paths`, but it cannot exclude branch or tag references by name.

## Proposed configuration

```yaml
ignore:
  branches:
    - ^release/legacy/
    - ^support/6\.
  tags:
    - ^experimental-
    - ^v0\.
```

Both properties contain regular-expression patterns. A reference is excluded when any configured pattern matches.

## Matching semantics

- Branch patterns are matched against the origin-independent friendly branch name. For example, `^release/` matches both `release/example` and `origin/release/example`.
- Tag patterns are matched against the friendly tag name without the `refs/tags/` prefix.
- Matching is case-insensitive, consistent with reference-name handling.
- Users can anchor expressions with `^` and `$` when a whole-name match is required.
- Invalid regular expressions fail configuration validation with a useful error.

## Calculation semantics

- `ignore.branches` excludes matching non-current branches from branch-based version-source discovery, including main-branch and release-branch discovery.
- The current or explicitly targeted branch always remains eligible for calculation, even when a broad expression such as `.*` matches it.
- Ignoring a branch must not globally exclude every commit reachable from that branch. Git commits can be shared by ignored and non-ignored branches, so doing so would incorrectly remove shared history.
- `ignore.tags` excludes matching tags as version-source candidates. It does not exclude the commit targeted by the tag.
- Existing `commits-before`, `sha`, and `paths` behavior remains unchanged.

## Repository-normalization semantics

- During build-server normalization, matching remote-tracking branches are not created or updated as local branches.
- The current or explicitly targeted branch is never skipped during preparation.
- Existing local references do not need to be deleted; calculation filtering ensures matching branches are not used as auxiliary version sources.
- Behavior must be consistent across the LibGit2Sharp and managed Git backends.

Filtering network fetches is out of scope. Git fetch refspecs use glob-style patterns rather than arbitrary .NET regular expressions, and CI systems often fetch before GitVersion starts. This feature can avoid GitVersion branch-normalization work after fetching, but it does not promise reduced network transfer.

## Implementation considerations

[...]

PR #5147, opened 2026-08-17
feat: support regex tag exclusions

arturcic:
## Description

Adds `ignore.tags`, a sequence of regular expressions for excluding tags from version-source discovery.

- Matches case-insensitively against the friendly tag name without `refs/tags/`.
- Uses OR semantics when multiple expressions are configured.
- Rejects invalid expressions during configuration validation.
- Keeps the tagged commit itself eligible for version calculation.
- Includes configuration parsing, builder, cache isolation, public API, schema, workflow snapshot, documentation, and integration coverage.

## Related issue

Part of #5130.

## Stack

1. [#5151](https://github.com/GitTools/GitVersion/pull/5151) - order configuration properties alphabetically
2. This PR - ignore tags
3. [#5148](https://github.com/GitTools/GitVersion/pull/5148) - ignore branches

Depends-On: GitTools/GitVersion#5151

## Validation

- `GitVersion.Configuration.Tests`: 107 passed.
- Focused `GitVersion.Core.Tests` with LibGit2Sharp: 7 passed.
- Focused `GitVersion.Core.Tests` with managed Git: 7 passed.
- BuildPrepare completed successfully with zero warnings and errors.
- `dotnet format src/GitVersion.slnx --no-restore` completed successfully.
- Regenerated only `schemas/7.0/GitVersion.configuration.json`.

## Checklist

- [x] Code follows the project style.
- [x] Documentation is updated.
- [x] Tests cover the change.
- [x] New and existing affected tests pass.

### DIFF
--- a/src/GitVersion.Core/Core/TaggedSemanticVersionRepository.cs
+++ b/src/GitVersion.Core/Core/TaggedSemanticVersionRepository.cs
@@ -1,4 +1,5 @@
 using System.Collections.Concurrent;
+using System.Globalization;
 using GitVersion.Configuration;
 using GitVersion.Extensions;
 using GitVersion.Git;
@@ -9,11 +10,11 @@ namespace GitVersion;
 internal sealed class TaggedSemanticVersionRepository(ILogger<TaggedSemanticVersionRepository> logger, IRepositoryStore repositoryStore) : ITaggedSemanticVersionRepository
 {
     private readonly ILogger<TaggedSemanticVersionRepository> logger = logger.NotNull();
-    private readonly ConcurrentDictionary<(IBranch, string, SemanticVersionFormat), IReadOnlyList<SemanticVersionWithTag>>
+    private readonly ConcurrentDictionary<(IBranch, string, SemanticVersionFormat, string), IReadOnlyList<SemanticVersionWithTag>>
         taggedSemanticVersionsOfBranchCache = new();
-    private readonly ConcurrentDictionary<(IBranch, string, SemanticVersionFormat), IReadOnlyList<(ICommit Key, SemanticVersionWithTag Value)>>
+    private readonly ConcurrentDictionary<(IBranch, string, SemanticVersionFormat, string), IReadOnlyList<(ICommit Key, SemanticVersionWithTag Value)>>
         taggedSemanticVersionsOfMergeTargetCache = new();
-    private readonly ConcurrentDictionary<(string, SemanticVersionFormat), IReadOnlyList<SemanticVersionWithTag>>
+    private readonly ConcurrentDictionary<(string, SemanticVersionFormat, string), IReadOnlyList<SemanticVersionWithTag>>
         taggedSemanticVersionsCache = new();
 
     private readonly IRepositoryStore repositoryStore = repositoryStore.NotNull();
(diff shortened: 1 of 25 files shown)


arturcic:
Hi @HHobeck, could you please take a look at this PR when you have a chance? Thank you!

HHobeck:
I need some time for the review. Anyway, what I was thinking about: Could we make case-insensitivity optional? Users who need it can easily add the modifier inline themselves, for example: `(?i)^experimental-`

src/GitVersion.Core/Core/TaggedSemanticVersionRepository.cs
```diff
@@ -134,4 +135,14 @@ IEnumerable<SemanticVersionWithTag> GetElements()
             }
         }
     }
+
+    private static string GetIgnoreCacheKey(IIgnoreConfiguration ignore)
```
HHobeck: Can you explain me for what this is good for? Why do we need to consider this in the cache? IMO the Ignore configuration is static and doesn't change.

src/GitVersion.Core/Core/TaggedSemanticVersionRepository.cs
HHobeck: Maybe change the filter function to IEnumerable<>

arturcic:
Thanks for raising this. GitVersion currently compiles configuration-supplied regexes through the shared `RegexPatterns.Cache`, which applies `RegexOptions.IgnoreCase`. This includes branch regexes, `tag-prefix`, version-in-branch patterns, bump-message patterns, custom merge-message formats, and `ignore.paths`.

I’d therefore prefer to keep `ignore.branches` and `ignore.tags` case-insensitive by default for consistency with the existing configuration behavior.

Users who need case-sensitive matching can disable case-insensitive matching inline with `(?-i)`, for example:

```yaml
ignore:
  tags:
    - '(?-i)^experimental-'\n```\n\nThis matches `experimental-...` but not `Experimental-...`.

File: src/GitVersion.Core/Core/TaggedSemanticVersionRepository.cs (hunk 1 above)
Comment (arturcic) [reply src/GitVersion.Core/Core/TaggedSemanticVersionRepository.cs]: You’re right—the ignore configuration is static per execution. I’ll remove it from the cache key and preserve the existing cache behavior.

**Comment by HHobeck** [comment]
> Thanks for raising this. GitVersion currently compiles configuration-supplied regexes through the shared `RegexPatterns.Cache`, which applies `RegexOptions.IgnoreCase`. This includes branch regexes, `tag-prefix`, version-in-branch patterns, bump-message patterns, custom merge-message formats, and `ignore.paths`.
> 
> I’d therefore prefer to keep `ignore.branches` and `ignore.tags` case-insensitive by default for consistency with the existing configuration behavior.
> 
> Users who need case-sensitive matching can disable case-insensitive matching inline with `(?-i)`, for example:
> 
> ```yaml
> ignore:
>   tags:
>     - '(?-i)^experimental-'\n```\n\nThis matches `experimental-...` but not `Experimental-...`.
> ```

Hmm okay. Can you document it? Fair enough if we want to have the behavior to be case insensitive. But why we have for example this in our configuration?

```yaml
tag-prefix: "[vV]?"
version-in-branch-pattern: "(?<version>[vV]?\\d+(\\.\\d+)?(\\.\\d+)?).*"
```

arturcic:
> [Hmm okay. Can you document it?](https://github.com/GitTools/GitVersion/pull/5147#issuecomment-5340452933)

@HHobeck Documented in a1c81bc80: the reference now explains the case-insensitive default and `(?-i)` opt-out, and every regex-specific schema description includes the same guidance. The `[vV]` defaults are redundant under the current behavior, so I left them unchanged as unrelated cleanup.

arturcic:
Hi @HHobeck, when you have a chance, could you please take another look at the updated PR? Your review feedback has been addressed. Thank you!

HHobeck (approved):

PR #5148, opened 2026-08-17
feat: support regex branch exclusions

### LINKED ISSUE #5130: [Feature] Support regex-based branch and tag exclusions in ignore configuration

arturcic:
## Description

Adds `ignore.branches`, a sequence of regular expressions for excluding auxiliary branches from version-source discovery and build-server normalization.

- Matches case-insensitively against branch names without the local or remote prefix.
- Uses OR semantics and treats local and remote-tracking names equivalently.
- Preserves the current branch and explicitly selected target branch.
- Keeps shared commits, shared tags, and configuration inheritance eligible.
- Avoids creating or updating ignored local branches during normalization while leaving existing local refs untouched.
- Includes configuration, public API, generated schema/workflows, documentation, calculation scenarios, and normalization coverage.

## Related issue

Closes #5130.

## Stack

1. [#5151](https://github.com/GitTools/GitVersion/pull/5151) - order configuration properties alphabetically
2. [#5147](https://github.com/GitTools/GitVersion/pull/5147) - ignore tags
3. This PR - ignore branches

Depends-On: GitTools/GitVersion#5147

## Validation

- `GitVersion.Configuration.Tests`: 113 passed.
- Focused `GitVersion.Core.Tests` with LibGit2Sharp: 29 passed.
- Focused `GitVersion.Core.Tests` with managed Git: 29 passed.
- BuildPrepare completed successfully with zero warnings and errors.
- `dotnet format src/GitVersion.slnx --no-restore` completed successfully.
- Regenerated only `schemas/7.0/GitVersion.configuration.json`.

## Checklist

- [x] Code follows the project style.
- [x] Documentation is updated.
- [x] Tests cover the change.
- [x] New and existing affected tests pass.

### DIFF
--- a/src/GitVersion.Core/Extensions/ConfigurationExtensions.cs
+++ b/src/GitVersion.Core/Extensions/ConfigurationExtensions.cs
@@ -116,6 +116,15 @@ internal static class ConfigurationExtensions
 
             return !ignoreConfig.IsEmpty ? source.Where(element => ShouldBeIncluded(element, ignoreConfig)) : source;
         }
+
+        public bool IsBranchIgnored(ReferenceName branchName)
+        {
+            ignoreConfig.NotNull();
+            branchName.NotNull();
+
+            return ignoreConfig.Branches.Any(expression =>
+                RegexPatterns.Cache.GetOrAdd(expression).IsMatch(branchName.WithoutRemote));
+        }
     }
 
     private static bool IsTagIgnored(string tagName, IIgnoreConfiguration ignore)
--- a/docs/input/docs/reference/configuration.md
+++ b/docs/input/docs/reference/configuration.md
@@ -174,6 +174,7 @@ branches:
 commit-date-format: yyyy-MM-dd
 commit-message-incrementing: Enabled
 ignore:
+  branches: []
   paths: []
   sha: []
   tags: []
@@ -211,7 +212,7 @@ update-build-number: true
 version-bump-reset-message: "=semver:"
 version-in-branch-pattern: "(?<version>[vV]?\\d+(\\.\\d+)?(\\.\\d+)?).*"
 ```
-<sup><a href='/docs/workflows/GitFlow/v1.yml#L1-L169' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/GitFlow/v1.yml' title='Start of snippet'>anchor</a></sup>
+<sup><a href='/docs/workflows/GitFlow/v1.yml#L1-L170' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/GitFlow/v1.yml' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
 The supported built-in configuration for the `GitHubFlow` workflow (`workflow: GitHubFlow/v1`) looks like:
@@ -301,6 +302,7 @@ branches:
 commit-date-format: yyyy-MM-dd
 commit-message-incrementing: Enabled
 ignore:
+  branches: []
   paths: []
   sha: []
   tags: []
@@ -338,7 +340,7 @@ update-build-number: true
 version-bump-reset-message: "=semver:"
 version-in-branch-pattern: "(?<version>[vV]?\\d+(\\.\\d+)?(\\.\\d+)?).*"
 ```
-<sup><a href='/docs/workflows/GitHubFlow/v1.yml#L1-L118' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/GitHubFlow/v1.yml' title='Start of snippet'>anchor</a></sup>
+<sup><a href='/docs/workflows/GitHubFlow/v1.yml#L1-L119' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/GitHubFlow/v1.yml' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
 The preview built-in configuration (experimental usage only) for the `TrunkBased` workflow (`workflow: TrunkBased/preview1`) looks like:
@@ -417,6 +419,7 @@ branches:
 commit-date-format: yyyy-MM-dd
 commit-message-incrementing: Enabled
 ignore:
+  branches: []
   paths: []
   sha: []
   tags: []
@@ -450,7 +453,7 @@ update-build-number: true
 version-bump-reset-message: "=semver:"
 version-in-branch-pattern: "(?<version>[vV]?\\d+(\\.\\d+)?(\\.\\d+)?).*"
 ```
-<sup><a href='/docs/workflows/TrunkBased/preview1.yml#L1-L103' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/TrunkBased/preview1.yml' title='Start of snippet'>anchor</a></sup>
+<sup><a href='/docs/workflows/TrunkBased/preview1.yml#L1-L104' title='Snippet source file'>snippet source</a> | <a href='#snippet-/docs/workflows/TrunkBased/preview1.yml' title='Start of snippet'>anchor</a></sup>
 <!-- endSnippet -->
 
 The details of the available options are as follows:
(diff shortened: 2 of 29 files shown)


arturcic:
Hi @HHobeck, could you please take a look at this PR when you have a chance? Thank you!

asbjornu (approved):
I think this looks superb! 👍🏼

src/GitVersion.Core/Extensions/ConfigurationExtensions.cs
```diff
@@ -116,6 +116,27 @@ public IEnumerable<ICommit> Filter(ICommit[] source)
...
+            ignoreConfig.NotNull();
+            branchName.NotNull();
+
+            var friendlyName = GetFriendlyBranchNameWithoutRemote(branchName);
+            return ignoreConfig.Branches.Any(expression =>
+                RegexPatterns.Cache.GetOrAdd(expression).IsMatch(friendlyName));
+        }
+    }
+
+    private static string GetFriendlyBranchNameWithoutRemote(ReferenceName branchName)
```
HHobeck: Please use branch.Name.WithoutOrigin (needs to be renamed to WithoutRemote IMO). The method BranchName::RemoveOrigin needs to be changed as well:

```csharp
private string RemoveRemote()
{
    if (IsRemoteBranch && !IsPullRequest)
    {
        var separatorIndex = Friendly.IndexOf('/');

        if (separatorIndex >= 0)
        {
            return Friendly[(separatorIndex + 1)..];
        }
    }

    return Friendly;
}
```

HHobeck (changes requested):

src/GitVersion.Core/Extensions/ConfigurationExtensions.cs
arturcic: Implemented in 3c8c8ec23 as a separate reviewable commit: `WithoutOrigin` is now `WithoutRemote` as an intentional v7 breaking change, arbitrary remote names are stripped centrally, all consumers are updated, and local/origin/upstream/pull-request coverage was added. Could you please take another look?

arturcic:
@HHobeck when you have time please have a new look

HHobeck (approved):

HHobeck (changes requested):
Could you change the configuration for GitFlow, GitHubFlow and TrunkbasedWorkflow and ignore all branches who are not permitted?

arturcic (commented):
@HHobeck, regarding [your review comment](https://github.com/GitTools/GitVersion/pull/5148#pullrequestreview-4972839696): could you clarify which branches should be considered "not permitted"? Each workflow currently has an `unknown` catch-all, while adding all configured branch patterns to `ignore.branches` would exclude the workflow’s normal version sources.

arturcic:
> Could you change the configuration for GitFlow, GitHubFlow and TrunkbasedWorkflow and ignore all branches who are not permitted?

@HHobeck, could you clarify which branches should be considered "not permitted"? Each workflow currently has an `unknown` catch-all, while adding all configured branch patterns to `ignore.branches` would exclude the workflow’s normal version sources.

HHobeck:
> > Could you change the configuration for GitFlow, GitHubFlow and TrunkbasedWorkflow and ignore all branches who are not permitted?
> 
> @HHobeck, could you clarify which branches should be considered "not permitted"? Each workflow currently has an `unknown` catch-all, while adding all configured branch patterns to `ignore.branches` would exclude the workflow’s normal version sources.

Good point. These two concepts seem to be clashing. But is it more like if your current branch is not known, it will still be used to calculate the version? Or what happens if your current branch is ignored?

arturcic:
> Good point. These two concepts seem to be clashing. But is it more like if your current branch is not known, it will still be used to calculate the version? Or what happens if your current branch is ignored?

Codex researched the current behavior and helped draft this clarification:

An unmatched current branch uses the workflow’s `unknown` configuration and is calculated normally. If the current branch matches `ignore.branches`, it is still protected during normalization and calculated normally. The option only prevents matching branches from being considered as version sources when calculating another branch; it does not disable calculation for the checked-out branch.

HHobeck:
> Codex researched the current behavior and helped draft this clarification:
> 
> An unmatched current branch uses the workflow’s `unknown` configuration and is calculated normally. If the current branch matches `ignore.branches`, it is still protected during normalization and calculated normally. The option only prevents matching branches from being considered as version sources when calculating another branch; it does not disable calculation for the checked-out branch.

So it would make sense to ignore the not known branch, correct? It was just an idea to improve performance. If you think it is not necessary please go ahead and leave it like it is.

arturcic:
Let's keep it like that, please approve

HHobeck (approved):

arturcic:
@mergifyio refresh

Issue #5228, opened 2026-09-18
[ISSUE]: next-version must include tag-prefix


judgeaxl:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched [issues](https://github.com/GitTools/GitVersion/issues) to ensure it has not already been reported

### GitVersion package

GitVersion.Portable

### GitVersion version

6.8.2

### Operating system

Windows

### What are you seeing?

When configuring tag-prefix to something non-optional, next-version cannot be used without including that prefix.

In a monorepo I've got a tag-prefix per package, like "package-a/v" and "package-b/v". If I then set next-version to something normal like "1.4.5", I get an error saying "Failed to parse 1.4.5 into a Semantic Version".

If I change next-version to package-a/v1.4.5, I get the desired result.

### What is expected?

Since next-version is not a tag, it's a version, I don't see any reason for it to require the tag-prefix, it should just be read as-is as a version without stripping any tag.

### Steps to Reproduce

This can be reproduced with a minimal config in any repo with at least one commit, no tags needed:
```
tag-prefix: 'project-a/'
# fails
next-version: 1.2.3
# works
next-version: project-a/1.2.3
```

### Output log or link to your CI build (if appropriate).

```shell

```

asbjornu:
Yea, this makes sense. Pull requests welcome! :)

judgeaxl:
I feared you'd say that :) I don't have a .net environment setup, but I suppose it shouldn't be too hard to get something up and running. The fix itself ought to be reasonably simple.

PR #5236, opened 2026-09-20
Fix next-version parsing with tag prefix

Voyagerroc-Lab:
## What changed\n- Parse configured 
ext-version independently of the Git tag prefix.\n- Add an integration regression test for 
ext-version: 1.2.3 with 	ag-prefix: project-a/.\n\nValidation: the focused regression and tag-prefix integration tests pass; dotnet build and formatting verification pass.\n\nFixes #5228

## Summary by CodeRabbit

* **Bug Fixes**
  * Configured next versions are now recognized correctly even when the repository uses a tag prefix.
  * Next versions retain the expected semantic version format, including the commit-based suffix.


asbjornu (approved):
LGTM! 👍🏼

asbjornu (changes requested):
Thinking of it, we need to document this change as a breaking one.

arturcic:
Quoting [@asbjornu's review](https://github.com/GitTools/GitVersion/pull/5236#pullrequestreview-5261716123):

> Thinking of it, we need to document this change as a breaking one.

@Voyagerroc-Lab The implementation looks correct. Could you document the compatibility change in these places?

1. [`BREAKING_CHANGES.md`](https://github.com/GitTools/GitVersion/blob/main/BREAKING_CHANGES.md), under **Unreleased**: explain that `next-version` is now parsed independently of `tag-prefix`. Previously accepted prefixed values now fail parsing, including `v1.2.3` with the default `[vV]?` prefix, as well as `project-a/1.2.3` with a custom prefix.
2. [`docs/input/docs/migration/v6-to-v7.md`](https://github.com/GitTools/GitVersion/blob/main/docs/input/docs/migration/v6-to-v7.md): add the migration instruction to remove the prefix from `next-version` while retaining `tag-prefix` for Git tags. For example, change `project-a/1.2.3` or `v1.2.3` to `1.2.3`. Use the v7 `calculation` layout in any complete YAML example.
3. [`docs/input/docs/reference/mdsource/configuration.source.md`](https://github.com/GitTools/GitVersion/blob/main/docs/input/docs/reference/mdsource/configuration.source.md#next-version), in the **next-version** section: clarify that the value must be a version without a Git tag prefix and still follows `semantic-version-format` (Strict/Loose).

After updating the reference source, run `mdsnippets --write-header false` from `docs/input` and include the generated `docs/input/docs/reference/configuration.md` in the PR.

